PR #4, opened 2019-10-14
Readme

SGauvin:
Removed redundancy between the README.md and the wiki. Instructions for compiling Lethe are now only in the wiki and a link to the wiki has been added to the readme.


blaisb:
Good idea, this will prevent confusion from now on.

PR #5, opened 2019-11-13
Non linear solver merge

SGauvin:
Merge new non-linear solver class.
Conflicts with master have been solved, and every test passed.


blaisb:
Looks good. There might be minor changes upcoming as they are needed, but this was a much needed refactoring !

PR #6, opened 2019-11-16
Introduce parallel::fullydistributed::Triangulation [WIP]

peterrum:
There is a need to switch from
```cpp
parallel::distributed::Triangulation<dim> triangulation;
```
to
```cpp
std::shared_ptr<parallel::DistributedTriangulationBase<dim>> triangulation;
```
to be able to handle both distributed triangulations.

### DIFF
--- a/applications/initial_conditions/initial_conditions.cc
+++ b/applications/initial_conditions/initial_conditions.cc
@@ -55,8 +55,8 @@ void
 InitialConditionsNavierStokes<dim>::runTest()
 {
   const int initialSize = this->nsparam.mesh.initialRefinement;
-  GridGenerator::hyper_cube(this->triangulation, -1, 1);
-  this->triangulation.refine_global(initialSize);
+  GridGenerator::hyper_cube(*this->triangulation, -1, 1);
+  this->triangulation->refine_global(initialSize);
   // this->make_cube_grid(initialSize);
   this->setup_dofs();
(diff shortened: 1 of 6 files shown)


blaisb:
Seems like the majority of the changes are indentation (my fault, I must not have ran the indentation in the last merge I made from a student). Thanks for the addition, the usage of fully parallel distributed triangulations should really help for a lot of cases.
It seems this pull request breaks two tests :  
- initial_conditions/initial_conditions.debug
- Restart 

The assertion thrown is the following:
--------------------------------------------------------
An error occurred in line <279> of file </home/blab/soft/dealii_luca/dealii/source/distributed/fully_distributed_tria.cc> in function
    void dealii::parallel::fullydistributed::Triangulation<dim, spacedim>::create_triangulation(const std::vector<dealii::Point<spacedim> >&, const std::vector<dealii::CellData<dim> >&, const dealii::SubCellData&) [with int dim = 2; int spacedim = 2]
The violated condition was: 
    currently_processing_create_triangulation_for_internal_usage
Additional information: 
    Use the other create_triangulation() function to create triangulations of type parallel::fullydistributed::Triangulation.!
--------------------------------------------------------

Do you have any idea why that is? I looked at the changes and to me the core was that the triangulation was only an std::shared pointer so I am unsure of where the issue is. I would need that to work before accepting the pull request :)!

PR #7, opened 2019-11-20
Replace direct usage of p:d:t by DistributedTriangulationBase

peterrum:
As a preparation to introduce `parallel::fullydistributed::Triangulation`.

> 100% tests passed, 0 tests failed out of 30

### DIFF
--- a/include/solvers/navier_stokes_base.h
+++ b/include/solvers/navier_stokes_base.h
@@ -316,9 +316,9 @@ protected:
   const unsigned int n_mpi_processes;
   const unsigned int this_mpi_process;
 
-  parallel::distributed::Triangulation<dim> triangulation;
-  DoFHandler<dim>                           dof_handler;
-  FESystem<dim>                             fe;
+  std::shared_ptr<parallel::DistributedTriangulationBase<dim>> triangulation;
+  DoFHandler<dim>                                              dof_handler;
+  FESystem<dim>                                                fe;
 
   TimerOutput computing_timer;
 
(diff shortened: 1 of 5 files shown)


include/solvers/navier_stokes_base.h
```diff
@@ -376,11 +376,13 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
   , mpi_communicator(MPI_COMM_WORLD)
   , n_mpi_processes(Utilities::MPI::n_mpi_processes(mpi_communicator))
   , this_mpi_process(Utilities::MPI::this_mpi_process(mpi_communicator))
-  , triangulation(this->mpi_communicator,
-                  typename Triangulation<dim>::MeshSmoothing(
-                    Triangulation<dim>::smoothing_on_refinement |
-                    Triangulation<dim>::smoothing_on_coarsening))
-  , dof_handler(this->triangulation)
+  , triangulation(dynamic_cast<parallel::DistributedTriangulationBase<dim> *>(
```
peterrum: This is the place where an instance of `p:f:t` will be created.

include/solvers/navier_stokes_base.h
```diff
@@ -1266,11 +1268,18 @@ template <int dim, typename VectorType, typename DofsType>
 void
 NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_kelly()
 {
+  if (dynamic_cast<parallel::distributed::Triangulation<dim> *>(
```
peterrum: Adaptive mesh refinement is unfortunately not supported at the moment for `p:f:t`. However, there is already some work in that direction (see https://github.com/dealii/dealii/pull/8834).

include/solvers/navier_stokes_base.h
```diff
@@ -1586,7 +1592,9 @@ NavierStokesBase<dim, VectorType, DofsType>::read_checkpoint()
 
   try
     {
-      this->triangulation.load(filename.c_str());
+      if (auto tria = dynamic_cast<parallel::distributed::Triangulation<dim> *>(
+            this->triangulation.get()))
+        tria->load(filename.c_str());
```
peterrum: The function `load` is also not implemented yet...

include/solvers/navier_stokes_base.h
```diff
@@ -1883,8 +1891,13 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
   system_trans_vectors.prepare_for_serialization(sol_set_transfer);
 
-  std::string triangulationName = prefix + ".triangulation";
-  this->triangulation.save(prefix + ".triangulation");
+
+  if (auto tria = dynamic_cast<parallel::distributed::Triangulation<dim> *>(
+        this->triangulation.get()))
+    {
+      std::string triangulationName = prefix + ".triangulation";
+      tria->save(prefix + ".triangulation");
```
peterrum: The same for `save`...

include/solvers/navier_stokes_base.h
blaisb: Seems coherent

blaisb:
So if I get this one correctly, right now you only made it so that the triangulation is a shared vector to the base parallel distributed triangulation class (which is aimed at supporting both types of parallel triangulations)?
All test pass so i'll wait for your OK and then merge to master.
The code is going to undergo some change in the next few weeks to make it more readable and just improve the design.
Waiting for your go and then I'll merge !

peterrum:
Exactly! Ready to merge!

blaisb:
Done
Thanks! This will be a great addition.
We'll be coming up with a paper soon, we should keep in touch for that.

PR #10, opened 2020-01-13
Remove trivial setters and getters

SGauvin:
Came across this https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get
Couldn't not change it!
Also simplified some ifs ;)
Removed this->physics_solver everywhere by making a local variable to be able to do solver-> instead

### DIFF
--- a/include/core/skip_newton_non_linear_solver.h
+++ b/include/core/skip_newton_non_linear_solver.h
@@ -49,59 +49,54 @@ SkipNewtonNonLinearSolver<VectorType>::solve(
   if (consecutive_iters == 0 || is_initial_step || force_matrix_renewal)
     assembly_needed = true;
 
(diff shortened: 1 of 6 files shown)


blaisb:
I was always annoyed at myself for having made these stupid getters and setters. Totally agree with the idea of removing them.

include/core/skip_newton_non_linear_solver.h
```diff
@@ -45,63 +45,52 @@ SkipNewtonNonLinearSolver<VectorType>::solve(
   last_res                     = 1.0;
   current_res                  = 1.0;
 
-  bool assembly_needed = false;
-  if (consecutive_iters == 0 || is_initial_step || force_matrix_renewal)
-    assembly_needed = true;
+  bool assembly_needed =
```
blaisb: Much more elegant

PR #22, opened 2020-04-01
all the tests except the application test fixed, all the tests passed…

shahabgol:
All the tests, except the application test are fixed. They all also passed the valgrind successfully.
The problem with the normal_force test was in the test, not in the fine search. The problem was:
I tried to access the particles in contact information before and after the contact. Accessing this vector of maps is not possible before and after the contact.

### DIFF
--- a/source/dem/pp_nonlinear_force.cc
+++ b/source/dem/pp_nonlinear_force.cc
@@ -111,9 +111,17 @@ PPNonLinearForce<dim>::calculate_pp_contact_force(
           Tensor<1, dim> normal_force;
           normal_force = spring_normal_force + dashpot_normal_force;
 
-          double maximum_tangential_overlap =
-            physical_properties.friction_coefficient_particle *
-            normal_force.norm() / tangential_spring_constant;
+          double maximum_tangential_overlap;
+          if (tangential_spring_constant != 0.0)
+            {
+              maximum_tangential_overlap =
+                physical_properties.friction_coefficient_particle *
+                normal_force.norm() / tangential_spring_constant;
+            }
+          else
+            {
+              maximum_tangential_overlap = 0.0;
+            }
 
           // Check for gross sliding
           if (contact_information_iterator_second.tangential_overlap.norm() >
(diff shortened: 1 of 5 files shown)


source/dem/pp_nonlinear_force.cc
```diff
@@ -111,9 +111,17 @@ PPNonLinearForce<dim>::calculate_pp_contact_force(
           Tensor<1, dim> normal_force;
           normal_force = spring_normal_force + dashpot_normal_force;
 
-          double maximum_tangential_overlap =
-            physical_properties.friction_coefficient_particle *
-            normal_force.norm() / tangential_spring_constant;
+          double maximum_tangential_overlap;
+          if (tangential_spring_constant != 0.0)
```
blaisb, line 115: Are you sure that tangential_spring_constant is always positive? This variable is the stiffness of the spring?

tests/dem/normal_force.cc
```diff
@@ -102,24 +103,51 @@ test()
...
 
-      Integ1.integrate(particle_handler, g, dt);
-      if (pw_contact_information_iterator->second.normal_overlap > 0)
+      distance = length + particle->get_location()[0] -
+                 particle->get_properties()[DEM::PropertiesIndex::dp] / 2.0;
+      if (distance > 0.0)
         {
+          Integ1.integrate(particle_handler, g, dt);
+        }
+      else
```
blaisb, line 124: So the root of the cause was that you were trying to access this information when there was no contact I guess right?

source/dem/pp_nonlinear_force.cc
shahabgol: Thanks for reminding the particle IDs. After I recloned the code, I forgot to recorrect the IDs in two tests. I corrected it and repushed it.
It is the tangential stiffness of the spring. It should be positive, since it is defined as (for non-linear model for instance):
kt = 8 * shear_modulus * sqrt(r * normal_overlap)

tests/dem/normal_force.cc
shahabgol: Exactly. I defined a variable distance in the test (which is equal to the normal overlap), and if the distance is positive, all the required functions for force calculations are called.

blaisb: Seems good to me :)!

PR #27, opened 2020-04-15
cleared tests, fixed torques and cross products for 2d

shahabgol:
The tests are cleared, torque calculations fixed and  cross products for 2d is removed.
In the next step I want to add choosing the contact models to the parameter file, and add dem_2d solver.

### DIFF
--- a/applications_tests/dem_3d/packing3d_box.prm
+++ b/applications_tests/dem_3d/packing3d_box.prm
@@ -4,9 +4,9 @@
 # Simulation and IO Control
 #---------------------------------------------------
 subsection simulation control
-  set time step                 			 = 0.00001
-  set step end       					 = 200000
-  set n total              				 = 3000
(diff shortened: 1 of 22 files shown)


applications_tests/dem_3d/packing3d_box.prm
```diff
@@ -4,9 +4,9 @@
 # Simulation and IO Control
 #---------------------------------------------------
 subsection simulation control
-  set time step                 			 = 0.00001
-  set step end       					 = 200000
-  set n total              				 = 3000
+  set time step                 			 = 1e-5
+  set step end       					 = 2e5
+  set n total              				 = 1000
```
blaisb, line 9: The number of particle should only be controlled by the insertion mechanism. It hsould only appear once. We can discuss this tomorrow and do another pull request for that.

PR #42, opened 2020-06-12
Particle insertion is modified to use insert_global_particles, A bug …

shahabgol:
1, Particle insertion is modified to use insert_global_particles
2, A bug in particle-wall contact detection is corrected
3, Four application tests and a test is added to the code


include/dem/dem.h
```diff
@@ -201,8 +210,15 @@ class DEMSolver
...
-  const double dt       = parameters.simulationControl.dt;
+  double       DEM_time           = parameters.simulationControl.dt;
+  const double dt                 = parameters.simulationControl.dt;
+  unsigned int inserted_this_step = 0;
+  unsigned int remained_particles =
+    parameters.simulationControl.total_particle_number;
+  bool         insertion_step = 0;
+  unsigned int number_of_particles_x_direction, number_of_particles_y_direction,
+    number_of_particles_z_direction;
+
```
blaisb, line 221: I don't think these parameters should be stored by the DEM class. It should be the insertion class that knows these elements. Futher more, the number of particles in x , y and z direction are already stored in the parameters::insertion, why are we storing a second copy of them?

include/dem/dem.h
```diff
@@ -261,6 +277,7 @@ class DEMSolver
   ParticlePointLineFineSearch<dim>     particle_point_line_fine_search_object;
   ParticlePointLineForce<dim>          particle_point_line_contact_force_object;
   std::shared_ptr<Integrator<dim>>     integrator_object;
+  std::shared_ptr<Insertion<dim>>      insertion_object;
   std::shared_ptr<PPContactForce<dim>> pp_contact_force_object;
   std::shared_ptr<PWContactForce<dim>> pw_contact_force_object;
   PVDHandler                           pvdhandler;
```
blaisb, line 283: I am not sure the suffix object (*_object) adds anything to the understanding, but if you think this helps.

include/dem/insertion.h
```diff
@@ -54,30 +55,72 @@ template <int dim>
...
-         const Triangulation<dim> &       triangulation,
-         Particles::PropertyPool &        property_pool,
-         const DEMSolverParameters<dim> & dem_parameters) = 0;
+  insert(Particles::ParticleHandler<dim> &                particle_handler,
+         const parallel::distributed::Triangulation<dim> &triangulation,
+         const DEMSolverParameters<dim> &                 dem_parameters,
+         unsigned int &                                   inserted_this_step,
+         const unsigned int &                             nx,
+         const unsigned int &                             ny,
+         const unsigned int &                             nz,
```
blaisb, line 89: nx ny and nz should be given to the constructors of the sub classes. What happens if you have a class that does not use this notion to insert? The number of particles inserted and the insertion mechanism should be controlled internally. The DEM class should not be the one providing how much should be inserted this step. The insertion class should keep track of this.

include/dem/insertion.h
```diff
@@ -54,30 +55,72 @@ template <int dim>
...
+   * Carries out assigning the properties of inserted particles.
+   *
+   * @param dem_parameters DEM parameters declared in the .prm file
+   * @param inserted_this_step Number of particles that is going to be inserted
+   * at each insetion step. This value can change in the last insertion step to
+   * reach the desired number of particles
+   * @param inserted_so_far Number of particles are already inserted
+   */
+  std::vector<std::vector<double>>
+  assign_particle_propertis(const DEMSolverParameters<dim> &dem_parameters,
```
blaisb, line 103: assign_particle_properties with an e

include/dem/insertion.h
```diff
@@ -54,30 +55,72 @@ template <int dim>
...
+   * @param nx Number of insertion points in the x direction
+   * @param ny Number of insertion points in the y direction
+   * @param nz Number of insertion points in the z direction
+   */
+  virtual std::vector<Point<dim>>
+  assign_insertion_points(const DEMSolverParameters<dim> &dem_parameters,
+                          const unsigned int &            inserted_at_this_step,
+                          const unsigned int &            nx,
+                          const unsigned int &            ny,
+                          const unsigned int &            nz) = 0;
```
blaisb, line 123: This should be passed to the constructor of the derived classes and does not belong in the mother class.

include/dem/uniform_insertion.h
```diff
@@ -40,17 +40,13 @@ template <int dim>
...
-   * the insertion in a rectangular box is defined
-   *
-   * @param dem_parameters DEM parameters declared in the .prm file
-   */
-  UniformInsertion<dim>(const DEMSolverParameters<dim> &dem_parameters);
+  UniformInsertion<dim>(const DEMSolverParameters<dim> &dem_parameters,
+                        unsigned int &                  inserted_this_step,
+                        unsigned int &                  nx,
+                        unsigned int &                  ny,
+                        unsigned int &                  nz);
```
blaisb, line 47: I think it would be a lot better if the constructor took instead of the nx ny and nz and inserted_this_step only the dem parameters it needs (so only the Parameters::insertion). Then you can construct whatever you need from this. I don't understand why the constructor requires the inserted_this_step.

source/dem/dem.cc
```diff
@@ -105,6 +105,38 @@ DEMSolver<dim>::reinitialize_force(
...
+        number_of_particles_x_direction,
+        number_of_particles_y_direction,
+        number_of_particles_z_direction);
+    }
+  else
+    {
+      throw "The chosen insertion method is invalid";
+    }
+  return insertion_object;
+}
```
blaisb, line 138: I don't understand why instead of passing number_of_x etc. you don't pass directly an insertion parameter structure.

source/core/parameters_lagrangian.cc
```diff
@@ -317,6 +318,16 @@ namespace Parameters
...
+
+        const std::string insertion = prm.get("insertion_method");
+        if (insertion == "uniform")
+          insertion_method = InsertionMethod::uniform;
+        else if (insertion == "non_uniform")
+          insertion_method = InsertionMethod::non_uniform;
+        else
+          {
+            std::runtime_error("Invalid insertion method");
+          }
```
blaisb, line 330: This does not belong here. The insertion method should be in the insertion sub section, not the DEM model one.

include/dem/uniform_insertion.h
```diff
@@ -40,17 +40,9 @@ template <int dim>
...
-   * particles (maximum_particle_number) which can be inserted in the insertion
-   * box and prints a warning if the number of particles to be inserted at this
-   * step exceeds this calculated number Note that in the current version only
-   * the insertion in a rectangular box is defined
-   *
-   * @param dem_parameters DEM parameters declared in the .prm file
-   */
   UniformInsertion<dim>(const DEMSolverParameters<dim> &dem_parameters);
 
+  // UPDATE
```
blaisb, line 45: You should remove this line

tests/dem/insertion_uniform.cc
```diff
@@ -58,26 +58,25 @@ test()
...
+  dem_parameters.insertionInfo.y_min                     = -0.05;
+  dem_parameters.insertionInfo.z_min                     = -0.05;
+  dem_parameters.insertionInfo.x_max                     = 0.05;
+  dem_parameters.insertionInfo.y_max                     = 0.05;
+  dem_parameters.insertionInfo.z_max                     = 0.05;
+  dem_parameters.insertionInfo.inserted_this_step        = 10;
+  dem_parameters.insertionInfo.distance_threshold        = 2;
+  dem_parameters.physicalProperties.diameter             = 0.005;
+  dem_parameters.physicalProperties.density              = 2500;
+  dem_parameters.simulationControl.total_particle_number = 10;
```
blaisb, line 72: I do not understand why the total particle number is still in the simulation control. It should be in the insertion parameters...

PR #56, opened 2020-08-10
Nitsche ib

CaroleAnneDaunais:
For the Nitsche method, a velocity term has been added to allow for the solid mesh to have a velocity. The particles are still not moving, it will be added shortly.


tests/core/solid_base_22.cc
```diff
@@ -36,7 +36,7 @@ main(int argc, char *argv[])
 
       initlog();
 
-      Parameters::Nitsche                                        param;
+      auto param = std::make_shared<Parameters::Nitsche<2>>();
```
blaisb, line 39: Nice, this is actually a good programming habit

PR #60, opened 2020-08-17
Nitsche moving particles

CaroleAnneDaunais:
Added two functions to SolidBase, one that moves the particles of the solid, and one that moves the solid triangulation, so it can be pictured in paraview or in any other visualization platform as it moves within the fluid. Two tests has been made to ensure the methods works.


include/solvers/gls_nitsche_navier_stokes.h
```diff
@@ -46,6 +46,9 @@ class GLSNitscheNavierStokesSolver : public GLSNavierStokesSolver<spacedim>
                                const unsigned int degreeVelocity,
                                const unsigned int degreePressure);
 
+  virtual void
+  solve() override;
```
blaisb, line 50: Very good habit. Always override virtual functions, this way you are sure of what you are overriding

source/core/solid_base.cc
```diff
@@ -167,6 +174,79 @@ SolidBase<dim, spacedim>::get_solid_velocity()
...
+          double k3 = velocity->value(p2, comp_i);
+
+          Point<spacedim> p3 = particle_location;
+          p3[comp_i] += time_step * k3;
+          double k4 = velocity->value(p3, comp_i);
+
+          particle_location[comp_i] +=
+            time_step / 6 * (k1 + 2 * k2 + 2 * k3 + k4);
+        }
+      particle->set_location(particle_location);
```
blaisb, line 205: There is a mistake here. When comp_i=0 it is correct, but when comp_i=1, then you have already updated the particle location of comp_i=0. This means that you are taking the new x position when calculating your velocity. 
The solution is to calculate the velocity in a loop, but to do the operation on the points directly. k1,k2,k3, k4 should be of type tensor<1,dim>

source/core/solid_base.cc
```diff
@@ -167,6 +174,79 @@ SolidBase<dim, spacedim>::get_solid_velocity()
...
+                  double k4 = velocity->value(p3, comp_i);
+
+                  vertex_position[comp_i] +=
+                    time_step / 6 * (k1 + 2 * k2 + 2 * k3 + k4);
+                }
+              displacement[cell->vertex_index(i)] = true;
+            }
+        }
+    }
+}
```
blaisb, line 248: Same comment as above. The RK4 implementation here is incorrect. You cannot do it component by component this way.

PR #68, opened 2020-09-11
Force evaluatoin correction

Luckabarbeau:
modification of the force evaluation and augmentation of the maximum number of boundaries conditions.


include/core/newton_non_linear_solver.h
```diff
@@ -126,7 +126,7 @@ NewtonNonLinearSolver<VectorType>::solve(
                             << current_res << std::endl;
             }
 
-          if (current_res < 0.9 * last_res || last_res < this->params.tolerance)
+          if (current_res < this->params.relative_tolerance * last_res || last_res < this->params.tolerance)
```
blaisb, line 129: This should be corrected. I'll take care of it.

PR #75, opened 2020-10-11
ID removed from particle properties, writing vtu changed accordingly,…

shahabgol:
ID removed from particle properties
writing vtu changed accordingly
tests fixed


include/dem/dem_properties.h
```diff
@@ -37,28 +37,30 @@ namespace DEM
...
+    omega_y        = 13,
+    omega_z        = 14,
+    mass           = 15,
+    mom_inertia    = 16,
+    M_x            = 17,
+    M_y            = 18,
+    M_z            = 19,
+    displacement_x = 20,
+    displacement_y = 21,
+    displacement_z = 22,
```
blaisb, line 62: Why store each component of the displacement? storing the total displacement would be sufficient (and cheaper)

include/dem/visualization.h
```diff
@@ -122,5 +122,15 @@ class Visualization : public dealii::DataOutInterface<0, dim>
   std::vector<std::tuple<unsigned int, unsigned int, std::string>>
     vector_datasets;
 #  endif
+
+  /**
+   * Number of properties that are written in output files
+   */
+  const unsigned int number_of_properties_to_write = 16;
+
+  /**
```
blaisb, line 131: Why not write all properties? I don't see what is the gain here.

blaisb (commented):
Need clarification on the displacement property and the fact that now all properties are not outputted.

PR #79, opened 2020-10-16
Dem cleanup temp

shahabgol:
- Dynamic contact search frequency is added
- An application test is added for dynamic contact search


include/dem/find_contact_detection_frequency.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef find_contact_detection_frequency_h
+#  define find_contact_detection_frequency_h
+
+/**
+ * Carries out
+ *
+ * @param
+ */
+
```
blaisb, line 36: Missing documentation

source/dem/find_boundary_cells_information.cc
```diff
@@ -150,21 +148,48 @@ BoundaryCellsInformation<dim>::find_particle_point_and_line_contact_cells(
...
+  //               ++v)
+  //            {
+  //              boundary_vertices.insert(
+  //                {face->vertex_index(v), face->vertex(v)});
+  //            }
+  //        }
+  //    }
+
+  // Iterating over the active cells in the trangulation
+  for (const auto &cell : triangulation.active_cell_iterators())
```
blaisb, line 172: Delete now useless comments

source/dem/find_boundary_cells_information.cc
```diff
@@ -173,43 +198,55 @@ BoundaryCellsInformation<dim>::find_particle_point_and_line_contact_cells(
...
+  //      candidate_cells.size();
+  //           ++counter)
+  //        {
+  //          if (!candidate_cells[counter]->at_boundary())
+  //            {
+  //              boundary_cells_with_lines_or_points.push_back(
+  //                candidate_cells[counter]);
+  //            }
+  //        }
+  //    }
```
blaisb, line 232: Delete now useless comments

source/dem/find_contact_detection_frequency.cc
```diff
@@ -0,0 +1,69 @@
+#include <dem/find_contact_detection_frequency.h>
+
+using namespace dealii;
+
+template <int dim>
+bool
+find_contact_detection_frequency(
```
blaisb, line 7: I don't think this is the right name. This function finds if contact detection is necessary, not the frequency.

source/dem/find_contact_detection_frequency.cc
```diff
@@ -0,0 +1,69 @@
...
+                            maximum_particle_diameter / 2)
+    {
+      update_step = true;
+
+      for (auto &particle : particle_handler)
+        {
+          for (unsigned int d = 0; d < dim; ++d)
+            {
+              particle
+                .get_properties()[DEM::PropertiesIndex::displacement_x + d] = 0;
```
blaisb, line 47: Get the properties before the loop in an array view then you use. Otherwise you are getting thrice the cost.

source/dem/find_contact_detection_frequency.cc
```diff
@@ -0,0 +1,69 @@
...
+
+      for (unsigned int d = 0; d < dim; ++d)
+        {
+          particle_properties[DEM::PropertiesIndex::displacement_x + d] +=
+            particle_properties[DEM::PropertiesIndex::v_x + d] * dt;
+
+          max_displacement = std::max(
+            max_displacement,
+            std::abs(
+              particle_properties[DEM::PropertiesIndex::displacement_x + d]));
```
blaisb, line 31: This is weird. The real displacement is the norm of the displacement of the particles. Now you are looking the max displacement in every direction, it should be the norm of the displacement. The max should be outside of the loop. The formula used right now is wrong.

blaisb (changes requested):
The code for the dynamic contact calculation right now is wrong and additionally far from optimal. It gathers the properties far too much.

source/dem/find_contact_detection_frequency.cc
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+  if (max_displacement >= (minimum_cell_size - maximum_particle_diameter / 2) ||
+      max_displacement >= dynamic_contact_search_factor *
+                            (neighborhood_threshold - 1) *
+                            maximum_particle_diameter / 2)
```
blaisb, line 38: The first criterion should be minimun cell_size / 2  directly no? I don't understand why you remove maximum_particle diameter from it.

PR #80, opened 2020-10-22
Log precision cleanup

blaisb:
This pull request clears all the confusion surrounding the writing precision for forces and etc. to the log as the simulation is progressing.
Previously, the log precision was controlled by multiple parameters in various places. This clears everything and centralize the write precision at a single spot, reducing risk of confusion.

### DIFF
--- a/source/core/parameters.cc
+++ b/source/core/parameters.cc
@@ -269,10 +269,6 @@ namespace Parameters
                         "10",
                         Patterns::Integer(),
                         "Calculation frequency");
-      prm.declare_entry("display precision",
-                        "6",
-                        Patterns::Integer(),
-                        "Output frequency");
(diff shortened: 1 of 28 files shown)


source/core/parameters.cc
```diff
@@ -261,10 +269,6 @@ namespace Parameters
                         "10",
                         Patterns::Integer(),
                         "Calculation frequency");
-      prm.declare_entry("display precision",
-                        "6",
-                        Patterns::Integer(),
-                        "Output frequency");
```
acdaigneault, line 267: I removed this parameter in the forces subsection in the wiki

acdaigneault (approved):
Everything seems good! I just removed the "display precision" parameter in the wiki about parameter files.

PR #89, opened 2020-10-28
added example 8 - Ahmed body

jeannejoachim:
Ahmed body example (see wiki page example 8), with prm, geo and msh


blaisb (approved):
Great work. An amazing example!

PR #90, opened 2020-10-29
Dem moving wall

shahabgol:
- Moving boundaries, including rotation and sliding of boundaries are added
- Print_xyz now sorts and print particle information based on particle id
- Useless comments removed and bugs fixed in different sections


source/core/parameters_lagrangian.cc
```diff
@@ -523,8 +521,231 @@ namespace Parameters
...
+                        Patterns::Double(),
+                        "Rotational vector element in x direction");
+      prm.declare_entry("rotational vector y",
+                        "0.",
+                        Patterns::Double(),
+                        "Rotational vector element in y direction");
+      prm.declare_entry("rotational vector z",
+                        "0.",
+                        Patterns::Double(),
+                        "Rotational vector element in z direction");
```
blaisb, line 566: Seems to me like information is duplicated. You have a vector and a speed?

blaisb (approved):
Great :)!

PR #91, opened 2020-10-30
Added application-test

acdaigneault:
Improvement of the flow control algorithm.
New application-test for the flow control feature with a Poiseuille flow solved with the gls_navier_stokes_3d solver.


include/solvers/flow_control.h
```diff
@@ -26,6 +26,27 @@
...
+#include <deal.II/lac/full_matrix.h>
+#include <deal.II/lac/precondition_block.h>
+#include <deal.II/lac/solver_bicgstab.h>
+#include <deal.II/lac/solver_cg.h>
+#include <deal.II/lac/solver_gmres.h>
+#include <deal.II/lac/sparse_direct.h>
+#include <deal.II/lac/sparse_ilu.h>
+#include <deal.II/lac/sparse_matrix.h>
+#include <deal.II/lac/sparsity_tools.h>
+#include <deal.II/lac/vector.h>
```
blaisb, line 41: I don't think all these includes are required.
```suggestion
#include <deal.II/lac/vector.h>
```

include/solvers/flow_control.h
```diff
@@ -26,6 +26,27 @@
...
+
+// Lac - Trilinos includes
+#include <deal.II/lac/trilinos_parallel_block_vector.h>
+#include <deal.II/lac/trilinos_precondition.h>
+#include <deal.II/lac/trilinos_solver.h>
+#include <deal.II/lac/trilinos_sparse_matrix.h>
```
blaisb, line 47: Again those are superflous.
```suggestion

```

include/solvers/flow_control.h
```diff
@@ -34,21 +55,96 @@
...
+   *
+   * @param flow_control. The flow control parameters
+   *
+   * @param simulation_control. The simulation control parameters
+   *
+   * @param step_number. The current step
+   */
+  void
+  calculate_beta(const Parameters::DynamicFlowControl &flow_control,
                  const Parameters::SimulationControl & simulation_control,
```
blaisb, line 107: Why does calculate Beta require the parameter at every calculation?
You should pass the parameter in the constructor of the class and not at every call of calculate_beta. Otherwise it becomes superflous.

blaisb, line 107: You should pass the simulation control if you need it and not the Parameters::SimulationControl.
But do you really need all of those? I think if you just have the iteration number and the time-step you should be ok. This would be a lot better. Especially since you pass the step-number below.

include/solvers/flow_control.h
```diff
@@ -34,21 +55,96 @@
...
+  /**
+   * @brief get_flow_rate. This function gives the flow rate of the
+   * step time. Arguments are the same as calculate_flow_rate(...)
+   */
+  double
+  get_flow_rate(const DoFHandler<dim> &               dof_handler,
+                const VectorType &                    present_solution,
+                const Parameters::DynamicFlowControl &flow_control,
+                const Parameters::FEM &               fem_parameters,
+                const MPI_Comm &                      mpi_communicator);
```
blaisb, line 119: The getter should only return the last flow rate calculated. They should not calculate it.

include/solvers/flow_control.h
```diff
@@ -34,21 +55,96 @@
...
+   * calculate_beta(...)
+   */
+  Tensor<1, dim>
+  get_beta(const DoFHandler<dim> &               dof_handler,
+           const VectorType &                    present_solution,
+           const Parameters::DynamicFlowControl &flow_control,
+           const Parameters::SimulationControl & simulation_control,
+           const Parameters::FEM &               fem_parameters,
+           const unsigned int &                  step_number,
+           const MPI_Comm &                      mpi_communicator);
```
blaisb, line 133: Same as above.
The getter should only return the latest value of beta.

include/solvers/flow_control.h
```diff
@@ -57,104 +153,29 @@ class FlowControl
   double         beta_n1     = 0;
   double         flow_rate_n = 0;
   double         flow_rate_1n;
-  double         area = 0;
+  double         area;
```
blaisb, line 156: Do not give default value to variables like that outside of the constructor. This should be done in the constructor. If you want to set them to zero, set them in the constructor itself.

include/solvers/flow_control.h
```diff
@@ -57,104 +153,29 @@ class FlowControl
...
+
+  // Variables used to improve convergence
+  bool   adjusted         = false;
+  double threshold_factor = 1.01; // 1%
```
blaisb, line 160: Same as above. What does adjusted do? Please add detail on what variables do. Is this variable controlled from the parameter file? Is it really useful?

include/solvers/flow_control.h
```diff
@@ -34,21 +55,96 @@
...
+   * @param fem_parameters. The FEM parameters
+   *
+   * @param mpi_communicator. The mpi communicator information
+   */
+  void
+  calculate_flow_rate(const DoFHandler<dim> &               dof_handler,
+                      const VectorType &                    present_solution,
+                      const Parameters::DynamicFlowControl &flow_control,
+                      const Parameters::FEM &               fem_parameters,
+                      const MPI_Comm &                      mpi_communicator);
```
blaisb, line 91: I really feel this should be moved outside of the flow control class and be just a utilities function that takes into argument the boundary id and etc.
Then the flow control calculate_beta would take the flow_rate as an argument.

blaisb (changes requested):
The core is there and works well.
But there is some major refactoring to do with your get_beta and geta_flow_rate.
I really think that the flow rate calculation should be moved outside of the flow control and the flow control itself should just have the calculate beta and maybe a getter for beta for example.

include/solvers/flow_control.h
acdaigneault: Thanks, I forgot to delete the useless ones :)

include/solvers/flow_control.h
acdaigneault: Oh yeah, good idea!

include/solvers/flow_control.h
acdaigneault: Actually, I have no idea why I didn't see that while splitting the flow_rate and the beta calculation oops.

include/solvers/flow_control.h
acdaigneault: Ok! "Adjusted" allows to enable to set the force value to 0 to get closer to the desired flow rate value. 
It's important if the head loss is to big otherwise every time there's no force, the value get too low.
It's useful also to be set at true when the flow rate go down with no force and then get below the value. It means that the flow rate reached the value and now need to be controlled by a calculated beta to keep the desired flow rate.
I said "If flow is reached with no force, adjusted is set to true" at line 77 of the .cc file, but I will give more information.

acdaigneault:
The flow rate calculation is now separated of the FlowControl class.
The verbosity parameter was added to the flow control subsection.
New unit test for flow control.

blaisb (approved):
Great work! Just a tiny change on the constructor (that I will take care off) and this is ready for a merge :)!

PR #92, opened 2020-10-31
Example 9 is added, corrected dem examples

shahabgol:
Example 9 (rotating drum) is added
Other dem tests are updated


blaisb (approved):
Nice example!

PR #93, opened 2020-11-03
Time-step inspection is added

shahabgol:
- Inspection of input parameters is moved to a separate function. I will use this function for other inspections on input parameters.
- at() operator is replaced with [] operator for all the maps

### DIFF
--- a/source/dem/dem.cc
+++ b/source/dem/dem.cc
@@ -51,13 +51,6 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
   , repartition_frequency(parameters.model_parameters.repartition_frequency)
   , insertion_frequency(parameters.insertion_info.insertion_frequency)
   , physical_properties(parameters.physical_properties)
-  , rayleigh_time_step(
-      M_PI_2 * parameters.physical_properties.diameter *
-      sqrt(2 * parameters.physical_properties.density *
-           (2 + parameters.physical_properties.poisson_ratio_particle) *
-           (1 - parameters.physical_properties.poisson_ratio_particle) /
-           parameters.physical_properties.youngs_modulus_particle) /
-      (0.1631 * parameters.physical_properties.poisson_ratio_particle + 0.8766))
   , background_dh(triangulation)
 {
   // Change the behavior of the timer for situations when you don't want outputs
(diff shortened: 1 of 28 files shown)


blaisb:
@shahabgol Could you add a description of this PR so that the commit message can be more thorough?

source/dem/dem.cc
```diff
@@ -51,6 +51,13 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
   , repartition_frequency(parameters.model_parameters.repartition_frequency)
   , insertion_frequency(parameters.insertion_info.insertion_frequency)
   , physical_properties(parameters.physical_properties)
+  , rayleigh_time_step(
+      M_PI_2 * parameters.physical_properties.diameter *
+      sqrt(2 * parameters.physical_properties.density *
+           (2 + parameters.physical_properties.poisson_ratio_particle) *
+           (1 - parameters.physical_properties.poisson_ratio_particle) /
+           parameters.physical_properties.youngs_modulus_particle) /
+      (0.1631 * parameters.physical_properties.poisson_ratio_particle + 0.8766))
```
blaisb, line 60: The calculation of the rayleigh time-step should be move to a small function.
Eventually we might want to calculate multiple critical time-step (hertz, etc). So instead of hardcoding it in the constructor, it would be better to move that to a small function and to call that function in the constructor.

source/dem/dem.cc
```diff
@@ -117,6 +124,20 @@ DEMSolver<dim>::print_initial_info()
...
+            << "% of Rayleigh time step" << std::endl;
+      pcout << "It is recommended to decrease the time-step" << std::endl;
+    }
+  else if (simulation_control->get_time_step() < 0.01 * rayleigh_time_step)
+    {
+      pcout << "Warning: DEM time-step is "
+            << (simulation_control->get_time_step() / rayleigh_time_step) * 100
+            << "% of Rayleigh time step" << std::endl;
+      pcout << "It is recommended to increase the time-step" << std::endl;
+    }
```
blaisb, line 140: It it a  good idea to suggest increasing or decreasing the time step. However, I would print the value of the fraction no-matter what.
You could print the % of the Rayleigh time step. Then if the value is >0.15 or <0.01 then you could print the warning. It would be clearer and this way the user would always be aware of the Rayleigh time step !

PR #95, opened 2020-11-05
Postprocessing velocities (average velocities)

acdaigneault:
This branch adds the AverageVelocities class which calculates the time-averaged velocities and pressure so far. 
This PR contains the modifications of the flow control branch because a new implementations (bulk velocity) in the FlowControl class is needed for the nondimensionalization of the average velocities.


examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -44,7 +44,18 @@ subsection flow control
...
+
+#---------------------------------------------------
+# Post-Processing
+#---------------------------------------------------
+subsection post-processing
+    set calculate average velocities    = true
+    set initial time 		        = 0.1
+    set boundary id    	           	= 0
+    set flow direction 		        = 0
+    set nondimensionalization           = true
```
blaisb, line 58: Is it really necessary to do the dimensionless analysis from the code directly? What prevents doing it from outside directly into paraview?

examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -98,7 +109,7 @@ end
 subsection non-linear solver
   set tolerance                   = 1e-5
   set max iterations              = 10
-  set verbosity                   = verbose
+  set verbosity                   = quiet
```
blaisb, line 112: Keep the verbosity for the examples. Those are good things to take a habit of looking at!

include/solvers/flow_control.h
```diff
@@ -26,6 +26,27 @@
 #include <deal.II/fe/fe_values.h>
 #include <deal.II/fe/mapping_q.h>
 
+// Lac
```
blaisb, line 29: This part will be fixed when we finish the other PR.

include/solvers/postprocessing_velocities.h
```diff
@@ -0,0 +1,156 @@
...
+
+  if (simulation_control->get_step_number() == 0)
+  {
+    // Reinitializing vectors with zeros and dt at t = 0
+    sum_velocity_dt.reinit(locally_owned_dofs,
+                           mpi_communicator);
+    average_velocities.reinit(locally_owned_dofs,
+                              mpi_communicator);
+  }
+  else if (abs(total_time) < 1e-6 || total_time > 0)
```
blaisb, line 114: I'm not sure I understand || here? If total_time >0 it should be sufficient, why the abs check?
Time is also always positive no?

include/solvers/postprocessing_velocities.h
```diff
@@ -0,0 +1,156 @@
...
+  {
+    // Generating average velocities at each time from initial time
+    inv_range_time = 1. / (total_time + dt);
+
+    VectorType velocity_dt(locally_owned_dofs,
+                           mpi_communicator);
```
blaisb, line 120: Why reconstruct it every time?

include/solvers/postprocessing_velocities.h
```diff
@@ -0,0 +1,156 @@
...
+    velocity_dt.equ(dt, local_evaluation_point);
+    sum_velocity_dt += velocity_dt;
+
+    if (simulation_control->is_output_iteration())
+      average_velocities.equ(inv_range_time, sum_velocity_dt);
```
blaisb, line 125: This function should not have access to the simulation control. Only the time step.

source/solvers/navier_stokes_base.cc
```diff
@@ -868,6 +864,50 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
...
+    {
+      // If Flow Control is disable, it calculates the bulk velocity through
+      // this class
+      if (!nsparam.flow_control.enable_flow_control &&
+          nsparam.post_processing.nondimensionalization)
+        {
+          nsparam.flow_control.id_flow_control =
+            nsparam.post_processing.id_flow_control;
+          nsparam.flow_control.flow_direction =
+            nsparam.post_processing.flow_direction;
```
blaisb, line 878: I really don't like this "dimensionless-ness". It couples too strongly various elements of the code.

blaisb (commented):
I think this will be better merged after you have completed the changes for the flow control :). It is a good idea to separate the way the class behaves from the flow control class.

blaisb:
> I think this will be better merged after you have completed the changes for the flow control :). It is a good idea to separate the way the class behaves from the flow control class and the simulation control class.

examples/07-periodic_hills/periodic_hills.prm
acdaigneault: Ok I will remove it, I didn't know we could do it in Paraview.

examples/07-periodic_hills/periodic_hills.prm
acdaigneault: Actually, I have commited the .prm file but I didn't want to, so it's not the final .prm for the example, it was only for testing.
But I will keep the verbosity for the final .prm, I promise :)

include/solvers/postprocessing_velocities.h
acdaigneault: The condition is once the initial time is reached do ... 
But with doubles I'm never sure that my condition is gonna be good if I don't use a tolerance because total_time is a subtraction. There's absolutely no chance total_time is going to be below 0? Or I guess if I test it and it works, it should works everytime.

include/solvers/postprocessing_velocities.h
acdaigneault: Now I think I know how to reset the values to zero, didn't know when I wrote that :)

source/solvers/navier_stokes_base.cc
acdaigneault: Yes, I will remove it

include/solvers/navier_stokes_base.h
```diff
@@ -324,6 +324,9 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   TableHandler                                 kinetic_energy_table;
   AverageVelocities<dim, VectorType, DofsType> average_velocities;
 
+  // Tolerance to compare doubles
+  double epsilon = 1e-6;
+
```
blaisb: No need to define that as a class member. You can define it locally where you need it. It is not a critical parameter.

source/solvers/postprocessing_velocities.cc
```diff
@@ -1,15 +1,66 @@
...
 
-template class
-AverageVelocities<2, TrilinosWrappers::MPI::Vector, IndexSet>;
 
-template class
-AverageVelocities<3, TrilinosWrappers::MPI::Vector, IndexSet>;
+template <int dim, typename VectorType, typename DofsType>
+AverageVelocities<dim, VectorType, DofsType>::AverageVelocities()
+{
+  epsilon = 1e-6;
```
blaisb, line 7: Like above, no need for this to be a class member.

blaisb (approved):
Looking good. Very minor comments. It would be good to remove the commits from this PR that relate to flow control though. We can look at this together if you want ;).
It seems the tests are failing right now though!

acdaigneault:
The master was merged into the branch and conflicts were resolved.

examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -79,7 +87,7 @@ end
 # FEM
 #---------------------------------------------------
 subsection FEM
-    set velocity order            = 2
```
blaisb, line 82: Is there a reason for this other than speed? It's ok if the example takes a while to run. I just wanted to keep the example using high-order mapping

include/core/parameters.h
```diff
@@ -578,7 +584,7 @@ namespace Parameters
     // Volumetric flow rate (L^3/t)
     double flow_rate_0;
 
-    // Id of the boundary where the flow come from
+    // Id of the boundary where the flow inlet
```
blaisb, line 587: What does "where the flow inlet" mean?

include/solvers/postprocessing_velocities.h
```diff
@@ -0,0 +1,91 @@
...
+   * @param post_processing. The parameters to start the processing
+   *
+   * @param locally_owned_dofs. The owned dofs
+   *
+   * @param mpi_communicator. The mpi communicator information
+   */
+  void
+  calculate_average_velocities(
+    const VectorType &                        local_evaluation_point,
+    const std::shared_ptr<SimulationControl> &simulation_control,
```
blaisb, line 72: This should be a regular pointer. No need to pass the shared pointer around as a shared pointer.

include/solvers/postprocessing_velocities.h
```diff
@@ -0,0 +1,91 @@
...
+class AverageVelocities
+{
+public:
+  /**
+   * @brief calculate_average_velocities. This function calculates time-averaged
+   * velocities and pressure with dof vector with no ghost cell.
+   *
+   * @param local_evaluation_point. The vector solutions with no ghost cells
+   *
+   * @param simulation_control. The simulation information (time)
```
blaisb, line 61: Why do you pass the entire simulation_control if the only things you need are the time? I am just wondering.

source/core/parameters.cc
```diff
@@ -337,6 +337,29 @@ namespace Parameters
...
+
+      prm.declare_entry("boundary id",
+                        "0",
+                        Patterns::Integer(),
+                        "Boundary id of the inlet flow");
+
+      prm.declare_entry("flow direction",
+                        "0",
+                        Patterns::Integer(),
+                        "Flow direction at flow inlet");
```
blaisb, line 360: I do not understand why those are needed for average velocities. I really don't think it's worth it to add all these parameters when all they will do is allow you to divide by a constant which could easily be done in paraview or elsewhere.

source/solvers/navier_stokes_base.cc
```diff
@@ -871,6 +869,33 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
...
 
+  if (this->nsparam.post_processing.calculate_average_velocities)
+    {
+      double epsilon = 1e-6;
+      // Reinitiation of the average_solution vector at the first iteration.
+      if (firstIter)
+        {
+          this->average_solution.reinit(this->locally_owned_dofs,
+                                        this->locally_relevant_dofs,
+                                        this->mpi_communicator);
```
blaisb, line 880: This could be re-inited in setup_dofs, but it is not a bad idea to have put it there also.

source/solvers/navier_stokes_base.cc
```diff
@@ -871,6 +869,33 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
         }
     }
 
+  if (this->nsparam.post_processing.calculate_average_velocities)
+    {
+      double epsilon = 1e-6;
```
blaisb, line 874: Epsilon should be declared just straight above where it is used.

source/solvers/navier_stokes_base.cc
```diff
@@ -1100,6 +1127,27 @@ NavierStokesBase<dim, VectorType, DofsType>::write_output_results(
...
+        average_data_component_interpretation(
+          dim, DataComponentInterpretation::component_is_part_of_vector);
+      average_data_component_interpretation.push_back(
+        DataComponentInterpretation::component_is_scalar);
+
+      data_out.add_data_vector(this->average_solution,
+                               average_solution_names,
+                               DataOut<dim>::type_dof_data,
+                               average_data_component_interpretation);
+    }
```
blaisb, line 1149: Great!

examples/07-periodic_hills/periodic_hills.prm
acdaigneault: I know, as I said, this is not my final .prm file, but I still need to change it because of the new parameters :)

include/core/parameters.h
acdaigneault: Boundary id at inlet

source/solvers/postprocessing_velocities.cc
```diff
@@ -0,0 +1,61 @@
...
+  const DofsType &                          locally_owned_dofs,
+  const MPI_Comm &                          mpi_communicator)
+{
+  double epsilon = 1e-6;
+  double total_time =
+    simulation_control->get_current_time() - post_processing.initial_time;
+  double dt = simulation_control->get_time_step();
+
+  // When averaging velocities begins (time = initial time)
+  if (abs(total_time) < epsilon)
```
blaisb, line 19: This is not a safe check. What happens if initial_time=10 and your time step brings you to 10.01. This check will fail. You inherently assume that you will start averaging "pile poil" at a multiple of your time step. This is impossible with adaptative time-stepping. You need to find a more robust mechanism.

include/solvers/postprocessing_velocities.h
acdaigneault: I think I saw it like that somewhere in the code and I thought it was the only way to have simulation_control->get_current_time() and the other getters

source/solvers/postprocessing_velocities.cc
```diff
@@ -0,0 +1,61 @@
...
+
+template class AverageVelocities<3, TrilinosWrappers::MPI::Vector, IndexSet>;
+
+template class AverageVelocities<2,
+                                 TrilinosWrappers::MPI::BlockVector,
+                                 std::vector<IndexSet>>;
+
+template class AverageVelocities<3,
+                                 TrilinosWrappers::MPI::BlockVector,
+                                 std::vector<IndexSet>>;
```
blaisb, line 61: Great!

tests/solvers/average_velocities_01.cc
```diff
@@ -0,0 +1,137 @@
...
+          simulation_control->integrate();
+
+          // Break if the next time from integrate() is the same because
+          // time will never get over the time end, but the average velocities
+          // at this time is wanted.
+          if (abs(time - simulation_control->get_current_time()) < epsilon)
+            break;
+
+          time = simulation_control->get_current_time();
+        }
```
blaisb, line 108: It is better to put the test itself in a test function instead of directly in the main file.

blaisb (changes requested):
Be careful. I added some comments. There are some places where you logic would fail if you don't use a constant time step. If I were you, I would add a second unit test where you test with a variable time-step simulation control. This would really allow you to see if everything is behaving like you want it to. Great work. This is very close to completion!

source/core/parameters.cc
acdaigneault: I thougth I removed it. this is a mistake, those parameters are deleted in the .h.

include/solvers/postprocessing_velocities.h
acdaigneault: It can be changed

source/solvers/postprocessing_velocities.cc
acdaigneault: Oh ok, did not think about it, I'm not very familiar with the adaptative time-stepping.

tests/solvers/average_velocities_01.cc
acdaigneault: It means line 71 to 107 should be in a function?

source/solvers/navier_stokes_base.cc
acdaigneault: It is used in the else if and this is right before the if, I can not really declare closer where it is use. In this case, don't you think it's better to use 1e-6 instead of epsilon?

README.md
```diff
@@ -26,6 +26,7 @@ Contributors (in alphabetical order):
 - Antoine Avrit
 - Lucka Barbeau
 - Valérie Bibeau
+- Audrey Collard-Daigneault
```
blaisb, line 29: Yay!

include/solvers/postprocessing_velocities.h
```diff
@@ -87,5 +91,8 @@ class AverageVelocities
   VectorType velocity_dt;
   VectorType sum_velocity_dt;
   VectorType average_velocities;
+
+  bool   average_calculation;
+  double real_initial_time;
```
blaisb, line 96: Cool idea. Well done

blaisb (approved):
I must say you are becoming a natural at this. Great work!

PR #99, opened 2020-11-18
Reynolds stress

acdaigneault:
This branch adds a Reynolds stress calculation. We can display Reynolds stresses (<u'u'>, <v'v'>, <w'w'>,<u'v'>,<v'w'>,<w'u'>)  in Paraview like other solutions.
It works for 2d/3d and for both gls_navier_stokes and gd_navier_stokes solvers.
There's 2 unit tests, the first one tests 3d with Trilinos vectors and the second one tests 2d with Trilinos block vectors with adaptive time steps.


include/solvers/postprocessing_velocities.h
```diff
@@ -92,7 +118,28 @@ class AverageVelocities
   VectorType sum_velocity_dt;
   VectorType average_velocities;
 
+  VectorType fluctuation;
+  VectorType reynolds_stress_dt;
+  VectorType sum_reynolds_stress_dt;
```
blaisb, line 123:
```suggestion
  VectorType sum_reynolds_stress_dt; +x
```

examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -6,18 +6,18 @@
 subsection simulation control
   set method                       = bdf1
   set number mesh adapt            = 0
-  set output name                  = periodic_hills-output
+  set output name                  = periodic_hills_re5600_dt1E-1-output
```
blaisb, line 9: No need for such a complex output name. You can just call it periodic_hill_output  or just hill_output or whatever. Make it simple :)

examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -6,18 +6,18 @@
...
+  set output frequency             = 2
+  set time end                     = 10
 end
 
 #---------------------------------------------------
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
-    set kinematic viscosity        = 1E-3
+    set kinematic viscosity        = 1.95E-05 # Re = 5600
```
blaisb, line 20: Yay!

examples/07-periodic_hills/periodic_hills.prm
```diff
@@ -53,7 +53,8 @@ end
 #---------------------------------------------------
 subsection post-processing
     set calculate average velocities    = true
-    set initial time 		        = 0.5
+    set calculate reynolds stress       = true
+    set initial time 		        = 100
```
blaisb, line 57: This would be how it would be done in the real case? For the tutorial it would be better to have it done sooner, even if this will not be the "real" way it is done. The idea  is still to be able to run the tutorial for a limited amount of time like 30-minutes or 1 hour.

include/core/parameters.h
```diff
@@ -238,6 +238,9 @@ namespace Parameters
     // Enable velocity post-processing
     bool calculate_average_velocities;
 
+    // Enable reynolds stress post-processing
+    bool calculate_reynolds_stress;
+
```
blaisb, line 243: I wonder if we should really add this. I mean, when you calculate averages you will most likely calculate_average_velocities we will also want the Reynolds stresses. I'm not sure we need an additional parameter for this shenanigan :)

source/solvers/postprocessing_velocities.cc
```diff
@@ -43,18 +54,121 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
...
+      average_velocities.equ(inv_range_time, sum_velocity_dt);
+    }
 }
 
+// Since Trilinos vectors and block vectors data doesn't have the same
+// structure, those vectors need to be processed in different ways.
+// Same about the dimension, where 2D solutions are stored in (u,v,p) groups
+// for vectors and [(u,v)][p] for block vectors and where 3D solutions
+// are stored in (u,v,w,p) groups for vectors and [(u,v,w)][p] for
+// block vectors.
```
blaisb, line 82: Love this

source/solvers/postprocessing_velocities.cc
```diff
@@ -43,18 +54,121 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
...
 template <int dim, typename VectorType, typename DofsType>
-const VectorType
-AverageVelocities<dim, VectorType, DofsType>::get_average_velocities()
+void
+AverageVelocities<dim, VectorType, DofsType>::calculate_reynolds_stress(
+  const VectorType &local_evaluation_point,
+  const bool &      is_output_iteration)
 {
-  return average_velocities;
+  if constexpr (std::is_same_v<VectorType, TrilinosWrappers::MPI::Vector>)
```
blaisb, line 89: Ohh wow. Popping up the fancy constexpr? This is the good stuff.

source/solvers/postprocessing_velocities.cc
```diff
@@ -43,18 +54,121 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
...
+    {
+      const unsigned int begin_index =
+        local_evaluation_point.local_range().first;
+      const unsigned int end_index =
+        local_evaluation_point.local_range().second;
+
+      for (unsigned int i = begin_index; i < end_index; i++)
```
blaisb, line 96: Why loop with increments of one then? You could loop by increments of dim+1 : " i+= dim+1". C++ allows thats. You can do whatever you want to i.

source/solvers/postprocessing_velocities.cc
```diff
@@ -43,18 +54,121 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
...
+      unsigned int end_index =
+        local_evaluation_point.block(0).local_range().second;
+      unsigned int size_velocity_block = reynolds_stress.block(0).size();
+
+      for (unsigned int i = begin_index; i < end_index; i++)
+        {
+          if (dim == 2 && (i + 2) % 2 == 0)
+            {
+              reynolds_stress_dt.block(1)[i / 2 + size_velocity_block] =
+                fluctuation.block(0)[i] * fluctuation.block(0)[i + 1] * dt;
```
blaisb, line 155: Great, this is legit.

source/solvers/postprocessing_velocities.cc
```diff
@@ -12,26 +12,37 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
   const Parameters::PostProcessing &post_processing,
   const double &                    current_time,
   const double &                    time_step,
+  const bool &                      is_output_iteration,
```
blaisb, line 15: The idea here I guess is to only calculate average when you really need it right? Why not calculate it then when we ask for it instead? (in the getter) this would prevent this argument being passed here. To me it is weird to have a calculation routine depending on an i/o argument.

source/solvers/postprocessing_velocities.cc
```diff
@@ -12,26 +12,37 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_average_velocities(
...
       sum_velocity_dt.reinit(locally_owned_dofs, mpi_communicator);
       average_velocities.reinit(locally_owned_dofs, mpi_communicator);
+
+      // Reinitialisation of reynolds stress vectors if calculation is enable
+      if (post_processing.calculate_reynolds_stress)
+        {
+          fluctuation.reinit(locally_owned_dofs, mpi_communicator);
+          reynolds_stress_dt.reinit(locally_owned_dofs, mpi_communicator);
+          sum_reynolds_stress_dt.reinit(locally_owned_dofs, mpi_communicator);
+          reynolds_stress.reinit(locally_owned_dofs, mpi_communicator);
```
blaisb, line 44: RIght now this means that you average velocity calculation won't work with dynamic mesh adaptation. You should at least document that because when we refine the mesh we need to "reinit the vectors and transfer the solution". It's not mandatory right now, but clearly this support should be added eventually.

blaisb (approved):
Looks very good to me. Very good work.
Things to do:

- Add documentation to the Wiki
- Verify that you really need to pass is_output_iteration every calculation or not (why not calculate it only when you need a getter). Additionnaly, don't you always need to have the average velocity freshen up for reynold stresses?
- No need for two bools, just calculate both Reynolds stresses and average, no need for two bools I think
- Clearly document that the routine will fail if mesh adaptation is used. This is something we can fix in a next PR by adding a "adapt to new mesh" routine to the averaging class. We can look at this together
- Finally, this is very very good work!

examples/07-periodic_hills/periodic_hills.prm
acdaigneault: I should be more careful about committing the final .prm in my last push, my bad!

source/solvers/postprocessing_velocities.cc
acdaigneault: Do you think I should add a condition where if there's a dynamic mesh adaptation all vectors reinitializes at each iteration?

source/solvers/postprocessing_velocities.cc
acdaigneault: My idea was to only calculate the average_velocities = sum__velocities_dt / dt when we want the data out or if reynolds stress is enable (because it needs the average velocities at each iteration). If calculate_reynolds_stress is removed, average_velocities should be calculated everytime so no needs to have an boolean of something in the getter to calculate it anymore.

source/solvers/postprocessing_velocities.cc
acdaigneault: Because if begin_index is not 0, dim+1, 2(dim+1), ... i will never get the right value for the vector indexation. I know it's kinda weird, it's because of the <u'v'> calculation. I tried it and it didn't work, but I will try again to make sure.

acdaigneault: I did try and it did not work, but I must have done something wrong because it actually works mhm.

source/solvers/navier_stokes_base.cc
```diff
@@ -876,19 +876,23 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
...
       if (firstIter)
         {
+          if (nsparam.mesh_adaptation.type ==
+              Parameters::MeshAdaptation::Type::none)
+            AssertThrow(true,
+                        ExcMessage(
+                          "Time-averaging velocities and calculating reynolds "
+                          "stresses are currently unavailable for mesh "
+                          "adaptation."));
+
```
blaisb, line 895: This is not correct. You want to assert throw nsparam.mesh_adaptation.type == none.
What this means is AssertThrow(nsparam.mesh_adaptation.type==none,...) If the condition is not true, it will throw and error and stop execution of the program. Right now this is working because you are throwing true, but true will always be true and thus always pass.

blaisb (approved):
Just a tiny mistake with the AssertThrow, but I will fix it before merging :) !
Great work! Looking forward to launching this on Niagara :)! Can't wait

source/solvers/navier_stokes_base.cc
acdaigneault: Ohhhhhh yeah I get it!

acdaigneault:
> Just a tiny mistake with the AssertThrow, but I will fix it before merging :) !
> Great work! Looking forward to launching this on Niagara :)! Can't wait

Me too :) !!

blaisb:
Some elements still need to be fixed :

1. Reynold stresses should be interpreted as tensors and not as vector + scalaer
2. Two components of the Reynolds stresses are missing. The full tensor should be calculated.
3. There are some places where you pass your vectors by copies instead of by reference. For example 
```
  const VectorType
  get_average_velocities()
  {
    return average_velocities;
  }

  /**
   * @brief get_reynolds_stress. Gives the time-averaged normal Reynolds
   * stresses and shear stress
   */
  const VectorType
  get_reynolds_stress()
  {
    return reynolds_stress;
  }
```

acdaigneault:
Last commits implement the Reynolds stresses independent components instead of only normal Reynolds stresses and shear stress.
There's new objets related to it as dof_hanfler_rs and fe_rs. There are needed to allow data out because que number of independent components is not the same as the number of dof per node.

blaisb (approved):
Ahh, like a melody to my eyes.
Good work. I will merge it tomorrow once Actions shenanigans are finished with.

PR #104, opened 2020-11-25
Refactoring unit tests

acdaigneault:
This is some minor refactoring on the unit tests to standardize format.
Shahab, I added you as reviewer because I added try blocks in your dem unit tests, you may want to take a look.
Jeanne, I think you did not do unit test so far, maybe it will help to see what it looks like. :)

### DIFF
--- a/tests/core/newton_non_linear_solver_01.cc
+++ b/tests/core/newton_non_linear_solver_01.cc
@@ -1,15 +1,3 @@
-#include <deal.II/lac/full_matrix.h>
-#include <deal.II/lac/lapack_full_matrix.h>
-#include <deal.II/lac/sparse_direct.h>
-#include <deal.II/lac/vector.h>
-
-#include <core/newton_non_linear_solver.h>
-#include <core/physics_solver.h>
-
-#include <iostream>
-#include <memory>
-
-#include "../tests.h"
 #include "non_linear_test_system_01.h"
 
 /**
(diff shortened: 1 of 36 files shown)


tests/core/non_linear_test_system_01.h
```diff
@@ -17,18 +17,8 @@
...
-#include <deal.II/lac/sparse_direct.h>
-#include <deal.II/lac/vector.h>
-
-#include <core/newton_non_linear_solver.h>
-#include <core/non_linear_solver.h>
 #include <core/physics_solver.h>
 
-#include <iostream>
-#include <memory>
-
```
blaisb, line 31: Are you sure? To me it seems like the .h would not compile without these includes. Or it makes it depend from the includes of another file?

tests/core/newton_non_linear_solver_01.cc
```diff
@@ -1,29 +1,13 @@
...
-#include <deal.II/lac/sparse_direct.h>
-#include <deal.II/lac/vector.h>
-
-#include <core/newton_non_linear_solver.h>
-#include <core/physics_solver.h>
-
-#include <iostream>
-#include <memory>
-
-#include "../tests.h"
```
blaisb, line 12: Are these includes really non-necessary? Then where are they drawn from? I am always of cross-include dependency. Clearly they are not necessary in the .cc, but you are sure they are not needed in the .h?

blaisb (commented):
Very good work (and thank you so much). Just fix a bit the overdeletion of includes and then this would be ready for merge from my point of view.

shahabgol (approved):
Perfect. These cross-includes are really confusing. I should do the same and clear them on the DEM side.

shahabgol (approved):

shahabgol (approved):

blaisb (approved):
Very good! Love the template. If this passes CI I'll merge it straight away

PR #105, opened 2020-11-27
9.2 compatibility

blaisb:
The last merged PR broke compatibility with 9.2 without us noticing.
This PR uses pragmas to detect the deal.II version and fix this issue.


acdaigneault:
But it still should not compile for deal.II version 9.3.0-pre.

blaisb:
> But it still should not compile for deal.II version 9.3.0-pre.

It will compile for 9.3.0-pre which is up to date. 9.3.0-pre has not been crystallised yet, so this moniker means a constantly changing version of deal.II.

PR #107, opened 2020-11-28
Add update ghosts mechanism to DEM

blaisb:
This pull request adds the update ghost mechanism to the DEM code. This mechanism makes the execution in parallel significantly more efficient (at least 5x) because it prevents the recreation of the ghost particles multimap and, instead, re-uses the particle data. To use the content of this PR, deal.II 9.3-pre must be installed with, at least, a version dating from early november.


shahabgol (approved):
perfect.

PR #108, opened 2020-12-02
Reynolds stress v2

acdaigneault:
This PR fixes the current Reynolds stresses calculation.
Now this calculation sets up 2 vectors (Reynolds normal stress vector & Reynolds shear stress vector) with the solution DofHandler. The turbulent kinetic energy is also calculated.
There's a minor issue. There's dummy components in the data_out (dummy_rss and dummy_rss_2d) since there is 3 (2d) or 4 (3d) dofs per node and 1 (2d) or 3 (3d) components per node for the Reynolds shear stress vector.
It also fixes the bug about ghost cells in debug mode.


blaisb (approved):
Good work, a very interesting PR and well done. the dummy_rss is the annoying part but we have to live with it. :)!

PR #109, opened 2020-12-03
Add checkpoints for FlowControl and AverageVelocities

acdaigneault:
This branch implements saving and getting checkpoints for flow control (last flow rate, beta coefficient and some useful variables for the algorithm) and for the average velocities/Reynolds stresses calculation (sum vectors with value*dt and some other variables).


applications_tests/gls_navier_stokes_3d/CMakeLists.txt
```diff
@@ -1,5 +1,25 @@
...
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0005.pvtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0010.pvtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0015.pvtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0000.0000.vtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0005.0000.vtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0010.0000.vtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
+file(COPY poiseuille_restart_files/poiseuille_restart-output.0015.0000.vtu DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/poiseuille_restart.${CMAKE_LOWER_BUILD_TYPE})
```
blaisb, line 21: This is a lot of binary files to keep track of. Maybe keep only 2 or 3 instead of 4? This is very minor anyway

include/core/grids.h
```diff
@@ -72,6 +72,7 @@ read_mesh_and_manifolds(
                                                           triangulation,
   const Parameters::Mesh &                                mesh_parameters,
   const Parameters::Manifolds &                           manifolds_parameters,
+  const bool &                                            restart,
```
blaisb, line 75: No criticism intended here, but why is this inserted in the middle?

source/solvers/flow_control.cc
```diff
@@ -111,14 +111,36 @@ FlowControl<dim>::calculate_beta(const std::pair<double, double> &flow_rate,
...
 
+template <int dim>
+void
+FlowControl<dim>::read(std::string prefix)
+{
+  std::string   filename = prefix + ".flowcontrol";
+  std::ifstream input(filename.c_str());
+  if (!input)
+    {
+      throw("Unable to open file");
```
blaisb, line 135: You have to throw an exception here and not just throw a string.
For example `throw std::runtime_error("Text")`
Or you can use AssertThrow
https://www.dealii.org/current/doxygen/deal.II/group__Exceptions.html#gafc0ca7ad85b3ebd64e8e51689ac85caf

source/solvers/flow_control.cc
```diff
@@ -111,14 +111,36 @@ FlowControl<dim>::calculate_beta(const std::pair<double, double> &flow_rate,
...
+    }
+  std::string buffer;
+  std::getline(input, buffer);
+  input >> buffer >> beta_n;
+  input >> buffer >> flow_rate_1n;
+  input >> buffer >> no_force;
+  input >> buffer >> threshold_factor;
```
blaisb, line 142: I always forget if it is better to close the file or if the end of the scope of the function will close it automatically anyway....

source/solvers/gd_navier_stokes.cc
```diff
@@ -630,6 +630,29 @@ GDNavierStokesSolver<dim>::setup_dofs()
   system_matrix.reinit(sparsity_pattern);
   pressure_mass_matrix.reinit(sparsity_pattern.block(1, 1));
 
+  if (this->nsparam.post_processing.calculate_average_velocities)
+    {
+      AssertThrow(this->nsparam.mesh_adaptation.type ==
```
blaisb, line 635: Yay!

source/solvers/navier_stokes_base.cc
```diff
@@ -876,36 +877,15 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
...
 
-  // The average_solution and reynolds_stresses vectors are reinitialized at the
-  // first iteration.
-  // Since the reynolds_stress (rs) vector doesn't have the same number of
-  // elements than the number of dofs, it requires its own dof_handler object
-  // even if data is not dofs. It allows to know locally owned elements and
-  // to output data.
   // The average velocities and reynolds stresses are calculated when the
   // time reaches the initial time. (time >= initial time) with 1e-6 as
   // tolerance.
```
blaisb, line 882: Maybe the tolerance should be 1e-6 * dt. This way it would be made dimensionless in a way.

source/solvers/navier_stokes_base.cc
```diff
@@ -876,36 +877,15 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess(bool firstIter)
...
-            this->locally_relevant_dofs,
-            this->fe.n_dofs_per_vertex(),
-            this->mpi_communicator);
-        }
       // Calculate average velocities when the time reaches the initial time.
       // time >= initial time with the epsilon as tolerance.
-      else if (simulation_control->get_current_time() >
-               (nsparam.post_processing.initial_time - 1e-6))
+      if (simulation_control->get_current_time() >
+          (nsparam.post_processing.initial_time - 1e-6))
```
blaisb, line 888:
```suggestion
          (nsparam.post_processing.initial_time - 1e-6 * dt))
```

blaisb: Not sure if dt is already defined there

source/solvers/postprocessing_velocities.cc
```diff
@@ -77,15 +74,14 @@ AverageVelocities<dim, VectorType, DofsType>::calculate_reynolds_stresses(
...
-      k_dt              = &reynolds_normal_stress_dt.block(1);
-      k_index           = [](unsigned int i) { return i / dim; };
+      begin_index    = local_evaluation_point.block(0).local_range().first;
+      end_index      = local_evaluation_point.block(0).local_range().second;
+      local_solution = &local_evaluation_point.block(0);
+      local_average  = &average_velocities.block(0);
+      rns_dt         = &reynolds_normal_stress_dt.block(0);
+      rss_dt         = &reynolds_shear_stress_dt.block(0);
+      k_dt           = &reynolds_normal_stress_dt.block(1);
+      k_index        = [](unsigned int i) { return i / dim; };
```
blaisb, line 84: What is this initialization? Maybe add a comment to make it easier to understand :). I sense some fancy lambda shenanigans here.

blaisb (commented):
Some minor changes there and there, but very good.

include/core/grids.h
acdaigneault: My mood of the day  :p.
Honestly, I don't know. I think this is because I feel restart is a parameter as mesh_parameters and manifolds_parameters so I put it straight after them even if not the same object after all.

source/solvers/flow_control.cc
acdaigneault: Ok, I just copy/paste it from Simulation control, it should be modified there too then.

source/solvers/navier_stokes_base.cc
acdaigneault: Ok!

source/solvers/postprocessing_velocities.cc
acdaigneault: Ok! 
This is because for Trilinos vectors, (example with 3d, so (u,v,w,p)), k is going to be at the fourth position of the Reynolds normal stress vector (<u'u'>,<v'v'>,<v'v'>,k). Fourth position -> k_index = i + dim (i = first position of the uvwp package)
For Trinilos block vectors ([u,v,w][p]), k is going to be at the first position of the block 1. First position -> k_index = i / dim. (i = first position of the uvw package in block 0).
I use lambda because this is not just a coefficient that can help me to get the k position in its vector but 2 differents operator with dim and i. 
And I mean, why not use some C++ fancy stuff :p

applications_tests/gls_navier_stokes_3d/CMakeLists.txt
acdaigneault: You mean about the pvtu/vtu files? Since average velocities can not be verify with the output text file, they can be useful to observe the flow on Paraview.

source/solvers/flow_control.cc
acdaigneault: "Note that any open file is automatically closed when the ifstream object is destroyed." :D

PR #112, opened 2020-12-06
Refactor GLS assembly

blaisb:
-Refactor GLS assembly to make it faster by using local variables. Lead
to a increase of 15% in assembly performance.
-Fix some test output due to a slight weird change
- Added a new linear solver which is not necessarily better, but it has
some interesting properties

This PR should be merged after Audrey's


tonielgeitani (commented):
@blaisb Should I change the gls_vans.cc for mat to match that of the new gls_navier_stokes.cc?

blaisb:
> @blaisb Should I change the gls_vans.cc for mat to match that of the new gls_navier_stokes.cc?

Not necessary for now. we can keep that in mind when we need more performance, but it's really not that necessary for the time being.

source/solvers/gls_navier_stokes.cc
```diff
@@ -1545,6 +1550,62 @@ GLSNavierStokesSolver<dim>::solve_system_direct(const bool   initial_step,
...
+
+  SolverControl solver_control(this->nsparam.linear_solver.max_iterations,
+                               linear_solver_tolerance,
+                               true,
+                               true);
+
+  TrilinosWrappers::SolverTFQMR::AdditionalData solver_parameters(false);
+
+
+  TrilinosWrappers::SolverTFQMR solver(solver_control, solver_parameters);
```
acdaigneault, line 1585: As in line 1544 and in the deal.ii documentation, I think you don't need to initialize solver_parameters since the TrilinosWrappers::SolverTFQMR::AdditionalData argument is declared and = false in the function.

source/solvers/gls_navier_stokes.cc
```diff
@@ -1545,6 +1550,62 @@ GLSNavierStokesSolver<dim>::solve_system_direct(const bool   initial_step,
...
+                 *ilu_preconditioner);
+
+    if (this->nsparam.linear_solver.verbosity != Parameters::Verbosity::quiet)
+      {
+        this->pcout << "  -Iterative solver took : "
+                    << solver_control.last_step() << " steps " << std::endl;
+      }
+  }
+  constraints_used.distribute(completely_distributed_solution);
+  auto &newton_update = this->get_newton_update();
```
acdaigneault, line 1605: initialization :) ?

source/solvers/gls_navier_stokes.cc
```diff
@@ -509,43 +515,39 @@ GLSNavierStokesSolver<dim>::assembleGLS()
...
+                  bdf_coefs[0] * velocity + bdf_coefs[1] * p1_velocity +
+                  bdf_coefs[2] * p2_velocity + bdf_coefs[3] * p3_velocity;
 
 
               if (is_sdirk_step1(scheme))
-                strong_residual +=
-                  sdirk_coefs[0][0] * present_velocity_values[q] +
-                  sdirk_coefs[0][1] * p1_velocity_values[q];
+                strong_residual += sdirk_coefs[0][0] * velocity +
+                                   sdirk_coefs[0][1] * p1_velocity_values[q];
```
acdaigneault, line 536:
```suggestion
                                   sdirk_coefs[0][1] * p1_velocity;
```

acdaigneault (approved):
Good :)

PR #116, opened 2020-12-27
Fix minor bug in linear force

shahabgol:
A minor bug in calculation of linear particle-particle and particle-wall is fixed.


blaisb (approved):
Seems good to me. Once CI finish i'll merge :).

PR #117, opened 2020-12-28
Fix a bug in velocity verlet integration

shahabgol:
A bug in velocity verlet integration is fixed.


blaisb (approved):
Seems perfect to me. Will wait for CI then merge.

PR #119, opened 2020-12-30
Dem gear3 integration

shahabgol:
Gear3 integration method is added to DEM.
A unit test is added for testing Gear3 integration.
Minor changes.


applications_tests/dem_3d/sliding_restart_files/sliding_restart.simulationcontrol
```diff
@@ -4,5 +4,5 @@ dt_1 1e-05
 dt_2 1e-05
 dt_3 1e-05
 CFL  0
-Time 0.15
-Iter 15000
+Time 0.25
+Iter 25000
```
blaisb, line 8: Why make this test longer? Won't this just make it more unstable?

shahabgol: Yes, it could become more unstable. I will correct it.

include/dem/dem_properties.h
```diff
@@ -37,28 +37,31 @@ namespace DEM
...
+    omega_z          = 14,
+    mass             = 15,
+    mom_inertia      = 16,
+    M_x              = 17,
+    M_y              = 18,
+    M_z              = 19,
+    displacement     = 20,
+    acc_derivative_x = 21,
+    acc_derivative_y = 22,
+    acc_derivative_z = 23,
```
blaisb, line 63: This is used to store a finite difference of the acceleration I guess? It would be good to have some sort of comments or something here to explain what this is.

include/dem/gear3_integrator.h
```diff
@@ -0,0 +1,71 @@
...
+
+#ifndef gear3_integrator_h
+#  define gear3_integrator_h
+
+/**
+ * Implementation of a gear3 scheme for the integration
+ * of the particle motion. Note that reinitilization of force and torque is also
+ * integrated into integration class
+ *
+ * @note
```
blaisb, line 35: Can you write the formula of the integration method here and put a reference? It's not like explciit Euler which is trivial. We should also add the formula for the verlet integrator and explicite euler anyway.

blaisb: This is a third order method right?

tests/dem/integration_gear3.output
```diff
@@ -0,0 +1,2 @@
+
+DEAL::The new position of the particle in z direction after 1.00000e-05 seconds is: -9.81000e-10
```
blaisb, line 2: Identical to Verlet for this case, which is to be expected :)!

tests/dem/integration_gear3.cc
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * ---------------------------------------------------------------------
+
+ *
+ * Author: Shahab Golshan, Polytechnique Montreal, 2019-
+ */
+
+/**
+ * @brief This test checks the performance of the velocity verlet integrator
+ * class.
```
blaisb, line 22:
```suggestion
 * @brief This test checks that the Gear3 integrator can accurately integrate the trajectory of a particle with
 a constant force such that d^2 x / dt^2 = - g
 * class.
```

blaisb (commented):
Very good. few corrections there and there and missing the unit test that is non-trivial (harmonic oscillator). Then it will be very good!

blaisb (commented):
Just correct the test for integrator accuracy then you are good to go. :)

include/dem/explicit_euler_integrator.h
```diff
@@ -49,18 +49,29 @@ class ExplicitEulerIntegrator : public Integrator<dim>
...
-   * the acceleration with the explicit Euler method
+   * Carries out the prediction (pre_force) integration calculations.
+   *
+   * @param particle_handler The particle handler whose particle motion we wish
+   * to integrate
+   * @param time_step The value of the time step used for the integration
+   */
+  virtual void
+  integrate_pre_force(Particles::ParticleHandler<dim> &particle_handler,
+                      double                           time_step) override;
```
blaisb, line 60: Shouldnt the Body force be passed as an argument here also for generality? it's kinda weird to have different arguments for both. I know Verlet does not use it but still.

include/dem/integrator.h
```diff
@@ -48,20 +48,31 @@ class Integrator
...
+   * After calculation of new acting forces on particles, integrate_post_force
+   * updates the integrate_pre_force results.
+   *
+   * @param particle_handler The particle handler whose particle motion we wish
+   * to integrate
+   * @param time_step The value of the time step used for the integration
+   */
+  virtual void
+  integrate_pre_force(Particles::ParticleHandler<dim> &particle_handler,
+                      double                           time_step) = 0;
```
blaisb, line 61: Make interface identical for pre and post :)

tests/dem/integration_schemes_accuracy.output
```diff
@@ -1,10 +1,4 @@
...
+DEAL::Explicit Euler is a 1 order integration scheme
+DEAL::Velocity Verlet is a 2 order integration scheme
+DEAL::Gear3 is a 5 order integration scheme
```
blaisb, line 4: So much better :)

blaisb: It would be better to leave the actual double values instead of rounding. Or maybe give both values (the ratio and the rounded order). This system is deterministic so the result won't be compiler dependent, it's not like DEM.

tests/dem/post_collision_velocity.output
```diff
@@ -1,2 +1,2 @@
 
-DEAL::Coefficient of restitution is 0.9 and the velocity of particle before collision is 0.1, the velocity of particle after collision is: 0.0899723
+DEAL::Coefficient of restitution is 0.9 and the velocity of particle before collision is 0.1, the velocity of particle after collision is: 0.0882268
```
blaisb, line 2: Energy conservation is now lower? Did you change the time step?

blaisb (commented):
Some very minor changes to do, but this is close to being merged :)

tests/dem/post_collision_velocity.output
shahabgol: no, I changed nothing. It was strange for me too.

blaisb (approved):
To merge once CI passes :)

PR #124, opened 2021-01-20
Modify assemble_matrix_and_rhs of heat_transfer

jeannejoachim:
Local pre-loading of gradients and values for faster computation. Validated with ctests

Fix issue with gls_nitsche

Temporal integration with heat transfer is now working properly.


blaisb (approved):
Good work. Much easier to read and understand.

PR #126, opened 2021-01-22
Add stabilization to thermal solver + test

blaisb:
Add stabilization for High Peclet number in the heat transfer equation.
Adds a high-pe test demonstrating this capacity.

### DIFF
--- a/source/solvers/heat_transfer.cc
+++ b/source/solvers/heat_transfer.cc
@@ -110,7 +110,7 @@ HeatTransfer<dim>::assemble_system(
                              quadrature_formula,
                              update_values | update_gradients |
                                update_quadrature_points | update_JxW_values |
-                               update_hessians | update_JxW_values);
+                               update_hessians);
 
   auto &evaluation_point = this->get_evaluation_point();
 
@@ -290,10 +290,8 @@ HeatTransfer<dim>::assemble_system(
 
 
                           // Weak form for : - k * laplacian T + rho * cp *
-                          // u
-                          // *
-                          //                      gradT - f -
-                          //                      grad(u)*grad(u) =0
+                          //                  u * gradT - f -
+                          //                  mu*tau:grad(u) =0
                           cell_matrix(i, j) +=
                             (thermal_conductivity * grad_phi_T_i *
                                grad_phi_T_j +
--- a/source/solvers/heat_transfer.cc
+++ b/source/solvers/heat_transfer.cc
@@ -56,6 +56,8 @@ HeatTransfer<dim>::assemble_system(
 
   const double rho_cp = density * specific_heat;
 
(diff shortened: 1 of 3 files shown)


source/solvers/heat_transfer.cc
```diff
@@ -102,7 +109,8 @@ HeatTransfer<dim>::assemble_system(
   FEValues<dim>     fe_values_ht(fe,
                              quadrature_formula,
                              update_values | update_gradients |
-                               update_quadrature_points | update_JxW_values);
+                               update_quadrature_points | update_JxW_values |
+                               update_hessians | update_JxW_values);
```
jeannejoachim, line 113: why is `update_JxW_values` repeated?

source/solvers/heat_transfer.cc
```diff
@@ -217,38 +241,82 @@ HeatTransfer<dim>::assemble_system(
...
+              // Calculation of the magnitude of the velocity for the
+              // stabilization parameter
+              const double u_mag = std::max(velocity.norm(), 1e-12);
+
+              // Calculation of the GLS stabilization parameter. The
+              // stabilization parameter used is different if the simulation is
+              // steady or unsteady. In the unsteady case it includes the value
+              // of the time-step
+              const double tau =
+                is_steady(time_stepping_method) ?
```
jeannejoachim, line 257: I didn't know this formulation! Is it faster than a simple if/else loop?

source/solvers/heat_transfer.cc
```diff
@@ -217,38 +241,82 @@ HeatTransfer<dim>::assemble_system(
...
+
+
 
-                          // Weak form for : - k * laplacian T + rho * cp * u *
-                          //                      gradT - f - grad(u)*grad(u) =0
+                          // Weak form for : - k * laplacian T + rho * cp *
+                          // u
+                          // *
+                          //                      gradT - f -
+                          //                      grad(u)*grad(u) =0
```
jeannejoachim, line 296: why is this equation on 5 lines rather than 2 as before?

jeannejoachim (approved):
Thank you for adding this feature so quickly! I am looking forward to see how it will impact our simulations :)

source/solvers/heat_transfer.cc
blaisb: That's a mistake, i'll fix it.

source/solvers/heat_transfer.cc
blaisb: Error due to automatic indentation. i'll fix that.

source/solvers/heat_transfer.cc
blaisb: It more importantly allow you to declare a variable whose content depend on an if. This way it can be const.

PR #128, opened 2021-01-29
Dem load balancing

shahabgol:
Load balancing is modified to cover single-step, frequent, and dynamic methods
An application test is added for dynamic load balancing
Load balance now reports the minimum and maximum number of particles and grids on processors


source/dem/dem.cc
```diff
@@ -322,27 +351,107 @@ DEMSolver<dim>::setup_background_dofs()
...
+  const unsigned int maximum_particle_number_on_proc =
+    Utilities::MPI::max(particle_handler.n_locally_owned_particles(),
+                        mpi_communicator);
+  const unsigned int minimum_particle_number_on_proc =
+    Utilities::MPI::min(particle_handler.n_locally_owned_particles(),
+                        mpi_communicator);
+  const unsigned int maximum_cell_number_on_proc =
+    Utilities::MPI::max(triangulation.n_active_cells(), mpi_communicator);
+  const unsigned int minimum_cell_number_on_proc =
+    Utilities::MPI::min(triangulation.n_active_cells(), mpi_communicator);
```
blaisb, line 378: Instead of reporting min and max, I would report Min - Max and Avg.
You can use the following function :+1: 
https://www.dealii.org/current/doxygen/deal.II/structUtilities_1_1MPI_1_1MinMaxAvg.html
This way in a single MPI call you get all the statistics and you can report them.

blaisb (commented):
Just change the Min/Max to Min/Max/Avg and I think this is good to go! Good job. CLearly this will allow us to work more intelligently on load balancing.

source/dem/dem.cc
shahabgol: Done!

blaisb (approved):

PR #129, opened 2021-01-31
Nitsche improvements

blaisb:
- Check that dynamic mesh adaptation works with Nitsche IB
- Adapt weak formulation to ensure asymptotic convergence at order 1
- Add a check for particle number and a warning if that is broken
- Add capacity to "sub time-step" particles in order to make sure that they are never lost.


source/core/solid_base.cc
```diff
@@ -144,17 +148,46 @@ SolidBase<dim, spacedim>::setup_particles()
...
+  //    if (cell->is_locally_owned())
+  //      all_boxes.emplace_back(cell->bounding_box());
+  //  const auto tree        = pack_rtree(all_boxes);
+  //  const auto local_boxes = extract_rtree_level(tree, 1);
+
+  //   global_fluid_bounding_boxes =
+  //    Utilities::MPI::all_gather(mpi_communicator, local_boxes);
+
+  // Naive global bounding boxes
+  // Obtaining global bounding boxes
```
jeannejoachim, line 164: remove dead bodies?

source/core/solid_base.cc
```diff
@@ -144,17 +148,46 @@ SolidBase<dim, spacedim>::setup_particles()
...
 
   std::vector<std::vector<BoundingBox<spacedim>>> global_fluid_bounding_boxes;
-  global_fluid_bounding_boxes =
-    Utilities::MPI::all_gather(mpi_communicator, local_boxes);
+
+  if (auto tria =
```
jeannejoachim, line 168: Could this if/else loop be documented? I think that this "if" condition checks for parallel computing, but as I understand it the "if" and the "else" block both use locally owned cells

source/core/solid_base.cc
```diff
@@ -198,35 +232,59 @@ template <int dim, int spacedim>
 void
 SolidBase<dim, spacedim>::integrate_velocity(double time_step)
 {
-  for (auto particle = solid_particle_handler->begin();
-       particle != solid_particle_handler->end();
-       ++particle)
+  const unsigned int sub_particles_iterations = param->particles_sub_iterations;
```
jeannejoachim, line 235: maybe document why we need these particles sub iterations?

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -65,9 +65,9 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::assemble_nitsche_restriction()
   Function<spacedim> *solid_velocity = solid.get_solid_velocity();
 
   // Penalization terms
-  const auto penalty_parameter =
-    1.0 / GridTools::minimal_cell_diameter(*this->triangulation);
-  double beta = this->simulation_parameters.nitsche->beta;
+  // const double h = GridTools::minimal_cell_diameter(*this->triangulation);
+  // const auto   global_penalty_parameter = 1.0 / (h);
```
jeannejoachim, line 69: remove dead bodies?

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -333,7 +353,6 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
         }
 
       this->postprocess(false);
-      // postprocess_ht();
```
jeannejoachim, line 336: remove if not needed

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -349,18 +368,10 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
         }
 
       this->finish_time_step();
-      if (this->simulation_parameters.multiphysics.heat_transfer)
```
jeannejoachim, line 352: remove if statement if not needed (because it is now managed by multiphysics?)

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -349,18 +368,10 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
...
-        {
-          //          finish_time_step_ht();
-        }
     }
   if (this->simulation_parameters.test.enabled)
     solid.print_particle_positions();
   this->finish_simulation();
-  if (this->simulation_parameters.multiphysics.heat_transfer)
```
jeannejoachim, line 360: remove statement

jeannejoachim (commented):
Very cool additions to the Nitsche solver, thank you! I am looking forward to use it on our industrial case :)

source/solvers/gls_nitsche_navier_stokes.cc
blaisb: When a statement is in red it means this is a line that has been removed. I cannot remove what has already been removed ^^!

PR #132, opened 2021-02-04
Add GGLS stabilization

blaisb:
Add GGLS stabilization for transient cases that are very stiff.
This stabilizes the time derivative of the gradient in a very efficient manner.
The test is a bit dumb, because it does not really produce an output to compare, but  I am really unable to find a test that has an analytical solution at the moment. Still, it's not bad because someone can just open the output of the test and have a ready idea if things are working ok. Not ideal, but not "the worst".


jeannejoachim (approved):
Very useful stabilization and clear implementation! I tested it on our industrial case : it reduces substantially temperature oscillations and gives a much better estimation of min/max temperatures. Thank you!

PR #134, opened 2021-02-14
Dem parallel optimization

shahabgol:
Decrease the number of particles' properties to improve the performance of the code in parallel mode.


include/dem/dem.h
```diff
@@ -364,6 +375,12 @@ class DEMSolver
   PVDHandler                           particles_pvdhandler;
   const unsigned int                   standard_deviation_multiplier;
 
+  std::unordered_map<int, Tensor<1, dim>> momentum;
+  std::unordered_map<int, Tensor<1, dim>> force;
+  std::unordered_map<int, double>         displacement;
+  std::unordered_map<int, double>         MOI;
+  std::unordered_map<int, Tensor<1, dim>> acceleration;
```
blaisb, line 382: Make the key of these map to be ParticleIndex:
https://www.dealii.org/current/doxygen/deal.II/namespacetypes.html#ae4ea18e3efb31f0312a9e754873d71b8
This way, it will be more coherent. Additionally, particle indices are unsigned int and not int, this should prevent some weird casting that might occur and make access potentially slightly faster.

include/dem/dem_properties.h
```diff
@@ -37,31 +37,16 @@ namespace DEM
...
+    type         = 0,
+    dp           = 1,
+    v_x          = 2,
+    v_y          = 3,
+    v_z          = 4,
+    omega_x      = 5,
+    omega_y      = 6,
+    omega_z      = 7,
+    mass         = 8,
+    n_properties = 9,
```
blaisb, line 49: So little properties. Miam

include/dem/explicit_euler_integrator.h
```diff
@@ -69,11 +69,20 @@ class ExplicitEulerIntegrator : public Integrator<dim>
...
-                       Tensor<1, dim>                   body_force,
-                       double                           time_step) override;
+  integrate_post_force(
+    Particles::ParticleHandler<dim> &        particle_handler,
+    Tensor<1, dim>                           body_force,
+    double                                   time_step,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
+    std::unordered_map<int, Tensor<1, dim>> &force,
+    std::unordered_map<int, double> &        MOI,
+    std::unordered_map<int, Tensor<1, dim>> &acceleration) override;
```
blaisb, line 85: I can understand why you store the force in a structure. But why acceleration? I mean, force = acceleration * mass. You can just store one of the two no?

include/dem/pp_contact_force.h
```diff
@@ -64,8 +66,10 @@ class PPContactForce
       &local_adjacent_particles,
     std::unordered_map<int,
```
blaisb, line 67: Again the key of that map should be a particle index.
Everywhere the key should be particle index

include/dem/pp_linear_force.h
```diff
@@ -68,8 +70,10 @@ class PPLinearForce : public PPContactForce<dim>
       &local_adjacent_particles,
     std::unordered_map<int,
                        std::unordered_map<int, pp_contact_info_struct<dim>>>
-      &           ghost_adjacent_particles,
-    const double &dt) override;
+      &                                      ghost_adjacent_particles,
+    const double &                           dt,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
```
blaisb, line 75: momentum is u * mass? Not sure it is worth it to store it...

include/dem/pp_nonlinear_force.h
```diff
@@ -68,8 +70,10 @@ class PPNonLinearForce : public PPContactForce<dim>
       &adjacent_particles,
     std::unordered_map<int,
                        std::unordered_map<int, pp_contact_info_struct<dim>>>
-      &           ghost_adjacent_particles,
-    const double &dt) override;
+      &                                      ghost_adjacent_particles,
+    const double &                           dt,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
+    std::unordered_map<int, Tensor<1, dim>> &force) override;
```
blaisb, line 76: Again, key of the map is wrong type.

include/dem/pw_linear_force.h
```diff
@@ -63,12 +63,16 @@ class PWLinearForce : public PWContactForce<dim>
...
    */
   virtual void
   calculate_pw_contact_force(
     std::unordered_map<int, std::map<int, pw_contact_info_struct<dim>>>
-      &           pw_pairs_in_contact,
-    const double &dt) override;
+      &                                      pw_pairs_in_contact,
+    const double &                           dt,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
+    std::unordered_map<int, Tensor<1, dim>> &force) override;
```
blaisb, line 75: Correct map keys

include/dem/pw_nonlinear_force.h
```diff
@@ -63,12 +63,16 @@ class PWNonLinearForce : public PWContactForce<dim>
...
    */
   virtual void
   calculate_pw_contact_force(
     std::unordered_map<int, std::map<int, pw_contact_info_struct<dim>>>
-      &           pw_pairs_in_contact,
-    const double &dt) override;
+      &                                      pw_pairs_in_contact,
+    const double &                           dt,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
+    std::unordered_map<int, Tensor<1, dim>> &force) override;
```
blaisb, line 75: Map keys

include/dem/velocity_verlet_integrator.h
```diff
@@ -74,11 +74,20 @@ class VelocityVerletIntegrator : public Integrator<dim>
...
-                       Tensor<1, dim>                   body_force,
-                       double                           time_step) override;
+  integrate_post_force(
+    Particles::ParticleHandler<dim> &        particle_handler,
+    Tensor<1, dim>                           body_force,
+    double                                   time_step,
+    std::unordered_map<int, Tensor<1, dim>> &momentum,
+    std::unordered_map<int, Tensor<1, dim>> &force,
+    std::unordered_map<int, double> &        MOI,
+    std::unordered_map<int, Tensor<1, dim>> &acceleration) override;
```
blaisb, line 90: Correct map keys

include/dem/visualization.h
```diff
@@ -133,7 +129,7 @@ class Visualization : public dealii::DataOutInterface<0, dim>
   /**
    * Number of properties that are written in output files
    */
-  const unsigned int number_of_properties_to_write = 16;
+  const unsigned int number_of_properties_to_write = 9;
```
blaisb, line 132: Why is this hardcoded? Couldnt we just assume that we write all properties using the dem_parameters enum?

source/dem/dem.cc
```diff
@@ -485,6 +491,25 @@ DEMSolver<dim>::insert_particles()
...
+  // Clearing the container first
+  MOI.clear();
+
+  for (auto &particle : particle_handler)
+    {
+      auto &particle_properties = particle.get_properties();
+      MOI.insert({particle.get_id(),
+                  0.4 * particle_properties[DEM::PropertiesIndex::mass] *
+                    (particle_properties[DEM::PropertiesIndex::dp] * 0.5) *
+                    (particle_properties[DEM::PropertiesIndex::dp] * 0.5)});
```
blaisb, line 509: >
```suggestion
                  0.1 * particle_properties[DEM::PropertiesIndex::mass] *
                    (particle_properties[DEM::PropertiesIndex::dp]) *
                    (particle_properties[DEM::PropertiesIndex::dp])});
```

source/dem/find_contact_detection_step.cc
```diff
@@ -3,33 +3,33 @@
...
-      // displacement is calculated
+      // If last step was a sorting into subdomains step, the displcement
+      // is reinitialized and then the new displacement is calculated
 
-      // Reinitilizing contact_detection_step
-      contact_detection_step = 0;
+      // Clearing displacement
+      displacement.clear();
 
       for (auto &particle : particle_handler)
```
blaisb, line 26: Why is the loop duplicated in the if? I don't see the difference here.

source/dem/pp_contact_force.cc
```diff
@@ -128,30 +128,30 @@ PPContactForce<dim>::update_contact_information(
 template <int dim>
 void
 PPContactForce<dim>::apply_force_and_torque_real(
```
blaisb, line 130: Have you considered moving this function to the .h and inlining it? It might be pretty efficient.

source/dem/velocity_verlet_integrator.cc
```diff
@@ -36,39 +32,53 @@ VelocityVerletIntegrator<dim>::integrate_pre_force(
...
 
       for (int d = 0; d < dim; ++d)
         {
           // Calculate the acceleration
-          particle_properties[PropertiesIndex::acc_x + d] =
-            g[d] + particle_properties[PropertiesIndex::force_x + d] /
-                     particle_properties[PropertiesIndex::mass];
+          particle_acceleration[d] =
+            g[d] +
+            particle_force[d] / particle_properties[PropertiesIndex::mass];
```
blaisb, line 61: Quick question. How often do we divide by mass compared to multiply by it? it might be a good idea to store 1/mass instead if over the entire code we only divide by mass. What's your take on this? This is mini, but imagine if we do it 1M * 1000k whatever...

tests/dem/integration_euler.cc
```diff
@@ -92,21 +95,22 @@ test()
...
   pit->get_properties()[DEM::PropertiesIndex::omega_y] = 0;
   pit->get_properties()[DEM::PropertiesIndex::omega_z] = 0;
   // mass and moment of inertia
-  pit->get_properties()[DEM::PropertiesIndex::mass]        = 1;
-  pit->get_properties()[DEM::PropertiesIndex::mom_inertia] = 1;
+  pit->get_properties()[DEM::PropertiesIndex::mass] = 1;
+
+  std::unordered_map<int, Tensor<1, dim>> momentum;
+  std::unordered_map<int, Tensor<1, dim>> force;
+  std::unordered_map<int, double>         MOI;
```
blaisb, line 107: Don't forget about the keys

blaisb (commented):
Some minor comments, but good ideas for some minor optimizations

include/dem/dem.h
shahabgol: changed to ParticleIndex

include/dem/explicit_euler_integrator.h
shahabgol: We need the previous step's acceleration in the Velocity Verlet integration. I changed it from a global container to a local container.

include/dem/pp_linear_force.h
shahabgol: Similar to force, we calculate momentum in the particle-particle and particle-wall contact force calculations. Then in the integration, we integrate it to calculate omega. So we need this momentum.

source/dem/velocity_verlet_integrator.cc
shahabgol: very good idea, applied.

source/dem/pp_contact_force.cc
shahabgol: done.

applications_tests/dem_2d/circle_with_floating_walls.mpirun=2.output
```diff
@@ -33,23 +33,22 @@ id, type, dp , v_x  , v_y  , v_z  , omega_x, omega_y, omega_z,
...
-37 0 0.00500 0.016 0.000 0.000 0.0 0.0 0.0 
-38 0 0.00500 0.017 0.000 0.000 0.0 0.0 0.0 
-39 0 0.00500 0.019 0.000 0.000 0.0 0.0 0.0 
-40 0 0.00500 0.016 0.000 0.000 0.0 0.0 0.0 
-41 0 0.00500 0.018 0.000 0.000 0.0 0.0 0.0 
+37 0 0.00500 0.000 0.000 0.000 0.0 0.0 0.0 
+38 0 0.00500 0.000 0.000 0.000 0.0 0.0 0.0 
+39 0 0.00500 -0.001 0.000 0.000 0.0 0.0 0.0 
+40 0 0.00500 0.000 0.000 0.000 0.0 0.0 0.0 
+41 0 0.00500 0.000 0.000 0.000 0.0 0.0 0.0
```
blaisb: Whydid these change so much?

blaisb (commented):
Good to go
If I understand correctly, right now in the post-processing if you use Verlet integrator, the velocity and the position won't be at the same time?
I think something is off with your first time step. Have you verified with the analytical solution for a falling particle? for the first iteration you should call half step before contacts than call regular integration. Otherwise right now the first step you only do half step for position but nothing? This is confusing

source/dem/velocity_verlet_integrator.cc
```diff
@@ -30,10 +31,14 @@ VelocityVerletIntegrator<dim>::integrate_half_step_location(
...
+          half_step_velocity[d] =
+            particle_properties[PropertiesIndex::v_x + d] +
+            0.5 * particle_acceleration[d] * dt;
+
           // Update particle position using half-step velocity
-          particle_position[d] +=
-            particle_properties[PropertiesIndex::v_x + d] * dt +
-            0.5 * dt * dt * particle_acceleration[d];
+          particle_position[d] += half_step_velocity[d] * dt +
+                                  0.5 * dt * dt * particle_acceleration[d];
```
blaisb, line 41:
```suggestion
          particle_position[d] += half_step_velocity[d] * dt;
```

blaisb: This is the real formula for velocity verlet.
You also need to update the velocity itself. The half-step velocity has to be stored.

source/dem/velocity_verlet_integrator.cc
```diff
@@ -37,8 +37,7 @@ VelocityVerletIntegrator<dim>::integrate_half_step_location(
             0.5 * particle_acceleration[d] * dt;
```
blaisb, line 37: You need to store that velocity and modify the particle velocity with it.

PR #135, opened 2021-02-17
Force + Torque on particles calculation

blaisb:
Add capacity to calculate the force and the torque on the Nitsche particles.
Tested in a taylor-couette case
Torque converges perfectly fine

image


jeannejoachim (approved):
Thank you for this addition!

PR #136, opened 2021-02-23
Nitsche multiple solids

blaisb:
Enable multiple solids in Nitsche simulation.
Refactors the construction of the Nitsche solid by seperating Solids parameters from the Nitsche parameter itself.


include/solvers/nitsche.h
```diff
@@ -34,53 +34,40 @@ using namespace dealii;
...
-    std::string force_output_name;
-    std::string torque_output_name;
-
-
-
     // Particle motion integration parameters
     unsigned int particles_sub_iterations;
+
+    // information for force calculation
+    Point<dim> cor; // Center of rotation used for torque calculation
```
jeannejoachim, line 60: I find this variable name misleading, as `cor` can also be used for a correction factor (at least this is what came first in mind for be). Maybe use "center" instead, so that it is more transparent for the rest of the code? (I know this variable has not been added in this PR)

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -610,69 +657,89 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
...
-          std::shared_ptr<Particles::ParticleHandler<spacedim>> solid_ph =
-            solid.get_solid_particle_handler();
-          output_solid_particles(solid_ph);
-          output_solid_triangulation();
+          for (unsigned int i_solid = 0; i_solid < solid.size(); ++i_solid)
+            {
+              std::shared_ptr<Particles::ParticleHandler<spacedim>> solid_ph =
+                solid[i_solid]->get_solid_particle_handler();
+              output_solid_particles();
+              output_solid_triangulation();
```
jeannejoachim, line 665: `output_solid_particles` and `output_solid_triangulation` methods both also have a loop over the `i_solid`, so I do not understand the loop here. I think either this loop on `solve` method is unnecessary, or the `i_solid` argument should be passed to the output methods and their loop other `i_solid` should be removed. I would vote for keeping the loop on solve() and passing `i_solid` to the output methods, as they are named `output_solid_xxx` and not `output_solids_xxx` + it would be coherent with the `print_particle_position` method called after for `solid[i_solid]`.

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -610,69 +657,89 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
...
-
-  write_vtu_and_pvd<0, spacedim>(pvdhandler_solid_particles,
-                                 *(&particles_out),
-                                 folder,
-                                 solution_name,
-                                 time,
-                                 iter,
-                                 group_files,
-                                 this->mpi_communicator);
+  for (unsigned int i_solid = 0; i_solid < solid.size(); ++i_solid)
```
jeannejoachim, line 685: remove redundant loop here and pass `i_solid` to the method

source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -610,69 +657,89 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::solve()
...
-
-  write_vtu_and_pvd<dim>(pvdhandler_solid_triangulation,
-                         data_out,
-                         folder,
-                         solution_name,
-                         time,
-                         iter,
-                         group_files,
-                         this->mpi_communicator);
+  for (unsigned int i_solid = 0; i_solid < solid.size(); ++i_solid)
```
jeannejoachim, line 716: remove redundant loop here and pass `i_solid` to the method

jeannejoachim (approved):
The addition of multiple Nitsche solids is a real asset! Also externalizing the Solid part in Nitsche class really makes this clearer. Finally the speed gain from modifying the bounding box is really impressive. I think there is just a loop in `output_solid_xxx` methods that is redundant with the loop in the `solve` method of Nitsche Solver, so a small modification is needed here before merging. Thank you!

include/solvers/nitsche.h
blaisb: I'll rename it center_of_rotation it will be clearer

source/solvers/gls_nitsche_navier_stokes.cc
blaisb: I'll keep the loop outside and pass the i_solid to the output methods ^^! Good catch

PR #138, opened 2021-02-26
Refactoring dem

shahabgol:
Move some functions into separate files to decrease the length of dem class


include/dem/checkpointing.h
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Carries out reading and writing dem checkpoints
+ *
+ * @note
+ *
+ * @author Shahab Golshan, Polytechnique Montreal 2019-
+ */
+
+template <int dim>
+class Checkpointing
```
blaisb, line 48: Clean stuff. Maybe class should be called DEMCheckpointing?

include/dem/print_initial_information.h
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Author: Shahab Golshan, Polytechnique Montreal, 2019
+ */
+
+#include <dem/dem_solver_parameters.h>
+
+using namespace std;
+
+#ifndef print_initial_information_h
+#  define print_initial_information_h
```
blaisb, line 25:
```suggestion
#define print_initial_information_h
```

include/dem/checkpointing.h
blaisb: Actually, why does this need to be a class? Could it not be just two functions? Class logic does not appear to be used here.

source/dem/read_mesh.cc
```diff
@@ -0,0 +1,66 @@
+#include <dem/read_mesh.h>
+
+template <int dim>
+void
+read_mesh(const DEMSolverParameters<dim> &           parameters,
+          const ConditionalOStream &                 pcout,
+          parallel::distributed::Triangulation<dim> &triangulation,
+          double &triangulation_cell_diameter)
```
blaisb, line 8: How is this function different than the read mesh function that is used in the core lethe library?

blaisb (commented):
Some minor comments:
- Checkpointing in it's present state does not need to be a class
- Can you see what is the difference between your read dem mesh and the one in the core library? If they are similar or if the core can be adaptated instead, we should use the one in the core. If there is a difference that make it not possible, then we can keep two but this will be a bit confusing...

include/dem/checkpointing.h
shahabgol: done

blaisb (approved):

PR #142, opened 2021-03-04
Add checkpointing for Nitsche solid particles

jeannejoachim:
Add checkpointing (write/read) and refactor solid setup

Implementation generalized for multiple Nitsche solids.
Checked with two solids and heat transfer resolution on.
Fix small issue with mesh adaptation on Nitsche solid.
Add test for Nitsche restart 22, test with multiple solids (2 bars) and heat transfer solved. Comparison on the particle location, as for impeller test.

For now checkpointing only apply to the solid particles (used in calculation). Checkpointing for solid triangulation (used in visualizations) is not possible for now, an issue will be opened on this matter.


include/core/solid_base.h
```diff
@@ -140,8 +172,6 @@ class SolidBase
 
   const unsigned int degree_velocity;
   unsigned int       initial_number_of_particles;
-
-  bool setup_done = false;
```
blaisb, line 144: Good riddance :)!

blaisb (approved):
Good to go! No comments! Good idea to add two tests.

PR #151, opened 2021-04-06
Correct bug in parallel checkpoints

shahabgol:
Correct a bug in parallel checkpoints

### DIFF
--- a/include/dem/dem.h
+++ b/include/dem/dem.h
@@ -278,7 +278,7 @@ private:
   Particles::ParticleHandler<dim, dim> particle_handler;
   unsigned int                         contact_detection_step;
   unsigned int                         load_balance_step;
-  unsigned int                         checkpoint_step;
+  bool                                 checkpoint_step;
   Tensor<1, dim>                       g;
   double                               triangulation_cell_diameter;
 
--- a/include/dem/dem.h
+++ b/include/dem/dem.h
@@ -278,6 +278,7 @@ private:
   Particles::ParticleHandler<dim, dim> particle_handler;
   unsigned int                         contact_detection_step;
   unsigned int                         load_balance_step;
+  unsigned int                         checkpoint_step;
   Tensor<1, dim>                       g;
   double                               triangulation_cell_diameter;
 
--- a/include/dem/dem.h
+++ b/include/dem/dem.h
@@ -276,8 +276,8 @@ private:
   TimerOutput                          computing_timer;
   double                               smallest_contact_search_criterion;
   Particles::ParticleHandler<dim, dim> particle_handler;
-  unsigned int                         contact_detection_step;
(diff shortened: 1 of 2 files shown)


include/dem/dem.h
```diff
@@ -278,6 +278,7 @@ class DEMSolver
   Particles::ParticleHandler<dim, dim> particle_handler;
   unsigned int                         contact_detection_step;
   unsigned int                         load_balance_step;
+  unsigned int                         checkpoint_step;
```
blaisb, line 281: Why are those variables not bool instead of ints?

source/dem/dem.cc
```diff
@@ -679,6 +680,8 @@ DEMSolver<dim>::solve()
                       particle_handler);
 
       update_moment_of_inertia(particle_handler, MOI);
+
+      checkpoint_step = 1;
```
blaisb, line 684: Why not a boolean? then set to true

source/dem/dem.cc
```diff
@@ -722,8 +725,11 @@ DEMSolver<dim>::solve()
 
       // Sort particles in cells
       if (particles_insertion_step || load_balance_step ||
-          contact_detection_step)
+          contact_detection_step || checkpoint_step)
         {
+          // Reset checkpoint step
+          checkpoint_step = 0;
+
```
blaisb, line 732: Should be boolean, otherwise good

blaisb (commented):
Can you switch type from unsigned int to bool?

include/dem/dem.h
shahabgol: Changed to bool.

source/dem/dem.cc
shahabgol: Done.

source/dem/dem.cc
shahabgol: Done.

shahabgol:
> Can you switch type from unsigned int to bool?

Changed to bool.

include/dem/dem.h
```diff
@@ -278,6 +278,7 @@ class DEMSolver
   Particles::ParticleHandler<dim, dim> particle_handler;
   unsigned int                         contact_detection_step;
   unsigned int                         load_balance_step;
```
blaisb, line 280: Can you change these ones also while you are it :)

blaisb (commented):
Can you change the leftover ints while you are at it?
I'll merge afterwards.

include/dem/dem.h
shahabgol: done.

shahabgol:
> Can you change the leftover ints while you are at it?
> I'll merge afterwards.

load_balance_step and contact_detection_step types are changed to bool.

blaisb (approved):

PR #154, opened 2021-04-22
Implement basic multiphasic simulations (VOF)

jeannejoachim:
Free_surface new auxiliary physic with Volume Of Fluid method
GLSNavierStokes FreeSurface assemble system, with bdf1 scheme
Definition of multiple fluids in parameter file
ctest on hydrostatic case, with error calculated for velocity, pressure and phase

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -124,6 +124,9 @@ namespace Parameters
     // Enable output of the boundaries
     bool output_boundaries;
 
+    // Enable automatic folder creation if does not exist
+    bool make_folder;
+
     // Frequency of the log output to the terminal
     unsigned int log_frequency;
 
@@ -143,14 +146,23 @@ namespace Parameters
   };
 
   /**
-   * @brief PhysicalProperties - Define the possible physical properties.
-   * All continuum equations share the same physical properties object but only
-   * take the subset of properties they require
+   * @brief Fluid - Class for fluid definition
    */
-  struct PhysicalProperties
+  class Fluid
   {
+  public:
+    Fluid()
+    {}
+
+    void
+    declare_parameters(ParameterHandler &prm, unsigned int id);
+    void
+    parse_parameters(ParameterHandler &prm, unsigned int id);
+
     // Kinematic viscosity (nu = mu/rho) in units of L^2/s
     double viscosity;
+    // Dynamic viscosity (mu = nu*rho) in units of Pa s
+    double dynamic_viscosity;
     // volumetric mass density (rho) in units of kg/m^3
     double density;
     // specific heat capacity (cp) in J/K/kg
@@ -159,8 +171,39 @@ namespace Parameters
     double thermal_conductivity;
     // tracer diffusivity) in L^2/s
     double tracer_diffusivity;
+  };
 
-    static void
+  /**
+   * @brief PhysicalProperties - Define the possible physical properties.
+   * All continuum equations share the same physical properties object but only
+   * take the subset of properties they require
+   * Defined as a class with public attributes in order to use a non-static
+   * declare_paremeters methods (useful for multiple fluid simulations).
+   */
+  class PhysicalProperties
+  {
+  public:
+    PhysicalProperties()
+    {}
+
+    // Monophasic simulations parameters
+    // Kinematic viscosity (nu = mu/rho) in units of L^2/s
+    double viscosity;
+    // volumetric mass density (rho) in units of kg/m^3
+    double density;
+    // specific heat capacity (cp) in J/K/kg
+    double specific_heat;
+    // thermal conductivity (k) in W/m/K
+    double thermal_conductivity;
+    // tracer diffusivity in L^2/s
+    double tracer_diffusivity;
+
+    // Fluid objects for multiphasic simulations
+    std::vector<Fluid>        fluids;
+    unsigned int              number_fluids;
+    static const unsigned int max_fluids = 2;
+
+    void
     declare_parameters(ParameterHandler &prm);
     void
     parse_parameters(ParameterHandler &prm);
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -158,8 +158,6 @@ namespace Parameters
 
     // Kinematic viscosity (nu = mu/rho) in units of L^2/s
     double viscosity;
-    // Dynamic viscosity (mu = nu*rho) in units of Pa s
-    double dynamic_viscosity;
     // volumetric mass density (rho) in units of kg/m^3
     double density;
     // specific heat capacity (cp) in J/K/kg
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -124,9 +124,6 @@ namespace Parameters
     // Enable output of the boundaries
     bool output_boundaries;
 
-    // Enable automatic folder creation if does not exist
-    bool make_folder;
-
     // Frequency of the log output to the terminal
     unsigned int log_frequency;
 
--- a/include/core/utilities.h
+++ b/include/core/utilities.h
@@ -128,24 +128,5 @@ double
 calculate_point_property(const double phase,
                          const double property0,
                          const double property1);
-//{
-//  double property_eq = phase * property1 + (1 - phase) * property0;
-
-//  // Limit parameters value (patch)
-//  // TODO see if necessary after compression term is added in the
-//  // free_surface advection equation
-//  const double property_min = std::min(property0, property1);
-//  const double property_max = std::max(property0, property1);
-//  if (property_eq < property_min)
-//    {
-//      property_eq = property_min;
-//    }
-//  if (property_eq > property_max)
-//    {
-//      property_eq = property_max;
-//    }
-
-//  return property_eq;
-//}
 
 #endif
--- a/include/core/utilities.h
+++ b/include/core/utilities.h
@@ -124,9 +124,28 @@ make_table_tensors_scalars(
  *
  * @param property1 Property value for the fluid with index 1 (fluid for phase = 1)
  */
-double
+inline double
 calculate_point_property(const double phase,
                          const double property0,
-                         const double property1);
+                         const double property1)
+{
+  double property_eq = phase * property1 + (1 - phase) * property0;
+
+  // Limit parameters value (patch)
+  // TODO see if necessary after compression term is added in the
+  // free_surface advection equation
+  const double property_min = std::min(property0, property1);
+  const double property_max = std::max(property0, property1);
+  if (property_eq < property_min)
+    {
+      property_eq = property_min;
+    }
+  if (property_eq > property_max)
+    {
+      property_eq = property_max;
+    }
+
+  return property_eq;
+}
 
 #endif
--- a/include/core/utilities.h
+++ b/include/core/utilities.h
@@ -115,4 +115,37 @@ make_table_tensors_scalars(
 
 
 
+/**
+ * @brief Used in multiphasic simulations to calculates the equivalent properties for a given phase. Method called in quadrature points loop
+ *
+ * @param phase Phase value for the given quadrature point
+ *
+ * @param property0 Property value for the fluid with index 0 (fluid for phase = 0)
+ *
+ * @param property1 Property value for the fluid with index 1 (fluid for phase = 1)
+ */
+double
+calculate_point_property(const double phase,
+                         const double property0,
+                         const double property1);
+//{
+//  double property_eq = phase * property1 + (1 - phase) * property0;
+
+//  // Limit parameters value (patch)
+//  // TODO see if necessary after compression term is added in the
+//  // free_surface advection equation
+//  const double property_min = std::min(property0, property1);
+//  const double property_max = std::max(property0, property1);
+//  if (property_eq < property_min)
+//    {
+//      property_eq = property_min;
+//    }
+//  if (property_eq > property_max)
+//    {
+//      property_eq = property_max;
+//    }
+
+//  return property_eq;
+//}
+
 #endif
--- a/include/solvers/analytical_solutions.h
+++ b/include/solvers/analytical_solutions.h
@@ -51,7 +51,7 @@ namespace AnalyticalSolutions
   public:
     AnalyticalSolution()
       : enable(false)
-      , velocity(dim + 1)
+      , uvwp(dim + 1)
       , temperature(1)
       , tracer(1)
     {}
@@ -75,7 +75,7 @@ namespace AnalyticalSolutions
 
     Parameters::Verbosity verbosity;
     // Velocity components + pressure value
-    Functions::ParsedFunction<dim> velocity;
+    Functions::ParsedFunction<dim> uvwp;
 
     // Auxiliary physics
     Functions::ParsedFunction<dim> temperature;
--- a/include/solvers/analytical_solutions.h
+++ b/include/solvers/analytical_solutions.h
@@ -74,12 +74,15 @@ namespace AnalyticalSolutions
     }
 
     Parameters::Verbosity verbosity;
-    // Velocity components
+    // Velocity components + pressure value
     Functions::ParsedFunction<dim> velocity;
 
+    // Auxiliary physics
     Functions::ParsedFunction<dim> temperature;
 
     Functions::ParsedFunction<dim> tracer;
+
+    Functions::ParsedFunction<dim> phase;
   };
 } // namespace AnalyticalSolutions
 
--- a/include/solvers/free_surface.h
+++ b/include/solvers/free_surface.h
@@ -32,6 +32,10 @@
 #include <deal.II/distributed/tria_base.h>
 
 #include <deal.II/fe/fe_q.h>
+#ifdef DEAL_II_WITH_SIMPLEX_SUPPORT
+#  include <deal.II/fe/fe_simplex_p.h>
+#endif
+
 #include <deal.II/fe/mapping_fe.h>
 #include <deal.II/fe/mapping_q.h>
 
--- a/include/solvers/free_surface.h
+++ b/include/solvers/free_surface.h
@@ -14,16 +14,16 @@
  * ---------------------------------------------------------------------
 
  *
- * Implementation of heat transfer as an auxiliary physics.
- * This heat equation is weakly coupled to the velocity field.
- * Equation solved:
- * rho * Cp * (dT/dt + u.gradT) = k div(gradT) + nu/rho * (gradu : gradu)
+ * Implementation of free surface with a Volume of Fluid method.
+ * Two fluid formulation. The phase indicator "phase" is equal to 0
+ * in one fluid and 1 in the other. The free surface is located
+ * where "phase" is equal to 0.5.
  *
- * Author: Bruno Blais and Jeanne Joachim, Polytechnique Montreal, 2020-
+ * Author: Jeanne Joachim, Polytechnique Montreal, 2021
  */
 
-#ifndef lethe_heat_transfer_h
-#define lethe_heat_transfer_h
+#ifndef lethe_free_surface_h
+#define lethe_free_surface_h
 
 #include <deal.II/base/convergence_table.h>
 #include <deal.II/base/quadrature_lib.h>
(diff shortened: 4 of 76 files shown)


include/solvers/simulation_parameters.h
```diff
@@ -135,6 +135,73 @@ class SimulationParameters
...
+    //              {
+    //                // if error other than "already exists"
+    //                std::cerr << "Could not create Output folder:  "
+    //                          << strerror(errno) << std::endl;
+    //              }
+    //          }
+    //        else
+    //          std::cout << "Output folder created: "
+    //                    << simulation_control.output_folder << std::endl;
+    //      }
```
blaisb, line 163: What to do about this block?

include/solvers/simulation_parameters.h
```diff
@@ -135,6 +135,73 @@ class SimulationParameters
...
+
+    particlesParameters.ib_force_output_file =
+      simulation_control.output_folder +
+      particlesParameters.ib_force_output_file;
+
+    nitsche->force_output_name =
+      simulation_control.output_folder + nitsche->force_output_name;
+
+    nitsche->torque_output_name =
+      simulation_control.output_folder + nitsche->torque_output_name;
```
blaisb, line 190: I don't really like this idea of "overwriting" the simulation parameters with additional variable. It's like if we parse something and we overwrite it locally. I think it would be better to use the output_folder manually where we write instead. This would be more coherent with the rest of the library.

include/solvers/simulation_parameters.h
```diff
@@ -135,6 +135,73 @@ class SimulationParameters
...
+        std::cerr
+          << "--------------------------------" << std::endl
+          << "WARNING inconsistency in .prm : " << std::endl
+          << "    in subsection multiphysics, free surface = "
+          << multiphysics.free_surface << std::endl
+          << "    but in subsection physical properties, number of fluids = "
+          << physical_properties.number_fluids << std::endl
+          << "    -> change for number of fluids = 2" << std::endl
+          << "--------------------------------" << std::endl;
+      }
```
blaisb, line 204: cerr stops the simulation completely right? Generally we try to AssertThrow or just Throw an exception instead of cerr, but maybe here it makes sense. I'm not sure which design philosophy is better here.

include/solvers/tracer.h
```diff
@@ -14,10 +14,7 @@
  * ---------------------------------------------------------------------
 
  *
- * Implementation of heat transfer as an auxiliary physics.
- * This heat equation is weakly coupled to the velocity field.
- * Equation solved:
- * rho * Cp * (dT/dt + u.gradT) = k div(gradT) + nu/rho * (gradu : gradu)
+ * Implementation of tracer as an auxiliary physics.
```
blaisb, line 17: Would be good to add the equation of the tracer here.

include/solvers/tracer.h
```diff
@@ -208,40 +205,43 @@ class Tracer : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
...
+  void
   solve_linear_system(const bool initial_step,
                       const bool renewed_matrix = true);
 
 
   /**
    * @brief Getter methods to get the private attributes for the physic currently solved
+   * NB : dof_handler and present_solution are already passed to the
+   * multiphysics interface at the end of the setup_dofs method
+   * //TODO Refactor
```
blaisb, line 217: Is it still necessary to refactor this?

source/core/parameters.cc
```diff
@@ -219,11 +240,95 @@ namespace Parameters
...
+      density           = prm.get_double("density");
+      dynamic_viscosity = prm.get_double("dynamic viscosity");
+      viscosity         = prm.get_double("kinematic viscosity");
+      if (dynamic_viscosity == 0)
+        // if the dynamic_viscosity is not given in prm
+        dynamic_viscosity = density * viscosity;
+      if (dynamic_viscosity != 0 && viscosity == 1)
+        // if the dynamic_viscosity is given in prm, and not the kinematic
+        // viscosity
+        viscosity = dynamic_viscosity / density;
```
blaisb, line 327: I think it would be important to check that density > 0 :)!

blaisb (commented):
First part of the review. I'll finish later this week :)!

blaisb:
@jeannejoachim CI is breaking right now. Seems there's still something to fix.

include/core/parameters.h
```diff
@@ -124,6 +124,9 @@ namespace Parameters
     // Enable output of the boundaries
     bool output_boundaries;
 
+    // Enable automatic folder creation if does not exist
+    bool make_folder;
+
```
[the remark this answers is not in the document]
jeannejoachim: No you're right it was a past test, I am going to remove it!

include/core/utilities.h
```diff
@@ -115,4 +115,37 @@ make_table_tensors_scalars(
...
+ * @param phase Phase value for the given quadrature point
+ *
+ * @param property0 Property value for the fluid with index 0 (fluid for phase = 0)
+ *
+ * @param property1 Property value for the fluid with index 1 (fluid for phase = 1)
+ */
+double
+calculate_point_property(const double phase,
+                         const double property0,
+                         const double property1);
```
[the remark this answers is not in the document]
jeannejoachim: I did that at first, but for the `utilities` I get multiple compilation error:
`/usr/bin/ld.gold:-1: error: error: ../../source/solvers/liblethe-solvers.a(navier_stokes_base.cc.o): multiple definition of 'calculate_point_property(double, double, double)'
 ../../source/solvers/liblethe-solvers.a(gls_navier_stokes.cc.o):-1: error: previous definition here`
 
So I assumed that I should not implement a method in the `utilities.h` ^^" Is there something else I should do?

include/solvers/analytical_solutions.h
```diff
@@ -74,12 +74,15 @@ namespace AnalyticalSolutions
     }
 
     Parameters::Verbosity verbosity;
-    // Velocity components
+    // Velocity components + pressure value
     Functions::ParsedFunction<dim> velocity;
```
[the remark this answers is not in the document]
jeannejoachim: Totally! I will change it to `uvwp` to be consistent with how I changed the parameter file

include/solvers/free_surface.h
```diff
@@ -0,0 +1,317 @@
...
+ * The full text of the license can be found in the file LICENSE at
+ * the top level of the Lethe distribution.
+ *
+ * ---------------------------------------------------------------------
+
+ *
+ * Implementation of free surface with a Volume of Fluid method.
+ * Two fluid formulation. The phase indicator "phase" is equal to 0
+ * in one fluid and 1 in the other. The free surface is located
+ * where "phase" is equal to 0.5.
```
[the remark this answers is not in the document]
jeannejoachim: I used the canonical definition I saw in all articles (from 0 to 1, with 0.5 being considered to be where the free surface is located). I do not remember seeing a -1 value in the articles, though there would be no problem using this value if you wish!

include/solvers/free_surface.h
```diff
@@ -0,0 +1,317 @@
...
+    //      {
+    //        // for simplex meshes
+    //        fe = std::make_shared<FE_SimplexP<dim>>(
+    //          simulation_parameters.fem_parameters.tracer_order);
+    //        mapping         = std::make_shared<MappingFE<dim>>(*fe);
+    //        cell_quadrature = std::make_shared<QGaussSimplex<dim>>(fe->degree
+    //        + 1);
+    //      }
+    //    else
+    //#endif
```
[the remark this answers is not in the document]
jeannejoachim: certainly, I will see to that! Should I had a simplex test at this point or we wait for a fuller implementation?

include/solvers/simulation_parameters.h
jeannejoachim: These are what worked on my computer + a test to get it CI-compatible, but we might need to use another C++ version as discussed together. It will be cleaner to remove it for the PR, I will keep it on my local branch to remember that I want to put it in a upcoming PR :)

include/solvers/simulation_parameters.h
jeannejoachim: Do you mean, to ask the user to manually give the "output folder" in multiple subsection? I think it would be very heavy for the user, and a recipe for disaster if you want to launch multiple simulations with sub-directories. 
Or do you mean, we should parse through the code/have a getter method for the `simulation_control.output_folder` so that the `output_name` is directly defined properly in the other parameter objects (`post_processing`, `nitsche` and so on)?

I must say I have great plans for improving folder's management with Lethe ^^ In a later PR, I would like to create automatically a copy of the `.prm` in the `output_folder`, and create a new subfolder in the `output_folder` with the Paraview visualizations. This will really make it easier to check previously launched simulations, and from then we could easily make parametric simulations automatic with an external script (like, you have another script which writes .prm for each parameter, then launches all the simulations one after the other, and the results are neatly stored in their folder). I could tell you more about that ^^

include/solvers/simulation_parameters.h
jeannejoachim: No here the simulation is not stopped, it is just a specialized kind of `cout` message. I will go with the Assertion then, thank you :)

include/solvers/tracer.h
jeannejoachim: ok done!

source/core/parameters.cc
jeannejoachim: yes sure! Done ;)

include/core/utilities.h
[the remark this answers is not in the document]
blaisb: That's weird. Are the include guards (#ifdef) of the function correctly implemented? It should not happen. Otherwise I'll check it tomorrow.

include/solvers/simulation_parameters.h
blaisb: What I mean is that I would not merge the output path and the output file name within the parameter itself. I would keep the two parameters seperate and carry them everywhere. I want to avoid "overwriting" a parameter once the parameter file has been read. It gives a very weird use experience (that parameters are being overwritten). I would just carry both the path and the output name and use it where you need it. E.g. When you want to write the force file, the file name would be = param.output_path + param.force_file_name.

source/solvers/gls_navier_stokes.cc
```diff
@@ -445,9 +450,9 @@ GLSNavierStokesSolver<dim>::assembleGLS()
               const double JxW = fe_values.JxW(q);
 
               // Calculation of the GLS stabilization parameter. The
-              // stabilization parameter used is different if the simulation is
-              // steady or unsteady. In the unsteady case it includes the value
-              // of the time-step
+              // stabilization parameter used is different if the simulation
+              // is steady or unsteady. In the unsteady case it includes the
+              // value of the time-step
```
blaisb, line 455: These are weird changes? Were they caused by the automatic indentation?

source/solvers/gls_navier_stokes.cc
```diff
@@ -818,6 +823,528 @@ GLSNavierStokesSolver<dim>::assembleGLS()
...
+                physical_properties.fluids[1].dynamic_viscosity);
+
+              // BB temporary
+              // Limitations for cases where air becomes liquid
+              if (density_eq < density_eq_m1)
+                density_eq_m1 = 0;
+
+              // BB temporary
+              // define epsilon below which gravity is not applied
+              double epsilon_alpha = 1e-6;
```
blaisb, line 1043: Should we remove those?

source/solvers/gls_navier_stokes.cc
```diff
@@ -961,9 +1488,17 @@ GLSNavierStokesSolver<dim>::assemble_matrix_and_rhs(
     {
       if (time_stepping_method ==
           Parameters::SimulationControl::TimeSteppingMethod::bdf1)
-        assembleGLS<true,
-                    Parameters::SimulationControl::TimeSteppingMethod::bdf1,
-                    Parameters::VelocitySource::VelocitySourceType::none>();
+        { // TEMPORARY free surface implementation
```
blaisb, line 1491: Is this still temporary?

source/solvers/heat_transfer.cc
```diff
@@ -44,19 +43,18 @@ void
...
-  const double dynamic_viscosity = viscosity * density;
+  double density              = physical_properties.density;
+  double specific_heat        = physical_properties.specific_heat;
+  double thermal_conductivity = physical_properties.thermal_conductivity;
+  double viscosity            = physical_properties.viscosity;
 
-  const double rho_cp = density * specific_heat;
-
-  const double alpha = thermal_conductivity / rho_cp;
+  double dynamic_viscosity = viscosity * density;
```
blaisb, line 55: Is this still coherent with the new physical properties? I feel this "Having two viscosity" definition is gonna be painful if it is used elsewhere. Maybe it would be simpler to revert to just having kinematic viscosity and density defined and let the dynamic viscosity be dynamically refined when needed.

source/solvers/heat_transfer.cc
```diff
@@ -923,7 +990,8 @@ HeatTransfer<dim>::solve_linear_system(const bool initial_step,
   if (simulation_parameters.linear_solver.verbosity !=
       Parameters::Verbosity::quiet)
     {
-      this->pcout << "  -Iterative solver took : " << solver_control.last_step()
+      this->pcout << "  Heat Transfer : " << std::endl
+                  << "  -Iterative solver took : " << solver_control.last_step()
```
blaisb, line 994: I don't think it should be at every iterative solver solution that we should announce the physics. Identifying which physics is being solved for should be done by the multiphysics interface (when it calls solve) instead of every solution of a linear system. That would be clearer. You would get something like
Fluid (default, no need to announce)
Heat transfer 
 Newton iteration 0
    - Iterative solver took 
  Newton iteration 1
    - Iterative solver took
Free Surface
Etc.

source/solvers/navier_stokes_base.cc
```diff
@@ -509,6 +509,8 @@ NavierStokesBase<dim, VectorType, DofsType>::iterate()
         simulation_parameters.simulation_control.method, false, false);
       multiphysics->solve(simulation_parameters.simulation_control.method,
                           false);
+      //      PhysicsSolver<VectorType>::solve_non_linear_system(
+      //        simulation_parameters.simulation_control.method, false, false);
```
blaisb, line 513:
```suggestion
```

blaisb (commented):
Some things there and there but nothing  major. There's still some good confusion with the duplication of viscosity (kinematic and dynamic), especially in the heat transfer solver. The more I think about it, the more i'd rather revert to defining only kinematic viscosity and density and then letting the solver refined dynamic viscosity when needed. Otherwise, it's going to be a major source of confusion down the road. Otherwise nothing major.

include/core/utilities.h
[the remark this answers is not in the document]
jeannejoachim: Ok done, I just had to explicitely inline it! Thanks

include/solvers/simulation_parameters.h
jeannejoachim: Alright, I will see to that

source/solvers/free_surface.cc
```diff
@@ -0,0 +1,762 @@
...
+      Parameters::SimulationControl::TimeSteppingMethod::bdf3)
+    bdf_coefs = bdf_coefficients(3, time_steps_vector);
+
+  if (time_stepping_method ==
+        Parameters::SimulationControl::TimeSteppingMethod::sdirk22_1 ||
+      time_stepping_method ==
+        Parameters::SimulationControl::TimeSteppingMethod::sdirk33_1)
+    {
+      throw std::runtime_error(
+        "SDIRK schemes are not supported by heat transfer physics");
```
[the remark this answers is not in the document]
jeannejoachim: good catch!

source/solvers/gls_navier_stokes.cc
jeannejoachim: certainly, I do not recall having chanced these!

source/solvers/gls_navier_stokes.cc
jeannejoachim: For the condition on `density_eq_m1`, right now we have `density_eq_m1 = density_eq` anyway, so it is never true. This would require further testing to consider a non-constant density as we had some issues with that, so I will remove it for now.

I think we should keep the principle of `epsilon_alpha`, so that gravity is not applied on the air part. Although, maybe we should instead consider a condition on the fluid density? If we make a simulation with two fluids that are not air, we may want to have gravity applied to both... What do you think?

source/solvers/gls_navier_stokes.cc
jeannejoachim: I guess no, if we keep this specialized assembler!

include/solvers/multiphysics_interface.h
```diff
@@ -65,6 +65,34 @@ class MultiphysicsInterface
     return active_physics;
   }
 
+  /**
+   * @brief Write physic solved in the terminal
+   *
+   * @param physics_id number associated with auxiliary physics in multiphysics.h
+   */
+  void
+  annonce_physics(const PhysicsID physics_id)
```
blaisb, line 74:
```suggestion
  announce_physics(const PhysicsID physics_id)
```

include/solvers/multiphysics_interface.h
```diff
@@ -65,6 +65,34 @@ class MultiphysicsInterface
...
+  {
+    if (physics_id == 1)
```
blaisb, line 76: Instead of using the raw numbers for the physics id, it's better to use the enum from the multiphysics namespace. This way things are coherent everywhere. 
E.g.  https://github.com/lethe-cfd/lethe/blob/master/include/core/multiphysics.h
PhysicsID::fluid_dynamics for example.

include/solvers/multiphysics_interface.h
```diff
@@ -81,10 +109,18 @@ class MultiphysicsInterface
...
+        // Annonce physic solved
+        if (verbosity != Parameters::Verbosity::quiet)
+          annonce_physics(iphys.first);
+
         solve_physics(iphys.first, time_stepping_method, force_matrix_renewal);
       }
     for (auto &iphys : block_physics)
       {
+        // Annonce physic solved
+        if (verbosity != Parameters::Verbosity::quiet)
```
blaisb, line 121: Which Parameters::Verbosity is that? Is there a general verbosity?

source/solvers/multiphysics_interface.cc
```diff
@@ -1,16 +1,17 @@
...
 
 
-
 template <int dim>
 MultiphysicsInterface<dim>::MultiphysicsInterface(
   const SimulationParameters<dim> &                            nsparam,
   std::shared_ptr<parallel::DistributedTriangulationBase<dim>> p_triangulation,
   std::shared_ptr<SimulationControl> p_simulation_control)
   : multiphysics_parameters(nsparam.multiphysics)
+  , verbosity(nsparam.linear_solver.verbosity)
```
blaisb, line 14: So are we sure we want to use the linear solver verbosity here? Maybe the non-linear solver verbosity would make more sense. Actually, maybe all we would need a single verbosity in the non-linear solver and if that verbosity is on, the iterative solver /linear solver verbosity would be on too.

blaisb (commented):
A few very minor comments left. Very very small.
So one final remark : Can you make sure that the examples still work with your changes? I mean, ensure that the examples do not break now with the uvwp for the analytical solution?
That's the main thing left to do IMO, but tomorrow we could merge :)!.

include/solvers/multiphysics_interface.h
jeannejoachim: Oh yes indeed! Thank you :)

source/solvers/multiphysics_interface.cc
jeannejoachim: Yes you're right, I will change it for `nonlinear_solver` verbosity! I would not mind generalizing the verbosity to the linear solver, but are we sure that it is not necessary to keep them apart? Also it can be confusing for the users that a parameter in the nonlinear solver section changes something for the linear solver. 

I would advise, if we want a general verbosity parameter for the solvers, to pass it to a more general subsection of the prm file (like simulation control). From there, we can have: solver verbosity = quiet / verbose. And it would enable or disable verbosity for both. Would this suits what you had in mind?

blaisb (approved):
Good to go. I'll merge it early next week Monday :).

source/solvers/multiphysics_interface.cc
blaisb: Yeah you are right. That would be the preferable route in the long term. Let's keep that in mind but keep it on the backburner for the moment. This PR has grown big enough as it is!

PR #155, opened 2021-04-24
Add sphere example

cenr20:
Add parameter and mesh files for Example 4 - 3D Flow around a Sphere


blaisb:
@cenr20 can  you remove the .msh file from the PR? I prefer to only leave .geo when possible, it makes for much lighter PR.
:)!

cenr20:
Done!

PR #157, opened 2021-04-28
Add Nitsche example and fix some issues 23 nitsche

blaisb:
This PR adds a Nitsche example to the Wiki and it fixes some scaling issues in the Nitsche_23 solver, making the solver a lot more robust than it used to be. The penalty parameter was not scaling correctly as the mesh was refined when the IB was a surface.


source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -104,8 +104,9 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::assemble_nitsche_restriction()
           else if (dim == 3)
             h_cell = pow(6 * cell->measure() / M_PI, 1. / 3.) /
                      this->velocity_fem_degree;
-          const double penalty_parameter = 1. / (h_cell * h_cell);
-          const auto & dh_cell =
+          const double penalty_parameter =
+            1. / std::pow(h_cell * h_cell, double(dim) / double(spacedim));
```
jeannejoachim, line 108: would you have a reference for this?

jeannejoachim (approved):
Thank you for adding these examples! They are gorgeous on the wiki as well :)

applications_tests/gls_nitsche_navier_stokes_23/couette_23.output
```diff
@@ -30,15 +30,15 @@ Steady iteration :        2/2
...
 +------------------------------------------+
 Solid ID    f_x      f_y       f_z    
-       0 -1.996979 0.000000 0.000000 
-cells error_velocity  error_pressure  
-   64 1.9041e-03    - 1.6541e-13    - 
-  512 2.6039e-04 2.87 1.5172e-13 0.12 
+       0 -1.990689 0.000000 0.000000 
+cells error_velocity   error_pressure  
+   64 7.8763e-03    - 1.3625e-13     - 
+  512 1.6457e-03 2.26 1.7555e-13 -0.37
```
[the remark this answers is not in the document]
blaisb: The couette example has been changed, it's much more "tall" now, so the cylinder being bigger leads to a bigger error. No biggie here :). I wanted to make this example less reliant on external files.

source/solvers/gls_nitsche_navier_stokes.cc
blaisb: So previously the penalty parameter is scaled as 1/h^2. But the issue is that in dim==2 spacedim==3, it is integrated over an area. So the penalty is extremely strong. It's what made the solver very unstable when the mesh was being refined.
Here we regularize the penalty parameter so that it is scaled correctly. We have a larger error, but the convergence is recovered correctly. It's kind-off a tradeoff. I don't have a reference because we are kinda inventing this... :)

PR #160, opened 2021-05-11
Revise parallel insertion, Fix minor bugs

shahabgol:
- Change parallel insertion algorithm. Insertion locations are now divided between the processes with a different algorithm.
- Fix a minor bug in write_xyz function.


applications_tests/dem_2d/circle_with_floating_walls.mpirun=2.output
```diff
@@ -7,20 +7,21 @@ Starting simulation with Lethe/DEM on 2 processors
 18 particles of type 0 were inserted, 0 particles of type 0 remaining
 ***************************************************************** 
 id, type, dp, x, y, z 
-0 0 0.00500 -0.0113 -0.0964
+0 0 0.00500 -0.0114 -0.0964
```
blaisb, line 10: What's the reason for the results changing?

source/dem/dem.cc
```diff
@@ -482,11 +482,12 @@ DEMSolver<dim>::finish_simulation()
            processor_number < n_mpi_processes;
            ++processor_number)
         {
+          MPI_Barrier(MPI_COMM_WORLD);
+
           if (this_mpi_process == processor_number)
             {
               visualization_object.print_xyz(particle_handler, pcout);
             }
-          MPI_Barrier(MPI_COMM_WORLD);
```
blaisb, line 489: Why the move of the Barrier. I don't really get it. It should work with the barrier afterwards, not before since what you want is to ensure that all processor have the time to write. Here what could happen is that the last processor does not have the time to write  output

source/dem/non_uniform_insertion.cc
```diff
@@ -62,25 +66,43 @@ NonUniformInsertion<dim>::insert(
         Utilities::MPI::all_gather(communicator, my_bounding_box);
 
       // Distbuting particles between processors
```
blaisb, line 68:
```suggestion
      // Calculate the number of particles to insert for each processor
```

source/dem/non_uniform_insertion.cc
```diff
@@ -62,25 +66,43 @@ NonUniformInsertion<dim>::insert(
...
-      if (Utilities::MPI::this_mpi_process(communicator) !=
-          (Utilities::MPI::n_mpi_processes(communicator) - 1))
+      if (this_mpi_process != (n_mpi_process - 1))
         this->inserted_this_step_this_proc =
-          floor(this->inserted_this_step /
-                Utilities::MPI::n_mpi_processes(communicator));
+          floor(this->inserted_this_step / n_mpi_process);
       else
```
blaisb, line 72: You could replace this if/else structure by just have everyone equal to           f
foor(this->inserted_this_step / n_mpi_process); then overwriting the result with an if((this_mpi_process == (n_mpi_process - 1))). It would actually make it easier to read.

source/dem/non_uniform_insertion.cc
```diff
@@ -62,25 +66,43 @@ NonUniformInsertion<dim>::insert(
...
+        }
+      else
+        {
+          int begin_element =
+            this->inserted_this_step - this->inserted_this_step_this_proc;
+          insertion_points_on_proc =
+            std::vector<Point<dim>>(global_insertion_points.cbegin() +
+                                      begin_element,
+                                    global_insertion_points.cend());
+        }
```
blaisb, line 105: Wait, so all processor calculate the entire std::vector of points, but then only takes a sub-part of it? Why? This seems expensive for no reason. Imagine you have 1M particles, but 1000 core. Each core will create a vector of 1M points just to use 1K points of it? You could calculate only the points necessary by obtaining the i,j,k id from the global particle number using modulo and integer division... It would be much cleaner than to carry the global vector all around.

source/dem/non_uniform_insertion.cc
```diff
@@ -151,50 +169,41 @@ NonUniformInsertion<dim>::assign_insertion_points(
           (dim == 3) ? this->number_of_particles_z_direction : 1;
         for (unsigned int k = 0; k < dim_nz; ++k)
```
blaisb, line 170: I'm not a fan of this idea of "recalculating all points" even if you insert a subsection of them. This is going to be very expensive for no reason. You can deduce the local k, j, i from the  index and avoid all of this mega looping.

blaisb (commented):
I added some comments. I don't like the fact that all points are calculated for all cores. You should only calculate the points that you need. Otherwise it will eventually break down. Imagine if you have 10M particles and 1k core, it will require 1000x more time/memory than needed. You can calculate i,j,k from the particle start and end number using modulo and integer division. We can talk about it if you want.

applications_tests/dem_2d/circle_with_floating_walls.mpirun=2.output
shahabgol: The insertion points are a little different in parallel

source/dem/dem.cc
shahabgol: the problem is: with having barrier at the end, the first process does not necessarily calls write_xyz function first. then in the function, the file header (which writes "id, type, x, y, z, vx, vy, vz" before writing the particles' information) is not written in the first row. for instance the output file becomes something like:
 0 2 0.1 0.3 2 1
 1 3 0.2 0.5 2 2
id, type, x, y, z, vx, vy, vz
2 1 0.1 0.4 0.1 0.5
3 1 0.4 0.5 0.1 0.6

source/dem/non_uniform_insertion.cc
shahabgol: revised

source/dem/non_uniform_insertion.cc
shahabgol: revised

source/dem/non_uniform_insertion.cc
shahabgol: done

include/dem/pp_nonlinear_force.h
```diff
@@ -106,9 +106,10 @@ class PPNonLinearForce : public PPContactForce<dim>
                 const Tensor<1, dim> & /*normal_contact_vector*/)
   {
     Tensor<1, dim> rolling_resistance;
-    rolling_resistance[0] = 0;
-    rolling_resistance[1] = 0;
-    rolling_resistance[2] = 0;
+    for (int d = 0; d < dim; ++d)
+      {
+        rolling_resistance[d] = 0;
+      }
```
blaisb, line 112:
```suggestion
        rolling_resistance[d] = 0;
```

include/dem/pw_linear_force.h
```diff
@@ -103,9 +103,10 @@ class PWLinearForce : public PWContactForce<dim>
                 const Tensor<1, dim> & /*normal_contact_vector*/)
   {
     Tensor<1, dim> rolling_resistance;
-    rolling_resistance[0] = 0;
-    rolling_resistance[1] = 0;
-    rolling_resistance[2] = 0;
+    for (int d = 0; d < dim; ++d)
+      {
+        rolling_resistance[d] = 0;
+      }
```
blaisb, line 109:
```suggestion
        rolling_resistance[d] = 0;
```

include/dem/pw_nonlinear_force.h
```diff
@@ -103,9 +103,10 @@ class PWNonLinearForce : public PWContactForce<dim>
                 const Tensor<1, dim> & /*normal_contact_vector*/)
   {
     Tensor<1, dim> rolling_resistance;
-    rolling_resistance[0] = 0;
-    rolling_resistance[1] = 0;
-    rolling_resistance[2] = 0;
+    for (int d = 0; d < dim; ++d)
+      {
+        rolling_resistance[d] = 0;
+      }
```
blaisb, line 109:
```suggestion
        rolling_resistance[d] = 0;
```

source/dem/input_parameter_inspection.cc
```diff
@@ -60,8 +60,10 @@ input_parameter_inspection(const DEMSolverParameters<dim> &dem_parameters,
     }
 
   // Insertion parameters checking
-  if (parameters.insertion_info.random_number_range >=
-      (0.5 * (parameters.insertion_info.distance_threshold - 1)))
+  if (parameters.insertion_info.insertion_method ==
+        Parameters::Lagrangian::InsertionInfo::InsertionMethod::non_uniform &&
+      parameters.insertion_info.random_number_range >=
+        (0.5 * (parameters.insertion_info.distance_threshold - 1)))
```
blaisb, line 66: Can you put the content of line 66 in a variable with a given name. Otherwise it's difficult to understand what it is supposed to mean.

source/dem/non_uniform_insertion.cc
```diff
@@ -47,6 +47,10 @@ NonUniformInsertion<dim>::insert(
       ConditionalOStream pcout        = {
         std::cout, Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0};
 
+      auto this_mpi_process = Utilities::MPI::this_mpi_process(communicator);
+      auto n_mpi_process    = Utilities::MPI::n_mpi_processes(communicator);
+
+
```
blaisb, line 53:
```suggestion
```

source/dem/non_uniform_insertion.cc
```diff
@@ -122,82 +163,65 @@ NonUniformInsertion<dim>::create_random_number_container(
...
-                      }
-
-                    insertion_positions.push_back(position);
-                    particle_counter++;
-                  }
-              }
-          }
-      }
+  std::vector<int> insertion_index;
+  insertion_index.reserve(3);
```
blaisb, line 201:
```suggestion
  insertion_index.resize(3);
```
This should be a resize here because you want to contain 3 elements and you are not pushing back. Reserve is only when you want to pre-allocate without changing size.

https://stackoverflow.com/questions/13029299/stdvectorresize-vs-stdvectorreserve

source/dem/non_uniform_insertion.cc
```diff
@@ -122,82 +163,65 @@ NonUniformInsertion<dim>::create_random_number_container(
...
+template <>
+void NonUniformInsertion<2>::find_insertion_location_nonuniform(
+  Point<2> &                                   insertion_location,
+  const unsigned int &                         id,
+  const double &                               random_number1,
+  const double &                               random_number2,
+  const Parameters::Lagrangian::InsertionInfo &insertion_information)
+{
+  std::vector<int> insertion_index;
+  insertion_index.reserve(2);
```
blaisb, line 175:
```suggestion
  insertion_index.resize(2);
```

This should be a resize here because you want to contain 3 elements and you are not pushing back. Reserve is only when you want to pre-allocate without changing size.

https://stackoverflow.com/questions/13029299/stdvectorresize-vs-stdvectorreserve

You could also use the constructor for this I guess.

source/dem/uniform_insertion.cc
```diff
@@ -63,24 +66,45 @@ UniformInsertion<dim>::insert(
...
+                 this_mpi_process * this->inserted_this_step_this_proc;
+               id < (this_mpi_process + 1) * this->inserted_this_step_this_proc;
+               ++id)
+            {
+              find_insertion_location_uniform(insertion_location,
+                                              id,
+                                              dem_parameters.insertion_info);
+              insertion_points_on_proc.push_back(insertion_location);
+            }
+        }
```
blaisb, line 106: I'm not sure this is ideal way to set this up. You could just set the variables for the start and the end of the loop using the if/else statement, then do the same loop for everyone and have them same function call. Would make understanding this easier.

source/dem/uniform_insertion.cc
```diff
@@ -105,70 +129,64 @@ UniformInsertion<dim>::insert(
...
-                      }
-
-                    insertion_positions.push_back(position);
-                    particle_counter++;
-                  }
-              }
-          }
-      }
+  std::vector<int> insertion_index;
+  insertion_index.reserve(2);
```
blaisb, line 139:
```suggestion
  insertion_index.resize(2);
```

source/dem/uniform_insertion.cc
```diff
@@ -105,70 +129,64 @@ UniformInsertion<dim>::insert(
...
+}
+
+template <>
+void UniformInsertion<3>::find_insertion_location_uniform(
+  Point<3> &                                   insertion_location,
+  const unsigned int &                         id,
+  const Parameters::Lagrangian::InsertionInfo &insertion_information)
+{
+  std::vector<int> insertion_index;
+  insertion_index.reserve(3);
```
blaisb, line 161:
```suggestion
  insertion_index.resize(3);
```

blaisb (commented):
Very minor stuff. Good PR and good job. Can you just apply the comments and I can merge it tomorrow?

source/dem/input_parameter_inspection.cc
shahabgol: done.

source/dem/non_uniform_insertion.cc
shahabgol: corrected. thank you for the tip.

source/dem/non_uniform_insertion.cc
shahabgol: corrected.

source/dem/uniform_insertion.cc
shahabgol: done.

shahabgol:
Thank you for the comments, especially the one on finding insertion's first and last id in parallel. All the comments are applied.

blaisb (approved):

PR #161, opened 2021-05-11
Iterative solver try catch

Luckabarbeau:
Modified how the fill level of the preconditioner is changed for the iterative solver when an error occurs. Applied this strategy to all iterative solver.

### DIFF
--- a/include/solvers/gls_navier_stokes.h
+++ b/include/solvers/gls_navier_stokes.h
@@ -142,13 +142,13 @@ private:
    * Set-up AMG preconditioner
    */
   void
-  setup_AMG(int temp_precond_fill_amg);
+  setup_AMG(const int current_amg_ilu_preconditioner_fill_level);
 
   /**
    * Set-up ILU preconditioner
    */
   void
-  setup_ILU(int temp_precond_fill_ilu);
+  setup_ILU(const int current_ilu_preconditioner_fill_level);
 
 
   /**
--- a/include/solvers/gls_navier_stokes.h
+++ b/include/solvers/gls_navier_stokes.h
@@ -162,8 +162,6 @@ private:
   std::shared_ptr<TrilinosWrappers::PreconditionILU> ilu_preconditioner;
   std::shared_ptr<TrilinosWrappers::PreconditionAMG> amg_preconditioner;
 
-
-
   const bool   SUPG        = true;
   const double GLS_u_scale = 1;
 };
--- a/include/solvers/gls_navier_stokes.h
+++ b/include/solvers/gls_navier_stokes.h
@@ -142,13 +142,13 @@ private:
    * Set-up AMG preconditioner
    */
   void
-  setup_AMG();
+  setup_AMG(int temp_precond_fill_amg);
 
   /**
    * Set-up ILU preconditioner
    */
   void
-  setup_ILU();
+  setup_ILU(int temp_precond_fill_ilu);
 
 
   /**
@@ -161,6 +161,8 @@ private:
   SparsityPattern                                    sparsity_pattern;
   std::shared_ptr<TrilinosWrappers::PreconditionILU> ilu_preconditioner;
   std::shared_ptr<TrilinosWrappers::PreconditionAMG> amg_preconditioner;
+
+
 
   const bool   SUPG        = true;
   const double GLS_u_scale = 1;
--- a/source/solvers/gls_navier_stokes.cc
+++ b/source/solvers/gls_navier_stokes.cc
@@ -1852,17 +1852,16 @@ GLSNavierStokesSolver<dim>::solve_linear_system(const bool initial_step,
 
 template <int dim>
 void
-GLSNavierStokesSolver<dim>::setup_ILU(int current_ilu_preconditioner_fill_level)
+GLSNavierStokesSolver<dim>::setup_ILU(const int current_ilu_preconditioner_fill_level)
 {
   TimerOutput::Scope t(this->computing_timer, "setup_ILU");
 
-  const double ilu_fill =current_ilu_preconditioner_fill_level;
   const double ilu_atol =
     this->simulation_parameters.linear_solver.ilu_precond_atol;
   const double ilu_rtol =
     this->simulation_parameters.linear_solver.ilu_precond_rtol;
   TrilinosWrappers::PreconditionILU::AdditionalData preconditionerOptions(
-    ilu_fill, ilu_atol, ilu_rtol, 0);
+          current_ilu_preconditioner_fill_level, ilu_atol, ilu_rtol, 0);
 
   ilu_preconditioner = std::make_shared<TrilinosWrappers::PreconditionILU>();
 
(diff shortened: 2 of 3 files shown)


include/solvers/gls_navier_stokes.h
```diff
@@ -142,13 +142,13 @@ class GLSNavierStokesSolver
    * Set-up AMG preconditioner
    */
   void
-  setup_AMG();
+  setup_AMG(int temp_precond_fill_amg);
```
blaisb, line 145:
```suggestion
  setup_AMG(int amg_ilu_preconditioner_fill);
```
Don't hesitate to use variable name that have a more extensive meaning. Additionnally, the fill level here is not temporary, it's the fill level used for the pre-conditioner itself.

include/solvers/gls_navier_stokes.h
```diff
@@ -142,13 +142,13 @@ class GLSNavierStokesSolver
...
 
   /**
    * Set-up ILU preconditioner
    */
   void
-  setup_ILU();
+  setup_ILU(int temp_precond_fill_ilu);
```
blaisb, line 151: Same as above.

source/solvers/gls_navier_stokes.cc
```diff
@@ -1852,12 +1852,11 @@ GLSNavierStokesSolver<dim>::solve_linear_system(const bool initial_step,
...
 template <int dim>
 void
-GLSNavierStokesSolver<dim>::setup_ILU()
+GLSNavierStokesSolver<dim>::setup_ILU(int temp_precond_fill_ilu)
 {
   TimerOutput::Scope t(this->computing_timer, "setup_ILU");
 
-  const double ilu_fill =
-    this->simulation_parameters.linear_solver.ilu_precond_fill;
+  const double ilu_fill =temp_precond_fill_ilu;
```
blaisb, line 1859: Is this variable still necessary? Why not just use the function argument?

include/solvers/gls_navier_stokes.h
blaisb: Additionnaly. Should be a const argument if it does not change.

source/solvers/gls_navier_stokes.cc
```diff
@@ -1952,9 +1950,8 @@ GLSNavierStokesSolver<dim>::solve_system_GMRES(const bool   initial_step,
   // the solver crash. if a change happened on the fill level it will go back to
   // it's original value at the end of the restart process.
 
-  unsigned int max_iter = 20;
-  unsigned int original_fill =
-    this->simulation_parameters.linear_solver.ilu_precond_fill;
+  unsigned int max_iter = 3;
+  int temp_precond_fill_ilu=this->simulation_parameters.linear_solver.ilu_precond_fill;
```
blaisb, line 1954:
```suggestion
  const unsigned int max_iter = 3;
  const unsigned int temp_precond_fill_ilu=this->simulation_parameters.linear_solver.ilu_precond_fill;
```

source/solvers/gls_navier_stokes.cc
```diff
@@ -2020,17 +2017,16 @@ GLSNavierStokesSolver<dim>::solve_system_GMRES(const bool   initial_step,
         }
       catch (std::exception &e)
         {
-          this->simulation_parameters.linear_solver.ilu_precond_fill += 1;
+          temp_precond_fill_ilu += 1;
           this->pcout
             << " GMRES solver failed! Trying with higher preconditioner fill level. New fill = "
```
blaisb, line 2022:
```suggestion
            << " GMRES solver failed! Trying with a higher preconditioner fill level. New fill = "
```

source/solvers/gls_navier_stokes.cc
```diff
@@ -2042,49 +2038,70 @@ GLSNavierStokesSolver<dim>::solve_system_BiCGStab(
   const bool   renewed_matrix)
 {
   TimerOutput::Scope t(this->computing_timer, "solve");
-  auto &             system_rhs          = this->system_rhs;
-  auto &             nonzero_constraints = this->nonzero_constraints;
+    unsigned int max_iter = 3;
+    int temp_precond_fill_ilu=this->simulation_parameters.linear_solver.ilu_precond_fill;
+    unsigned int iter    = 0;
+    bool         success = false;
```
blaisb, line 2044: If something is const, make it const so the user see it won't change.

blaisb: Also please use another name then temp_precond_fill_ilu. Maybe current_fill_level ?

source/solvers/gls_navier_stokes.cc
```diff
@@ -2094,56 +2111,79 @@ GLSNavierStokesSolver<dim>::solve_system_AMG(const bool   initial_step,
                                              const double relative_residual,
                                              const bool   renewed_matrix)
 {
-  auto &system_rhs          = this->system_rhs;
-  auto &nonzero_constraints = this->nonzero_constraints;
+    unsigned int max_iter = 3;
+    int temp_precond_fill_amg=this->simulation_parameters.linear_solver.amg_precond_ilu_fill;
+    unsigned int iter    = 0;
+    bool         success = false;
```
blaisb, line 2117: Make const what is const

source/solvers/gls_navier_stokes.cc
```diff
@@ -2187,51 +2228,73 @@ GLSNavierStokesSolver<dim>::solve_system_TFQMR(const bool   initial_step,
                                                const double relative_residual,
                                                const bool   renewed_matrix)
 {
-  auto &system_rhs          = this->system_rhs;
-  auto &nonzero_constraints = this->nonzero_constraints;
+    unsigned int max_iter = 3;
+    int temp_precond_fill_ilu=this->simulation_parameters.linear_solver.ilu_precond_fill;
+    unsigned int iter    = 0;
+    bool         success = false;
```
blaisb, line 2234: Make const what is

source/solvers/gls_navier_stokes.cc
```diff
@@ -2187,51 +2228,73 @@ GLSNavierStokesSolver<dim>::solve_system_TFQMR(const bool   initial_step,
...
+          }
+          constraints_used.distribute(completely_distributed_solution);
+          this->newton_update = completely_distributed_solution;
+          success             = true;
+        }
+        catch (std::exception &e)
+        {
+            temp_precond_fill_ilu += 1;
+            this->pcout
+                    << " TFQMR solver failed! Trying with higher preconditioner fill level. New fill = "
```
blaisb, line 2290:
```suggestion
                    << " TFQMR solver failed! Trying with a higher preconditioner fill level. New fill = "
```

blaisb (commented):
Very minor comments about variable naming and to add const when needed. Otherwise, good work.

source/solvers/gls_navier_stokes.cc
Luckabarbeau: I actually don't know I think this is a conversion from int to double  since the parameter fill returns an int.

blaisb: Can you verify if the function requires a double or an int? If it requires an int as an argument, it might be that the double is dynamically casted to an int. If that's the case, it should be an int from the get go and that just means I made a stupid mistake in the past.

blaisb (approved):
Awesome! Ready for merge

PR #165, opened 2021-05-26
New class for stencil

Luckabarbeau:
Created a new class for the immersed boundary stencil. This class allows a huge simplification of the code in the sharp_edge function and a small speed up in certain cases. This PR also includes a modification of the assembleGLS function of the GLS_sharp_navier_stokes solver. The modification is simply an update to make it on par with the standard assembleGLS function in the GLS_navier_stokes_solver. This update fixes an issue with BDF2, BDF3 , SDIRK2, SDIRK 3 integrator.


include/core/ib_stencils.h
```diff
@@ -0,0 +1,80 @@
...
+
+
+template<int dim>
+class IBStencils {
+
+public:
+
+    /**
+    * @brief
+    Return the number of point use in the interpolation excluding the dof itself base on the order of the stencil.
```
blaisb, line 36:
```suggestion
    Return the number of points used by the interpolation stencil excluding the dof itself for a stencil of a given order.
```

include/core/ib_stencils.h
```diff
@@ -0,0 +1,80 @@
...
+    /**
+    * @brief
+     * Define the points for the IB stencil, based on the order and the particle position as well as the DOF position
+     * Depending on the order, the output variable "point" change definition. In the case of stencil orders 1 to 4 the variable point returns the position of the DOF directly.
+     * In the case of high order stencil, it returns the position of the point that is on the IB.
+     * The variable "interpolation points" return the points used to define the cell used for the stencil definition and
+     * the locations of the points use in the stencil calculation.
+    *
+    * @param order, the stencil order.
+    * @param p, the IB particule that cut the cell.
```
blaisb, line 51:
```suggestion
    * @param p, the IB particle that cuts the cell.
```

include/core/ib_stencils.h
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+    * @brief
+     * Return the velocity of the IB used in the RHS of the equation
+    *
+    * @param p, the IB particule that cut the cell.
+    * @param dof_point, the support point of the DOF.
+    * @param component, the stencil component of the dof (vx=0,vy=1,vz=2).
+    */
+    virtual double vitesse_ib(IBParticle<dim> p,Point<dim> dof_point,unsigned int component);
```
blaisb, line 72:
```suggestion
    virtual double ib_velocity(IBParticle<dim> p,Point<dim> dof_point,unsigned int component);
```

A bit of French here? :P

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -181,6 +183,17 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
     find_cell_around_point_with_neighbors(const typename DoFHandler<dim>::active_cell_iterator &cell,
                                        Point<dim>             point);
 
+    /**
+    * @brief
+    Return a bool that describes  if a cell contains a specific point
+    *
+    * @param cell , The initial cell. We want to check if the point is inside.
```
blaisb, line 190:
```suggestion
    * @param cell , The initial cell for which we want to check if the point is inside.
```

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -181,6 +183,17 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+    *
+    * @param point, The point that we check
```
blaisb, line 192:
```suggestion
    * @param point, The point that we wish to check
```

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,272 @@
...
+
+
+        return coef;
+    }
+
+    std::vector<double> coef(2);
+    coef[0] = 9;
+    coef[1] = -8;
+
+    return coef;
```
blaisb, line 75: What is this weird other scenario?

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,272 @@
...
+}
+
+template<int dim>
```
blaisb, line 78: Maybe the .h would require a bit more explanation here?

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,272 @@
...
+                 ((dof_point -
+                   p.position) /
+                  (dof_point -
+                   p.position)
+                          .norm())[0] +
+                 p.velocity[1];
+
+        }
+    }
+    if (dim == 3) {
```
blaisb, line 205: Did you apply automatic indentation here ? That look sa bit messed up. Also, if you use a cross product,  why not use the cross product functionnality of deal.II? It's already there.

blaisb (commented):
Close to being good.  However, you should apply the indent all script and make sure that your calculation of the velocity at the dof points uses a proper cross product instead of hardcoding it. It will be much easier to understand ! :)

source/core/ib_stencils.cc
Luckabarbeau: apply both of those recommendation

source/core/ib_stencils.cc
Luckabarbeau: added some comments

include/core/ib_stencils.h
```diff
@@ -0,0 +1,89 @@
...
+#include <deal.II/fe/mapping_q1.h>
+
+#include <deal.II/lac/trilinos_parallel_block_vector.h>
+#include <deal.II/lac/trilinos_vector.h>
+
+#include <core/ib_particle.h>
+
+
+#ifndef LETHE_IB_STENCILS_H
+#  define LETHE_IB_STENCILS_H
```
blaisb, line 23: We don't use capital letters anymore
Can you fix that
```suggestion
#ifndef lethe_ib_stencils_h
#  define lethe_ib_stencils_h
```

include/core/ib_stencils.h
```diff
@@ -0,0 +1,89 @@
...
+   * @param order, the stencil order.
+   */
+  virtual std::vector<double>
+  coefficients(unsigned int order);
+
+  /**
+   * @brief
+   * Return the velocity of the IB used in the RHS of the equation
+   *
+   * @param p, the IB particule that cut the cell.
```
blaisb, line 80:
```suggestion
   * @param p, the IB particle that cuts the cell.
```

blaisb: Be careful about spelling overall

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -60,165 +61,193 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+
+  /**
+   * @brief
+   * Allow a refinement around each of the particles.
+   * The zone where the cells will be refined is defined by a ring in 2D and a
+   *shell in 3D. The outside and inside radius of the ring\shell is defined in
+   *relation to the diameter of the particle by the immersed boundaries
+   *parameter: "refine mesh inside radius factor" and "refine mesh outside
+   *radius factor". These factors multiply the radius of the particle to define
+   *the outside and inside radius of the ring\shell.
```
blaisb, line 159: Did you apply automatic indentation?

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -60,165 +61,193 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+   */
+  void
+  clear_line_in_matrix(
+    const typename DoFHandler<dim>::active_cell_iterator &cell,
+    unsigned int                                          dof_index);
+
+
+  // Modified version of assemble_matrix_and_rhs to include the presence of
+  // extra steps. For more detail see the same function in the
+  // gls_navier_stokes.h solver.
```
blaisb, line 235: Use /* */ syntax for multiline comments.

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,235 @@
...
+
+
+template <int dim>
+unsigned int
+IBStencils<dim>::nb_points(unsigned int order)
+{
+  // The number of points used in the stencil excluding the DOF is equal to the
+  // order.
+  unsigned int nb_points = order;
+  // In the case where the cell is used directly a=to to find the solution at
```
blaisb, line 16:
```suggestion
  // In the case where the cell is used directly to to find the solution at
```

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,235 @@
...
+    nb_points = 1;
+  return nb_points;
+}
+
+template <int dim>
+std::vector<double>
+IBStencils<dim>::coefficients(unsigned int order)
+{
+  // Initialize the coefficient vector
+  std::vector<double> coef;
```
blaisb, line 28: Can you resize here?
The size is order+1 except if order>4. You could set the size here for everyone once and for all.

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,235 @@
...
+      coef.resize(1);
+      coef[0] = 1;
+    }
+
+  return coef;
+}
+
+template <int dim>
+std::tuple<Point<dim>, std::vector<Point<dim>>>
+IBStencils<dim>::points(unsigned int    order,
```
blaisb, line 76: IBStencils is plural or should it be singular?

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,235 @@
...
+        }
+      if (component == 1)
+        {
+          // vy in 2D
+          v_ib = v_rot[1] + p.velocity[1];
+        }
+    }
+  if (dim == 3)
+    {
+      ;
```
blaisb, line 206:
```suggestion
      
```

source/core/ib_stencils.cc
```diff
@@ -0,0 +1,235 @@
...
+      radius[0]          = p.radius * ((dof_point - p.position) /
```
blaisb, line 207: What is radius here? Is it like a distance vector? Radius to me is a particle property no?

blaisb (commented):
Another batch of comments. Nothing too major

include/solvers/gls_sharp_navier_stokes.h
Luckabarbeau: yeah this is after the indentation function call

include/solvers/gls_sharp_navier_stokes.h
Luckabarbeau: Those comments in .cc must be from // to /* */ ?

source/core/ib_stencils.cc
Luckabarbeau: I put plural because there are many possible stencils. But in fact only one is used so  I don’t really know.

source/core/ib_stencils.cc
Luckabarbeau: it's the vector from the center of the particle to the surface close to the DOF.  should it be reference as radial vector instead

include/solvers/gls_sharp_navier_stokes.h
blaisb: Yeah it is better to use /* than //

source/core/ib_stencils.cc
blaisb: Singular then my friend. A quick refactor

source/core/ib_stencils.cc
blaisb: Radial_vector (in lower case) is a very good way to name it.

source/core/ib_stencils.cc
```diff
@@ -25,34 +25,29 @@ std::vector<double>
 IBStencils<dim>::coefficients(unsigned int order)
 {
   // Initialize the coefficient vector
-  std::vector<double> coef;
+  std::vector<double> coef(order+1);
```
blaisb, line 28: Be careful though. Here you should  not make coef order +1 if the order is more than 4 no? In this case, better to resize in if.

blaisb (commented):
Just micro comments left.  Shouldn’t take more than a few minutes.

source/core/ib_stencils.cc
Luckabarbeau: applied

source/core/ib_stencils.cc
Luckabarbeau: i put a resize only in the case of order 5 or more

blaisb (approved):

PR #169, opened 2021-06-01
Reduced cut cell evaluation

Luckabarbeau:
Reduce the number of times we need to evaluate if a cell is cut by creating a map with the key being the cell and to a bool that define if the cell is cut or not.

### DIFF
--- a/include/solvers/gls_sharp_navier_stokes.h
+++ b/include/solvers/gls_sharp_navier_stokes.h
@@ -164,8 +164,9 @@ private:
 
   /**
    * @brief
-   *This function create a map with the key being the cell and what it contains a bool to define if a cell is cut and the particle that cut it if it's the case.
-   * This map contains is referred as cutted_cells_map.
+   *This function create a map with the key being the cell and what it contains
+   *a bool to define if a cell is cut and the particle that cut it if it's the
+   *case. This map contains is referred as cutted_cells_map.
    */
   void
   cutted_cells_mapping();
@@ -259,10 +260,10 @@ private:
 private:
   std::map<unsigned int,
            std::set<typename DoFHandler<dim>::active_cell_iterator>>
-                               vertices_to_cell;
-    std::map<typename DoFHandler<dim>::active_cell_iterator,
-            std::tuple<bool,unsigned int>>
-            cutted_cells_map;
+    vertices_to_cell;
+  std::map<typename DoFHandler<dim>::active_cell_iterator,
+           std::tuple<bool, unsigned int>>
+                               cutted_cells_map;
   const bool                   SUPG        = true;
   const bool                   PSPG        = true;
   const double                 GLS_u_scale = 1;
--- a/include/solvers/gls_sharp_navier_stokes.h
+++ b/include/solvers/gls_sharp_navier_stokes.h
@@ -164,6 +164,14 @@ private:
 
   /**
    * @brief
+   *This function create a map with the key being the cell and what it contains a bool to define if a cell is cut and the particle that cut it if it's the case.
+   * This map contains is referred as cutted_cells_map.
+   */
+  void
+  cutted_cells_mapping();
+
+  /**
+   * @brief
    * Return a bool to define if a cell is cut by an IB particle and the local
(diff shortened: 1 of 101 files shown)


include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -159,6 +159,15 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
   void
   refine_ib();
 
+  /**
+   * @brief
+   *This function create a map with the key being the cell and what it contains
+   *a bool to define if a cell is cut and the particle that cut it if it's the
+   *case. This map contains is referred as cutted_cells_map.
```
blaisb, line 166:
```suggestion
   *This function create a map (cut_cells_map) that indicates if a cell is cut.
```
cut in the past is cut, not cutted

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -159,6 +159,15 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+   */
+  void
+  cutted_cells_mapping();
```
blaisb, line 169:
```suggestion
  generate_cut_cells_map();
```

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -248,7 +257,10 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
 private:
   std::map<unsigned int,
            std::set<typename DoFHandler<dim>::active_cell_iterator>>
-                               vertices_to_cell;
+    vertices_to_cell;
+  std::map<typename DoFHandler<dim>::active_cell_iterator,
+           std::tuple<bool, unsigned int>>
```
blaisb, line 262: Why is the element a tuple instead of just a bool? What is this unsigned int parameter for? Can you document it above the map?

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -66,6 +68,39 @@ GLSSharpNavierStokesSolver<dim>::vertices_cell_mapping()
...
+  const unsigned int dofs_per_cell = this->fe->dofs_per_cell;
+  std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
+
+  // // Loop on all the cells and check if they are cut.
+  for (const auto &cell : cell_iterator)
+    {
+      if (cell->is_locally_owned() || cell->is_ghost())
+        {
+          bool         cell_is_cut;
+          unsigned int p;
```
blaisb, line 93: What is this p? I don't get it. You make a tuple with it but what is it for?
Why is it often set to std::ignore? I don't get it.... Either you can remove it or you need to explain what it is for...

blaisb (commented):
The idea is good, but I really don't understand where this tuple(bool,uint) is coming from. Can you explain and document? it never seems to have a value.

source/solvers/gls_sharp_navier_stokes.cc
Luckabarbeau: I added the information in the code. p is the particle id of the particle that cut the cell. this info is only used in the sharp edge function. All other calls to this map only need to know if the cell is cut or not and doesn’t care about what particle cuts it. In these cases I used std::ignore.

blaisb: Then call it particle_id or ib_particle_id or sharp_particle_id, but not p! :)! P does not mean anything to a normal reader

blaisb (commented):
Final last comment.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -90,13 +90,14 @@ GLSSharpNavierStokesSolver<dim>::cutted_cells_mapping()
       if (cell->is_locally_owned() || cell->is_ghost())
         {
           bool         cell_is_cut;
-          unsigned int p;
+          unsigned int p = 0;
```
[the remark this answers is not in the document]
Luckabarbeau: Was not necessary to do so. Return to previous calls and simply declare p.

PR #171, opened 2021-06-02
Fix parallel void fraction bounding + grad-div stab for VANS

blaisb:
This PR fixes two things:
- the issue with the parallel bounding of the void fraction
- It adds grad-div stabilization for the vans equation to ensure mass conservation when there are shocks. Right now this is disabled (the bool is set to false) because I did not want to break all the tests. If we deem that this implementation is superior, then we can enable it.
This settles my first part of this project. @tonielgeitani work on the complete stress tensor and @shahabgol work on the void fraction with smoothing will be next :)

### DIFF
--- a/include/fem-dem/gls_vans.h
+++ b/include/fem-dem/gls_vans.h
@@ -184,6 +184,7 @@ private:
   const bool   PSPG        = true;
   const bool   SUPG        = true;
   const bool   DCDD        = true;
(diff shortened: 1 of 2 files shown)


tonielgeitani (approved):
Everything is clear for me :)

shahabgol (approved):

PR #175, opened 2021-06-07
Fix DEM tests + Remove particles warning

blaisb:
Fixes two bugs in DEM tests that would lead to assertion throw
Remove warning related to particles


shahabgol (approved):
It was a short and neat PR, I have no comments here.

PR #177, opened 2021-06-08
RPT Photon count calculation with Monte Carlo method

acdaigneault:
This branch adds the first step to an extension of Lethe for RPT. It allows calculating photon counts calculation with the Monte Carlo method.

- The application to use this feature is rpt_3d.
- A complete example is provided with all parameters needed for the calculation. (rpt_count_calculation.prm)

- There is 4 unit tests : 

1. **position_parameters** to assure position parameters are proceed correctly in any different cases related to the particle position regarding the detector position.
2. **solid_angle** verifies angles for the same reason than position_parameters.
3. **reactor_detector_path_lengths** to assure path lengths calculation are proceed correctly.
4. **count** to assure that the count is always the same with a fixed seed number for random number generator and iteration number.

- How does the code work :
The RPTParameters and RPTCalculationParameters handle declaration and parsing parameters while particles and detectors positions are assigned in the RPT class with assign_particle_positions() and assign_detector_positions() methods.
The RadioParticle and Detector classes allow to set positions and some parameters to objects prior be inserted in the ParticleDetectorInteractions class. The photon count calculation is done for one particle and one detector in there. 
Meanwhile, it is the RPT class that calls the calculate() method in the ParticleDetectorInteractions class to get the count for every particle-detector pair. This class also handles printing results in the terminal or exporting them, in .csv files if enable.


examples/rpt/count_calculation/rpt_count_calculation.prm
```diff
@@ -0,0 +1,37 @@
...
+    set reactor radius       			 = 0.1
+    set peak-to-total ratio  			 = 0.4
+    set sampling time        			 = 1
+end
+	
+# --------------------------------------------------
+# Initial value of parameters for tuning
+#---------------------------------------------------
+subsection parameter tuning
+	set tuning                           = false
```
blaisb, line 20: What is this option for?

examples/rpt/count_calculation/rpt_count_calculation.prm
```diff
@@ -0,0 +1,37 @@
...
+    set random number seed               = auto
+    set reactor radius       			 = 0.1
+    set peak-to-total ratio  			 = 0.4
+    set sampling time        			 = 1
+end
+	
+# --------------------------------------------------
+# Initial value of parameters for tuning
+#---------------------------------------------------
+subsection parameter tuning
```
blaisb, line 19: I would call this section parameters or rpt parameters or particle and detector parameters because these parameters will be  use for the dictionnary generation also. The tuning itself (if it becomes inside lethe) could go in another section :).

include/rpt/detector.h
```diff
@@ -0,0 +1,99 @@
...
+template <int dim>
+class Detector
+{
+public:
+  /**
+   * @brief Constructor for the Detector.
+   *
+   * @param detector_parameters All parameters related to the detector
+   *
+   * @param n ID number to the particle
```
blaisb, line 42:
```suggestion
   * @param n ID number of the particle
```
Is this  the id of the detector or the particle here?

include/rpt/parameters_rpt.h
```diff
@@ -0,0 +1,116 @@
...
+  struct RPTParameters
+  {
+    // Particle positions filename
+    std::string particle_positions_file;
+
+    // Enable to export counts result in a .csv file
+    bool export_counts;
+
+    // Number of Monte Carlo iterations for alpha and theta
+    unsigned int iteration_number;
```
blaisb, line 55:
```suggestion
    unsigned int n_monte_carlo_iterations
```
Use a variable name that is a bit more explicit :)

include/rpt/parameters_rpt.h
```diff
@@ -0,0 +1,116 @@
...
+
+    // Seed of the random number generator
+    int seed;
```
blaisb, line 58:
```suggestion
    unsigned int seed;
```
Do you want to allow for negative seed? (do they even make sense?)

include/rpt/parameters_rpt.h
```diff
@@ -0,0 +1,116 @@
...
+
+    // All parameters that are fixed by the user
+    double reactor_radius;
+    double peak_to_total_ratio;
+    double sampling_time;
```
blaisb, line 63: Add description of what each parameter means :)

include/rpt/particle_detector_interactions.h
```diff
@@ -0,0 +1,186 @@
...
+  get_h();
+
+  double
+  get_rho();
+
+  double
+  get_alpha(double n_alpha, double n_theta);
+
+  double
+  get_theta(double n_alpha, double n_theta);
```
blaisb, line 79: is this a getter or do these function calculate theta and alpha? In the latter case, they should not be named getters :)

include/rpt/particle_detector_interactions.h
```diff
@@ -0,0 +1,186 @@
...
+
+  double
+  get_detector_path_length(double n_alpha, double n_theta);
+
+  double
+  get_reactor_path_length(double n_alpha, double n_theta);
```
blaisb, line 85: Same as above.

source/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 MESSAGE(STATUS "Setting up Lethe library")
-FIND_PACKAGE(deal.II 9.0.0 QUIET
+FIND_PACKAGE(deal.II 9.2.0 QUIET
```
blaisb, line 2: Good catch.

source/rpt/particle_detector_interactions.cc
```diff
@@ -0,0 +1,454 @@
...
+  // Projection of the distance vector on the detector axis which is parallel to
+  // the z' axis Particle position in particle-detector coordinate frame (h =
+  // z'p)
+  h = scalar_product(face_detector_particle_distance, detector_orientation_z);
+
+  // x' coordinate of the particle-detector frame (e'x) (see equations 29 to 31)
+  Tensor<1, dim> Z = h * detector_orientation_z;
+  Tensor<1, dim> X = Z - face_detector_particle_distance;
+
+  if (X.norm() == 0) // When particle is aligned to the detector
```
blaisb, line 29: Not a good idea to compare floats/double using equation. Maybe define a threshold?

blaisb (approved):
Very good work. I can either merge the PR in it's current state and you can work on a follow up PR in the next week to fix some minor things or if you have time tomorrow, you can do the fixes now and I'll merge them in this PR.
Good work. It's a very good first version of a code.

examples/rpt/count_calculation/positions1.detector
```diff
@@ -0,0 +1,3 @@
+
+0.15 0 0.17
+0.17 0 0.17
```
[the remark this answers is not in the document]
acdaigneault: Yes, it works with many detectors, but it's less confusing with all the number in one line, I will change it.

examples/rpt/count_calculation/positions1.particle
```diff
@@ -0,0 +1,4 @@
+
```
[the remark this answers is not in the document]
acdaigneault: It needs to have a space to get the first line of position. I thought it was because of "getline" but after a quick search it shouldn't, I will look at it.

examples/rpt/count_calculation/rpt_count_calculation.prm
acdaigneault: It's useless right now, I did it prior adding the tuning part.

examples/rpt/count_calculation/rpt_count_calculation.prm
acdaigneault: Actually I did it because I thought we would tune all of those 5 parameters. If I keep tau, R, and mu_r in this section and move the other two to the section above (maybe mu_d to the detector section), is it better?

include/rpt/detector.h
acdaigneault: Typo, it's "of the detector" thanks

include/rpt/particle_detector_interactions.h
acdaigneault: Actually, those getter are for the unit tests because my function calculate_solid_angle() is private. I think I can just have "return alpha" but to have the angles (and not empty doubles) but calculate_position_parameters() & calculate_solid_angle have to be called. I can do it by hand in unit tests but I wanted to avoid someone use it but get issues because the methods were not call before. 
Should I keep it as is, change to only "return alpha/theta" or change my methods to public?

examples/rpt/count_calculation/rpt_count_calculation.prm
ghazalemir: I think moving the gama-rays emitted to rpt parameters, and mu_d to detector parameters is a good idea!

include/rpt/parameters_rpt.h
acdaigneault: I wanted to add an assert for that, but a seed number < 0 actually gave similar results. I can change it if you prefer.

examples/rpt/count_calculation/rpt_count_calculation.prm
```diff
@@ -0,0 +1,37 @@
...
+	set tuning                           = false	
+    set dead time       				 = 1e-5
+    set activity  						 = 2e6
+    set gamma-rays emitted        		 = 2
+    set attenuation coefficient reactor  = 10
```
ghazalemir, line 24: This attenuation coefficient is  related to reactor (not only reactor body) and material inside, I think reactor content is a better word but not the best representative maybe.

examples/rpt/count_calculation/rpt_count_calculation.prm
```diff
@@ -0,0 +1,37 @@
+# Listing of Parameters
```
ghazalemir, line 1: In general in this part maybe it is better to add the unit for non dimensionless parameters like radius which is in meter for example!

examples/rpt/count_calculation/rpt_count_calculation.prm
blaisb: I think it's better to keep all "tunable parameters" in the same section. Maybe just rename the section to use a better name?

examples/rpt/count_calculation/rpt_count_calculation.prm
blaisb: Good :)! Was just wondering.

include/rpt/parameters_rpt.h
blaisb: Nope that's ok. I was just wondering if seed needed ot be positive or not.
If they don't need to be, no worries.

include/rpt/particle_detector_interactions.h
blaisb: Nope that's good. I was just wondering if it was a getter or not.
You can maybe add a comment above those to say : "Getter used for unit testing purposes"
It's a good design what you did :).

examples/rpt/count_calculation/rpt_count_calculation.prm
blaisb: Yeah but there is no way to have units in the PRM file. I mean, we cannot enforce it. We will need to document it that the units have to be consistent (in SI).

examples/rpt/count_calculation/rpt_count_calculation.prm
blaisb: Let's lump all "RPT ish" parameters together for the time being. This will make life simpler for now.

blaisb (commented):
Nothing major to fix. Maybe we can merge as is and then you can do another PR when you want to reshuffle the parameters around? These are things that @ghazalemir and you could try to fully agree upon before? What do you think?

examples/rpt/count_calculation/positions1.particle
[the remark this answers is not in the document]
acdaigneault: Done :)

examples/rpt/count_calculation/rpt_count_calculation.prm
acdaigneault: I did explain a bit more as comments in the code itself, but good point, it will be important to explain it also in documentation.

examples/rpt/count_calculation/rpt_count_calculation.prm
acdaigneault: I will in the documentation and/or the example of the wiki

include/rpt/particle_detector_interactions.h
```diff
@@ -0,0 +1,187 @@
...
+
```
ghazalemir, line 86: Thanks for clear and brief annotation!

acdaigneault:
I did minor modifications and changed related comments as resolved. We should discuss about the remaining comments to agree about it, but I will change it in the next branch, is that ok?

source/rpt/particle_detector_interactions.cc
```diff
@@ -0,0 +1,452 @@
...
+
+  // Detector orientation vector parallel to z' in particle-detector coordinate
+  // (e'z)
+  Tensor<1, dim> face_middle_vector =
+    detector_face_position - detector_middle_position;
+  detector_orientation_z = face_middle_vector / face_middle_vector.norm();
+
+  // Projection of the distance vector on the detector axis which is parallel to
+  // the z' axis Particle position in particle-detector coordinate frame (h =
+  // z'p)
```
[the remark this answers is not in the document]
acdaigneault: Your right, it's correct in the code, but not in the comment :)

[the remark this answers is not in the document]
acdaigneault: h = z'p = Xdp e'z no?

PR #179, opened 2021-06-10
Add refinement connections

blaisb:
Right now the particle_handler is only connected to the partition signal of the triangulation, it should also be connected to the refinement signals. This PR fixes that. It is not relevant for our current version of deal.II, but will be for future versions.


shahabgol (approved):
I have no comments.

blaisb:
> I have no comments.

Is that not a comment? :p !

PR #181, opened 2021-06-10
Use vector rather than unordered map for force/momentum/disp

kronbichler:
Following on some of the discussion I had with @blaisb, I started playing with some of the data structures. As far as I can see, the access to force, momemtum, MOI, displacement is always through the `particle->get_id()` number, so I would intuitively say it is best to use a vector. More precisely, it seems that the index range is limited and we actually fill out all of it (except for holes due to ghosts), so a vector should be the best data type.

This PR is work in progress, because there are a few tests failing and I have not checked carefully. Before I go on, I would like to discuss with you about the general direction or whether there is some other ideas. Note that I did not check that particle point contact candidates and friends yet, where a vector seems less obvious (but I did not check closely yet).

Besides correctness, there is also some cleanup that I still need to do: For example, I now misused the function `update_moment_of_inertia` to resize the vectors (as I do only want to check the largest id once, and I need to step in here), but it really should be in a more descriptively named function or somewhere else. I can think about it more if nobody else has an immediate idea.

As to why I do it: I don't have precise timings, but the assembly code improved quite a bit in the various force functions. For example, running the `dem_3d/rotation_drum.prmg` test case with 2 MPI ranks I see 774b instructions with master and 734b instructions with my patch. Also in other functions involving these containers I see 5-20% improvements.


blaisb:
Very good change @kronbichler. Thank you :). could you rebase so that I can launch the CI? our github actions was just updated to allow running for forks

source/dem/dem.cc
```diff
@@ -352,18 +352,26 @@ template <int dim>
...
+  unsigned int max_particle_id = 0;
+  for (const auto &particle : particle_handler)
+    max_particle_id = std::max(max_particle_id, particle.get_id());
+
+  // We resize force and momentum every time we sort the particles
+  // into subdomains
+  momentum.resize(max_particle_id + 1);
+  force.resize(max_particle_id + 1);
+  displacement.resize(max_particle_id + 1);
+  MOI.resize(max_particle_id + 1);
```
blaisb, line 366: Is this not dangerous in parallel? For example, if a processor has particle id 8M, won't that reserve an 8M vector?

source/dem/find_contact_detection_step.cc
```diff
@@ -4,31 +4,26 @@ using namespace dealii;
...
+  double       max_displacement       = 0;
+  unsigned int contact_detection_step = 0;
 
   // Updating displacement
   for (auto &particle : particle_handler)
     {
-      auto &particle_properties   = particle.get_properties();
+      auto &particle_properties = particle.get_properties();
+      if (displacement.size() <= particle.get_id())
+        displacement.resize(particle.get_id() + 1);
```
blaisb, line 26: Again, can't that pause a problem when id is large? Maybe there is something I don't understand, but isnt there an issue with sizing vectors with particle id?

blaisb (commented):
I think this is a very good change, but there are two questions I would like to address (see below) regarding sizing of the vectors.
Thanks @kronbichler for your contribution

source/dem/dem.cc
kronbichler: You are right - I misread the deal.II documentation, I thought they were local to each MPI process. But the solution is to make the accessor classes in deal.II aware of an MPI-local index - it does exist because this is what is used for the property pool and other access information inside deal.II, so it is just a matter of finding it (or exposing it in deal.II's classes). Let me have a look at it later today or tomorrow.

kronbichler: I had a short look now: I think we would like to get the information inside `get_handle()` as a way to have indexing in MPI-local index space, https://github.com/dealii/dealii/blob/b4e88620cab2396f6daf8d36c78553ecc6a3301d/include/deal.II/particles/particle_accessor.h#L327-L328. We don't want to make that function public, but I think it would make a lot of sense to add a function `get_local_index()` (or `get_local_id()`) to `ParticleAccessor` that returns the number in MPI-local index space. I don't know if we want to make the numbers a tight map in the sense that we keep the numbers contiguous or if we simply use the number that is used inside the property pool. For our purposes, the latter would suffice, as we update it whenever the particles move around. Before going more, two short questions:
- In order to satisfy backward compatibility with Lethe, we will need to make this conditional and have a translation layer for the older version. Any other comments on this topic?
- Do we need to carry over the forces, momentum, etc. when the particle assignments to cells change? I don't think so, because we have https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/dem.cc#L749 and similar statements for the rest. This would mean we do not need to be particularly careful in keeping the numbering alive from Lethe's end, and we could take whatever numbers deal.II gives us.

With the closer look I realized what confused me to make this wrong assumption, it is the fact that you use `unsigned int particle_one_id = ...` e.g. in https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/pp_nonlinear_force.cc#L188 which I interpreted as the local index, as it differs from `types::particle_index` in the context I use to work with (big parallel machines with 64 bit indices). In any case, this suggestion was also among the optimizations we introduced when writing the `LinearAlgebra::distributed::Vector` class in deal.II: When we work with some array structures that are located on the current MPI process, we really do not want to go through the index translation from the global space and then somehow back again, but we should immediately use the local one. Of course, the global index is the more convenient one because it helps us to synchronize the view of different MPI processes on the index, but the internal data structures should be designed differently.

blaisb: Please find my comments below

> I had a short look now: I think we would like to get the information inside `get_handle()` as a way to have indexing in MPI-local index space, https://github.com/dealii/dealii/blob/b4e88620cab2396f6daf8d36c78553ecc6a3301d/include/deal.II/particles/particle_accessor.h#L327-L328. We don't want to make that function public, but I think it would make a lot of sense to add a function `get_local_index()` (or `get_local_id()`) to `ParticleAccessor` that returns the number in MPI-local index space. I don't know if we want to make the numbers a tight map in the sense that we keep the numbers contiguous or if we simply use the number that is used inside the property pool. For our purposes, the latter would suffice, as we update it whenever the particles move around. Before going more, two short questions:
> 
> * In order to satisfy backward compatibility with Lethe, we will need to make this conditional and have a translation layer for the older version. Any other comments on this topic?

You mean to ensure backward compatibility with deal.II 9.3 right? In this case yeah, we would need to have a translation layer. I don't like it very much when there are too much #ifdef lying around or when the behavior changes drastically from 9.3 to 10.0pre, but to be honest, there is no way around this. We had the same thing happen to use when we introduced update_ghost_particles. The deal.II particle library is still immature, this is bound to happen.

> * Do we need to carry over the forces, momentum, etc. when the particle assignments to cells change? I don't think so, because we have https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/dem.cc#L749
>    and similar statements for the rest. This would mean we do not need to be particularly careful in keeping the numbering alive from Lethe's end, and we could take whatever numbers deal.II gives us.
We do not. Maybe @shahabgol can comment on this, but as far as I know, the only thing we need to keep alive and coherent in a contact in pretty much the tangential overlap  (because it carries over in a contact history) and this is kept in a seperate structure which is defined on a id-id basis, the force and the other contact properties do not need to be ept.

> 
> With the closer look I realized what confused me to make this wrong assumption, it is the fact that you use `unsigned int particle_one_id = ...` e.g. in
> 
> https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/pp_nonlinear_force.cc#L188
> 
> which I interpreted as the local index, as it differs from `types::particle_index` in the context I use to work with (big parallel machines with 64 bit indices). In any case, this suggestion was also among the optimizations we introduced when writing the `LinearAlgebra::distributed::Vector` class in deal.II: When we work with some array structures that are located on the current MPI process, we really do not want to go through the index translation from the global space and then somehow back again, but we should immediately use the local one. Of course, the global index is the more convenient one because it helps us to synchronize the view of different MPI processes on the index, but the internal data structures should be designed differently.

Using unsigned int is a mistake on our part, types::particle_index should have been used from the get go. I fully agree with you that having a local id as a 32bit index would be extremely relevant.  I don't think we need to keep that number contiguous, but having some sort of way indexing the particles "Locally" instead of globally would actually allow us to simplify a lot of the stuff happening because there would be a reliable way to create a vector-like container for the local particles.

kronbichler: Sorry that I can only come back to this one now, but I think we should progress now. I will open a PR with a suitable function in deal.II now and then we can see how we can integrate the new functionality in a backward-compatible way.

blaisb: Absolutely. In the worst case, we can make the 9.3 version "backward compatible" without being optimal and provide the correct documentation.

kronbichler:
I rebased and updated the PR to the current deal.II master (and checked with the 9.3 release as well). Let's see what the CI says regarding possible bugs.

blaisb:
> I rebased and updated the PR to the current deal.II master (and checked with the 9.3 release as well). Let's see what the CI says regarding possible bugs.

It seems like the CI is loving your PR.
I'm leaving for the weekend, but it's the next item on my todo for Sunday :). Thank you so much for taking the time.

source/dem/dem.cc
shahabgol: > Please find my comments below
> 
> > I had a short look now: I think we would like to get the information inside `get_handle()` as a way to have indexing in MPI-local index space, https://github.com/dealii/dealii/blob/b4e88620cab2396f6daf8d36c78553ecc6a3301d/include/deal.II/particles/particle_accessor.h#L327-L328. We don't want to make that function public, but I think it would make a lot of sense to add a function `get_local_index()` (or `get_local_id()`) to `ParticleAccessor` that returns the number in MPI-local index space. I don't know if we want to make the numbers a tight map in the sense that we keep the numbers contiguous or if we simply use the number that is used inside the property pool. For our purposes, the latter would suffice, as we update it whenever the particles move around. Before going more, two short questions:
> > 
> > * In order to satisfy backward compatibility with Lethe, we will need to make this conditional and have a translation layer for the older version. Any other comments on this topic?
> 
> You mean to ensure backward compatibility with deal.II 9.3 right? In this case yeah, we would need to have a translation layer. I don't like it very much when there are too much #ifdef lying around or when the behavior changes drastically from 9.3 to 10.0pre, but to be honest, there is no way around this. We had the same thing happen to use when we introduced update_ghost_particles. The deal.II particle library is still immature, this is bound to happen.
> 
> > * Do we need to carry over the forces, momentum, etc. when the particle assignments to cells change? I don't think so, because we have https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/dem.cc#L749
> >   
> >   and similar statements for the rest. This would mean we do not need to be particularly careful in keeping the numbering alive from Lethe's end, and we could take whatever numbers deal.II gives us.
> >   We do not. Maybe @shahabgol can comment on this, but as far as I know, the only thing we need to keep alive and coherent in a contact in pretty much the tangential overlap  (because it carries over in a contact history) and this is kept in a seperate structure which is defined on a id-id basis, the force and the other contact properties do not need to be ept.
> 
> > With the closer look I realized what confused me to make this wrong assumption, it is the fact that you use `unsigned int particle_one_id = ...` e.g. in
> > https://github.com/lethe-cfd/lethe/blob/59b11a72dfde82d9d1d48b646bf9dfff4c517e93/source/dem/pp_nonlinear_force.cc#L188
> > 
> > which I interpreted as the local index, as it differs from `types::particle_index` in the context I use to work with (big parallel machines with 64 bit indices). In any case, this suggestion was also among the optimizations we introduced when writing the `LinearAlgebra::distributed::Vector` class in deal.II: When we work with some array structures that are located on the current MPI process, we really do not want to go through the index translation from the global space and then somehow back again, but we should immediately use the local one. Of course, the global index is the more convenient one because it helps us to synchronize the view of different MPI processes on the index, but the internal data structures should be designed differently.
> 
> Using unsigned int is a mistake on our part, types::particle_index should have been used from the get go. I fully agree with you that having a local id as a 32bit index would be extremely relevant. I don't think we need to keep that number contiguous, but having some sort of way indexing the particles "Locally" instead of globally would actually allow us to simplify a lot of the stuff happening because there would be a reliable way to create a vector-like container for the local particles.

We need the force and momentum in the integration to obtain the acceleration, velocity, position, and angular velocities on the particles. Plus, we need them in the visualization. Other than that, we don't need force and momentum in the next time-step.

source/dem/find_contact_detection_step.cc
```diff
@@ -4,32 +4,29 @@ using namespace dealii;
...
+  double       max_displacement       = 0;
+  unsigned int contact_detection_step = 0;
 
   // Updating displacement
   for (auto &particle : particle_handler)
     {
-      auto &particle_properties   = particle.get_properties();
+      auto &particle_properties = particle.get_properties();
+#if DEAL_II_VERSION_GTE(10, 0, 0)
+      auto &particle_displacement = displacement[particle.get_local_index()];
```
shahabgol, line 26: This is the main reason I've used unordered_map instead of vector. When the container size becomes large, finding an element is faster in maps. Do you think using vectors is a suitable choice?

shahabgol (commented):
I only have one comment. I chose unordered_map, because finding an element in the maps (if the container size is large) is faster. In the DEM, we are generally dealing with 1-10 million particles. So I guess finding elements in such a large container would be time-consuming.
Other than this, thank you VERY much @kronbichler for your help. I learned a lot by reading your PR. :)

blaisb (approved):
To me this looks real good. Let's just wait for the results from @shahabgol benchmark on Cedar to see how much of an impact this has for more particles and larger cases :)

blaisb:
@kronbichler It seems like our CI is github actions setup is still struggling with PRs from fork. The pull_request flag is actually the master branch checkup.

The following tests break:
	  1 - dem_2d/circle_restart.release (Failed)
	 24 - dem_3d/sliding_restart.release (Failed)
	115 - dem/integration_euler.release (Failed)
	117 - dem/integration_schemes_accuracy.release (Failed)
	118 - dem/integration_velocity_verlet.release (Failed)
	119 - dem/normal_force.release (Failed)
	120 - dem/particle_particle_contact_force_linear.release (Failed)
	121 - dem/particle_particle_contact_force_nonlinear.release (Failed)
	122 - dem/particle_particle_contact_on_two_processors.mpirun=2.release (Failed)
	124 - dem/particle_point_contact.release (Failed)
	125 - dem/particle_wall_contact_force_linear.release (Failed)
	126 - dem/particle_wall_contact_force_nonlinear.release (Failed)
	129 - dem/post_collision_velocity.release (Failed)
	130 - dem/two_particles_multiple_contacts_parallel.mpirun=2.release (Failed)


It might be a lot of painful work for you to fix those on your own. I think there are two choices:
1- You can create a branch directly in Lethe and cherry pick your two commits. This way I could work on it directly without needing to fork your branch. @shahabgol will also be able to help.
2- I can try to help you directly in your branch by making a PR agaisnt it.

Sorry for the CI issue, it's still something I'm trying to get a better grasp off and most people contributing in Lethe have not done it from forks :).

blaisb:
@kronbichler By the way. I saw very good speed-up due to those changes in cases which are CPU-bound (ex: 123s to 93s for the same case). These might be edge cases, because others had little speedup (5-10% max), but this is an amazing move in the right direction. @shahabgol and I will help you with fixing the test and moving on to a quick merge.

source/dem/find_contact_detection_step.cc
blaisb: There is no use of std::find() here so there is no searching. Only the [] operator is used. Consequently, the cost here should be lower and lower as the number of particles decreases. The reason why we used an unordered map in the past was that we did not have any way to number the local particles, hence we had to access them through the ID which is not a continuous numbering scheme.

kronbichler:
I should mention that the code for the deal-9.3 release wastes memory because I allocate the large vectors (as done in my original PR). For deal.II master I only allocate the necessary space, apart from the usual gaps in the particle storage. Thus, we need to determine whether that degradation for deal-9.3 is acceptable. Doing that more efficiently requires additional code because we need to carry along a translation layer between the function calls.

kronbichler:
I looked into one of the tests and indeed it will be some work because we need to adjust the interfaces of the tests and appropriately resize things. It is certainly not too much work, but I guess we can identify the best way forward together.

blaisb:
> I looked into one of the tests and indeed it will be some work because we need to adjust the interfaces of the tests and appropriately resize things. It is certainly not too much work, but I guess we can identify the best way forward together.

@kronbichler Let me take care of the tests. You seem to have fixed the restart one, I can take care of the rest of them.
As for the deal.II 9.3 performance degradation, I don't think it's a major issue. We try to use the master version and I can deal with this inconsistency for a year between 9.3 and 10.0

blaisb:
@kronbichler The last commits fix all the tests. I'll check if everything runs well in deal.II 9.3 and then I'll wait for @shahabgol results on the multicore benchmark. Most likely will merge tomorrow morning

kronbichler:
@blaisb thanks for fixing the tests - I think this is worth it because it helps with nicer data structures. The speedups you're reporting are pretty good, and match with what I saw on one of your test cases (3d_rotating_drum run on my laptop). As you mentioned in one of the top posts, we might also give the data structure for the contact pairs another thought. I'd be happy to take a closer look at those parts to see if we have alternatives there - I'm mostly thinking of the `DynamicSparsityPattern` class we have in deal.II.

blaisb:
> @blaisb thanks for fixing the tests - I think this is worth it because it helps with nicer data structures. The speedups you're reporting are pretty good, and match with what I saw on one of your test cases (3d_rotating_drum run on my laptop). As you mentioned in one of the top posts, we might also give the data structure for the contact pairs another thought. I'd be happy to take a closer look at those parts to see if we have alternatives there - I'm mostly thinking of the `DynamicSparsityPattern` class we have in deal.II.

@kronbichler I think that even with a 0% speed-up I would consider this PR really great, because it leads to much cleaner data structure. However, I am sure it provides increased performance as the number of particles/core increases, since this is where the unordered map data structure became a poor choice. :) Greatly looking forward to your contributions.

blaisb:
I confirm that this passes all test using deal.II 9.3. @shahabgol I'll be waiting for your benchmark results then I merge :).

PR #190, opened 2021-06-23
Add module that calculates forces and torques due by particle/wall collision over the boundary

EduardsBlandin:


include/dem/dem.h
```diff
@@ -263,6 +263,9 @@ class DEMSolver
   void
   write_output_results();
 
+  void
```
blaisb, line 266: Add documentation above the function explaining what it does.

source/core/parameters_lagrangian.cc
```diff
@@ -598,6 +598,74 @@ namespace Parameters
...
+      calculate_force_torque = prm.get_bool("calculation");
+      const std::string display = prm.get("Output mode");
+      if (display == "none")
+        force_torque_display_method = ForcesAndTorquesDisplay::none;
+      else if (display == "terminal")
+        force_torque_display_method = ForcesAndTorquesDisplay::terminal;
+      else if (display == "file")
+        force_torque_display_method = ForcesAndTorquesDisplay::file;
+      else if (display == "both")
+        force_torque_display_method = ForcesAndTorquesDisplay::both;
```
blaisb, line 653: Adapt with comment above

source/dem/dem.cc
```diff
@@ -671,6 +671,44 @@ DEMSolver<dim>::write_output_results()
...
 }
 
+template <int dim>
+void
+DEMSolver<dim>::write_forces_torques_output_results()
+{
+  std::map<unsigned int, Tensor<1, dim>> force_on_walls;
+  std::map<unsigned int, Tensor<1, dim>> torque_on_walls;
+  force_on_walls=pw_contact_force_object->get_force();
+  torque_on_walls=pw_contact_force_object->get_torque();
```
blaisb, line 681:
```suggestion
  std::map<unsigned int, Tensor<1, dim>> force_on_walls  = pw_contact_force_object->get_force();
  std::map<unsigned int, Tensor<1, dim>> torque_on_walls = pw_contact_force_object->get_torque();
```

source/dem/dem.cc
```diff
@@ -671,6 +671,44 @@ DEMSolver<dim>::write_output_results()
...
+  if ( (parameters.forces_torques.force_torque_display_method ==
+      Parameters::Lagrangian::ForceTorqueOnWall<
+        dim>::ForcesAndTorquesDisplay::terminal) ||
+      (parameters.forces_torques.force_torque_display_method ==
+      Parameters::Lagrangian::ForceTorqueOnWall<
+        dim>::ForcesAndTorquesDisplay::both) )
```
blaisb, line 687: This would be changed by the proposed modification above. Always assume you are writing to file.

source/dem/dem.cc
```diff
@@ -671,6 +671,44 @@ DEMSolver<dim>::write_output_results()
...
+    {
+      for (const auto &it : force_on_walls)
+      {
+        pcout     << "Boundary " << it.first << " :\n"
+                  << "F = " << it.second << "\nM = " << torque_on_walls[it.first]
+                  << "\n\n";
```
blaisb, line 693: F should be Force
M should be Torque

source/dem/dem.cc
```diff
@@ -671,6 +671,44 @@ DEMSolver<dim>::write_output_results()
...
+    {
+      std::string name_output_file = parameters.forces_torques.force_torque_output_name;
+
+    }
+}
+
+
+
+
+
```
blaisb, line 711:
```suggestion
```

blaisb (commented):
Good! Two things are missing:
- You are missing an application test
- I did not see where the code outputs the force and torque to a file. To output it to a file, I would use the TableHandler object of deal.II. You can look at the NavierStokesBase class to see how it can be used
-  Important : Your code does not take into account the parallel computing aspect. When you sum the force acting on the boundaries, you are only summing them on a single core. But what happens if a boundary is spread onto multiple cores? You need to do an MPI sum over all the cores for all torques and forces applied on the boundary. This is done for example in the L2 error calculation or in the dynamic contact detection if  you want examples.

PR #193, opened 2021-06-27
Modify DEM application tests

shahabgol:
Modify two DEM application tests to reproduce a recent bug.


applications_tests/dem_3d/CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_LOWER_BUILD_TYPE)
 DEAL_II_PICKUP_TESTS()
 
 if (CMAKE_BUILD_TYPE STREQUAL "Debug" )
-       	set_tests_properties(dem_3d/packing_in_box.debug PROPERTIES TIMEOUT 3000) 
+       	set_tests_properties(dem_3d/packing_in_box.mpirun=1.debug PROPERTIES TIMEOUT 3000) 
+       	set_tests_properties(dem_3d/packing_in_box.mpirun=2.debug PROPERTIES TIMEOUT 3000)
```
blaisb, line 18: Why does the test need to be more than 3000s? The bug could be produced with a single particle, shouldnt the test be runnable in less than 300s?

blaisb (commented):
How long are the new tests? Try to avoid introducing tests that take more than 100-200s in debug mode.

PR #195, opened 2021-06-29
Ib pressure and hanging nodes

Luckabarbeau:
This pull request make possible the simulation of sharp interface immersed boundaries that cross cells that have hanging nodes in Q1Q1 element. The modification made the modified clear line function useless. This function was removed. This PR also includes a small improvement on how IB are imposed. We now check if the IB condition was already imposed on a DOF from the calculations in another cell.

### DIFF
--- a/source/solvers/gls_sharp_navier_stokes.cc
+++ b/source/solvers/gls_sharp_navier_stokes.cc
@@ -2021,7 +2021,7 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
                           double entries=local_entrie[j].second;
 
                           interpolation+=this->evaluation_point(col)*entries;
-                          this->system_matrix.set(local_dof_indices[i],col,-entries*sum_line);
+                          this->system_matrix.set(local_dof_indices[i],col,entries*sum_line);
 
                       }
 
--- a/source/solvers/gls_sharp_navier_stokes.cc
+++ b/source/solvers/gls_sharp_navier_stokes.cc
@@ -90,6 +90,7 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
   const unsigned int dofs_per_cell = this->fe->dofs_per_cell;
   std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
 
+
   // // Loop on all the cells and check if they are cut.
   for (const auto &cell : cell_iterator)
     {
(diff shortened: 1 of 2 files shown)


source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -90,6 +90,7 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
   const unsigned int dofs_per_cell = this->fe->dofs_per_cell;
   std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
 
+
```
blaisb, line 93:
```suggestion
```

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -150,7 +151,7 @@ GLSSharpNavierStokesSolver<dim>::point_inside_cell(
         this->mapping->transform_real_to_unit_cell(cell, point);
       const double dist = GeometryInfo<dim>::distance_to_unit_cell(p_cell);
       // if the cell contains the point, the distance is equal to 0
-      if (dist == 0)
+      if (dist <= 1e-12)
```
blaisb, line 154: Can you change that so that the tolerance is a multiple of the cell size? Like 1e-6 * cell_size or something?

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1801,6 +1750,7 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
     dt = 1;
 
 
+
```
blaisb, line 1753:
```suggestion
```

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1824,7 +1774,25 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
           sum_line = sum_line / dt;
           // Clear the line in the matrix
           unsigned int inside_index = local_dof_indices[dim];
-          clear_line_in_matrix(cell, inside_index);
+          // Check on which DOF of the cell to impose the pressure. If the dof
+          // is on a hanging node, the pressure cannot be imposed there. So we
+          // just go to the next pressure DOF of the cell.
```
blaisb, line 1779:
```suggestion
          // Check on which DOF of the cell to impose the pressure. If the dof
          // is on a hanging node, it is already constrained and
          // the pressure cannot be imposed there. So we just go to the next pressure DOF of the cell.
```

blaisb (commented):
Good, Can you do the small suggested changes and I'll merge?

source/solvers/gls_sharp_navier_stokes.cc
Luckabarbeau: it's on the unite cell so it's always one

PR #196, opened 2021-07-02
Add module to calculate forces and torques applied over the boundary by the particles

EduardsBlandin:


blaisb:
Why do so many tests fail now?

blaisb:
@shahabgol I'll let you review and I will review afterwards

shahabgol:
> @shahabgol I'll let you review and I will review afterwards

ok, I'll review the PR today.

include/dem/pw_contact_force.h
```diff
@@ -129,6 +147,26 @@ class PWContactForce
...
+   * with the member class boundary index which has the keys as information
+   */
+  std::map<unsigned int, Tensor<1, dim>>
+  initialize();
+
+  /** This function sums all the forces and torques from all the
+   * MPI processes
+   */
+  void
+  mpi_correction();
```
shahabgol, line 168: Do you think this is a decent name for this function? Are we correcting something here?

source/core/parameters_lagrangian.cc
```diff
@@ -598,6 +598,70 @@ namespace Parameters
...
+      else if (verbose == "verbose")
+        force_torque_verbosity = Verbosity::verbose;
+      else
+        {
+          throw(std::runtime_error("Invalid verbosity choice "));
+        }
+      force_torque_output_name =
+        prm.get("Output name of force and torque's file (string)");
+      output_frequency =
+        prm.get_integer("Which output's creation frequency? (int)");
```
shahabgol, line 657: I think it is better to reword the sentence within the quotation marks.

source/dem/dem.cc
```diff
@@ -449,6 +451,11 @@ DEMSolver<dim>::particle_wall_contact_force()
   pw_contact_force_object->calculate_pw_contact_force(
     pw_pairs_in_contact, simulation_control->get_time_step(), momentum, force);
 
+  forces_boundary_information[simulation_control->get_step_number()] =
```
shahabgol, line 454: Don't we need to call these functions only if the calculate_force_torque parameter is true in the parameter handler file?

source/dem/dem.cc
```diff
@@ -499,6 +506,56 @@ DEMSolver<dim>::finish_simulation()
             }
         }
     }
+
```
shahabgol, line 509: I think it is better to move all these new functions in the dem.cc to separate files.

source/dem/pw_linear_force.cc
```diff
@@ -94,6 +95,13 @@ PWLinearForce<dim>::PWLinearForce(
       calculate_rolling_resistance_torque =
         &PWLinearForce<dim>::viscous_resistance;
     }
+
+  this->calculate_force_torque_on_boundary =
```
shahabgol, line 99: Again shouldn't we call this function only when it is required? The same comment for the nonlinear force.

source/dem/pw_contact_force.cc
```diff
@@ -97,6 +97,57 @@ PWContactForce<dim>::update_contact_information(
...
+  const Point<dim>   point_contact)
+{
+  if (calculate_force_torque_on_boundary == true)
+    {
+      force_on_walls[boundary_id] = force_on_walls[boundary_id] - add_force;
+      if (dim == 2)
+        {
+          Tensor<1, dim> r = point_contact - center_mass_container;
+          torque_on_walls[boundary_id][2] =
+            torque_on_walls[boundary_id][2] -
```
shahabgol, line 114: I think torque is meaningless in two-dimensional simulations. What do you think @blaisb ?

shahabgol (commented):
Some comments in the architecture of the added changes. First of all, the functions corresponding to force calculations on boundaries should only be called when they are needed (in the parameter handler file). Furthermore, I'm not sure whether we need torque calculations in two-dimensional simulations.

source/dem/pw_contact_force.cc
blaisb: Well torque in 2D simulation is just the torque acting on the third axis, so it had a meaning even if it's a bit of a trivial meaning.

source/core/parameters_lagrangian.cc
blaisb: Avoid question marks and etc in the text file also. You can look in the wiki for other examples

include/dem/pw_contact_force.h
```diff
@@ -102,7 +114,9 @@ class PWContactForce
                                           Tensor<1, dim>,
                                           Tensor<1, dim>> &forces_and_torques,
                          Tensor<1, dim> &                  particle_momentum,
-                         Tensor<1, dim> &                  particle_force)
+                         Tensor<1, dim> &                  particle_force,
+                         Point<dim> &point_on_boundary = 0,
```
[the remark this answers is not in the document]
EduardsBlandin: These two information are known by the derived class, but not by the mother class, so i needed to pass them as arguments. They are optional, so no CI problem. It is used to call calculate_force_torque_on_boundary_function that we gonna change for the name.

include/dem/pw_contact_force.h
```diff
@@ -115,6 +129,10 @@ class PWContactForce
     // Calculation of total force
     Tensor<1, dim> total_force = normal_force + tangential_force;
 
+    calculate_force_torque_on_boundary_function(boundary_id,
```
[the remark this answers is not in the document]
EduardsBlandin: The way i thought was to put the if inside each function that is optional.

include/dem/pw_contact_force.h
EduardsBlandin: We are correcting a fake result that comes from mpi operations, so i thought that mpi_correction was a good idea but maybe too large in signification. @shahabgol , do you have a better idea like mpi_correction as suffix + force_torque_member ?

source/core/parameters_lagrangian.cc
EduardsBlandin: force_torque_output_name =
        prm.get("Name file");
      output_frequency =
        prm.get_integer("Output frequency");
Better?

source/dem/dem.cc
EduardsBlandin: Is that necessary, like the member class would stay at zero tensor, yeah a big map is gonna be created of zero tensor. Would it be  significant?

source/dem/dem.cc
EduardsBlandin: Like i create a a new dem function that i insert in the finish_simulation function. Or i insert it after the finish_simulation in the solve function?

source/dem/pw_linear_force.cc
EduardsBlandin: These are member class and do not call any function. Do you mean that they should not be initialized if calculate_force_and_torque boolean is false?

source/dem/dem.cc
EduardsBlandin: I still don't catch it, which new functions? And your comment point to dem.cc file so these function are already in ...

shahabgol: no look at input_parameter_inspection function for example. It is defined in a separate file (input_parameter_inspection.cc and .h) and not inside the dem.cc

shahabgol:
I think we can merge this PR now.

applications_tests/dem_3d/force_boundary_box.prm
```diff
@@ -0,0 +1,88 @@
...
+# Mesh
+#---------------------------------------------------
+subsection mesh
+    set type                 				            = dealii
+    set grid type      	     				            = hyper_cube
+    set grid arguments       				            = 0:0.1:true
+    set initial refinement   				            = 4
+end
+
+subsection Forces and Torques
```
blaisb, line 80: We put subsection names in lower case letter and not capital
```suggestion
subsection forces and torques
```

applications_tests/dem_3d/force_boundary_box.prm
```diff
@@ -0,0 +1,88 @@
...
+    set calculation               	                                            = true
+    set verbosity                                                               = verbose
+    set Name file                                            	     		    = force_and_torque
```
blaisb, line 83:
```suggestion
    set filename                                            	     		    = force_and_torque
```

applications_tests/dem_3d/force_boundary_box.prm
blaisb: Additionally maybe a title like particles forces might be a better name. We always inherently assume that forces and torque go under the same section.

applications_tests/dem_3d/force_boundary_box.prm
```diff
@@ -0,0 +1,88 @@
...
+    set Output frequency                                                        = 2000
+    set Coordinate x of center of mass                                          = 0.05
+    set Coordinate y of center of mass                                          = 0.05
+    set Coordinate z of center of mass                                          = 0
```
blaisb, line 87: Remove capital letters
There is already a center of rotation subsection, you could re-use it
https://github.com/lethe-cfd/lethe/wiki/Understanding-the-Parameter-file-(.prm)#boundary-conditions---cfd
Or at least, make a subsection like center of mass  with x y z therein

applications_tests/dem_3d/torque_boundary_box.output
```diff
@@ -0,0 +1,43 @@
...
+Force = 0 0 0
+Moment = 0 0 0
+
+Boundary 3 :
+Force = 0 0 0
+Moment = 0 0 0
+
+Boundary 4 :
+Force = 0 0 -0.01309
+Moment = 0 -0.0005318 0
```
blaisb, line 27: I am really not a fan of this output format, but this can be modified later.
I would have imagined something like
Boundary_id  f_x f_y f_z t_x t_y t_z
0 x x x x x
1 x x x x x

etc.

include/core/parameters_lagrangian.h
```diff
@@ -231,6 +231,39 @@ namespace Parameters
...
+     */
+    template <int dim>
+    class ForceTorqueOnWall
+    {
+    public:
+      // Enable force post-processing
+      bool calculate_force_torque;
+
+      // Choosing how the outputs is gonna be displayed
+      enum class Verbosity
```
blaisb, line 246: The Verbosity enum already exists in parameter. Why recode it?

include/dem/pw_contact_force.h
```diff
@@ -157,7 +195,12 @@ class PWContactForce
   std::map<types::particle_index, double> effective_coefficient_of_restitution;
   std::map<types::particle_index, double> effective_coefficient_of_friction;
   std::map<types::particle_index, double>
-    effective_coefficient_of_rolling_friction;
+                                         effective_coefficient_of_rolling_friction;
+  std::map<unsigned int, Tensor<1, dim>> force_on_walls;
+  std::map<unsigned int, Tensor<1, dim>> torque_on_walls;
+  bool                                   calculate_force_torque_on_boundary;
+  Point<dim>                             center_mass_container;
+  std::vector<types::boundary_id>        boundary_index;
```
blaisb, line 203: Good!

include/dem/pw_linear_force.h
```diff
@@ -60,9 +60,10 @@ class PWLinearForce : public PWContactForce<dim>
                                                    boundary_translational_velocity,
     const std::unordered_map<unsigned int, double> boundary_rotational_speed,
     const std::unordered_map<unsigned int, Tensor<1, dim>>
-                                    boundary_rotational_vector,
-    const double                    triangulation_radius,
-    const DEMSolverParameters<dim> &dem_parameters);
+                                          boundary_rotational_vector,
+    const double                          triangulation_radius,
+    const DEMSolverParameters<dim> &      dem_parameters,
+    const std::vector<types::boundary_id> boundary_index = {0});
```
blaisb, line 66: Why does the boundary_index need to have a default value. Shouldnt the default vector be  an empty vector? It's not clear to me.

source/dem/dem.cc
blaisb: Has this been addressed?

source/dem/dem.cc
```diff
@@ -500,6 +510,56 @@ DEMSolver<dim>::finish_simulation()
...
+                  table.add_value("Force y",
+                                  forces_boundary_information[j][i][1]);
+                  table.add_value("Force z",
+                                  forces_boundary_information[j][i][2]);
+                  table.add_value("Moment x",
+                                  torques_boundary_information[j][i][0]);
+                  table.add_value("Moment y",
+                                  torques_boundary_information[j][i][1]);
+                  table.add_value("Moment z",
+                                  torques_boundary_information[j][i][2]);
```
blaisb, line 543: Should be Torque and not moment

source/dem/pw_nonlinear_force.cc
```diff
@@ -193,6 +204,7 @@ PWNonLinearForce<dim>::calculate_pw_contact_force(
             }
         }
     }
+  this->mpi_correction_over_calculation_of_force_torque();
```
blaisb, line 207:
```suggestion
  this->mpi_correction_over_calculation_of_forces_and_torques();
```

blaisb (commented):
Good work. Only aesthethics remain and you can fix those easily
Can you rebase your PR also?

include/core/parameters_lagrangian.h
EduardsBlandin: I don't wanna have a verbosity dependence. I mean that i wanna be verbose for force and torque and not for other parameters.

applications_tests/dem_3d/torque_boundary_box.output
EduardsBlandin: Yeah std:setw is doable. Do i do it in this prm? @blaisb

applications_tests/dem_3d/force_boundary_box.prm
EduardsBlandin: I choose boundary forces because particle forces may be confusing

include/dem/pw_linear_force.h
EduardsBlandin: I wanted to have an empty vector as default but didn't find, will retry for next commit ^^

applications_tests/dem_3d/torque_boundary_box.output
blaisb: I'd suggest using a TableHandler like is done in the other force output of Lethe instead of this approach. This can be done in a next PR.

include/core/parameters_lagrangian.h
blaisb: No what I meant is that enum already exists elsewhere. There is no need to recreate the enum, you can use it as it is.
Look here : https://github.com/lethe-cfd/lethe/blob/master/include/core/parameters.h#L46
It already exists. Use it...

include/dem/pw_linear_force.h
blaisb: Make it an empty vector, yeah.

include/core/parameters_lagrangian.h
EduardsBlandin: perfect

EduardsBlandin:
Ok, each comment has been processed. If no CI problems, should be able to merge :)

source/dem/output_force_torque_calculation.cc
```diff
@@ -0,0 +1,39 @@
...
+      table.add_value("Tz", torque_on_walls[it.first][2]);
+    }
+  table.set_precision("Fx", 9);
+  table.set_precision("Fy", 9);
+  table.set_precision("Fz", 9);
+  table.set_precision("Tx", 9);
+  table.set_precision("Ty", 9);
+  table.set_precision("Tz", 9);
+
+  table.write_text(std::cout);
```
blaisb, line 28: You need to be careful, this won't work in parallel except if you call this function only from proc 0. Can you check that this is how you have done it now?

blaisb (commented):
One final comment to verify and this should be good for merge

source/dem/output_force_torque_calculation.cc
EduardsBlandin: This is how i did it :)

PR #199, opened 2021-07-05
Steady bdf in sharp ib

Luckabarbeau:
Update assembly in the sharp ib solver to the latest gls_navier_stokes assembly. Implemented the steady_bdf method in sharp interface gls solver.


blaisb:
Can you confirm whether applying the modification everywhere causes an issue at very very low tolerance?Otherwise you can apply it everywhere and modify the cavity test if it makes the model significantly more robust.

blaisb:
@Luckabarbeau can you apply automatic indentation? It seems a lot of lines are changed for nothing

blaisb (approved):
Good job, ready for merge after CI

PR #200, opened 2021-07-06
Dem list insertion

blaisb:
This PR provides a new DEM insertion mechanism using a list of x, y and z positions. This mechanisms works in parallel and application_tests are provided.
The mechanism is very simple. A list of x, y and z values are given. There specific positions are then used to insert particles. This is particularly interesting for testing purposes. This insertion mechanism could eventually be extended such that a list of velocity in x,y,z directions could be provided but I leave that for a future PR.


source/dem/dem.cc
```diff
@@ -519,6 +520,13 @@ DEMSolver<dim>::set_insertion_type(const DEMSolverParameters<dim> &parameters)
         std::make_shared<NonUniformInsertion<dim>>(parameters,
                                                    maximum_particle_diameter);
     }
+  else if (parameters.insertion_info.insertion_method ==
+           Parameters::Lagrangian::InsertionInfo::InsertionMethod::list)
+    {
+      insertion_object =
+        std::make_shared<ListInsertion<dim>>(parameters,
+                                             maximum_particle_diameter);
```
EduardsBlandin, line 528: Why do you need maximum_particle_diameter parameter for this insertion type?

source/dem/list_insertion.cc
```diff
@@ -0,0 +1,130 @@
...
+  const auto &list_z = dem_parameters.insertion_info.list_z;
+
+
+  Assert(list_x.size() == list_y.size(),
+         ListSizeCoherence(list_x.size(), list_y.size()));
+  if (dim == 3)
+    {
+      Assert(list_y.size() == list_z.size(),
+             ListSizeCoherence(list_y.size(), list_z.size()));
+    }
```
EduardsBlandin, line 37: How are you sure that in 2D, he would use x and y list?

source/dem/dem.cc
EduardsBlandin, line 528: Why do you need maximum_particle_diameter parameter for this insertion type?

source/dem/list_insertion.cc
```diff
@@ -0,0 +1,130 @@
...
+
+  // Generate vector of insertion position
+  for (unsigned int i = 0; i < list_x.size(); ++i)
+    {
+      if (dim == 2)
+        insertion_points.emplace_back(Point<dim>({list_x[i], list_y[i]}));
```
EduardsBlandin, line 43: Same as above

EduardsBlandin (approved):
Awsome. Just have to add this insertion type in Lethe's wiki. For the velocity list, it's a good idea.

source/dem/list_insertion.cc
blaisb: Because in deal.II 2D simulations only have the x and y coordinate. The Z list has no meaning in 2D because points are 2D.

source/dem/dem.cc
blaisb: We don't, good catch.

source/dem/list_insertion.cc
```diff
@@ -0,0 +1,127 @@
...
+      MPI_Comm communicator = triangulation.get_communicator();
+      auto this_mpi_process = Utilities::MPI::this_mpi_process(communicator);
+      const unsigned int n_particles_to_insert_this_proc =
+        this_mpi_process == 0 ? n_total_particles_to_insert : 0;
+
+      std::vector<Point<dim>> insertion_points_on_proc_this_step;
+      insertion_points_on_proc_this_step.reserve(
+        n_particles_to_insert_this_proc);
+
+      // Because the list insertion is made to insert only a few particles
```
shahabgol, line 82: Shouldn't we provide a limit for the number of particles inserted by list insertion? Maybe we can define a warning in input_parameter_inspection function for it.

shahabgol (commented):
I just have one small suggestion. Other than that the PR is good to be merged.

source/dem/list_insertion.cc
blaisb: Well, theoretically there is no limitations. You could insert 100 particles this way and it would not be an issue other than the fact that it would be very tedious to set-up the parameter file.

shahabgol (approved):

PR #202, opened 2021-07-06
Implement simplex gmsh solid with Nitsche solver

oguevremont:
Change mapping, quadrature, and FE for simplex cases
Specify simplex status for solid_base tests
Added simplex Nitsche test
Added fluid mapping


applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
```diff
@@ -0,0 +1,4 @@
+SetFactory("OpenCASCADE");
+Sphere(1) = {0, 0, 0, 0.5, -Pi/2, Pi/2, 2*Pi};
+Physical Surface(0) = {1};
+Physical Volume(0) = {1};
```
blaisb, line 4: I question if it is a good idea to keep the geo file or not in the application test, but it's such a small file, why not...

applications_tests/gls_nitsche_navier_stokes_33/sphere.msh
```diff
@@ -0,0 +1,7561 @@
+$MeshFormat
```
blaisb, line 1: This seems like a very heavy mesh, does it have to be this big. Accuracy is not that important for an application test like this?

source/core/solid_base.cc
```diff
@@ -79,7 +103,72 @@ template <int dim, int spacedim>
...
 SolidBase<dim, spacedim>::setup_triangulation(const bool restart)
 {
-  if (param->solid_mesh.type == Parameters::Mesh::Type::gmsh)
+  if (param->solid_mesh.type == Parameters::Mesh::Type::gmsh &&
+      param->solid_mesh.simplex)
+    {
+      // TODO - Serial triangulation should only be created on one processor...
+      // Get communicator from triangulation
+      // auto communicator     = triangulation->get_communicator();
+      // auto this_mpi_process = Utilities::MPI::this_mpi_process(communicator);
```
blaisb, line 112: If this is not necessary, delete

blaisb (commented):
Few minor comments. Very good work. On my side once my comments are answered, it's good to go :)!

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
oguevremont: I don't have a strong opinion on this. I kept it because if we ever need to remake the mesh it could be useful, but at the same time it will be in the history.

applications_tests/gls_nitsche_navier_stokes_33/sphere.msh
oguevremont: I made it lighter and the coarse results are indeed the same

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
blaisb: No worries, keep the geo file. Your rationale is good here.

blaisb:
@oguevremont 
It seems like you have severe reproducibility issues between deal.II 9.3.0 and deal.II master. This is most likely because the quadrature were slightly changed in simplices between the two. Is there a way to minimize the number of digits shown or put yourself in a scenario where things are a bit more robust?

oguevremont:
@blaisb I reduced the number of digits again and it should be fine. I had not updated my dealii version, but now I have, and I'll do it for all future PRs at least.

blaisb (approved):
Good to go on my side. @jeannejoachim or @lpsaavedra  any comments?

source/core/solid_base.cc
```diff
@@ -104,12 +190,16 @@ SolidBase<dim, spacedim>::setup_triangulation(const bool restart)
...
     {
       solid_tria->refine_global(param->solid_mesh.initial_refinement);
       // Initialize dof handler for solid
-      FE_Q<dim, spacedim> fe(1);
-      solid_dh.distribute_dofs(fe);
+      solid_dh.distribute_dofs(*fe);
+    }
+  else if (!restart)
+    {
+      // Initialize dof handler for solid
```
lpsaavedra, line 201: Don't you need to globally refine the triangulation in this case as well?

lpsaavedra (commented):
Seems good to me... only one small question

applications_tests/gls_nitsche_navier_stokes_33/noslip33_simplex.prm
```diff
@@ -0,0 +1,148 @@
...
+	set beta = 10
+  	set calculate torques on solid = true
+  	set verbosity = verbose
+  	set number of solids = 1
+
+  	subsection nitsche solid 0
+	  	subsection mesh
+	  		set simplex = true
+		  	set type = gmsh
+			set file name = ../../sphere.msh
```
jeannejoachim, line 81: could deal.ii mesh generator works for simplex as of now? If yes, I think it would be better to use it rather than an external .msh. If no, perhaps leave a note (on the git issues tab?) so it can be changed when it does.

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
jeannejoachim: Maybe put is in Lethe-utils?

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
```diff
@@ -0,0 +1,5 @@
...
+Mesh.MeshSizeFactor = 2;
```
jeannejoachim, line 5: Have you checked that this is well taken into account during the mesh generation? To my experience with GMSH, mesh control options can be discarded by OpenCascade. This would not be troublesome here, but if you need to have a higher control over your mesh you may have to switch to the default generator. Just keep that in mind :)

source/core/solid_base.cc
```diff
@@ -79,7 +103,68 @@ template <int dim, int spacedim>
 void
 SolidBase<dim, spacedim>::setup_triangulation(const bool restart)
 {
-  if (param->solid_mesh.type == Parameters::Mesh::Type::gmsh)
+  if (param->solid_mesh.type == Parameters::Mesh::Type::gmsh &&
```
jeannejoachim, line 106: Would not it be clearer/faster to have cascade ifs? Like :
```
if (param->solid_mesh.type == Parameters::Mesh::Type::gmsh)
  {
  if (param->solid_mesh.simplex)
   {...}
  else
   {...}
  }
if (param->solid_mesh.type == Parameters::Mesh::Type::dealii)
  {
  if (param->solid_mesh.simplex)
   {...}
  else
   {...}
  }
```
This would have several advantages : 

- removing the `param->!solid_mesh.simplex`, as it is the logical complementary of `param->solid_mesh.simplex` (better to write "else" for that),
- in terms of speed I believe it is faster, as you have an upper layer of 2 conditions, then of 2 sub-conditions, rather than potentially 4 conditions to check (it would be on Python, but I am not fully aware of speed designs on C++ yet),
- maybe remove redundancy in the code if any, like here if you have common things to do if the mesh if of type gmsh, be it simplex of not. Or the other way around : things to do if the mesh is simplex, be it generated by gmsh of dealii (and then you will have to invert the if conditions : put the `simplex` condition in the upper layer and `mesh.type` as a sub-condition)

source/core/solid_base.cc
```diff
@@ -104,12 +190,16 @@ SolidBase<dim, spacedim>::setup_triangulation(const bool restart)
...
-  if (!restart)
+  if (!restart && !param->solid_mesh.simplex)
     {
       solid_tria->refine_global(param->solid_mesh.initial_refinement);
       // Initialize dof handler for solid
-      FE_Q<dim, spacedim> fe(1);
-      solid_dh.distribute_dofs(fe);
+      solid_dh.distribute_dofs(*fe);
+    }
+  else if (!restart)
```
jeannejoachim, line 199: Same as before. This says that `!restart` is the logical complement of `!restart && !param->solid_mesh.simplex`, when really here you have two sub-cases for a `!restart` condition. This decreases the code readability in my point of view. Of course now the conditions are close enough to each other so that we can get it, but it might not be the case in the future. So I would prefer :
```
if (!restart)
  if(!param->solid_mesh.simplex)
    {...}
  else
    {...}
```

jeannejoachim (commented):
Very cool implementation! Done neatly, though I think some condition handling should be changed to improve readability and maybe reduce code redundancy. I like that you added multiple tests to cover everything, good work!

applications_tests/gls_nitsche_navier_stokes_33/noslip33_simplex.prm
blaisb: That's a good remark, because in theory you could use deal.II meshes with simplices. Maybe not with this test case, but eventually yeah. This would be a good "TODO".

source/core/solid_base.cc
blaisb: I don't think speed matters here, but readability does. I like the idea of making a cascade, it increases readability. In this case Olivier, could you also refactor it in the source/core/grid so that it follows the same idea?

oguevremont: @jeannejoachim I really like the idea of the cascade. I didn't think about it so that's why I didn't do it that way, but you are right regarding the readability. And Bruno, I'll do it there too.

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
oguevremont: I checked when making the geo file, but I'll double check.

applications_tests/gls_nitsche_navier_stokes_33/noslip33_simplex.prm
oguevremont: It didn't work when I coded it, but maybe it's because I had an old version of dealii, so I'll make sure, and if not possible for now I'll make a note of it for later (TODO or issues)

applications_tests/gls_nitsche_navier_stokes_33/sphere.geo
oguevremont: With the latest versions of gmsh there is no problem

blaisb:
@oguevremont Once you have applied @jeannejoachim suggestions for the mesh parsing, just ping me. This will be ready for merge.

source/core/solid_base.cc
oguevremont: With the simplex triangulation it is not possible for now, but I can do it when/if it is implemented in dealii

blaisb (approved):

PR #203, opened 2021-07-08
Add update-golden.tl utility

paulapatience:
This utility updates Lethe's golden files, which is unobvious to do with
a command-line one-liner because the expected and golden files' paths
are vexatiously inconsistent, e.g., for suitable values of APP, TEST,
MPIRUN and N,
  "applications_tests/APP/TEST.release/MPIRUN=N/output"
becomes
  "applications_tests/APP/TEST.MPIRUN=N.output".

* contrib/utilities/update-golden.tl: New file.

* CMakeLists.txt: Search for the TXR executable and store its path in
TXR_EXECUTABLE if found. If found, add...
(update-golden): ...this new custom target which runs update-golden.tl
with the proper arguments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,8 +53,15 @@ IF(PROTOTYPES)
     ADD_SUBDIRECTORY(prototypes)
  ENDIF()
 
-
-
-
-
-
+find_program(TXR_EXECUTABLE
+  NAMES txr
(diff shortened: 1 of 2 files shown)


blaisb (approved):
Nothing to say here. A nice PR, although difficult for me to review the TXR part :)

PR #204, opened 2021-07-09
Dem triangulation motion

shahabgol:
This PR adds motion of triangulation to the DEM solver. It uses deal.II functions GridTools::shift and GridTools::rotate for rotating and sliding (transnational) motions of the triangulation. Using the combination of these two functions, we can have transnational-rotational motions of the grid in the solver. 
To this end, we need to update the position of the point on the boundary face and its normal vector in the particle-wall contact list. update_boundary_points_and_normal_vectors_in_contact_list function in GridMotion class takes care of this issue. 
Three application tests (box_grid_rotation, box_grid_slide and box_grid_slide_rotation) have been added to the code. The application test box_grid_slide_rotation is defined in parallel and with load-balancing to investigate the performance of grid motion in parallel mode.


blaisb:
@shahabgol  You will need to rebase

shahabgol:
> @shahabgol You will need to rebase

Done! :)

source/dem/grid_motion.cc
```diff
@@ -0,0 +1,110 @@
...
+{
+  GridTools::shift(shift_vector, triangulation);
+}
+
+template <>
+void GridMotion<2>::move_grid_translational_rotational(
+  parallel::distributed::Triangulation<2> &triangulation)
+{
+  GridTools::shift(shift_vector, triangulation);
+  GridTools::rotate(rotation_angle, triangulation);
```
blaisb, line 70: This will need to be documented clearly right ? Because shift + rotate is not the same as rotate + shift.

source/dem/localize_contacts.cc
```diff
@@ -31,6 +31,7 @@ localize_contacts(
                        std::tuple<Particles::ParticleIterator<dim>,
                                   Tensor<1, dim>,
                                   Point<dim>,
+                                  unsigned int,
```
blaisb, line 34: This is for boundary id? Is there not a type of the boundary id?

blaisb: https://www.dealii.org/developer/doxygen/deal.II/namespacetypes.html#aed8813fee8c8a2edcc6005e6a48c321a

I know they are the same, but it makes readibility better.

blaisb (commented):
Very good. I only have a few minor comments that will be easy to address.

source/dem/dem.cc
```diff
@@ -254,6 +258,11 @@ DEMSolver<dim>::load_balance()
 
   boundary_cell_object.build(triangulation, parameters.floating_walls);
 
+  if (parameters.grid_motion.motion_type !=
+      Parameters::Lagrangian::GridMotion<dim>::MotionType::none)
+    boundary_cell_object.update_boundary_info_after_grid_motion(
+      updated_boundary_points_and_normal_vectors);
+
```
EduardsBlandin, line 265: Why do you need to make the update during the load balance?

include/dem/grid_motion.h
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * The constructor sets up the grid motion function defined in the parameter
+   * handler, and calculates the rotation angle for rotational and
+   * translational-rotational motions.
+   *
+   * @param dem_parameters DEM parameters defined using the parameter handler
+   * @param dem_time_step DEM time-step
+   */
+  GridMotion(const DEMSolverParameters<dim> &dem_parameters,
+             const double &                  dem_time_step);
```
EduardsBlandin, line 56: dem_time_step is not something you already find in dem_parameters?

source/dem/dem.cc
```diff
@@ -842,6 +861,18 @@ DEMSolver<dim>::solve()
...
+      // their normal vectors here. The localize_contact class deletes the
+      // particle-wall contact candidate if it exists in the contact list. As a
+      // result, when we update the points on boundary faces and their normal
+      // vectors, localize_contact deletes it from the output of broad search
+      // and they are not updated in the contact force calculations.
+      if (parameters.grid_motion.motion_type !=
+          Parameters::Lagrangian::GridMotion<dim>::MotionType::none)
+        grid_motion_object
+          ->update_boundary_points_and_normal_vectors_in_contact_list(
+            pw_pairs_in_contact, updated_boundary_points_and_normal_vectors);
```
EduardsBlandin, line 874: The update is obliged to be done twice in the while boucle?

include/dem/boundary_cells_info_struct.h
```diff
@@ -41,7 +41,7 @@ struct boundary_cells_info_struct
   unsigned int boundary_id;
 
   // ID of boundary face in the boundary cell
-  unsigned int boundary_face_id;
+  unsigned int global_face_id;
```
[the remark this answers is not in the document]
shahabgol: It is the global face id, so I corrected it.

source/dem/dem.cc
shahabgol: When we call load-balancing, all the structures related to triangulation including finding cell neighbors and boundary cells should be rebuilt.

source/dem/dem.cc
shahabgol: No, these are two different functions. update_boundary_info_after_grid_motion updates the normal vectors and vertices. update_boundary_points_and_normal_vectors_in_contact_list applies these updated points in the particle-wall contact list.

EduardsBlandin (approved):
Awesome. Have just one question which you'll see. Bruno questions and comments are good. Good job ^^

include/dem/boundary_cells_info_struct.h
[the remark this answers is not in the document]
blaisb: But what does global face id mean? I don't understand. Is it the ID of a face of a triangulation?

blaisb:
Can you address the issue with the boundary_id type and then I could merge this?

shahabgol:
> Can you address the issue with the boundary_id type and then I could merge this?

Yes, there is a problem in sliding_rotational motion. I sent you a video. If it is acceptable, I will take care of the rest.

source/dem/localize_contacts.cc
shahabgol: done

include/dem/boundary_cells_info_struct.h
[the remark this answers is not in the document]
shahabgol: face_index() returns the global id of the face in triangulation. each face in dealii triangulation has a unique global id.

shahabgol:
@blaisb All the comments are applied. Can you please take another look? :)

EduardsBlandin:
@shahabgol I think that it would be better to have a mother class where you have your update triangulation function and some members, and that your constant motion should be a children of this class. This way, in future developments, people would be able to have differents motion possibility (like functional motion->depends of function, forces motion->depends of forces, hazardous motion->depends of random parameters and yours constant motion-> depends of constant parameters).

shahabgol:
> @shahabgol I think that it would be better to have a mother class where you have your update triangulation function and some members, and that your constant motion should be a children of this class. This way, in future developments, people would be able to have differents motion possibility (like functional motion->depends of function, forces motion->depends of forces, hazardous motion->depends of random parameters and yours constant motion-> depends of constant parameters).
> What do you think?

The method I used is very similar. We have a base function (move_grid) which is rewritten by the children functions (move_grid_rotational, move_grid_translational). New functions can easily be added similar to the method I've used.

blaisb (approved):

PR #206, opened 2021-07-12
Stored all velocity (velocity, and p1 to p3) in a vector called velocity

CaroleAnneDaunais:


blaisb (commented):
Very good first change!
The next step would be to remove the p1, p2, etc. variables and use a vector directly from the get go. Do you want to do this in this PR?

CaroleAnneDaunais:
> Very good first change!
> The next step would be to remove the p1, p2, etc. variables and use a vector directly from the get go. Do you want to do this in this PR?

Yes definitely!

blaisb:
Very cool & much cleaner.
I suggest that once CI pass I merge? @CaroleAnneDaunais

CaroleAnneDaunais:
Yes I was waiting the the CI, it will be ready to merge!

PR #207, opened 2021-07-13
New force evaluation on ib

Luckabarbeau:
Change how forces are evaluated on IB. Now uses face cell projections on the IB surface.


blaisb (approved):
Much, much, much cleaner than before...

PR #208, opened 2021-07-13
Dem refactor boundary

shahabgol:
This PR refactors the class the BoundaryMotion to a more general BCDEM inside the DEM parameter handler.
DEM boundaries can now be defined as outlet, fixed_wall, rotational and transnational. If defined as outlet, the particles may leave the triangulation from these outlets.


applications_tests/dem_3d/rotating_in_ball.prm
```diff
@@ -91,11 +91,11 @@ subsection mesh
...
 
 #---------------------------------------------------
-# Boundary Motion
+#  DEM Boundary Conditions
 #---------------------------------------------------
-subsection boundary motion
-  set number of boundary motion         		= 1
-    subsection moving boundary 0
+subsection DEM boundary conditions
+  set number of boundary conditions         		= 1
```
blaisb, line 97: Why is there only one boundary condition set? does the default behavior is to assume all boundaries without anything assigned to them are walls?

blaisb (approved):
Only one comment which is not very much a comment. But just to be sure. I'm ok with the current behavior, but it will need to be documented well.

applications_tests/dem_3d/rotating_in_ball.prm
EduardsBlandin: That's exactly that. And if the user wanna still put a boundary condition as wall, the option exists as well. (as default)

shahabgol: Yes, as @EduardsBlandin mentioned, the default condition for the boundaries is fixed_wall.

PR #209, opened 2021-07-14
Smoothed void fraction

tonielgeitani:


examples/cfd-dem/flow_in_bed.prm
```diff
@@ -0,0 +1,141 @@
...
+    set type                 				= dealii
+    set grid type            				= subdivided_cylinder
+    set grid arguments       				= 16:0.01:0.1
+    set initial refinement   				= 1
+end
+
+#---------------------------------------------------
+# Void Fraction
+#---------------------------------------------------
+subsection void fraction
```
blaisb, line 50: I think this section is a bit poorly named right now. One section is more like CFD-DEM parameters, and the void fraction is another thing right? Maybe just rename the entire section cfd-dem coupling?

blaisb (commented):
Can you just change my minor comment and I will merge afterwards?
I think this will  be enough for your Article 1. When the assembly refactoring is finished, we can use this occasion to rewrite your assembly to be all neat and tidy. What do you think?&

blaisb (commented):
You merged instead of rebasing I think. I'll just fix indentation and push then your branch will be  merged. Just be careful about rebasing correctly next time.

PR #210, opened 2021-07-15
Allow solid mesh translation at initiation

oguevremont:
Add mesh translation parameters
Add exception for trying to move fluid mesh
Add solid mesh translation


oguevremont:
Since there is no point doing the translation of a fluid mesh if there is no solid mesh, we can say that only the solid mesh is movable. The reason for this choice is that moving the fluid mesh instead makes it deformed, so I made it throw an exception if an attempt is made.
@blaisb Would such a small and simple PR justify a test?

PR #211, opened 2021-07-19
Packed bed example

tonielgeitani:
This PR adds a 3D packed bed test for the VANS equations with 10,000 particle.  The end time of the test was reduced drastically since in Debug mode the test always timed out.


applications_tests/gls_vans_3d/packed_bed.prm
```diff
@@ -0,0 +1,146 @@
+# Listing of Parameters
+# ---------------------
+# --------------------------------------------------
+# Simulation and IO Control
+#---------------------------------------------------
+subsection simulation control
+  set method                  = bdf1
+  set number mesh adapt       = 0
+  set output name              = result_
+  set output frequency	  	= 1
```
shahabgol, line 10: I think we are not supposed to write output files in application tests. It's better to increase the output frequency, so that the test doesn't write any output files.

applications_tests/gls_vans_3d/packed_bed.output
```diff
@@ -0,0 +1,24 @@
...
+Transient iteration : 3        Time : 0.006    Time step : 0.002    CFL : 0.2453174007
+*****************************************************************
+
+*****************************************************************
+Transient iteration : 4        Time : 0.008    Time step : 0.002    CFL : 0.2122950793
+*****************************************************************
+
+*****************************************************************
+Transient iteration : 5        Time : 0.01     Time step : 0.002    CFL : 0.224382112
+*****************************************************************
```
shahabgol, line 24: We should write some simulation information at the end of the test. For example, you can see the poiseuille3d application test, for which cells error_velocity and error_pressure are written in the output file.

shahabgol (commented):
The test is working fine, I checked it on my machine. I just think that the test should not write any outputs, and we should write some simulation information in the output file. By comparing this information, we would be sure that the test results are identical on different machines.

blaisb:
> The test is working fine, I checked it on my machine. I just think that the test should not write any outputs, and we should write some simulation information in the output file. By comparing this information, we would be sure that the test results are identical on different machines.

Yeah that's a great idea. What do you suggest in terms of term? Maybe we could add an additional post-processing tool that calculates pressure drop between two boundary ids and use that as the verification? what do you think @tonielgeitani

shahabgol:
> > The test is working fine, I checked it on my machine. I just think that the test should not write any outputs, and we should write some simulation information in the output file. By comparing this information, we would be sure that the test results are identical on different machines.
> 
> Yeah that's a great idea. What do you suggest in terms of term? Maybe we could add an additional post-processing tool that calculates pressure drop between two boundary ids and use that as the verification? what do you think @tonielgeitani

Yes, bed pressure drop and average voidage in the packing region are good, in my opinion.

tonielgeitani:
Yes of course. I agree. I will start working on the average voidage and the pressure drop

shahabgol:
Very good work. I just have two small comments.
1, The test still writes one output
2, We should add the unit of the pressure drop after reporting the value

shahabgol (approved):

include/core/parameters_cfd_dem.h
```diff
@@ -88,6 +88,9 @@ namespace Parameters
     bool      full_stress_tensor;
     double    reference_velocity;
     DragModel drag_model;
+    bool      post_processing;
+    int       inlet_boundary_id;
+    int       outlet_boundary_id;
```
blaisb, line 93:
```suggestion
    unsigned int       inlet_boundary_id;
    unsigned int       outlet_boundary_id;
```

blaisb (commented):
Minor thing to fix, then ready for merge :)

PR #212, opened 2021-07-20
Add scratch/copy assembly mechanism for GLS Navier Stokes Solver

blaisb:
This PR Introduces the scratch/copy data mechanism to assemble equations in Lethe instead of using a traditional assembly approach. Right now only the GLS Navier-Stokes assembly is done this way, but the transition for the rest will be easy.


oguevremont (commented):
Good job!

include/core/sdirk.h
```diff
@@ -87,5 +87,83 @@ number_of_intermediary_stages(
...
+           method ==
+             Parameters::SimulationControl::TimeSteppingMethod::sdirk22_2)
+    return 1;
+
+  else // if (method ==
+       // Parameters::SimulationControl::TimeSteppingMethod::sdirk33_3)
+    return 2;
+}
+
+inline unsigned int
```
CaroleAnneDaunais, line 108: Could also be a const attribute of the class

tonielgeitani (approved):

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -0,0 +1,201 @@
...
+  std::vector<std::vector<Tensor<1, dim>>> grad_phi_p;
+
+  // Values at previous time step for transient schemes
+  // std::vector<std::vector<Tensor<1, dim>>> velocity_values = {
+  //  std::vector<Tensor<1, dim>>(n_q_points),
+  //  std::vector<Tensor<1, dim>>(n_q_points),
+  //  std::vector<Tensor<1, dim>>(n_q_points)};
+
+  //  std::vector<double> time_steps_vector =
+  //    this->simulation_control->get_time_steps_vector();
```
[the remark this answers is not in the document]
blaisb: Nope it will be deleted. Thanks!

include/solvers/copy_data.h
```diff
@@ -0,0 +1,187 @@
...
+ * storing the information calculated using the assembly of stabilized
+ * scalar equations. Like the CopyData class, this class is used to initialize,
+ * zero (reset) and store the cell_matrix and the cell_rhs.
+ * Contrary to the regular CopyData class, this class
+ * also stores the strong_residual and the strong_jacobian of the equation being
+ * assembled. This is useful for equations that implement residual-based
+ * stabilization such as SUPG. This class is specialized for single component
+ * equations because the strong jacobian is stored using a Vector<double>
+ **/
+class StabilizedMethodsCopyData
```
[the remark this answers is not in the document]
blaisb: I thought about it, but then I did not know what would be shared between the classes so I went agaisnt it. It is something to keep in mind though.

oguevremont (approved):

include/solvers/copy_data.h
```diff
@@ -0,0 +1,187 @@
...
+  CopyData(const unsigned int n_dofs)
+    : local_matrix(n_dofs, n_dofs)
+    , local_rhs(n_dofs)
+    , local_dof_indices(n_dofs){};
+
+  /**
+   * @brief Resets the cell_matrix and the cell_rhs to zero
+   */
+  void
+  zero()
```
lpsaavedra, line 55: Could we change the name of these functions to `reset()` or `set_to_zero()`? I find it more intuitive than just `zero()`

blaisb: Absolutely.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -0,0 +1,675 @@
...
+                velocity_gradient_x_phi_u_j[j] * phi_u_i +
+                grad_phi_u_j_x_velocity[j] * phi_u_i - div_phi_u_i * phi_p_j +
+                // Continuity
+                phi_p_i * div_phi_u_j;
+
+              // PSPG GLS term
+              local_matrix_ij += tau * (strong_jac * grad_phi_p_i);
+
+
+              // Jacobian is currently incomplete
```
lpsaavedra, line 134: Does this mean that the SUPG tau term is not there? Maybe it would be a good idea to specify it here?

blaisb: The SUPG tau term is there but the jacobian of the tau itself is not included.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -0,0 +1,675 @@
...
+
+          // PSPG GLS term
+          local_rhs(i) += -tau * (strong_residual * grad_phi_p_i) * JxW;
+
+          // SUPG GLS term
+          if (SUPG)
+            {
+              local_rhs(i) +=
+                -tau * (strong_residual * (grad_phi_u_i * velocity)) * JxW;
+            }
```
lpsaavedra, line 246: Could we use the same strategy as with the matrix i.e. writing to a double variable and the writing to the `local_rhs(i)` once? Or it does not speed up the computation at all in this case?

blaisb: You are right we could. It will most likely have a more minor effect, but worth trying! I'll add this.

include/solvers/navier_stokes_assemblers.h
```diff
@@ -0,0 +1,119 @@
...
+
+  virtual void
+  assemble_matrix(NavierStokesScratchData<dim> &        scratch_data,
+                  StabilizedMethodsTensorCopyData<dim> &copy_data) override;
+  virtual void
+  assemble_rhs(NavierStokesScratchData<dim> &        scratch_data,
+               StabilizedMethodsTensorCopyData<dim> &copy_data) override;
+
+  const bool SUPG = true;
+  const bool PSPG = true;
```
lpsaavedra, line 60: Where is this PSPG variable used?

source/solvers/navier_stokes_assemblers.cc
lpsaavedra: Ah ok! I did not understand the comment, do you think it would be relevant to rewrite it? Something like: it does not include the jacobian of the tau?

include/solvers/navier_stokes_assemblers.h
blaisb: It is not used! Good catch :)

include/solvers/copy_data.h
blaisb: Done, will show in next push.

blaisb:
I don't think this is fully ready for a merge.
@jeannejoachim  Would you be able to test 1-2 free surface cases to see if this does not make things go bonkers? I did not with the unit test, but it's not that sensitive of  a test.

I'll finish documentation and I'll use this opportunity to clean the solvers.

include/solvers/navier_stokes_vof_assemblers.h
```diff
@@ -0,0 +1,77 @@
...
+
+  virtual void
+  assemble_matrix(NavierStokesScratchData<dim> &        scratch_data,
+                  StabilizedMethodsTensorCopyData<dim> &copy_data) override;
+  virtual void
+  assemble_rhs(NavierStokesScratchData<dim> &        scratch_data,
+               StabilizedMethodsTensorCopyData<dim> &copy_data) override;
+
+  const bool SUPG = true;
+  const bool PSPG = true;
```
lpsaavedra, line 48: We are not using this variable, should we remove it?

blaisb: Yup, will do :)!

include/core/bdf.h
```diff
@@ -74,5 +95,24 @@ maximum_number_of_previous_solutions()
...
 
+/**
+ * @brief Returns the number of previous time steps for a BDF Scheme
+ *
+ */
+inline unsigned int
+number_of_previous_solutions(
+  Parameters::SimulationControl::TimeSteppingMethod method)
+{
+  if (method == Parameters::SimulationControl::TimeSteppingMethod::bdf1 ||
```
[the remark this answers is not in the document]
blaisb: Good idea. Done in next push.

include/core/sdirk.h
blaisb: True, but bdf is not a class per say right now. It could become eventually though. Let's leave it like that for the moment :).

source/solvers/navier_stokes_assemblers.cc
blaisb: This is what the message reads now : 
              // The jacobian matrix for the SUPG formulation
              // currently does not include the jacobian of the stabilization
              // parameter tau. Our experience has shown that does not alter the
              // number of newton iteration for convergence, but greatly
              // simplifies assembly.

source/solvers/navier_stokes_assemblers.cc
blaisb: Added :)!

include/solvers/navier_stokes_vof_assemblers.h
blaisb: Done

PR #213, opened 2021-07-20
Add std::flush to print_xyz function

shahabgol:
DEM application tests use print_xyz function to print and compare the outputs at the end of the test simulations.
Sometimes, the order of processes writing particles' info using this function mixes up, which leads to failed tests.
This PR uses std::flush in print_xyz function, in order to avoid buffering in std::cout and write the header and particles' info in order.

### DIFF
--- a/source/dem/visualization.cc
+++ b/source/dem/visualization.cc
@@ -94,6 +94,7 @@ Visualization<dim>::print_xyz(
     }
 
   pcout << "id, type, dp, x, y, z " << std::endl;
+  std::flush(std::cout);
 
   for (auto &iterator : local_particles)
     {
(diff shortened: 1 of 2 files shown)


source/dem/visualization.cc
```diff
@@ -94,6 +94,7 @@ Visualization<dim>::print_xyz(
     }
 
   pcout << "id, type, dp, x, y, z " << std::endl;
```
blaisb, line 96: why is this the only I/o with pcout? Maybe this one should be also with std::cout instead. Just manually done with the proc 0.

blaisb (commented):
one minor comment on the I/O

source/dem/visualization.cc
shahabgol: I checked it with std::cout, as you said. But again proc1 calls the print_xyz function first in insert_list_2d test and the header is written at the end of the output file.

blaisb: Hmm this is weird. Might be a CI-only related issue.

PR #214, opened 2021-07-21
Add vector resizing after load-balancing

shahabgol:
This PR fixes a bug in the DEM solver.
After calling load-balancing, the displacement vector is resized, similar to the way we do it after the insertion.


blaisb (approved):
Yesssss

PR #217, opened 2021-07-28
Load GMSH simplex grid on one processor

blaisb:
Simplex meshes are now read on a single core instead of multiple cores. This significantly reduces memory usage.

### DIFF
--- a/source/core/grids.cc
+++ b/source/core/grids.cc
@@ -28,30 +28,27 @@ attach_grid_to_triangulation(
     {
       if (mesh_parameters.simplex)
         {
-          // TODO - Serial triangulation should only be created on one
-          // processor... Get communicator from triangulation auto communicator
-          // = triangulation->get_communicator(); auto this_mpi_process =
-          // Utilities::MPI::this_mpi_process(communicator);
-
-          Triangulation<dim, spacedim> basetria(
-            Triangulation<dim, spacedim>::limit_level_difference_at_vertices);
-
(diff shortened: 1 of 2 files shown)


source/core/grids.cc
```diff
@@ -28,30 +28,27 @@ attach_grid_to_triangulation(
     {
       if (mesh_parameters.simplex)
         {
-          // TODO - Serial triangulation should only be created on one
-          // processor... Get communicator from triangulation auto communicator
-          // = triangulation->get_communicator(); auto this_mpi_process =
-          // Utilities::MPI::this_mpi_process(communicator);
-
-          Triangulation<dim, spacedim> basetria(
-            Triangulation<dim, spacedim>::limit_level_difference_at_vertices);
```
oguevremont, line 37: Will this flag be needed again when/if simplex refinement is available? If yes, we'll have to keep it in mind

oguevremont (approved):
Good improvement to the mesh reading

source/core/grids.cc
blaisb: I am not sure to be honest. The flag  was already there.... Good question. I suggest we leave it at that and just wait?

oguevremont: Yes I agree, and for simplex refinement we'll know where to look in case it is needed

peterrum: If you want you can pass in the type of smoothing via the setting parameter.

PR #222, opened 2021-07-30
Add a motion type where the particle collision force moves the boundary (only translation)

EduardsBlandin:
It moves the boundary depending of the forces applied by the particles. Right now, it only translates the triangulation. Cannot work with load-balancing.
Changed a little bit the prm format, i changed the ci test format accordingly. Like we don't need to set three times for only one variable right now. So set ax = 1, set ay=2,set az=3 becomes set a = 1,2,3.
A test could be add as well. 
Next pull would be to use this structure to a specific motion case.


blaisb:
Test are not passing at the moment @EduardsBlandin  :)

include/core/parameters_lagrangian.h
```diff
@@ -256,6 +256,21 @@ namespace Parameters
       // Center of mass
       Point<dim> point_center_mass;
 
+      // Boundary Inertia
+      Tensor<1,3> boundary_inertia;
```
blaisb, line 260: What is boundary inertia? Is it the moment of inertia of the solid triangulation?

include/core/parameters_lagrangian.h
```diff
@@ -256,6 +256,21 @@ namespace Parameters
       // Center of mass
       Point<dim> point_center_mass;
```
blaisb, line 257: Should be renamed to center_of_mass or triangulation_center_of_mass

include/core/parameters_lagrangian.h
```diff
@@ -256,6 +256,21 @@ namespace Parameters
...
 
+      // Boundary Inertia
+      Tensor<1,3> boundary_inertia;
+
+      // Boundary mass
+      double boundary_mass;
```
blaisb, line 263: Triangulation mass?

include/core/parameters_lagrangian.h
```diff
@@ -350,22 +365,21 @@ namespace Parameters
...
       // Translational velocity of the moving grid
       Tensor<1, dim> grid_translational_velocity;
 
       // Rotational speed of rotating grid in rad/s
-      double grid_rotational_speed;
-
-      // Rotational axis of rotating grid. Similar to deal.II, we use 0=x axis,
-      // 1=y axis, 2=z axis.
-      unsigned int grid_rotational_axis;
+      Tensor<1,3> grid_rotational_speed;
```
blaisb, line 382: You introduce a rotation tensor? That's a good idea.

include/dem/grid_motion.h
```diff
@@ -109,14 +111,49 @@ class GridMotion
...
     parallel::distributed::Triangulation<dim> &triangulation);
 
+  /**
+   * Carries out translational motion of the triangulation
+   *
+   * @param triangulation Triangulation
+   */
+  void
+  move_grid_due_particles_forces(
+    parallel::distributed::Triangulation<dim> &triangulation);
```
blaisb, line 121: The naming choice here is weird. Is this moving the grid, or integrating the equation of motion for the grid? Use complete sentences or wording when establishing function names.

include/dem/grid_motion.h
```diff
@@ -109,14 +111,49 @@ class GridMotion
...
+
+  /**
+   * Update member information like boundary_forces
+   */
+  void
+  update_parameters_before_motion();
```
blaisb, line 127: Need a better explanation of the function content. I don't understand what the function does from the comments.

include/dem/grid_motion.h
```diff
@@ -109,14 +111,49 @@ class GridMotion
...
+
+  /**
+   * Update member information like inertia tensor,
+   * center of mass coordinate, etc..
+   */
+  void
+  update_parameters_after_motion();
```
blaisb, line 134: Same as above. Specify everything that is updated. Be more descriptive, you must help people to understand what you say.

include/dem/grid_motion.h
```diff
@@ -109,14 +111,49 @@ class GridMotion
...
-  unsigned int rotation_axis;
-
-  Tensor<1, dim> shift_vector;
+  Tensor<1,3> rotation_angle;
+  Tensor<1,dim> shift_vector;
+
+  const double boundary_mass;
+  Tensor<1,3> boundary_inertia;
+  Tensor<1,dim> boundary_forces;
+  Tensor<1,dim> boundary_torques;
```
blaisb, line 152: Why are those called boundary force/masse/torque. Are they not the information about the triangulation or the container?

include/dem/grid_motion.h
```diff
@@ -109,14 +111,49 @@ class GridMotion
...
+  const double dt;
+  Tensor<1,dim>     boundary_translational_velocity;
+  Tensor<1,3> boundary_rotational_velocity;
+  std::shared_ptr<PWContactForce<dim>> pw_contact_force_object;
```
blaisb, line 156: Why is dt stored as a class member?

include/dem/pw_contact_force.h
```diff
@@ -80,6 +80,15 @@ class PWContactForce
     return torque_on_walls;
   }
 
+  /** This function is used to update the center of mass container member class
+   * @param Tensor would be added to the member class to make the update
+   */
```
blaisb, line 85: Revise English here.

include/dem/pw_contact_force.h
```diff
@@ -80,6 +80,15 @@ class PWContactForce
...
+  void
+  update_center_mass(const Tensor<1, dim> Tensor)
```
blaisb, line 87:
```suggestion
  update_center_of_mass(const Tensor<1, dim> Tensor)
```

include/dem/pw_contact_force.h
```diff
@@ -80,6 +80,15 @@ class PWContactForce
...
+  {
+    center_mass_container += Tensor;
```
blaisb, line 89: What is this Tensor? Is it the displacement? if so, why not call it displacement?

source/core/parameters_lagrangian.cc
```diff
@@ -605,21 +606,33 @@ namespace Parameters
...
-      prm.declare_entry("y",
-                        "0",
-                        Patterns::Double(),
-                        "Y coordinate of center of mass");
-      prm.declare_entry("z",
+      prm.declare_entry("center of mass coordinate",
+                        "0,0,0",
+                        Patterns::List(Patterns::Double(),3,3),
+                        "Coordinate of center of mass");
+      prm.declare_entry("boundary inertia",
```
blaisb, line 613:
```suggestion
      prm.declare_entry("moment of inertia",
```

source/core/parameters_lagrangian.cc
```diff
@@ -605,21 +606,33 @@ namespace Parameters
...
-      prm.declare_entry("x",
-                        "0",
-                        Patterns::Double(),
-                        "X coordinate of center of mass");
-      prm.declare_entry("y",
-                        "0",
-                        Patterns::Double(),
-                        "Y coordinate of center of mass");
-      prm.declare_entry("z",
+      prm.declare_entry("center of mass coordinate",
```
blaisb, line 609:
```suggestion
      prm.declare_entry("center of mass",
```

source/core/parameters_lagrangian.cc
```diff
@@ -605,21 +606,33 @@ namespace Parameters
...
+      prm.declare_entry("boundary inertia",
+                        "0,0,0",
+                        Patterns::List(Patterns::Double(),3,3),
+                        "Boundary inertia around x,y,z axis");
+      prm.declare_entry("boundary mass",
                         "0",
                         Patterns::Double(),
-                        "Z coordinate of center of mass");
-      prm.leave_subsection();
+                        "Boundary mass in kg");
```
blaisb, line 620: What is boundary mass? Is it the mass of the solid body that is moving?

blaisb (commented):
First part of the review. I'll finish later.

include/core/parameters_lagrangian.h
EduardsBlandin: Yes, it was a redundant information.

include/dem/grid_motion.h
EduardsBlandin: like move_grid_due_to_particles_forces is a better name @blaisb ?

include/dem/pw_contact_force.h
EduardsBlandin: Didn't think about that name too much. I will call it displacement!

source/core/parameters_lagrangian.cc
EduardsBlandin: Yes

include/core/parameters_lagrangian.h
EduardsBlandin: I changed to triangulation_inertia

include/dem/grid_motion.h
EduardsBlandin: Next commit

include/dem/grid_motion.h
EduardsBlandin: Next commit

include/dem/grid_motion.h
EduardsBlandin: Yes, i would add triangulation in theirs names, is that enough @blaisb ?

include/dem/grid_motion.h
EduardsBlandin: To be applied in the Newton's theorem. Should it be an input?

include/dem/pw_contact_force.h
EduardsBlandin: Just see would to will. Is there anything else @blaisb ?

include/core/parameters_lagrangian.h
```diff
@@ -254,7 +254,22 @@ namespace Parameters
...
+      Point<dim> triangulation_center_mass;
+
+      // Triangulation inertia
+      Tensor<1, 3> triangulation_inertia;
+
+      // Boundary mass
+      double triangulation_mass;
+
+      // Enable moving boundary due to particles forces collision with walls
+      bool enable_moving_boundary;
```
shahabgol, line 266: I think it would be better if you chose a better name for this variable like: "free_boundary_motion"

include/dem/grid_motion.h
```diff
@@ -54,8 +55,9 @@ class GridMotion
    * @param dem_parameters DEM parameters defined using the parameter handler
    * @param dem_time_step DEM time-step
    */
-  GridMotion(const DEMSolverParameters<dim> &dem_parameters,
-             const double &                  dem_time_step);
+  GridMotion(const DEMSolverParameters<dim> &     dem_parameters,
+             const double &                       dem_time_step,
+             std::shared_ptr<PWContactForce<dim>> pw_contact_force_object = {});
```
shahabgol, line 60: Does the GridMotion class need to have access to the pw_contact_force_object? or can we just send the forces from the pw_contact_force to the GridMotion?

source/core/parameters_lagrangian.cc
```diff
@@ -605,21 +606,33 @@ namespace Parameters
...
+                        Patterns::List(Patterns::Double(), 3, 3),
+                        "Boundary inertia around x,y,z axis");
+      prm.declare_entry("triangulation mass",
                         "0",
                         Patterns::Double(),
-                        "Z coordinate of center of mass");
-      prm.leave_subsection();
+                        "Triangulation mass in kg");
 
+      prm.enter_subsection("triangulation motion");
```
shahabgol, line 622: The same comment here: maybe it's better to call this type of motion "free motion"

source/dem/grid_motion.cc
```diff
@@ -51,6 +79,96 @@ GridMotion<dim>::move_grid_translational(
...
+  unsigned int this_mpi_process(
+    Utilities::MPI::this_mpi_process(MPI_COMM_WORLD));
+
+  if (this_mpi_process == 0)
+    {
+      Tensor<1, dim> translational_velocity_one_step_time_further;
+      Tensor<1, 3>   rotational_velocity_one_step_time_further;
+
+      translational_velocity_one_step_time_further =
+        (dt / triangulation_mass) * triangulation_forces +
```
shahabgol, line 109: what if triangulation_mass or intertia are zero? you need to throw a specific error for these conditions

shahabgol (commented):
I just have a few comments on the naming and some questions. Good job.

applications_tests/dem_3d/box_grid_rotation.prm
```diff
@@ -96,6 +96,5 @@ end
 #---------------------------------------------------
 subsection grid motion
     set motion type                 			= rotational
-    set grid rotational speed            		= 1
-    set grid rotational axis		            	= 0
+    set grid rotational speed            		= 1,0,0
```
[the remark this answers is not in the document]
EduardsBlandin: Hm, i think it would be great if the users can use other vectors as (0.2,0.4,0.1).

include/core/parameters_lagrangian.h
EduardsBlandin: Yeah, it would be up to you. Cannot make any change in this branch anymore. But yes, good idea.

source/dem/grid_motion.cc
EduardsBlandin: Yes that's really good point. I let the value be nan for [0] and [2] vector coordinate. I fixed the inertia tensor to be three of length and as default (0,0,0), maybe just a message telling that information in x or y or z axis is wrong but not throwing an error. Like i still wanna do it even with zero information for inertia (for full zero, we throw an error). But for triangulation_mass, if it's zero, yes we should throw an error. Good point again.

include/dem/grid_motion.h
EduardsBlandin: Yes, i think it's true !

shahabgol:
@EduardsBlandin OK, I'll take care of the rest of the works. Can you update me on what is left to be done? And also the experimental data. Can you please share the data with me?

PR #223, opened 2021-08-03
RPT reconstruction of positions

acdaigneault:
This branch allows to do find unknown positions of a radioactive particule for RPT with counts data from experimental.

How it works : 
- Creation of a refine mesh of the reactor vessel
- Determination of vertices positions of each coarse cells & count calculation => store in map
- Searching candidate cells that may contains the particle (in loop for every particles positions)
- Determination of vertices positions of each candidate cells at refined cell level & count calculation for new vertices => store in map
- Searching candidate until the more refined level is reach or if there's no more candidate at refined cell level (may happen since counts have noise because of Monte Carlo)
- If there's still many candidates at the more refined cell level, it calculates the error for each cell with least squared method (it calculates the sum of (error for the counts of unknown position with each vertex for each detector)^2) and keep the cell with lower error.   
- Assignation of the position of the best cell center position to the unknows particle

Everytime there's a new vertex, the count result is stored in a map so it allows to avoided double count calculation for the same vertex.


blaisb (commented):
First batch of comments

include/rpt/rpt_cell_reconstruction.h
```diff
@@ -0,0 +1,169 @@
...
+   * @param level Level of the current cells of the mesh
+   *
+   * @param parent_cell_indexes List of the parent cells which are candidate.
+   *                            Is {-1} by default when executing coarse mesh
+   *                            (no parent cells)
+   */
+  std::vector<int>
+  find_cells(unsigned int         level,
+             std::vector<double> &particle_reconstruction_counts,
+             std::vector<int>     parent_cell_indexes = {-1});
```
blaisb, line 119: Why int? cells are of type types::global_cell_index  (see https://www.dealii.org/current/doxygen/deal.II/classCellAccessor.html)

acdaigneault: Indexes the way I use it seem to be int.
image
image

examples/rpt/count_calculation/positions.particle
```diff
@@ -0,0 +1,150 @@
+0	0	0.02
```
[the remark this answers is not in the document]
acdaigneault: Done :)

applications_tests/rpt_cell_reconstruction_3d/unknown_positions.prm
```diff
@@ -0,0 +1,53 @@
...
+    set activity  						 = 2e6
+    set attenuation coefficient reactor  = 10
+end
+
+
+# --------------------------------------------------
+# Particle reconstruction
+#---------------------------------------------------
+subsection reconstruction
+	set reconstruction                          = true
```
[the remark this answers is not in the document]
acdaigneault: Removed

include/rpt/parameters_rpt.h
```diff
@@ -128,6 +132,21 @@ namespace Parameters
     parse_parameters(ParameterHandler &prm);
   };
 
+  struct RPTReconstructionParameters
+  {
+    // Enable tuning parameters
```
ghazalemir, line 137: Enable tuning the parameter or reconstruct the position?

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,454 @@
...
+  // Seed the random number generator
+  srand(parameters.rpt_param.seed);
+}
+
+template <int dim>
+void
+RPTCellReconstruction<dim>::execute_cell_reconstruction()
+{
+  assign_detector_positions(); // Assign detector position & create detectors
+  read_counts(); // Read reconstruction counts of unknown particle positions
```
ghazalemir, line 31: Read **experimental counts** of unknown particle position I think can be a better explanation!

source/rpt/parameters_rpt.cc
```diff
@@ -186,3 +200,47 @@ Parameters::DetectorParameters::parse_parameters(ParameterHandler &prm)
...
+void
+Parameters::RPTReconstructionParameters::declare_parameters(
+  ParameterHandler &prm)
+{
+  prm.enter_subsection("reconstruction");
+  {
+    prm.declare_entry("reconstruction",
+                      "false",
+                      Patterns::Bool(),
+                      "Enable position reconstruction <true|false>");
```
blaisb, line 213: Is this parameter necessary now that this uses another executable?

blaisb (commented):
I have one file left to review, will do once I get back :)

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,454 @@
...
+  create_grid(); // Read a grid for the reactor vessel
```
ghazalemir, line 32: **Create** a grid for the reactor vessel.

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,501 @@
...
+RPTCellReconstruction<dim>::find_vertices_positions(
+  unsigned int     level,
+  std::vector<int> parent_cell_indexes /* {-1} */)
+{
+  TimerOutput::Scope t(computing_timer, "finding_vertices_positions");
+
+  // Find the positions associated to the vertices of cell
+  for (const auto &cell : triangulation.cell_iterators_on_level(level))
+    {
+      bool parent_is_candidate = false;
```
ghazalemir, line 186: The only case that parent_is_candidate=false happens is when the algorithm wants to find the vertices of the coarse mesh at the beginning?

source/rpt/rpt.cc
```diff
@@ -159,14 +139,62 @@ RPT<dim>::assign_detector_positions()
       detectors.push_back(detector);
     }
 }
+
+template <int dim>
+void
+RPT<dim>::export_data()
```
blaisb, line 145: Good work putting this into a clear function.

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+  Point<dim>       reconstruction_position;
+  double           volume;
+  bool             still_new_candidates = true;
+  std::string      status;
+
+  // Find the first cell candidates from coarse mesh
+  cells_indexes = find_cells(level, particle_reconstruction_counts);
+
+  // Find the positions with the more refine candidate cell
+  while (level < parameters.reconstruction_param.reactor_refinement &&
```
blaisb, line 112: You can query the number of levels from the triangulation itself. It will be safer.
https://www.dealii.org/current/doxygen/deal.II/classTriangulation.html
 use the member method n_levels () const

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+  if (cells_indexes.size() > 1)
+    {
+      best_cell =
+        find_best_cell(level, particle_reconstruction_counts, cells_indexes);
+
+      // Update status
+      if (status.empty())
+        status = "cost_function";
+      else
+        status += "|cost_function";
```
blaisb, line 151: what? Why is the status stored as a string?  There needs to be a comment to explain this :). This is used to print afterwards right?

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+{
+  TimerOutput::Scope t(computing_timer, "finding_vertices_positions");
+
+  // Find the positions associated to the vertices of cell
+  for (const auto &cell : triangulation.cell_iterators_on_level(level))
+    {
+      bool parent_is_candidate = false;
+
+      // Check if parent cell of current level cell is a candidate
+      if (cell->level() > parameters.reconstruction_param.coarse_mesh_level)
```
blaisb, line 235: The coarse mesh is level is the coarsest level from which the search is started I guess?

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+              if (particle_reconstruction_counts[j] >= min &&
+                  particle_reconstruction_counts[j] <= max)
+                candidate_detector[j] = true;
+              else
+                candidate_detector[j] = false;
+            }
+
+          // If reconstruction counts are in ranges of cell for every detector,
+          // cell in candidate
+          if (std::all_of(candidate_detector.begin(),
```
blaisb, line 359: I love this so much

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+  std::string   sep;
+  std::string   filename =
+    parameters.reconstruction_param.reconstruction_positions_file;
+  myfile.open(filename);
+
+  // Headers with space or comma
+  if (filename.substr(filename.find_last_of(".") + 1) == ".dat")
+    {
+      myfile
+        << "unknown_positions_id volumes particle_positions_x particle_positions_y particle_positions_z cell_level";
```
blaisb, line 474: What does unknown_position mean here? The position reconstructed from the detectors?

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+            status_id = 3;
+          else if (cell_status[i_particle] == "wrong_cell")
+            status_id = 4;
+          else if (cell_status[i_particle] == "cost_function|wrong_cell")
+            status_id = 5;
+          else if (cell_status[i_particle] == "parent_cell|wrong_cell")
+            status_id = 6;
+          else if (cell_status[i_particle] ==
+                   "parent_cell|cost_function|wrong_cell")
+            status_id = 7;
```
blaisb, line 534: Maybe use a switch case instead of bunch of ifs? Very interesting way of setting status though. You could have used a mask instead of a long string (the ifs would have been faster), but I don't think that's a big deal.

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+                 << distance << std::endl;
+        }
+      else
+        myfile << std::endl;
+    }
+}
+
+template <int dim>
+void
+RPTCellReconstruction<dim>::assign_detector_positions()
```
blaisb, line 547: This is also present in the rpt.cc file no? It would be better to move it to a seperate file and define it as a function instead.

source/rpt/rpt_cell_reconstruction.cc
```diff
@@ -0,0 +1,626 @@
...
+                             middle_point);
+
+      detectors.push_back(detector);
+    }
+}
+
+template <int dim>
+void
+RPTCellReconstruction<dim>::read_known_positions()
+{
```
blaisb, line 585: Same thing, I would move this function to a different file, this way all solvers can use it :).

blaisb (commented):
Good work. Only some minor/medium comments there and there. Good job. Some very nice solutions there and there.

source/rpt/rpt_cell_reconstruction.cc
acdaigneault: Yes it's better to change that. It's funny because I actually implemented the strings and then the status. 
By "mask", do you mean a id number associate to each case that can be processed with
 switch case or determined the status directly in the function?

source/rpt/rpt_cell_reconstruction.cc
acdaigneault: Good idea!

include/rpt/parameters_rpt.h
```diff
@@ -48,6 +48,9 @@ namespace Parameters
     // Particle positions filename
     std::string particle_positions_file;
 
+    // Show results in terminal during computation
+    bool verbose;
```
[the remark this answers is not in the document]
acdaigneault: Yeah, I was lazy on this one :p

source/rpt/rpt_cell_reconstruction.cc
blaisb: Well a mask is generally a lot of booleans in a single variable. For example 0000 means something false false false false, 0001 means false false false true, etc. This is a good way when you have multiple independent bools. I don't think this is required here.

source/rpt/rpt_cell_reconstruction.cc
acdaigneault: Yes, or the list of candidate cells is empty (fail to find a cell at a higher refined level)
image

acdaigneault: Will add explaination

acdaigneault:
I did all the modifications but I still need to test every examples and tests :)

blaisb:
Looks good to me! When you and @ghazalemir feel ready for a merge we can do it :)

PR #224, opened 2021-08-04
Dem post processing

shahabgol:
This PR adds post-processing calculations to the DEM solver.
Now, we can calculate average velocity of particles in each cell, as well as the granular temperature distribution.
I will add post-processing of tracer particles in a future PR.

### DIFF
--- a/source/core/parameters_lagrangian.cc
+++ b/source/core/parameters_lagrangian.cc
@@ -1078,12 +1078,12 @@ namespace Parameters
                           "Enable calculation of granular temperature.");
 
         prm.declare_entry("initial step",
-                          "0.0",
+                          "0",
                           Patterns::Integer(),
                           "Initial step to start Lagrangian post-processing.");
 
         prm.declare_entry("end step",
-                          "0.0",
+                          "0",
                           Patterns::Integer(),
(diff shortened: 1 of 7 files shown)


source/core/parameters_lagrangian.cc
```diff
@@ -1056,6 +1056,74 @@ namespace Parameters
...
+    {
+      prm.enter_subsection("post-processing");
+      {
+        Lagrangian_post_processing = prm.get_bool("Lagrangian post processing");
+        calculate_particles_average_velocity =
+          prm.get_bool("calculate particles average velocity");
+        calculate_granular_temperature =
+          prm.get_bool("calculate granular temperature");
+        initial_step              = prm.get_double("initial step");
+        end_step                  = prm.get_double("end step");
```
blaisb, line 1119: Why are those doubles? Should they not be unsigned integers since they are iteration numbers?

blaisb (commented):
A single small comment. Good job!

source/core/parameters_lagrangian.cc
shahabgol: Corrected. :)

blaisb (approved):

PR #227, opened 2021-08-07
fix bdf2 and add vf degree option

tonielgeitani:


blaisb (commented):
All good except one thing @tonielgeitani . Can you move the void fraction order declaration in the FEM section where all the other variables order are declared instead of in the void fraction section? I'd rather all the FEM order stuff be at the same location.

PR #233, opened 2021-08-18
Add example warming up a viscous dissipation

jeannejoachim:
Add option for viscous_dissipation (now disabled by default) and fix according tests.
Add prm and README files in subfolder in example/multiphysics.
See the wiki page of the example for more information.

### DIFF
--- a/examples/multiphysics/warming_up_viscous_fluid/warming_up_viscous_fluid.prm
+++ b/examples/multiphysics/warming_up_viscous_fluid/warming_up_viscous_fluid.prm
@@ -6,27 +6,26 @@
 # Simulation Control
 #---------------------------------------------------
 subsection simulation control
-  set method                  = steady
-  set number mesh adapt       = 2
-  set output name             = mms_22
-  set output frequency        = 1
-  set subdivision             = 1
+  set method                  = bdf2
+  set time step               = 0.05
+  set time end                = 7.0     
+  set output name             = warming-up
+  set output frequency        = 1       
 end
-
 #---------------------------------------------------
 # FEM
 #---------------------------------------------------
 subsection FEM
     set velocity order        = 1
     set pressure order        = 1
-    set temperature order        = 1
+    set temperature order     = 2
 end
 
 #---------------------------------------------------
 # Timer
 #---------------------------------------------------
 subsection timer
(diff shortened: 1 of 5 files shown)


examples/multiphysics/warming_up_viscous_fluid/warming_up_viscous_fluid.prm
```diff
@@ -0,0 +1,149 @@
...
+end
+
+subsection boundary conditions heat transfer
+  set number                  = 2
+    subsection bc 0
+    set id = 0
+        set type          = convection
+	set h             = 0
+	set Tinf	  = 0
+    end
```
blaisb, line 120: Why a convection bc with h=0 & T_inf=0?

source/core/parameters_multiphysics.cc
```diff
@@ -29,6 +29,12 @@ Parameters::Multiphysics::declare_parameters(ParameterHandler &prm)
                       "false",
                       Patterns::Bool(),
                       "Free surface calculation <true|false>");
+
+    // subparameter for heat_transfer
+    prm.declare_entry("viscous dissipation",
+                      "false",
+                      Patterns::Bool(),
+                      "Viscous thermic dissipation calculation <true|false>");
```
blaisb, line 37:
```suggestion
                      "Viscous dissipation in heat equation <true|false>");
```

blaisb (commented):
Two tiny comments to address

examples/multiphysics/warming_up_viscous_fluid/warming_up_viscous_fluid.prm
jeannejoachim: For `bc 0` we have a thermic insulation condition, so `h=0` defines a "no heat transfer" condition. However `T_inf` does not have to be given. If you find this misleading we can remove it!

blaisb: Well it's just that you could have specified no boundary condition and it would have been the same (a neumann instead of a Robin with h=0), but it's equivalent, so let's leave it at that.

blaisb (approved):

PR #236, opened 2021-08-23
Add DCDD shock capture to free surface

blaisb:
This adds DCDD shock capturing to the free surface solver. It's a bit diffusive, but it really helps dam breaks test cases and the likes.


include/solvers/tracer.h
```diff
@@ -328,6 +328,7 @@ class Tracer : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
   // Tracer statistics table
   TableHandler statistics_table;
 
+  // Enable DCDD shock capturing scheme
   const bool DCDD = true;
```
jeannejoachim, line 332: so you'll want to implement it in tracer.cc as well?

blaisb: It is already implemented in the tracer. This is where I took it. It's been merged in a past PR.

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
+                                   pow(phase_gradients[q].norm() * h, order);
+
+              // Tensor<1, dim> s = velocity / (velocity.norm() + 1e-12);
+              Tensor<1, dim> r =
+                phase_gradients[q] / (phase_gradients[q].norm() + 1e-12);
+
+              // const Tensor<2, dim> k_corr      = (r * s) * outer_product(s,
+              // s);
+              const Tensor<2, dim> rr          = outer_product(r, r);
+              const Tensor<2, dim> dcdd_factor = rr; // - k_corr;
```
jeannejoachim, line 249: so remove the variable `rr`? `dcdd_factor = outer_product(r, r);`

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
+
+
+              const double d_vdcdd =
```
jeannejoachim, line 252: what is it? could you comment that?

source/solvers/free_surface.cc
```diff
@@ -267,6 +305,21 @@ FreeSurface<dim>::assemble_system(
                           cell_matrix(i, j) +=
                             tau * strong_jacobian *
                             (grad_phi_phase_i * velocity_values[q]) * JxW;
+
+                          if (DCDD)
+                            {
+                              cell_matrix(i, j) +=
```
jeannejoachim, line 311: why put it in two sets of equations? I guess it is for code clarity, and it's a good idea, but then could you comment on the two equations? (state what they are, or give a reference)

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
+              for (unsigned int k : fe_values_fs.dof_indices())
+                {
+                  phi_phase[k]      = fe_values_fs.shape_value(k, q);
+                  grad_phi_phase[k] = fe_values_fs.shape_grad(k, q);
+                }
+
+
+              // Shock capturing viscosity term
+              const double order =
+                this->simulation_parameters.fem_parameters.tracer_order;
```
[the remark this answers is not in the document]
blaisb: it's a mistake, it should have been free_surface_order.

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
+              const double vdcdd = (0.5 * h) *
+                                   (velocity.norm() * velocity.norm()) *
+                                   pow(phase_gradients[q].norm() * h, order);
+
+              // Tensor<1, dim> s = velocity / (velocity.norm() + 1e-12);
+              Tensor<1, dim> r =
+                phase_gradients[q] / (phase_gradients[q].norm() + 1e-12);
+
+              // const Tensor<2, dim> k_corr      = (r * s) * outer_product(s,
+              // s);
```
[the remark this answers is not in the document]
blaisb: I'm keeping them just in case if this is required. I'm still debating whether we need this formulation or not.

source/solvers/free_surface.cc
blaisb: I'd like to keep the capacity for both approaches.

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
+              const double order =
+                this->simulation_parameters.fem_parameters.tracer_order;
+
+              const double vdcdd = (0.5 * h) *
+                                   (velocity.norm() * velocity.norm()) *
+                                   pow(phase_gradients[q].norm() * h, order);
+
+              // Tensor<1, dim> s = velocity / (velocity.norm() + 1e-12);
+              Tensor<1, dim> r =
+                phase_gradients[q] / (phase_gradients[q].norm() + 1e-12);
```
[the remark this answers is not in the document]
blaisb: It is to prevent division zero and  the occurence of NANs. I can add a variable.

source/solvers/free_surface.cc
```diff
@@ -215,6 +223,36 @@ FreeSurface<dim>::assemble_system(
...
 
+              // Gather the shape functions and their gradient
+              for (unsigned int k : fe_values_fs.dof_indices())
+                {
+                  phi_phase[k]      = fe_values_fs.shape_value(k, q);
+                  grad_phi_phase[k] = fe_values_fs.shape_grad(k, q);
+                }
+
+
+              // Shock capturing viscosity term
```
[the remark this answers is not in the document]
blaisb: Can't because then the variables will be defined in ifs and it won't be possible to compile.

source/solvers/free_surface.cc
blaisb: Merged into a single equation. It's just two equations because it's the jacobian of a product of two velocity functions.

blaisb:
@jeannejoachim All comments have been applied.
If you don't have more, i'll merge after the CI passes.

PR #237, opened 2021-08-25
Lethe grid tools

emilebergeron:
This is the first version of LetheGridTools, a namespace designed to add features related to triangulation manipulation. It is inspired by deal.ii's GridTools, and mostly provides methods designed to allow a user to select a specific set of cells in a triangulation. While the code is functional, is has not been profiled and will definitely be changed in the following weeks to improve performance. The API is also subject to change if needed. 

I'm creating this pull request mostly to avoid conflicts with the upcoming changes in indentation style.


include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
+//
+// Created by lucka on 2021-07-15.
+//
```
blaisb, line 3: Add license information.
Also remove author name. I am trying to move out of "file belonging to someone" mindset.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+
+#ifndef LETHE_LETHEGRIDTOOLS_H
+#define LETHE_LETHEGRIDTOOLS_H
```
blaisb, line 6: We use lower case letters for include guards now.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+#include <unordered_set>
+
+using namespace dealii;
+namespace LetheGridTools
+{
+    /**
+    * @brief
+    * Map the vertex index to the cell that includes that vertex.
+    * This map is used to find all the cell close to a specific vertex.
+    */
```
blaisb, line 26: Add comments in doxygen format to explain the arguments.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+    template <int dim>
+    void
+    vertices_cell_mapping(const DoFHandler<dim> &dof_handler,
+            std::map<unsigned int,std::set<typename DoFHandler<dim>::active_cell_iterator>> & vertices_cell_map);
+
+    template <int dim>
+    typename DoFHandler<dim>::active_cell_iterator
+    find_cell_around_point_with_tree(const DoFHandler<dim> &dof_handler,
+                                     Point<dim>             &point);
```
blaisb, line 35: Add comments for this part.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+      * @param cell , The initial cell. We suspect the point of being in one of the neighbours of this cell.
+      *
+      * @param point, The point that we want to find the cell that contains it
+      */
+
+    template <int dim>
+    typename DoFHandler<dim>::active_cell_iterator
+    find_cell_around_point_with_tree_with_guess(const DoFHandler<dim> &dof_handler,
+            const typename DoFHandler<dim>::active_cell_iterator &cell,
+                                     Point<dim>             &point);
```
blaisb, line 51: Is the point modified? If so explain why/how and if not, make it const.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+    find_cell_around_point_with_neighbors(const DoFHandler<dim> &dof_handler,
+            std::map<unsigned int,std::set<typename DoFHandler<dim>::active_cell_iterator>> &vertices_cell_map,
+            const typename DoFHandler<dim>::active_cell_iterator &cell,
+            Point<dim>                                           &point);
+
+    /**
+      * @brief
+      *Return a vector of cells around a cell including vertex neighbors
+      *
+      * @param cell , The initial cell. we want to know all the cells that share a vertex with this cell.
```
blaisb, line 64: Missing a parameter explanation here.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,150 @@
...
+      */
+    template <int dim>
```
blaisb, line 66: All functions below don't have documentation.

include/solvers/gls_sharp_navier_stokes.h
```diff
@@ -377,6 +377,9 @@ Return a bool that describes  if a cell contains a specific point
 
 
 
+  void
+  box_refine();
+
```
blaisb, line 382: Need comments.

include/solvers/navier_stokes_base.h
```diff
@@ -320,6 +320,13 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   void
   refine_mesh();
 
+  /**
+   * @brief Allow the initial refinement of all cells of the principal mesh that are partially
+   * contained in one of the cells of the box refinement mesh given in the parameter.
+   */
+  void
+  box_refine_mesh();
+
```
blaisb, line 329: Why is box_refine in gls_sharp and box_refine_mesh in the base class?

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+#include <deal.II/fe/fe_q.h>
+#include <cmath>
+
+
+template <int dim>
+void
+LetheGridTools::vertices_cell_mapping(const DoFHandler<dim> &dof_handler,
+        std::map<unsigned int,std::set<typename DoFHandler<dim>::active_cell_iterator>> & vertices_cell_map)
+{
+    // Find all the cells around each vertices
```
blaisb, line 18:
```suggestion
```
This belongs in the header file.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+    }
+}
+
+
+template <int dim>
+typename DoFHandler<dim>::active_cell_iterator
+LetheGridTools::find_cell_around_point_with_tree(const DoFHandler<dim> &dof_handler,
+                                 Point<dim>             &point)
+{
+    // find cell around point using tree properties instead of looping on all cell
```
blaisb, line 50:
```suggestion
```
Move to header.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+
+    // define stuff for the search
```
blaisb, line 52: Don't call it stuff, explain what it is.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+    MappingQ1<dim> map;
```
blaisb, line 53:
```suggestion
    MappingQ1<dim> mapping;
```
don't call it a map, otherwise it becomes confusing with the std::map container.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+    for (const auto &cell : cell_iterator)
+    {
+        try
+        {
+            const Point<dim, double> p_cell =
+                    map.transform_real_to_unit_cell(cell, point);
+
+            const double dist = GeometryInfo<dim>::distance_to_unit_cell(p_cell);
+
+            if (dist == 0)
```
blaisb, line 70: Don't do equation tests on double, this is a very fragile way of testing.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+            {
+                // cell on lvl 0 found
+                cell_0_found   = true;
+                best_cell_iter = cell;
```
blaisb, line 74: Why not break out of the loop if that happens then? Would prevent you from looping  a little bit.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+            }
+        }
+        catch (const typename MappingQGeneric<dim>::ExcTransformationFailed &)
+        {}
+    }
+
+    if (cell_0_found)
```
blaisb, line 81: Why this if? What happens if cell_0_found==false? The function does nothing? This is weird. It would be better to have if (!cell_0_found) return;  than to use this large encompassing if. This hinders readability.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+{
+    // find cell around point using tree properties instead of looping on all cell
+
+    // define stuff for the search
+    MappingQ1<dim> map;
+    const auto &   cell_iterator = dof_handler.cell_iterators_on_level(0);
+    unsigned int   max_childs    = GeometryInfo<dim>::max_children_per_cell;
+    typename DoFHandler<dim>::cell_iterator best_cell_iter;
+
+    bool cell_0_found = false;
```
blaisb, line 58:
```suggestion
    bool cell_on_level_0_found = false;
```
Use more descriptive names.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+typename DoFHandler<dim>::active_cell_iterator
+LetheGridTools::find_cell_around_point_with_tree(const DoFHandler<dim> &dof_handler,
+                                 Point<dim>             &point)
+{
+    // find cell around point using tree properties instead of looping on all cell
+
+    // define stuff for the search
+    MappingQ1<dim> map;
+    const auto &   cell_iterator = dof_handler.cell_iterators_on_level(0);
+    unsigned int   max_childs    = GeometryInfo<dim>::max_children_per_cell;
```
blaisb, line 55: Define this variable where you need it.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+        {}
+    }
+
+    if (cell_0_found)
+    {
+        // the cell on lvl 0 contain the point so now loop on the childs of
+        // this cell when we found the child of the cell that containt it we
+        // stop and loop if the cell is active if the cell is not active loop
+        // on the child of the cell. Repeat
+        unsigned int lvl = 0;
```
blaisb, line 87:
```suggestion
        unsigned int level = 0;
```

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+}
+
+template <int dim>
+typename DoFHandler<dim>::active_cell_iterator
+LetheGridTools::find_cell_around_point_with_neighbors(const DoFHandler<dim> &dof_handler,
+                                                      std::map<unsigned int,std::set<typename DoFHandler<dim>::active_cell_iterator>> &vertices_cell_map,
+        const typename DoFHandler<dim>::active_cell_iterator &cell,
+        Point<dim>                                            &point)
+{
+    // Find the cell around a point based on an initial cell.
```
blaisb, line 142:
```suggestion
    // Find the cell around a point based on an initial cell.
```
Move to header.

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+        }
+    }
+
+
+    return best_cell_iter;
+}
+
+template <int dim>
+typename DoFHandler<dim>::active_cell_iterator
+LetheGridTools::find_cell_around_point_with_neighbors(const DoFHandler<dim> &dof_handler,
```
blaisb, line 137: There is a structure callled GridToolsCache https://www.dealii.org/current/doxygen/deal.II/classGridTools_1_1Cache.html  . Have you thought about using it instead of passing stuff all around ?

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+                                                        const typename DoFHandler<dim>::active_cell_iterator &cell)
+{
+    // Find all the cells that share a vertex with a reference cell including the
+    // initial cell.
+    //std::unordered_set<typename DoFHandler<dim>::active_cell_iterator, LetheGridTools::hash_cell<dim>, LetheGridTools::equal_cell<dim>> neighbors_cells;
+    std::set<typename DoFHandler<dim>::active_cell_iterator> neighbors_cells;
+
+    // Loop over the vertices of the initial cell and find all the cells around
+    // each vertex and add them to the set of cells around the reference cell.
+    for (unsigned int i = 0; i < GeometryInfo<dim>::vertices_per_cell; i++)
```
blaisb, line 175: Be careful, this only works for the case of Hexahedral triangulation. It's not that big of a deal but it could eventually be. https://www.dealii.org/developer/doxygen/deal.II/group__simplex.html#details

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,982 @@
...
+    }
+    // Transform the set into a vector.
+    std::vector<typename DoFHandler<dim>::active_cell_iterator>
+            cells_sharing_vertices(neighbors_cells.begin(), neighbors_cells.end());
+    return cells_sharing_vertices;
+}
+
+template <int dim>
+std::vector<typename DoFHandler<dim>::active_cell_iterator>
+LetheGridTools::find_cells_in_cells(const DoFHandler<dim> &dof_handler_1,
```
blaisb, line 189: What is dof handler 1? This has to be documented a bit better.

blaisb (commented):
Review is still a WIP, but I already have multiple comments.

blaisb (commented):
Finished everything but the .cc of lethe grid tools.

source/core/lethegridtools.cc
emilebergeron: Do we have exception handling in Lethe? The situation that happens here is that if you give a point that is outside the triangulation, the function is not really valid in the sense that you can't return a valid cell iterator.

blaisb: Well, we do have some exception handling through deal.II, but we don't necessarily want to continue the simulation if things broke. What I would do is have an assertThrow here that would stop the simulation. At least it is useful in debug mode.

include/solvers/navier_stokes_base.h
Luckabarbeau: error now corrected

source/core/lethegridtools.cc
Luckabarbeau: good point

source/core/lethegridtools.cc
Luckabarbeau: corrected it now throw if the point is not in the level 0 of the mesh.

source/core/lethegridtools.cc
Luckabarbeau: This could be implemented in an upcoming modification with other modifications. But I wouldn't add this to the current PR.

source/core/lethegridtools.cc
Luckabarbeau: Good point I suggest we add this option in the next PR as well.

source/solvers/navier_stokes_base.cc
```diff
@@ -654,6 +659,187 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh()
     }
 }
 
+template <int dim, typename VectorType, typename DofsType>
+void
+NavierStokesBase<dim, VectorType, DofsType>::box_refine_mesh()
```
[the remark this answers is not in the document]
Luckabarbeau: At the moment the function in Lethe grid tool need a dof_handler but this could be changed in a future update and then moved to the mesh definition.

tests/core/lethe_grid_tool_mesh_cut_by_flat_2.cc
```diff
@@ -0,0 +1,180 @@
...
+
+    std::map<unsigned int,std::set<typename DoFHandler<2>::active_cell_iterator>> vertice_to_cell;
+
+    LetheGridTools::vertices_cell_mapping(dof_handler, vertice_to_cell);
+    std::vector<typename DoFHandler<2>::active_cell_iterator> cells_cut;
+
+    for (const auto &flat_cell :
+            flat_dof_handler.active_cell_iterators()) {
+        cells_cut = LetheGridTools::find_cells_around_flat_cell(dof_handler, flat_cell, vertice_to_cell);
+    }
```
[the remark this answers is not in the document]
Luckabarbeau: corrected

source/core/parameters.cc
```diff
@@ -750,6 +750,97 @@ namespace Parameters
...
+                              "GMSH file name");
+
+            prm.declare_entry("initial refinement",
+                              "0",
+                              Patterns::Integer(),
+                              "Initial refinement of the principal mesh");
+            prm.declare_entry("box initial refinement",
+                              "0",
+                              Patterns::Integer(),
+                              "Initial refinement of the box mesh");
```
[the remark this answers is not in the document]
Luckabarbeau: This is used in the case we are using a sphere or any curved mesh to define the box zone. In these cases if we want to have a smoother and rounder mesh refinement zone, we need to use a more refined mesh for the box. So we give the option to refine the mesh of the box. This is different from the refinement of the main mesh inside the box.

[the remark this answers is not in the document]
blaisb: Yeah but what I mean is that the parameter block you are using here, it could be a Mesh parameter block, it does not have to be a new type of block. You can use the same structure under a different variable name! Here you duplicated a bunch of code just for a synonym.

include/core/lethegridtools.h
```diff
@@ -0,0 +1,245 @@
...
+  std::vector<typename DoFHandler<dim>::active_cell_iterator>
+  find_cells_around_cell(
+    std::map<unsigned int,
+             std::set<typename DoFHandler<dim>::active_cell_iterator>>
+      &                                                   vertices_cell_map,
+    const typename DoFHandler<dim>::active_cell_iterator &cell);
+
+  /**
+   * @brief
+   *Return a vector of cells around a flat cell in spacedim. A flat cell as a
```
blaisb, line 102:
```suggestion
   *Return a vector of cells around a flat cell in spacedim. A flat cell has a
```

include/core/lethegridtools.h
```diff
@@ -0,0 +1,245 @@
...
+  find_cells_around_flat_cell(
+    const DoFHandler<dim> &                                        dof_handler,
+    const typename DoFHandler<dim - 1, dim>::active_cell_iterator &cell,
+    std::map<unsigned int,
+             std::set<typename DoFHandler<dim>::active_cell_iterator>>
+      &vertices_cell_map);
+
+  /**
+   * @brief
+   *Return a vector of cells around a edge cell in spacedim. A edge cell as a
```
blaisb, line 121:
```suggestion
   *Return a vector of cells around an edge cell in spacedim. An edge cell as a
```

include/core/lethegridtools.h
```diff
@@ -0,0 +1,245 @@
...
+   */
+  template <int dim>
+  std::vector<typename DoFHandler<dim>::active_cell_iterator>
+  find_cells_in_cells(
+    const DoFHandler<dim> &                               dof_handler,
+    const typename DoFHandler<dim>::active_cell_iterator &cell);
+
+  /**
+   * @brief
+   * Return a bool that say if a cell is cut by a flat cell
```
blaisb, line 156:
```suggestion
   * Return a bool that indicates if a cell is cut by a flat cell
```

include/core/lethegridtools.h
```diff
@@ -0,0 +1,245 @@
...
+  struct equal_cell
+  {
+    std::size_t
+    operator()(const typename DoFHandler<dim>::active_cell_iterator &cell1,
+               const typename DoFHandler<dim>::active_cell_iterator &cell2)
+      const noexcept
+    {
+      return cell1->global_active_cell_index() ==
+             cell2->global_active_cell_index();
+    }
```
blaisb, line 239: Add comments to define these hashing functions

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -590,17 +505,37 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                                        fe_face_projection_values.JxW(q) *
+                                        fe_face_projection_values.shape_value(
+                                          i, q);
+                                      local_weight +=
+                                        fe_face_projection_values.shape_value(
+                                          i, q);
+                                      /*std::cout<<"i_face_tensor "<<
+                                         local_face_tensor[i]<<" fe  "<<
+                                         fe_face_projection_values.shape_value(
+                                              i, q)<<std::endl;*/
```
blaisb, line 526:
```suggestion
 
```

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,905 @@
+//
+// Created by lucka on 2021-07-15.
+//
+
```
blaisb, line 4:
```suggestion
```
We don't give individual names to file anymore

source/core/lethegridtools.cc
```diff
@@ -0,0 +1,905 @@
...
+    {
+      bool inside_cell = active_neighbors_set[i]->point_inside(point);
+      if (inside_cell)
+        {
+          return active_neighbors_set[i];
+        }
+    }
+  // The cell is not found near the initial cell so we use the cell tree
+  // algorithm instead (much slower).
+  std::cout << "Cell not found around " << point << std::endl;
```
blaisb, line 171: Try to avoid using std::cout in the code, what happens if it running in parallel? It will print like crazy.

blaisb (approved):
I don't have specific algorithmic concerns, however I am not sure of the speed of the implementation. This is something that i suggest we benchmark in a next PR.
The following things need to be done:
1- Fix spelling in some of the comments
2- Apply automatic indentation

Then I can merge. Next step will be to benchmark this in a realistic situation.

tests/core/lethe_grid_tool_mesh_cut_by_flat_3.cc
```diff
@@ -0,0 +1,173 @@
...
+  data_out.attach_dof_handler(dof_handler);
+  data_out.add_data_vector(subdomain, "subdomain");
+  data_out.build_patches();
+  std::ofstream output("solution.vtu");
+  data_out.write_vtu(output);
+  DataOut<2, 3> flat_data_out;
+  flat_data_out.attach_dof_handler(flat_dof_handler);
+  flat_data_out.build_patches();
+  std::ofstream flat_output("flat_trig.vtu");
+  flat_data_out.write_vtu(flat_output);
```
blaisb, line 133: Do you really need to have these entire blocks into the #ifdef . Why not just have the declaration of the data_out and then remove code duplication? Same applies for the other test.

source/core/lethegridtools.cc
Luckabarbeau: This will append only in one process as it's related to only one point. So when there is a problem only one proc shows this message.

PR #240, opened 2021-09-01
Vans scratch assembly

tonielgeitani:
Migration of the vans_solver into an assembler mechanism


source/fem-dem/vans_assemblers.cc
```diff
@@ -0,0 +1,607 @@
...
+          const auto div_phi_u_i = scratch_data.div_phi_u[q][i];
+          double     local_rhs_i = 0;
+          for (unsigned int p = 0; p < number_of_previous_solutions(method) + 1;
+               ++p)
+            {
+              local_rhs_i -=
+                (bdf_coefs[p] * void_fraction[0] * (velocity[p] * phi_u_i) +
+                 bdf_coefs[p] * (void_fraction[p] * scratch_data.phi_p[q][i]));
+
+              if (cfd_dem.grad_div == true)
```
blaisb, line 427: Good :)

source/fem-dem/vans_assemblers.cc
```diff
@@ -0,0 +1,607 @@
...
+    }
+}
+
+template class GLSVansAssemblerBDF<2>;
+template class GLSVansAssemblerBDF<3>;
+
+
+template <int dim>
+void
+GLSVansAssemblerPFI<dim>::calculate_particle_fluid_interactions(
```
blaisb, line 444: So you made a single assembler for all of the particle_fluid_interaction. This is ok for now, but I think it would be best to eventually move to one assembler for each particle fluid interaction term.
One for each drag model, one for the pressure force, etc. You can do this in a next PR since this is already a big PR.

blaisb (approved):
Good job. Just be careful about how you assemble particle_fluid interaction. See my comment on this.

PR #241, opened 2021-09-13
Add kinsol nonlinear solver

lpsaavedra:
This PR introduces the kinsol nonlinear solver in Lethe as an alternative to the traditional newton solver. The goal is to have an efficient nonlinear solver that reduces the number of times the matrix is assembled and the preconditioner is computed. The main changes can be summarized as follows:

- The SkipNewtonNonLinearSolver class is replaced by the  KinsolNewtonNonLinearSolver class
- Physics Solver requires two separate functions for the assembly of the matrix and the rhs (in accordance with the new scratch assembly mechanism). 
- The setup of the preconditioners in the GLS Navier Stokes solver is now done after the matrix assembly (this should be taken into account when other derived solvers are migrated to the new assembly).

**Note:** It requires deal.ii to be compiled with Sundials version 5.7.


blaisb (commented):
Seems very good. Can you just fix the breaking KINSOL test, then we could move on to a merge?

PR #248, opened 2021-09-29
Forced cut mesh tests to output cells in increasing order

emilebergeron:
This fixes a bug caused by the fact that the order of the items in an unordered map is not guaranteed to be the same on all machines


blaisb (approved):
Good job! Makes the test significantly more portable. We will have to keep this in mind in the future.

PR #249, opened 2021-10-01
Table and check point

Luckabarbeau:
Add table read function and checkpoint for sharp immersed boundaries.


source/core/utilities.cc
```diff
@@ -94,7 +94,105 @@ make_table_tensors_scalars(
   return table;
 }
 
+void
+fill_table_from_file(TableHandler &    table,
+                     std::string       file,
+                     const std::string delimiter)
+{
+  std::string line;
```
blaisb, line 102: Can you declare line in the if itself if you don't need it outside of the if

source/core/utilities.cc
```diff
@@ -94,7 +94,105 @@ make_table_tensors_scalars(
...
+                {
+                  table.add_value(vector_of_column_names[i], line_of_data[i]);
+                }
+            }
+          catch (...)
+            {
+              // the line contains words we assume these are the column
+              vector_of_column_names = list_of_words_clean;
+            }
+        }
```
blaisb, line 138: This is a bit weird, why do it with this way where you test every line?
You know the first line contains the word, and then the rest of the lines contains the data. It would be a lot better to just read a first line, store the words, then read afterwards and store the data. The current way is worse, because for every single line you end up testing everything for no reason. This is also error prone.

source/core/utilities.cc
```diff
@@ -94,7 +94,105 @@ make_table_tensors_scalars(
...
+  std::ifstream myfile(file);
+  // open the file.
+  if (myfile.is_open())
+    {
+      std::vector<std::string> vector_of_column_names;
+      std::vector<double>      line_of_data;
+
+      while (std::getline(myfile, line))
+        {
+          // read the line and clean the resulting vector.
```
blaisb, line 164: Same comment as above. Read the first line, then read the rest.

tests/core/CMakeLists.txt
```diff
@@ -4,4 +4,13 @@ INCLUDE_DIRECTORIES(
...
+
+string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_LOWER_BUILD_TYPE)
+	file(COPY table_read_input.dat DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/table_read.${CMAKE_LOWER_BUILD_TYPE})
+
 DEAL_II_PICKUP_TESTS()
+
+
+
+
+
```
blaisb, line 16:
```suggestion
```

tests/core/table_read.cc
```diff
@@ -0,0 +1,101 @@
...
+// Lethe
+#include <core/utilities.h>
+
+#include <deal.II/numerics/data_out.h>
+
+// Tests (with common definitions)
+#include <../tests/tests.h>
+
+
+
```
blaisb, line 41: Add a comment block here that explains what the test does.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1951,6 +1943,253 @@ GLSSharpNavierStokesSolver<dim>::copy_local_rhs_to_global_rhs(
...
+}
+
+template <int dim>
+void
+GLSSharpNavierStokesSolver<dim>::read_checkpoint()
+{
+  this->GLSNavierStokesSolver<dim>::read_checkpoint();
+
+  TimerOutput::Scope t(this->computing_timer,
+                       "Reset Nitsche solid mesh and particles");
```
blaisb, line 2049:
```suggestion
                       "Reset Sharp-Edge particle information");
```

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1951,6 +1943,253 @@ GLSSharpNavierStokesSolver<dim>::copy_local_rhs_to_global_rhs(
...
+  std::vector<double> p_x =
+    restart_data.second[std::find(restart_data.first.begin(),
+                                  restart_data.first.end(),
+                                  "p_x") -
+                        restart_data.first.begin()];
+  std::vector<double> p_y =
+    restart_data.second[std::find(restart_data.first.begin(),
+                                  restart_data.first.end(),
+                                  "p_y") -
+                        restart_data.first.begin()];
```
blaisb, line 2084: I don't really like this copy paste of stuff. Maybe a good strategy could have been to have a map where they key is the string and the value is a reference to the vector. This way the std::find could be inside of a loop over the map.
In all cases, this is something we can try later to clean once this is merged in a follow-up PR.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1951,6 +1943,253 @@ GLSSharpNavierStokesSolver<dim>::copy_local_rhs_to_global_rhs(
...
+        }
+      // Write the table in the checkpoint file.
+      particles_information_table.write_text(output);
+    }
+}
+
+template <int dim>
+void
+GLSSharpNavierStokesSolver<dim>::read_checkpoint()
+{
```
blaisb, line 2045: Add an application_test for this, Just a restart of a single particle falling with a single iteration...

blaisb (commented):
Good work, but there are still stuff missing:
- Better reading/writing mechanism instead of try/catch every line
- Add an application_test for your restart! Otherwise it might break without you noticing.
That's why it's good to do smaller PRs :)!

source/core/utilities.cc
```diff
@@ -94,7 +94,113 @@ make_table_tensors_scalars(
...
+{
+  table.clear();
+  std::ifstream myfile(file);
+  // open the file
+  if (myfile.is_open())
+    {
+      std::string              line;
+      std::vector<std::string> vector_of_column_names;
+      std::vector<double>      line_of_data;
+      unsigned int             i = 0;
```
blaisb, line 110: Can you use another variable than "i" here. Because otherwise you have two i variables in the same functions. i understand that they are well defined because of {} encapsulation, but it really hinders readability.

source/core/utilities.cc
```diff
@@ -94,7 +94,113 @@ make_table_tensors_scalars(
...
+
+
+  std::ifstream myfile(file);
+  // open the file.
+  if (myfile.is_open())
+    {
+      std::string              line;
+      std::vector<std::string> column_names;
+      std::vector<double>      line_of_data;
+      unsigned int             i = 0;
```
blaisb, line 163: Same comment as above, use another variable line to make sure both can be distinguished.

blaisb (commented):
There is still an issue with the "i" index part. Please look into my comments.
Still missing an application test for the restart with sharp edge :)

PR #250, opened 2021-10-01
Get residual for nonlinear solver

Luckabarbeau:
Add a get_current_residual method in the physical solver and an override in the gls_shapr in order to take into account the particle residual. At the moment this variable is useless but upcoming change will make use of the particle residual.

### DIFF
--- a/include/core/newton_non_linear_solver.h
+++ b/include/core/newton_non_linear_solver.h
@@ -159,7 +159,7 @@ NewtonNonLinearSolver<VectorType>::solve(const bool is_initial_step)
             }
           last_alpha_res = current_res;
         }
-      
+
       global_res       = solver->get_current_residual();
       present_solution = evaluation_point;
       last_res         = current_res;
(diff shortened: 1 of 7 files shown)


include/core/newton_non_linear_solver.h
```diff
@@ -65,19 +65,28 @@ template <typename VectorType>
...
   last_res                     = 1e6;
   current_res                  = 1e6;
+  global_res                   = 1e6;
+
+  // current_res and global_res are different as one is defined based on the l2
+  // norm of the residual vector (current_res) and the other (global_res) is
+  // defined by the physical solver and may differ from the l2_norm of the
+  // residual vector. Only the global_res is compared to the tolerance in order
+  // to evaluate if the nonlinear system is solved. Only current_res is used for
+  // the alpha scheme as this scheme only applied to the matrix problem.
```
blaisb, line 82:
```suggestion
  // the alpha scheme as this scheme only monitors the convergence of the non-linear system of equation (the matrix problem).
```

blaisb (approved):
Good :)!

PR #251, opened 2021-10-01
Particle line bug fix

shahabgol:
This PR fixes a problem in DEM simulations using some of the triangulations created with gmsh.
If the triangulation created with gmsh had a diamond-shaped cell (with a boundary line), the particles that move from this cell toward the neighbour boundary cells were recognized as particle-wall candidates with a delay; that could lead to sudden jumps (like popcorn!) when using dynamic contact detection method.
In this PR, we add these diamond-shaped cell to particle-wall candidates with an imaginary face id and normal vectors of their boundary neighbour cell. We use this imaginary face id, because these cells do not have any boundary faces.

### DIFF
--- a/applications_tests/dem_3d/force_boundary_box.output
+++ b/applications_tests/dem_3d/force_boundary_box.output
@@ -3,6 +3,9 @@ DEM time-step is 4.64363% of Rayleigh time step
 Starting simulation with Lethe/DEM on 1 processors
 *****************************************************************
 
+Reading triangulation 
+
+Finished reading triangulation 
 *****************************************************************
 10 particles of type 0 were inserted, 0 particles of type 0 remaining
 *****************************************************************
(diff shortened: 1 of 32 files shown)


applications_tests/dem_3d/force_boundary_box.output
```diff
@@ -3,6 +3,9 @@ DEM time-step is 4.64363% of Rayleigh time step
 Starting simulation with Lethe/DEM on 1 processors
 *****************************************************************
 
+Reading triangulation 
+
+Finished reading triangulation 
 *****************************************************************
```
blaisb, line 9: So I guess you really wanted to add this line? :)

blaisb (approved):
Good job! Nothing to say. Very nice fix to a very specific edge case :)!

PR #255, opened 2021-10-04
Heat transfer fix

blaisb:
This PR fixes issues with the heat transfer Robin boundary condition assembler.
The PR restructures the way boundary terms are assembled.


shahabgol (approved):
Very clear to me.

PR #259, opened 2021-10-06
Dem diamond shaped cells

shahabgol:
This PR prints out a warning if the input triangulation has diamond shaped cells.
The normal vectors and points on boundary faces of these cells are not updated if the grid is moving.

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -500,6 +500,9 @@ namespace Parameters
     // Target size when automatically refining initial mesh
     double target_size;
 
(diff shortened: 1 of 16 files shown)


blaisb:
@shahabgol  Things appear to be breaking a bit. A unit test maybe does not have the parameter set-up?

shahabgol:
> @shahabgol Things appear to be breaking a bit. A unit test maybe does not have the parameter set-up?

Yes, it is fixed now.

PR #264, opened 2021-10-19
Nitsche time variable solid speed

blaisb:
This PR enables the use of time-dependent velocity for the Nitsche IB.
A test for this feature is added.
@jeannejoachim  this is the feature you needed  so i'll let you review it.

### DIFF
--- a/source/core/solid_base.cc
+++ b/source/core/solid_base.cc
@@ -486,11 +486,23 @@ SolidBase<dim, spacedim>::integrate_velocity(double time_step,
   // than unity
   for (unsigned int it = 0; it < sub_particles_iterations; ++it)
     {
-      // Set the time of the velocity function
       for (auto particle = solid_particle_handler->begin();
            particle != solid_particle_handler->end();
            ++particle)
         {
+          // Calculate the next position of the particle using the RK4
+          // time integration scheme
+          // x(t+dt) = x(t) + 1/6 (k1+2*k2+2*k3+k4)
+          // k1 = dt * v(t,x(t))
+          // k2 = dt * v(t+0.5*dt,x+k1/2)
+          // k3 = dt * v(t+0.5*dt,x+k2/2)
+          // k4 = dt * vt(t+dt,x+k3)
+          // The four stages (1 to 4) are built successively
+          // For each stage, the time of the function must be "reset"
+          // to ensure adequate evaluation of the RK4 scheme.
+          // See https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods for
+          // more details
+
           Point<spacedim> particle_location = particle->get_location();
 
           Tensor<1, spacedim> k1;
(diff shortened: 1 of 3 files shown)


source/core/solid_base.cc
```diff
@@ -504,13 +511,15 @@ SolidBase<dim, spacedim>::integrate_velocity(double time_step)
 
           Point<spacedim>     p3 = particle_location + time_step * k3;
           Tensor<1, spacedim> k4;
+          velocity->set_time(initial_time + time_step);
           for (unsigned int comp_i = 0; comp_i < spacedim; ++comp_i)
             k4[comp_i] = velocity->value(p3, comp_i);
 
           particle_location += time_step / 6 * (k1 + 2 * k2 + 2 * k3 + k4);
```
jeannejoachim, line 518: again, shouldn't we use `bdf_coeffs` rather than these `1,2,2,1` coefficients? If the scheme here is different than the one used for the fluid resolution, please add a comment to precise the scheme used way ahead, like with the command outside of the loop at l. 489 or just before `k1` initialization

source/core/solid_base.cc
```diff
@@ -482,18 +486,21 @@ SolidBase<dim, spacedim>::integrate_velocity(double time_step)
...
   for (unsigned int it = 0; it < sub_particles_iterations; ++it)
     {
+      // Set the time of the velocity function
       for (auto particle = solid_particle_handler->begin();
            particle != solid_particle_handler->end();
            ++particle)
         {
           Point<spacedim> particle_location = particle->get_location();
 
           Tensor<1, spacedim> k1;
```
jeannejoachim, line 496: I would indicate in a comment what these are (particles velocity at different time step?) + maybe use a more explicit variable name (such as `v1`, for velocity)

source/core/solid_base.cc
```diff
@@ -573,6 +586,7 @@ SolidBase<dim, spacedim>::move_solid_triangulation(double time_step)
 
               Point<spacedim>     p3 = vertex_position + time_step * k3;
```
jeannejoachim, line 587: `time_step/2`?

source/core/solid_base.cc
```diff
@@ -482,18 +486,21 @@ SolidBase<dim, spacedim>::integrate_velocity(double time_step)
...
           Point<spacedim> particle_location = particle->get_location();
 
           Tensor<1, spacedim> k1;
+          velocity->set_time(initial_time);
           for (unsigned int comp_i = 0; comp_i < spacedim; ++comp_i)
             k1[comp_i] = velocity->value(particle_location, comp_i);
 
           Point<spacedim>     p1 = particle_location + time_step / 2 * k1;
           Tensor<1, spacedim> k2;
+          velocity->set_time(initial_time + time_step / 2);
```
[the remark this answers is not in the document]
blaisb: The particle displacement in the Nitsche method is integrated with a fourth order Runge-Kutta scheme. The time-stepping scheme to advect the particle is different than the CFD one, so here we are recalculating the velocity of the particles for the intermediary RK steps.

source/core/solid_base.cc
blaisb: It's an implicit RK4 scheme. i'll ad a note regarding this. We advect the particles with an explicit scheme and not an implicit one since anyway the motion of the particle is decoupled from that of the fluid,

source/core/solid_base.cc
blaisb: I'm not sure I understand your comment. This function integrates the particle velocity, hence it's pretty clear that it is the particle velocities no?
I can remove line 489. It's confusing.

jeannejoachim: I think the key misunderstanding here is that I missed it is a classical RK4 scheme. I would put this comment at the beginning of this section (like, at line 489):

`// A classic Runge-Kutta scheme (RK4) is used for the particles velocities. See https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods for more details.`

Then at line 495, just before `k1` definition:

`// RK4 coefficients: k1, k2, k3, k4`

PR #265, opened 2021-10-20
[WIP] Significant upgrade of non-linear solver

blaisb:
Reduce number of RHS evaluation in the non-linear solver and add a few options which enable more flexibility.
Add a inexact jacobian newton solver that only renews the jacobian matrix when the solution is not progressing fast enough.
Still need to add a test so this is a WIP, but @Luckabarbeau and @lpsaavedra can you take a look at it?

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -382,8 +382,6 @@ namespace Parameters
     // Tolerance
     double tolerance;
 
-
-
     // Maximal number of iterations for the Newton solver
     unsigned int max_iterations;
 
--- a/tests/core/inexact_newton_iteration_non_linear_solver_01.cc
+++ b/tests/core/inexact_newton_iteration_non_linear_solver_01.cc
@@ -16,7 +16,7 @@ test()
 {
   Parameters::NonLinearSolver params{
     Parameters::Verbosity::verbose,
-    Parameters::NonLinearSolver::SolverType::newton,
+    Parameters::NonLinearSolver::SolverType::inexact_newton_iteration,
     Parameters::NonLinearSolver::KinsolStrategy::
       normal_newton, // kinsol strategy, not used in this case
     1e-8,            // tolerance
--- a/tests/core/inexact_newton_iteration_non_linear_solver_01.output
+++ b/tests/core/inexact_newton_iteration_non_linear_solver_01.output
@@ -1,0 +1,4 @@
+
+DEAL::Creating solver
+DEAL::Solving non-linear system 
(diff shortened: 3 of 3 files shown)


include/core/inexact_newton_iteration_non_linear_solver.h
```diff
@@ -0,0 +1,185 @@
...
+          if (current_res < this->params.step_tolerance * last_res ||
+              last_res < this->params.tolerance)
+            {
+              if (current_res > this->params.matrix_tolerance * last_res)
+                matrix_requires_assembly = true;
+              else
+                matrix_requires_assembly = false;
+
+              break;
+            }
```
Luckabarbeau, line 173: Maybe add a small comment on how this section works.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -1626,6 +1625,7 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
                           support_points[local_dof_indices[i]]);
                       typename DoFHandler<dim>::active_cell_iterator cell_2;
                       bool particle_close_to_wall = false;
+                      (void)particle_close_to_wall;
```
Luckabarbeau, line 1628: Not sure why a void is needed here.

Luckabarbeau (approved):
Very good PR  it will save a lot of useless calculation of the RHS.

include/core/parameters.h
```diff
@@ -381,15 +382,33 @@ namespace Parameters
...
+
 
     // Maximal number of iterations for the Newton solver
     unsigned int max_iterations;
 
     // Residual precision
     unsigned int display_precision;
 
+
+    // Force RHS recalculation at the beggining of every non-linear steps
```
lpsaavedra, line 394:
```suggestion
    // Force RHS recalculation at the beginning of every non-linear steps
```

include/core/parameters.h
```diff
@@ -381,15 +382,33 @@ namespace Parameters
...
+    // This is required if there is a fixed point component to the non-linear
+    // solver that is changed at the beggining of every newton iteration.
```
lpsaavedra, line 396:
```suggestion
    // solver that is changed at the beginning of every newton iteration.
```

lpsaavedra (approved):
Nice addition to the non-linear solvers and good solution as KINSOL is not working as it should at the moment. Found some typos in the documentation, but other than that it seems good. I also liked the distinction between current_res and global_res, it makes everything clearer.

PR #266, opened 2021-10-20
Make pressure drop calculations accessible to all NS solvers

oguevremont:


include/solvers/postprocessing_cfd.h
```diff
@@ -40,6 +40,39 @@
...
+ *
+ * @param evaluation_point The solution for which the CFL is calculated. The velocity field is assumed to be the first field.
+ *
+ * @param mpi_communicator The mpi communicator. It is used to reduce the CFL calculation.
+ *
+ * @param fe The finite element of the simulation
+ *
+ * @param cell_quadrature_formula The cell quadrature formula for the calculation
+ *
+ * @param face_quadrature_formula The cell quadrature formula for the calculation
```
oguevremont, line 57:
```suggestion
 * @param face_quadrature_formula The face quadrature formula for the calculation
```

include/solvers/postprocessing_cfd.h
```diff
@@ -40,6 +40,39 @@
...
 
+/**
+ * @brief Calculate the pressure drop between two boundaries
+ * @return Pressure drop of the flow between two boundaries of the domain
+ *
+ * @param dof_handler The dof_handler used for the calculation
+ *
+ * @param evaluation_point The solution for which the CFL is calculated. The velocity field is assumed to be the first field.
+ *
+ * @param mpi_communicator The mpi communicator. It is used to reduce the CFL calculation.
```
blaisb, line 51: You can actually get the communicator straight out of the dof_handler : see https://www.dealii.org/current/doxygen/deal.II/classDoFHandler.html
This way you could remove this extra argument. It could also be removed from the other functions and make the API cleaner :)

include/solvers/postprocessing_cfd.h
```diff
@@ -40,6 +40,39 @@
...
+ *
+ * @param fe The finite element of the simulation
```
blaisb, line 53: Same as above, you can get the FE straight out of the dof_handler actually :)

source/solvers/postprocessing_cfd.cc
```diff
@@ -31,6 +31,141 @@
...
+
+          for (unsigned int q = 0; q < face_n_q_points; ++q)
+            {
+              for (const auto &face : cell->face_iterators())
+                {
+                  if (face->at_boundary() &&
+                      (face->boundary_id() == outlet_boundary_id))
+                    {
+                      fe_face_values.reinit(cell, face);
+                      pressure_upper_boundary +=
```
blaisb, line 87:
```suggestion
                      pressure_outlet_boundary +=
```

source/solvers/postprocessing_cfd.cc
```diff
@@ -31,6 +31,141 @@
...
+                        present_pressure_values[q] * fe_face_values.JxW(q);
+
+                      upper_surface += fe_face_values.JxW(q);
+                    }
+
+                  if (face->at_boundary() &&
+                      (face->boundary_id() == inlet_boundary_id))
+                    {
+                      fe_face_values.reinit(cell, face);
+                      pressure_lower_boundary +=
```
blaisb, line 97:
```suggestion
                      pressure_inlet_boundary +=
```

source/solvers/postprocessing_cfd.cc
```diff
@@ -31,6 +31,141 @@
...
+
+  const unsigned int n_q_points      = cell_quadrature_formula.size();
+  const unsigned int face_n_q_points = face_quadrature_formula.size();
+
+  std::vector<double> present_pressure_values(n_q_points);
+
+  double pressure_upper_boundary = 0;
+  double upper_surface           = 0;
+  double pressure_lower_boundary = 0;
+  double lower_surface           = 0;
```
blaisb, line 67:
```suggestion
  double pressure_outlet_boundary = 0;
  double outlet_surface           = 0;
  double pressure_inlet_boundary = 0;
  double inlet_surface           = 0;
```

Better to use the inlet/outlet notation here because upper and lower don't make sense if your pipe is horizontal.

include/solvers/postprocessing_cfd.h
```diff
@@ -40,6 +40,39 @@
 #  include <core/boundary_conditions.h>
 #  include <core/parameters.h>
 
+/**
+ * @brief Calculate the pressure drop between two boundaries
```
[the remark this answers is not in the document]
oguevremont: Good point! I'll apply the 2nd suggestion

PR #273, opened 2021-10-28
Time dependant bc

Luckabarbeau:
Add the possibility to have time dependent boundary condition and add the possibility to control immersed boundary with time and space dependent function. Also add gravity as a physical property. this function can be time dependent and space dependent


blaisb:
Can you rebase, then I will review it ASAP :) ?

applications_tests/gls_sharp_navier_stokes_2d/moving_ib_gls.prm
```diff
@@ -18,6 +18,9 @@ end
 #---------------------------------------------------
 subsection physical properties
     set kinematic viscosity            = 0.1
+    subsection gravity
+        set Function expression =0;-9.81;0
```
blaisb, line 22: I'm not sure, but why do you need to make gravity different, it's the one you apply on the particles?

include/core/boundary_conditions.h
```diff
@@ -61,6 +61,7 @@ namespace BoundaryConditions
     // Number of boundary conditions
     unsigned int size;
     unsigned int max_size;
+    bool         time_dependant;
```
blaisb, line 64: Good idea. Don't forget to add this parameter to the wiki.

include/core/ib_particle.h
```diff
@@ -115,6 +114,11 @@ class IBParticle
   // iteration.
   Tensor<1, 3> omega_iter;
 
+  Functions::ParsedFunction<dim> *f_velocity;
+  Functions::ParsedFunction<dim> *f_position;
+  Functions::ParsedFunction<dim> *f_omega;
```
blaisb, line 119: Make them shared pointers instead of raw pointers. Raw pointers are very prone to mem leak. We avoid them (I don't think there are any left in Lethe)

include/core/parameters.h
```diff
@@ -195,6 +195,8 @@ namespace Parameters
     // tracer diffusivity in L^2/s
     double tracer_diffusivity;
 
+    Functions::ParsedFunction<3> *f_gravity;
```
blaisb, line 198: Use shared ptr

include/core/parameters.h
```diff
@@ -687,22 +696,12 @@ namespace Parameters
     bool                         assemble_navier_stokes_inside;
     std::string                  ib_force_output_file;
     double                       density;
-    Tensor<1, dim>               gravity;
-    double                       particle_nonlinear_tol;
-    double                       length_ratio;
 
+    double      particle_nonlinear_tol;
```
blaisb, line 700:
```suggestion
    double      particle_nonlinear_tolerance;
```

include/solvers/gls_navier_stokes.h
```diff
@@ -64,6 +64,12 @@ class GLSNavierStokesSolver
   virtual void
   setup_dofs_fd();
 
+  /**
+   * @brief update non zero constraint if the boundary is time dependant
+   */
+  void
+  update_boundary_condition();
```
blaisb, line 71:
```suggestion
  update_boundary_conditions();
```

include/solvers/gls_navier_stokes.h
```diff
@@ -108,6 +114,18 @@ class GLSNavierStokesSolver
   void
   setup_preconditioner();
 
+  /**
+   * @brief  defined the non zero constraints used to solved the problem.
```
blaisb, line 118:
```suggestion
   * @brief  define the non zero constraints used to solved the problem.
```

include/solvers/gls_navier_stokes.h
```diff
@@ -108,6 +114,18 @@ class GLSNavierStokesSolver
...
+   */
+  void
+  define_non_zero_constraints();
+
+  /**
+   * @brief defined the zero_constraints used to solved the problem.
```
blaisb, line 124:
```suggestion
   * @brief define the zero_constraints used to solved the problem.
```

include/solvers/simulation_parameters.h
```diff
@@ -59,7 +59,7 @@ class SimulationParameters
   AnalyticalSolutions::AnalyticalSolution<dim> *    analytical_solution;
   SourceTerms::SourceTerm<dim> *                    source_term;
   Parameters::VelocitySource                        velocity_sources;
-  Parameters::IBParticles<dim>                      particlesParameters;
+  Parameters::IBParticles<dim> *                    particlesParameters;
```
blaisb, line 62: Good idea here would be to transition this Parameter and the other parameters to shared pointers. It would be a small change and would make things cleaner in the code.

include/solvers/simulation_parameters.h
```diff
@@ -92,7 +92,8 @@ class SimulationParameters
     Parameters::LinearSolver::declare_parameters(prm);
     Parameters::PostProcessing::declare_parameters(prm);
     Parameters::DynamicFlowControl ::declare_parameters(prm);
-    particlesParameters.declare_parameters(prm);
+    particlesParameters = new Parameters::IBParticles<dim>;
```
blaisb, line 95: make shared ptr.

source/core/parameters.cc
```diff
@@ -214,6 +214,11 @@ namespace Parameters
                         Patterns::Double(),
                         "Tracer diffusivity");
 
+      prm.enter_subsection("gravity");
```
blaisb, line 217: So this gravity, it is applied on fluid or only on the particles? It's a bit unclear, because we already have a source term section. This will need to be really documented well that this gravity is only the gravity to be used in solid-fluid flows.

source/solvers/gls_navier_stokes.cc
```diff
@@ -292,6 +175,159 @@ GLSNavierStokesSolver<dim>::setup_dofs_fd()
                                    &this->present_solution);
 }
 
+template <int dim>
+void
+GLSNavierStokesSolver<dim>::update_boundary_condition()
```
blaisb, line 180:
```suggestion
GLSNavierStokesSolver<dim>::update_boundary_conditions()
```

source/solvers/navier_stokes_base.cc
```diff
@@ -1624,6 +1624,8 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
     }
 }
 
+
+
```
blaisb, line 1628:
```suggestion
```

blaisb (commented):
Some minor comments. Good work.

blaisb (approved):

PR #278, opened 2021-11-02
Add capacity to reuse matrix for other nlin prob.

blaisb:
This PR adds the capacity to reuse the matrix on a subsequent non-linear
problem as long as the matrix is considered to be sufficiently valid. This greatly speeds-up time-stepping scheme that have multiple stages such as the SDIRK ones.


lpsaavedra (approved):
It looks good to me! Why is the CI not working for dealii:master? Why was it suddenly cancelled?

blaisb:
> It looks good to me! Why is the CI not working for dealii:master? Why was it suddenly cancelled?

Good question, I will re-run it

PR #283, opened 2021-11-09
Debug build now depends on release build

technophil98:
In this PR, I queued the debug builds after the release builds. 

If a release build fails or any test ran against it fails, the associated debug build and tests won't be executed.
I kept the mechanism to run the same workflow on multiple versions of deal.ii. 

I also cleaned up the workflow file by removing obsolete formatting jobs and moved the first interaction in its own file.


technophil98:
Execution example: https://github.com/technophil98/lethe/actions/runs/1433270230

blaisb (approved):
Great job!

PR #285, opened 2021-11-15
Non newtonian carreau model

CaroleAnneDaunais:
The non Newtonian Carreau model is implemented within the GD solver, where the assembly of the matrix and right and side differ from the CORE assembler. Therefore, the non Newtonian solver constitutes an assembler, and replaces the CORE assembler for non Newtonian flows. The model needs to be specified in the parameter file, as well as the model parameters. For now, only the Carreau model is implemented, but there can be an infinite amount of non Newtonian models in the future. The non Newtonian assembler assembles the matrix and right and side with the non Newtonian viscosity given by the model.


source/core/parameters.cc
```diff
@@ -186,6 +186,51 @@ namespace Parameters
...
+                        "1",
+                        Patterns::Double(),
+                        "Viscosity at rest");
+      prm.declare_entry("viscosity_inf",
+                        "2",
+                        Patterns::Double(),
+                        "Viscosity for an infinite constraint");
+      prm.declare_entry("lambda", "1", Patterns::Double(), "Relaxation time");
+      prm.declare_entry("a", "1", Patterns::Double(), "Relaxation time");
+      prm.declare_entry("n", "1", Patterns::Double(), "Power parameter");
```
Luckabarbeau, line 209: If these parameters are only for the Carreau model, I would suggest putting them in a subsection specific to the Carreau model even though at this point in time you only have this model implemented.

include/core/rheological_model.h
```diff
@@ -0,0 +1,114 @@
...
+
+#ifndef lethe_rheological_model_h
+#define lethe_rheological_model_h
+
+#include <core/parameters.h>
+
+using namespace dealii;
+
+/**
+ * @brief RheologicalModel. Abstract class that allows to calculate the
```
blaisb, line 25: Good job, nice brief, love it.

include/solvers/navier_stokes_assemblers.h
```diff
@@ -294,6 +295,53 @@ class GDNavierStokesAssemblerCore : public NavierStokesAssemblerBase<dim>
...
+    std::shared_ptr<SimulationControl> simulation_control,
+    Parameters::PhysicalProperties     physical_properties,
+    const double                       gamma)
+    : simulation_control(simulation_control)
+    , physical_properties(physical_properties)
+    , gamma(gamma)
+  {
+    // if (physical_properties.non_newtonian_parameters.model ==
+    // Parameters::NonNewtonian::Model::Carreau)
+    //{
```
blaisb, line 313: Seems legit to leave it like that for the moment.

include/solvers/navier_stokes_assemblers.h
```diff
@@ -294,6 +295,53 @@ class GDNavierStokesAssemblerCore : public NavierStokesAssemblerBase<dim>
...
+   * @param copy_data (see base class)
+   */
+  virtual void
+  assemble_rhs(NavierStokesScratchData<dim> &        scratch_data,
+               StabilizedMethodsTensorCopyData<dim> &copy_data) override;
+
+
+  std::shared_ptr<SimulationControl>     simulation_control;
+  Parameters::PhysicalProperties         physical_properties;
+  double                                 gamma;
```
blaisb, line 341: What is gamma here?

source/solvers/gd_navier_stokes.cc
```diff
@@ -110,12 +110,24 @@ GDNavierStokesSolver<dim>::setup_assemblers()
...
-      // Core assembler
-      this->assemblers.push_back(
-        std::make_shared<GDNavierStokesAssemblerCore<dim>>(
-          this->simulation_control,
-          this->simulation_parameters.physical_properties,
-          gamma));
+      if (this->simulation_parameters.physical_properties.non_newtonian_flow)
+        {
+          // Core assembler with Non newtonian viscosity
+          this->assemblers.push_back(
```
blaisb, line 116: I love this notion of having a single assembler for all non-newtonian flow, this is very nice.

Luckabarbeau (approved):
Very nice PR only small comments on the prm structure for the non-Newtonian parameters

include/core/rheological_model.h
```diff
@@ -0,0 +1,114 @@
...
+  {}
+
+  /**
+   * @brief Returns the non-newtonian viscosity.
+   *
+   * @param shear_rate The shear rate tensor at the position of the
+   * considered quadratured point
+   */
+  virtual double
+  get_viscosity(const Tensor<2, dim> shear_rate) = 0;
```
blaisb, line 48: It would be better to pass a reference instead of passing a copy. This will prevent unnecessary copying of the tensor for reasons which are not necessary.
```suggestion
  get_viscosity(const Tensor<2, dim> & shear_rate) = 0;
```

blaisb (commented):
My only comment is to change the shear rate tensor to be passed by reference instead of copy. That's all
Good job. Well commented. Please continue working like this :). It's easy for me to review this.

source/core/parameters.cc
blaisb: Good idea @Luckabarbeau

include/core/rheological_model.h
```diff
@@ -0,0 +1,114 @@
...
+   * @brief Default constructor
+   */
+  RheologicalModel()
+  {}
+
+  /**
+   * @brief Returns the non-newtonian viscosity.
+   *
+   * @param shear_rate The shear rate tensor at the position of the
+   * considered quadratured point
```
lpsaavedra, line 45:
```suggestion
   * @param shear_rate The shear rate tensor at the position of the
   * considered quadrature point
```

include/core/rheological_model.h
```diff
@@ -0,0 +1,114 @@
...
+   */
+  virtual double
+  get_viscosity(const Tensor<2, dim> shear_rate) = 0;
+
+  /**
+   * @brief Returns the magnitude of the shear rate tensor given in parameter.
+   *
+   * @param shear_rate The shear rate tensor at the position of the
+   * considered quadratured point
```
lpsaavedra, line 54:
```suggestion
   * @param shear_rate The shear rate tensor at the position of the
   * considered quadrature point
```

include/solvers/navier_stokes_assemblers.h
CaroleAnneDaunais: gamma is the term in the continuity equation, the same used in the CORE assembler. It does not refer to that shear rate tensor.

source/core/parameters.cc
```diff
@@ -186,6 +186,51 @@ namespace Parameters
...
+      prm.declare_entry("viscosity_0",
+                        "1",
+                        Patterns::Double(),
+                        "Viscosity at rest");
+      prm.declare_entry("viscosity_inf",
+                        "2",
+                        Patterns::Double(),
+                        "Viscosity for an infinite constraint");
+      prm.declare_entry("lambda", "1", Patterns::Double(), "Relaxation time");
+      prm.declare_entry("a", "1", Patterns::Double(), "Relaxation time");
```
lpsaavedra, line 208: Are both of this relaxation time?

include/core/rheological_model.h
```diff
@@ -0,0 +1,114 @@
...
+                      (n - 1) / a);
+  }
+
+private:
+  double viscosity_0;
+  double viscosity_inf;
+  double lambda;
+  double a;
+  double n;
+};
```
lpsaavedra, line 111: I wonder if it would be useful to maybe add a comment with the reference of the paper where this model can be found in this form? I did not know the model and after a quick search in google is easy to find it but without the "a" parameter... do you think that would make the model clearer from a user perspective?

lpsaavedra (approved):
Very nice PR! Happy to have non newtonian models in Lethe now. 

I was wondering if it would be possible to change the name of the test from `mms_Carreau_gd` to  `mms_carreau_gd`  or `mms_non_newtonian_carreau_gd` as we usually don't use capital letters for the name of tests.

include/core/rheological_model.h
CaroleAnneDaunais: Thank you :)

source/core/parameters.cc
CaroleAnneDaunais: No, thank you for the comment.

source/solvers/gd_navier_stokes.cc
CaroleAnneDaunais: Thank you.

include/solvers/navier_stokes_assemblers.h
CaroleAnneDaunais: Ok I'll leave it like this.

include/core/rheological_model.h
CaroleAnneDaunais: This is a very good point. @blaisb what do you think? Also, how do we manage references in the software? Should we add it in the Lethe wiki? I could add a reference to the Morrison.

source/core/parameters.cc
CaroleAnneDaunais: For the name of the example, I used Carreau since in is the name of a person, but I can totally switch to carreau.

include/core/rheological_model.h
blaisb: You can add a reference to the morisson directly in the comment. Eventually when we have a more full-fledged documentation we will use this documentation also, but it's always good to keep it in the code.

source/core/parameters.cc
CaroleAnneDaunais: I fixed it in my most recent commit.

include/core/rheological_model.h
CaroleAnneDaunais: It is added.

blaisb (approved):

PR #292, opened 2021-11-24
Initial step bdf 2 3 refactor

Luckabarbeau:
Refactor the startup of bdf scheme. This allows sharp IB to use BDF 2 and BDF 3 scheme. There are now 3 options to start up these schemes. Multiple BDF step, which use a lower order BDF scheme to start, sdirk step that uses SDIRK to make the first few time steps, initial solution which use to initial solution to define the solution before the start of the simulation. 
This pull request also include 2 other small modifications:
Mini refactor of the linear solver try fail scheme.
A call to the base post-process_Fd for the sharp IB solver.


source/core/parameters.cc
```diff
@@ -1030,6 +1055,11 @@ namespace Parameters
                         "1",
                         Patterns::Integer(),
                         "amg smoother overlap");
+      prm.declare_entry(
+        "block linear solver fail",
+        "false",
+        Patterns::Bool(),
+        "A bool that allow the use to block linear solver to throw errors");
```
blaisb, line 1062: You should find a better name for this parameter. Like force linear solver continuation?
Something like that. Try to add a better description of it also :).

source/core/parameters.cc
```diff
@@ -166,6 +185,11 @@ namespace Parameters
                         Patterns::Selection("none|iteration|end"),
                         "Clock monitoring methods "
                         "Choices are <none|iteration|end>.");
+      prm.declare_entry(
+        "write time in error table",
+        "false",
+        Patterns::Bool(),
+        "Bool to define if the time is written in the error table");
```
oguevremont, line 192:
```suggestion
        "Boolean to define if the time is written in the error table");
```

source/core/simulation_control.cc
```diff
@@ -48,6 +55,116 @@ SimulationControl::is_output_iteration()
...
+void
+SimulationControl::update_assembly_method()
+{
+  if (iteration_number <= 1 &&
+      method == Parameters::SimulationControl::TimeSteppingMethod::bdf2 &&
+      bdf_start_method ==
+        Parameters::SimulationControl::BDFStartupMethods::multiple_step_bdf)
+    {
+      assembly_method = Parameters::SimulationControl::TimeSteppingMethod::bdf1;
+      set_current_time_step(initial_time_step * startup_timestep_scaling);
```
blaisb, line 67: One thing we could consider would be to abandon this startup_timestep_scaling parameter in a next PR and just use the regular time step and live with it.

source/core/parameters.cc
oguevremont, line 1062:
```suggestion
        "A boolean that allows the used block linear solver to throw errors");
```

source/core/simulation_control.cc
```diff
@@ -48,6 +55,116 @@ SimulationControl::is_output_iteration()
...
+           bdf_start_method ==
+             Parameters::SimulationControl::BDFStartupMethods::sdirk_step)
+    {
+      return Parameters::SimulationControl::TimeSteppingMethod::sdirk33;
+    }
+  else
+    {
+      return method;
+    }
+}
```
blaisb, line 166: I don't understand, why did you need an update_method and a get_method that both manage a bit the same thing? is it because of the time-step scaling?

source/core/simulation_control.cc
```diff
@@ -48,6 +55,116 @@ SimulationControl::is_output_iteration()
...
+             Parameters::SimulationControl::BDFStartupMethods::sdirk_step)
+    {
+      assembly_method =
+        Parameters::SimulationControl::TimeSteppingMethod::sdirk22;
+      set_suggested_time_step(initial_time_step);
+    }
+  else
+    {
+      assembly_method = method;
+    }
```
oguevremont, line 142: I think cascading "if"s would improve the readability, for example with level 1 - BDFStartupMethods, level 2 - TimeSteppingMethod, level 3 - iteration number.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -634,6 +635,12 @@ GLSSharpNavierStokesSolver<dim>::postprocess_fd(bool firstIter)
       this->write_output_results(this->present_solution);
     }
 
+  bool enable =
+    this->simulation_parameters.analytical_solution->calculate_error();
+  this->simulation_parameters.analytical_solution->set_enable(false);
+  NavierStokesBase<dim, TrilinosWrappers::MPI::Vector, IndexSet>::
+    postprocess_fd(firstIter);
```
blaisb, line 642: What is this? You do this to use the postprocess_fd without having it calculate error? 
Could you just not overload the calculate_L2_error and that would work? or does it not because the function is called from the base class?

source/solvers/navier_stokes_base.cc
```diff
@@ -535,117 +532,11 @@ NavierStokesBase<dim, VectorType, DofsType>::iterate()
...
-      this->simulation_control->set_assembly_method(
-        Parameters::SimulationControl::TimeSteppingMethod::bdf3);
-      PhysicsSolver<VectorType>::solve_non_linear_system(false);
-
-      multiphysics->solve(
-        Parameters::SimulationControl::TimeSteppingMethod::bdf3);
-
-      simulation_control->set_suggested_time_step(timeParameters.dt);
-    }
-}
```
blaisb, line 648: So happy to see this gone. Was crap code and my entire fault.

tests/core/simulation_control_05.cc
```diff
@@ -0,0 +1,177 @@
...
+  {
+    simulation_control_parameters.method =
+      Parameters::SimulationControl::TimeSteppingMethod::bdf2;
+    SimulationControlTransient simulation_control(
+      simulation_control_parameters);
+
+
+    // Constant time-stepping
+    deallog << "*************************************************" << std::endl;
+    deallog << "Constant time stepping - constant output" << std::endl;
```
blaisb, line 59: You should change this message to indicate what it is that you are testing... no?

tests/core/simulation_control_05.cc
```diff
@@ -0,0 +1,177 @@
...
+ * ---------------------------------------------------------------------
+
+*
+* Author: Bruno Blais, Polytechnique Montreal, 2020-
+*/
+
+/**
+ * @brief This test checks that the time-step round-off errors
+ * do not affect the end of the simulation
+ */
```
blaisb, line 23: DId you modify an existing or create a new one? Change the header and add an explanation of what the test does.

tests/core/simulation_control_05.output
```diff
@@ -0,0 +1,36 @@
...
+DEAL::Iteration : 3    Time : 3.00000
+DEAL::Iteration : 4    Time : 4.00000
+DEAL::*************************************************
+DEAL::Constant time stepping - constant output
+DEAL::*************************************************
+DEAL::Iteration : 0    Time : 0.00000
+DEAL::Iteration : 1    Time : 1.00000
+DEAL::Iteration : 2    Time : 2.00000
+DEAL::Iteration : 3    Time : 3.00000
+DEAL::Iteration : 4    Time : 4.00000
```
blaisb, line 36: This output will not help you debug the test if it crashes. You should have it print something worthwhile (like e.g the method used for the assembly)

blaisb (commented):
Good work. There are still some things left to address and fix (e.g. some tests and etc.). Your unit test is not good right now, it needs to test what you want to test.
You forgot some comments there and there also.

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -667,7 +674,21 @@ GLSSharpNavierStokesSolver<dim>::postprocess_fd(bool firstIter)
...
+              double total_time = 0;
+              for (auto it = summary.begin(); it != summary.end(); ++it)
+                {
+                  total_time += summary[it->first];
+                }
+              this->error_table.add_value("total_time", total_time);
+            }
         }
+      // Calculate pressure drop between two boundaries
+
```
oguevremont, line 691:
```suggestion
```

oguevremont (approved):
Nice work! A few comments missing here and there, but overall clear. The cascading if statements is a suggestion I got on another PR and I think it would apply well in this case, but it is not absolutely necessary.

include/core/sdirk.h
```diff
@@ -76,12 +76,14 @@ number_of_intermediary_stages(
   if (method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33_1 ||
       method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33_2 ||
       method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33_3 ||
-      method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33)
+      method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33 ||
+      method == Parameters::SimulationControl::TimeSteppingMethod::bdf3)
```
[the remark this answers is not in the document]
Luckabarbeau: BDF2 and BDF3 now require stage if they start up with sdirk so should i move that to another function i use that to initialize the solution stage vector for this case.

source/core/simulation_control.cc
Luckabarbeau: the two have been merge together

source/solvers/gls_sharp_navier_stokes.cc
Luckabarbeau: It is not possible to have calculate_L2_error virtual since it is from a base class

source/core/simulation_control.cc
Luckabarbeau: This would lead to duplication of the else statement so for the moment i think it would let it as it is for the time being.

source/core/parameters.cc
```diff
@@ -123,6 +131,17 @@ namespace Parameters
         {
           std::runtime_error("Invalid time stepping scheme");
         }
+      const std::string bdf_startup_string = prm.get("bdf startup method");
+      if (bdf_startup_string == "multiple step bdf")
+        bdf_startup_method = BDFStartupMethods::multiple_step_bdf;
+      else if (bdf_startup_string == "sdirk step")
```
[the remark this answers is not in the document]
Luckabarbeau: i would avoid using the bdf alone as it is used in other places in the code with a different signification (same fo sdirk).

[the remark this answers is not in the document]
blaisb: Good argument, agreed

blaisb (approved):

PR #293, opened 2021-11-24
Add buoyancy cfd dem

tonielgeitani:
This PR adds buoyancy force to the particle fluid interaction assemblers. It also fixes some units in the buoyancy, shear force and pressure force assemblers. All tests involving particles have had their outputs changed since the introduction of additional  pf interactions changes the results of these tests.


source/fem-dem/gls_vans.cc
```diff
@@ -631,42 +631,51 @@ GLSVANSSolver<dim>::setup_assemblers()
...
-
-  // Pressure Force
-  particle_fluid_assemblers.push_back(
-    std::make_shared<GLSVansAssemblerPressureForce<dim>>());
-
-  // Shear Force
-  particle_fluid_assemblers.push_back(
-    std::make_shared<GLSVansAssemblerShearForce<dim>>(
-      this->cfd_dem_simulation_parameters.cfd_parameters.physical_properties));
+  if (this->cfd_dem_simulation_parameters.cfd_dem.buoyancy_force == true)
```
blaisb, line 658: Good job.

source/fem-dem/vans_assemblers.cc
```diff
@@ -665,25 +668,25 @@ GLSVansAssemblerShearForce<dim>::calculate_particle_fluid_interactions(
...
+        viscosity * velocity_laplacians[particle_number];
 
       for (int d = 0; d < dim; ++d)
         {
           particle_properties[DEM::PropertiesIndex::fem_force_x + d] +=
             shear_force[d];
           undisturbed_flow_force[d] +=
-            shear_force[d] / particle_properties[DEM::PropertiesIndex::mass];
+            shear_force[d] /
+            (physical_properties.density * scratch_data.cell_volume);
```
blaisb, line 689: the shear force and the pressure force already have the right units, you don't need to divide by density because your viscosity is already divided by density (nu instead of mu).

blaisb (commented):
One tiny thing to fix, very good PR! One place your are dividing by density and this is not correct. Be careful that we always work with kinematic viscosity and 1/rho * grad P

blaisb (approved):

PR #294, opened 2021-11-26
Add parameter to stop simulation if Nitsche particles have been lost

jeannejoachim:
Tested on a simulation case.


blaisb (approved):
Cool idea. I love it.

PR #295, opened 2021-12-01
Fix PVD output refactor solve function

blaisb:
This PR fixes a bug where the pvdHandler would buffer a large number of time steps since it was being called by the DEM code. Additionnaly, some code duplication was removed.


tonielgeitani (approved):
Everything is clear for me. If I understand correctly, the CFD and DEM outputs are now written simultaneously by the function postprocess_fd.

PR #297, opened 2021-12-09
Conservation monitoring in the vof postprocess

jeannejoachim:
The surface occupied by one of the two fluids is calculated at every time step. This can be used for any simulation, and will be especially needed with wetting/peeling.

Outputs a table in free_surface_monitoring.dat. The user can choose which fluid to monitor through the `fluid index` parameter.

Checked for 2D simulations. Compatible with multiple MPI run.

### DIFF
--- a/include/solvers/vof.h
+++ b/include/solvers/vof.h
@@ -27,6 +27,7 @@
 
 #include <deal.II/base/convergence_table.h>
 #include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/table_handler.h>
 #include <deal.II/base/timer.h>
 
 #include <deal.II/distributed/solution_transfer.h>
@@ -405,7 +406,7 @@ private:
     previous_solutions_transfer;
 
   // Conservation Analysis
-  ConvergenceTable error_table_fs;
+  TableHandler volume_table_fs;
 
   // Enable DCDD shock capturing scheme
   const bool DCDD = true;
--- a/include/solvers/vof.h
+++ b/include/solvers/vof.h
@@ -17,14 +17,6 @@
 #ifndef lethe_VOF_h
 #define lethe_VOF_h
 
-#include <core/bdf.h>
-#include <core/simulation_control.h>
-
-#include <solvers/auxiliary_physics.h>
-#include <solvers/multiphysics_interface.h>
-#include <solvers/vof_assemblers.h>
-#include <solvers/vof_scratch_data.h>
-
 #include <deal.II/base/convergence_table.h>
 #include <deal.II/base/quadrature_lib.h>
 #include <deal.II/base/timer.h>
@@ -41,6 +33,13 @@
 #include <deal.II/lac/trilinos_vector.h>
 
 #include <deal.II/numerics/error_estimator.h>
+
+#include <core/bdf.h>
+#include <core/simulation_control.h>
+#include <solvers/auxiliary_physics.h>
+#include <solvers/multiphysics_interface.h>
+#include <solvers/vof_assemblers.h>
+#include <solvers/vof_scratch_data.h>
 
 
 template <int dim>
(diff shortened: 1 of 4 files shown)


include/solvers/vof.h
```diff
@@ -163,6 +170,11 @@ class VOF : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
    * physics. It does not concern the output of the solution using the
    * DataOutObject, which is accomplished through the attach_solution_to_output
    * function
+   *
+   * @param scratch_data The scratch data which is used to store
+   * the calculated finite element information at the gauss point.
+   * See the documentation for VOFScratchData for more
+   * information
```
blaisb, line 177: I think this block is not necessary as a header here no?
```suggestion
   *
   * @param scratch_data The scratch data which is used to store
   * the calculated finite element information at the gauss point.
   * See the documentation for VOFScratchData for more
   * information
```

include/solvers/vof.h
```diff
@@ -392,6 +404,9 @@ class VOF : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
     parallel::distributed::SolutionTransfer<dim, TrilinosWrappers::MPI::Vector>>
     previous_solutions_transfer;
 
+  // Conservation Analysis
+  ConvergenceTable error_table_fs;
```
blaisb, line 408: You can use a regular TableHandler here, because you are not monitoring an error or something that converges.

source/solvers/vof.cc
```diff
@@ -295,6 +295,55 @@ VOF<dim>::calculate_L2_error()
...
+                      if (q_scalar_values[q] < 0.5)
+                        volume += fe_values_fs.JxW(q) * q_scalar_values[q];
+                      break;
+                    }
+                  case 1:
+                    {
+                      if (q_scalar_values[q] > 0.5)
+                        volume += fe_values_fs.JxW(q) * q_scalar_values[q];
+                      break;
+                    }
```
blaisb, line 338: So you are essentially integrating what is behind in the interface instead of integrating the whole VOF field right? It's a good idea, but don't expect exact conservation here for sure.

source/solvers/vof.cc
```diff
@@ -364,6 +413,41 @@ VOF<dim>::postprocess(bool first_iteration)
...
+              error_table_fs.add_value(
+                "cells", this->triangulation->n_global_active_cells());
+            }
+          else
+            {
+              error_table_fs.add_value("time",
+                                       simulation_control->get_current_time());
+            }
+          error_table_fs.add_value("fluid_volume", volume);
+          error_table_fs.set_scientific("fluid_volume", true);
```
blaisb, line 443: See above comments on keeping a seperate table for vvolume conservation. You don't need an error table here you can use a regular table (and even that is only required if you want to print the entire volume conservation history, which you could want to :) ).

blaisb (commented):
A minor comment about the table used to store the data, otherwise, awesome PR.

include/solvers/vof.h
jeannejoachim: Maybe, I just corrected a typo here!

source/solvers/vof.cc
jeannejoachim: Yes that's it! I took the usual definition of the literature (the interface is considered located at `phase_value = 0.5`). Right now, with the diffusivity, conservation is not exact (as you say), but still gives a good idea of it's order of magnitude. I think it will work well with Shahab's interface sharpening, we will see if another criterion is needed at this point :)

source/solvers/vof.cc
jeannejoachim: maybe it won't be relevant in the long run, but for now yes I wanted to have the entire history!

Changed, thank you for the review :)

include/solvers/vof.h
jeannejoachim: done

blaisb (approved):
Good work. Ready for ze merge

PR #302, opened 2021-12-15
Multiphysics solve structure

jeannejoachim:
Mesh adapt on free surface would need to be tested on a less dump test than the hydrostatic one.
For now, structure is here for pre and post solve in multiphysics, but all physics remain post-solved.

### DIFF
--- a/include/solvers/multiphysics_interface.h
+++ b/include/solvers/multiphysics_interface.h
@@ -122,7 +122,7 @@ public:
         else
           {
             pcout
-              << "Resolution order is not specified for this auxiliary physic :"
+              << "Solution order is not specified for this auxiliary physic :"
               << std::endl
               << "will be solved after the fluid dynamics by default."
               << std::endl
--- a/include/solvers/multiphysics_interface.h
+++ b/include/solvers/multiphysics_interface.h
@@ -154,11 +154,13 @@ public:
         else
           {
             pcout
-              << "Resolution order is not specified for this auxiliary physic :"
+              << "Solution order is not specified for this auxiliary physic "
               << std::endl
               << "will be solved after the fluid dynamics by default."
               << std::endl
-              << "For clarity, please specify a resolution order in the map solve_pre_fluid defined in multiphysics_interface.h"
+              << "For clarity, please specify a solution order for this physic "
+              << std::endl
+              << "in the map solve_pre_fluid, member of multiphysics_interface.h"
               << std::endl
               << "-------------" << std::endl;
           }
--- a/include/solvers/multiphysics_interface.h
+++ b/include/solvers/multiphysics_interface.h
@@ -126,7 +126,7 @@ public:
               << std::endl
               << "will be solved after the fluid dynamics by default."
               << std::endl
-              << "For clarity, please specify a resolution order in the map solve_pre_fluid defined in multiphysics_interface.h"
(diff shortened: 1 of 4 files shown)


source/solvers/vof.cc
```diff
@@ -639,11 +639,11 @@ VOF<dim>::setup_dofs()
...
-  // the fluid at present iteration is solved before the free surface, and
-  // after percolate is called for the previous iteration, so at the time the
-  // getter is called solution_m2 = solution_m1
-  // TODO deactivated for now (inertia is considered with a constant density),
-  // see if needed / to be debugged
+  // the fluid at present iteration is solved AFTER the free surface (see map
+  // solve_pre_fluid defined in multiphysics_interface.h), and after percolate
+  // is called for the previous iteration.
+  // NB: for now, inertia in fluid dynamics is considered with a constant
+  // density (see if needed / to be debugged)
```
blaisb, line 646: This is not true right now because all flags are set to false.
We'll have to investigate which is better once we have more difficult test cases.
```suggestion
  // the fluid at present iteration is solved BEFORE the free surface (see map
  // solve_pre_fluid defined in multiphysics_interface.h), and after percolate
  // is called for the previous iteration.
  // NB: for now, inertia in fluid dynamics is considered with a constant
  // density (see if needed / to be debugged)
```

blaisb (commented):
Minor comments, good architecture, this will be very flexible for the future I love it.

include/solvers/multiphysics_interface.h
```diff
@@ -77,55 +75,149 @@ class MultiphysicsInterface
...
-              << "Heat Transfer" << std::endl
-              << "--------------" << std::endl;
+        pcout << "Heat Transfer" << std::endl << "--------------" << std::endl;
       }
     else if (physics_id == PhysicsID::tracer)
       {
-        pcout << "-------" << std::endl
-              << "Tracer" << std::endl
-              << "-------" << std::endl;
+        pcout << "Tracer" << std::endl << "-------" << std::endl;
```
shahabgol, line 82: I think keeping the message as it was before is a better idea. It separates the message from the upper line in the log.

include/solvers/multiphysics_interface.h
```diff
@@ -77,55 +75,149 @@ class MultiphysicsInterface
...
+                if (verbosity != Parameters::Verbosity::quiet)
+                  pcout << "-----------------------" << std::endl;
+              }
+          }
+        else
+          {
+            pcout
+              << "Resolution order is not specified for this auxiliary physic :"
+              << std::endl
+              << "will be solved after the fluid dynamics by default."
```
shahabgol, line 159: I don't understand this message. Can you please make it more clear?

include/solvers/multiphysics_interface.h
jeannejoachim: It is still the case! Now it is handled by the pre_solve and post_solve functions, because otherwise it was ugly for presolved physics. In the end, we still have a line before the announce message, no worries!

include/solvers/multiphysics_interface.h
blaisb: I think your suggestion is a good one. Anyway it's a bug that only developers should face :)!

blaisb:
Sounds good to me. Good for merge!

PR #306, opened 2021-12-19
Fluid identification refactoring

blaisb:
# Description of the problem

We currently use a confusing mechanism for the physical properties of the fluid. Originally, the code was not meant to have more than a single fluid, so all properties were indicated in the physical properties section. As we introduced the VOF model, things became confusing. We have fluid 0  and fluid 1 section for the fluid properties of the two fluids, but these sections are not used when the simulation is single phase. 

# Description of the solution

This PR makes the single and multiphase simulations coherent by introducing the usage of the fluid subsection even if simulations are single phase. Consequently, even single phase simulations use the fluid 0 subsection.

# How Has This Been Tested?

All tests had to be adapted to this change in properties. Consequently, this feature is fully tested

# Future changes

The next step will be the generalisation of the properties using physical properties model and to adapt the non-newtonian solver logic to this. I prefer to do it step-by-step. Once this will be finished, we will be able to use single and vof phase change and non-newtonian flows. Since this is a major breaking PR, I prefer to make it small.

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
+++ b/applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
@@ -43,6 +43,7 @@ end
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
+  set number of fluids = 1
   subsection fluid 0
     set tracer diffusivity = 1
   end
--- a/applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
+++ b/applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
@@ -43,7 +43,9 @@ end
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
-  set tracer diffusivity = 1
+  subsection fluid 0
+    set tracer diffusivity = 1
+  end
 end
 
 #---------------------------------------------------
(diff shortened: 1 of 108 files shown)


CaroleAnneDaunais (commented):
That makes a lot of sense. In a next pull request, do you intend to put the non newtonian parameters in the FLUIDS struct? Everything seems good to me!

applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
```diff
@@ -43,7 +43,9 @@ end
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
-  set tracer diffusivity = 1
+  subsection fluid 0
```
shahabgol, line 46: I know that the default value of the variable "number of fluids" is equal to 1, but I think it would be better to define this variable in 2-3 application tests as an example.

shahabgol (approved):
Very clean PR. I just have a small suggestion.
It also seems that the clang format is not correct.

blaisb:
> That makes a lot of sense. In a next pull request, do you intend to put the non newtonian parameters in the FLUIDS struct? Everything seems good to me!

Yeah exactly. This way we could automatically have a two fluid model with two different rheology and everything would work out of the box, it's the next logical step. The only part that would remain outside is the non-newtonian boolean.

applications_tests/gls_navier_stokes_2d/gls_tracer_simplex_diffusion_mms_2d.prm
blaisb: Very good suggestion. I will add it to the 2D tests just to be sure that it is clear.
I'll also modify the wiki documentaiton.

PR #307, opened 2021-12-20
Force and torque non newtonian

CaroleAnneDaunais:
# Description of the problem

The current forces and torque calculated as a postprocess tool use a constant viscosity. For non Newtonian flows, viscosity varies in space, therefore the viscosity should be calculated on each quadrature point when calculating the total force and torque. 

# Description of the solution

The methods calculate_forces() and calculate_torque() now calculated the value of viscosity for each quadrature point when the non_newtonian_flow flag is activated, and this viscosity is being used.

# How Has This Been Tested?

A new test has been added to the Grad-Div solver where the force field is calculated using a Carreau rheolgical model. 

- [x] application_tests/gs_navier_stokes_2d/cylinder_carreau_gd


source/solvers/postprocessing_cfd.cc
```diff
@@ -434,14 +435,30 @@ calculate_forces(
...
+  // Non newtonian properties
+  std::unique_ptr<RheologicalModel<dim>> rheological_model;
+  if (physical_properties.non_newtonian_flow)
+    {
+      // if (physical_properties.non_newtonian_parameters.model ==
+      // Parameters::NonNewtonian::Model::Carreau)
+      //{
+      rheological_model = std::make_unique<Carreau<dim>>(
+        physical_properties.non_newtonian_parameters);
+      //}
```
Luckabarbeau, line 451: Those these comments are still necessary?

source/solvers/postprocessing_cfd.cc
```diff
@@ -553,14 +577,29 @@ calculate_torques(
...
```
Luckabarbeau, line 593: Same as above

Luckabarbeau (commented):
Very good implementation. I wonder if the same could be directly implemented in the sharp_interface solver in this PR or should we do another PR?

blaisb:
> Very good implementation. I wonder if the same could be directly implemented in the sharp_interface solver in this PR or should we do another PR?

Another PR.
Small PRs are good PRs.

source/solvers/postprocessing_cfd.cc
CaroleAnneDaunais: I think we should keep them as long as there is not more than 1 rheological model

blaisb (commented):
I think this is very good, but it would be better to have the constant viscosity case being handled by a "constant" rheology model than anything else. It would prevent multiple ifs and make the code easier to understand (and directly extendable to different rheology model, especially if you created the shared ptr to the rheology model in a function).

blaisb (approved):

Luckabarbeau (approved):

source/solvers/postprocessing_cfd.cc
```diff
@@ -434,14 +435,30 @@ calculate_forces(
...
   const FESystem<dim, dim> fe = dof_handler.get_fe();
 
-  double viscosity = physical_properties.fluids[0].viscosity;
+  double viscosity;
+  if (!physical_properties.non_newtonian_flow)
+    viscosity = physical_properties.fluids[0].viscosity;
+
+  // Non newtonian properties
+  std::unique_ptr<RheologicalModel<dim>> rheological_model;
+  if (physical_properties.non_newtonian_flow)
```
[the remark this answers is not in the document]
CaroleAnneDaunais: Done! I generated a rheological_model.cc file since the methode model_cast() needs both Newtonian and Carreau classes to be instantiated before the definition of the method. I would really be much clearer if all methods from the rheological_model.h file were instantiated in the rheological_model.cc file (this should have been done in the first place, but I was very lazy I admit). A call to model_cast() should also replace the if conditions in the  GLSNavierStokesAssemblerNonNewtonianCore constructer in navier_stokes_assemblers.h.  Should I do it in this PR or in another one?

source/solvers/postprocessing_cfd.cc
```diff
@@ -434,14 +435,21 @@ calculate_forces(
 {
   const FESystem<dim, dim> fe = dof_handler.get_fe();
 
-  double viscosity = physical_properties.fluids[0].viscosity;
+  // Rheological model for viscosity properties
+  double viscosity;
+  // Cast rheological model to either a Newtonian model or one of the
+  // non Newtonian models according to the physical properties
+  std::shared_ptr<RheologicalModel<dim>> rheological_model =
+    RheologicalModel<dim>::model_cast(physical_properties);
```
blaisb, line 443: Good job, that's a very nice way of doing it.

blaisb (approved):
Good job!
Love the static cast for the model choice. Good job.

PR #309, opened 2021-12-22
Strong form nitsche ib

blaisb:
# Description of the problem

The terms arising from the Nitsche immersed boundary were not added to the strong form of the residual used in the stabilization. Consequently, as the time step became lower (and the stabilization parameter played a more important role), the solution would diverge in time and a slight loss of mass would occur.

# Description of the solution

The solution is to add the Nitsche term to the strong form along with it's jacobian. It fixes the issue with time convergence

# How Has This Been Tested?

This is very difficult to test specifically in an application test, but all tests add to be modified since this changed the residual. I think this is sufficient.

# Future changes

None

# Comments

It took a while for Jeanne and I to figure this one out.


source/solvers/gls_nitsche_navier_stokes.cc
```diff
@@ -137,6 +147,20 @@ GLSNitscheNavierStokesSolver<dim, spacedim>::assemble_nitsche_restriction()
...
+
+              const double tau =
+                this->simulation_control->get_assembly_method() ==
+                    Parameters::SimulationControl::TimeSteppingMethod::steady ?
+                  1. / std::sqrt(
+                         std::pow(2. * u_mag / h_cell, 2) +
+                         9 * std::pow(4 * viscosity / (h_cell * h_cell), 2)) :
+                  1. / std::sqrt(
+                         std::pow(sdt, 2) + std::pow(2. * u_mag / h_cell, 2) +
+                         9 * std::pow(4 * viscosity / (h_cell * h_cell), 2));
```
oguevremont, line 162: Would it be useful to put the tau calculation in a common function, since we use it in a few classes? 
If we push even further, could we put the calculation of the stabilization terms in a separate class that any solver that we want to stabilize could use? Just an idea, maybe for a future hackathon

oguevremont (commented):
Quite clear, good work!

source/solvers/gls_nitsche_navier_stokes.cc
blaisb: Good idea. This would be a great follow-up PR.

PR #311, opened 2021-12-31
Fixes some leftover compilation warnings

blaisb:
# Description of the problem

Some compilation warnings were still hanging around after previous PRs were merged

# Description of the solution

Each warning were fixed one by one. They were mostly unused variables warnings or no return in a non-void function.

# How Has This Been Tested?

No need for tests here.

# Comments

We need to start being a bit more careful about warnings :)


lpsaavedra (approved):
Seems good to me! I will be more careful with those warnings :)

include/core/rheological_model.h
```diff
@@ -102,7 +102,7 @@ class Newtonian : public RheologicalModel<dim>
    * parameter to fit the virtual method in RheologicalModel.
    */
   double
-  get_viscosity(const double &shear_rate_magnitude) override
+  get_viscosity(const double & /*shear_rate_magnitude*/) override
```
CaroleAnneDaunais, line 105: Right we can do this! We could also just take the comment off, and specify it in the method header

CaroleAnneDaunais (approved):
This is going to be much clearer now! :)

PR #320, opened 2022-01-03
Force and torque non newtonian sharp interface

CaroleAnneDaunais:
# Description of the problem

Prior to a previous branch merge, all forces and torque were calculated with a constant viscosity. The calculate_forces() and calculate_torque() methods were previously modified to get the right viscosity on the coordinates where the force and torque are calculated. This is not implemented for the sharp interface. 

# Description of the solution

I proceeded to a similar change in the force and torque calculation for the sharp interface.

# How Has This Been Tested?

- No new tests were added, the gls_sharp_navier_stokes_3d/steady_couette_sphere_poisson.debug test was used to verify the force calculation still works

# Comments

- Some grammar mistakes were fixed on the way, and the use of "mu" and "rho" was changed for "viscosity" and "density".


source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -287,7 +294,7 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
           unsigned int p;
           bool         cell_is_cut;
           std::tie(cell_is_cut, p) = cut_cells_map[cell];
-          // If the cell is cute
```
blaisb, line 290: I always liked cute cells

blaisb (approved):
Good work. Concise and clear, I love it.
 Thanks for fixing some spelling mistakes along the way.

source/solvers/gls_sharp_navier_stokes.cc
CaroleAnneDaunais: Me too to be honest. But then it's discrimination for not cute cells so

source/solvers/gls_sharp_navier_stokes.cc
```diff
@@ -470,8 +477,16 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                                      viscosity =
+                                        rheological_model->get_viscosity(
+                                          rheological_model
+                                            ->get_shear_rate_magnitude(
+                                              shear_rate));
                                       fluid_stress =
-                                        mu *
+                                        viscosity *
                                           (velocity_gradients[k] +
                                            transpose(velocity_gradients[k])) -
```
Luckabarbeau, line 491: We could reuse the shear_rate tensor defined previously defined

Luckabarbeau (commented):
We could add a test for when it uses a viscosity model that is not constant. Only to check if others modifications are impacting the evaluation of the force. The torque of the Taylor Couette case could be a good example.

source/solvers/gls_sharp_navier_stokes.cc
CaroleAnneDaunais: Yes I forgot! thank you

blaisb (approved):
Good idea to add an application test :)

Luckabarbeau (approved):

PR #321, opened 2022-01-03
Constructor calls model_cast() method

CaroleAnneDaunais:
# Description of the problem

The issue was very minor. In the GDNavierStokesAssemblerNonNewtonianCore and GLSNavierStokesAssemblerNonNewtonianCore constructors, we manually casted the rheological_model attribute by using a if case on the rheological model. Now that this method with the ifs already exists, we will use it.

# Description of the solution

- Change the if series with the existing method

# How Has This Been Tested?

- No tests are needed


blaisb (approved):
Good find. This is minor so I'll merge once CI passes.

PR #324, opened 2022-01-04
Folder structure explanation in sphinx doc

lpsaavedra:
# Description of the problem

The explanation of the folder structure of Lethe was only available on the wiki. 

# Description of the solution

Migrates the explanation of the folder structure to the new sphinx documentation and it updates the information regarding sub directories.

# How Has This Been Tested?

This does not require any test. The new section of the documentation compiles without warnings. 

# Future changes

None

# Comments

None

### DIFF
--- a/doc/source/structure.rst
+++ b/doc/source/structure.rst
@@ -1,0 +1,105 @@
+############################################
+Explanation of the folder structure
+############################################
+
+The folder structure of Lethe is relatively complex. Once you have cloned and compiled Lethe, it is normal that you might be confused as to the meaning of the various folders in the cloned directory (which we will refer to as the lethe folder). This section of the documentation aims at demystifying these elements.
+
+================
+The Lethe folder
+================
+In the Lethe folder, the following structure is present:
+
+* ``/applications``
+* ``/applications_tests``
+* ``/contrib``
+* ``/doc``
+* ``/examples``
+* ``/include``
+* ``/logo``
+* ``/prototypes``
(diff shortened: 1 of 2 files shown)


doc/source/structure.rst
```diff
@@ -0,0 +1,108 @@
...
+
+The contrib folder
+-------------------
+
+The ``/contrib`` folder houses various scripts that help for the maintenance of Lethe. It notably contains the ``contrib/utilities/indent-all`` script which can be used to automatically indent the entire Lethe source code to ensure that it adheres to the clang-format rules contained in the ``/contrib`` folder.
+
+The doc folder
+--------------
+
+The ``\doc`` folder contains the source files of this documentation page of Lethe. To contribute or compile the documentation on your own machine follow the instructions of the *contributing* tab.
```
blaisb, line 81: Do you know if it is possible to hyperlink to the contributing time from inside the documentation? a bit like how we can do references in a single file?
If it's not possible, let's not worry about it, but if @technophil98 you have an idea ? :).

blaisb (commented):
Awesome work, there is jhust a single hanging question and otherwise, this will be good to go. Thanks for migrating this

doc/source/structure.rst
lpsaavedra: Yes, it is possible! Just pushed a commit using the hyperlink :)

blaisb (approved):
Awesome work. Yay! Thanks for the help.

PR #325, opened 2022-01-04
Bug correction high order force sharp edgre

Luckabarbeau:
# Description of the problem

The evaluation of the force on particles using the sharp interface immersed boundary adds an error in the FE system used on the surface cell. Also, the code had an error in the projection of the solution on the surface cell for high-order elements since we use a spherical manifold.

# Description of the solution
We changed the fe_system used for the evaluation on the surface cell to the correct one and added an interpolation scheme to correctly map the fluid stress tensor on the surface cell for high order elements.

# How Has This Been Tested?
No test was added as this is simply a bug correction that improves the code's robustness and improves the precision for high order elements.


CaroleAnneDaunais (approved):
Good catch there, good PR. :)

PR #327, opened 2022-01-04
Power law model

CaroleAnneDaunais:
# Description of the problem

Lethe now has a block that deals with non newtonian flow problems. The Carreau model was the first rheological model to be added that deals with changing viscosity with a change in the local shear forces. Though the Carreau model appears to better represent most rheological flows such has plastic flow than the power-law model, this model is used in the majority of rheological analysis in the litterature. I added this second model to better validate with the litterature.

# Description of the solution

The power-law model such as decribed in [1] takes two parameters, a nominal viscosity K and a power index n. A third parameter is added (shear_rate_min) to avoid undefined values when calculating the viscosity (since viscosity = K * shear_rate ^ (n-1)). A subclass PowerLaw to the mother class RheologicalModel was added to handle viscosity calculation. The majority of the code added is very similar to what was done with the Carreau model.

# How Has This Been Tested?

The gd_navier_stokes_2d/mms2d_gd was used to generate the new test, now using the power-law model where n = 0.7 (shear-thinning) and the source term is changed due non-constant viscosity

- [x] gd_navier_stokes_2d/mms2d_powerlaw_gd

# Comments

The changes made in navier_stokes_base.cc and post_processors.h should have been done before, it is more of a code-cleaning issue than something related to the Power-law assembler.

[1] F.A. Morrison and A.P.C.E.F.A. Morrison.Understanding Rheology. Ray-mond  F.  Boyer  Library  Collection.  Oxford  University  Press,  2001.isbn:9780195141665.url:https://books.google.ca/books?id=YxnwwAEACAAJ.


source/core/parameters.cc
```diff
@@ -239,6 +239,33 @@ namespace Parameters
...
 
+  void
+  PowerLawParameters::declare_parameters(ParameterHandler &prm)
+  {
+    prm.enter_subsection("power-law");
+    {
+      prm.declare_entry("K", "1.0", Patterns::Double(), "Viscosity at rest");
+      prm.declare_entry("n", "0.5", Patterns::Double(), "Power parameter");
+      prm.declare_entry("shear rate min",
+                        "0.001",
```
CaroleAnneDaunais, line 250: This number is totally random, I think there's no such thing as a default minimum shear rate, it will really depend on K and n.

include/core/rheological_model.h
```diff
@@ -111,6 +111,46 @@ class Newtonian : public RheologicalModel<dim>
...
+  {
+    return shear_rate_magnitude > shear_rate_min ?
+             K * std::pow(shear_rate_magnitude, n - 1) :
+             K * std::pow(shear_rate_min, n - 1);
+  }
+
+private:
+  double K;
+  double n;
+  double shear_rate_min;
```
blaisb, line 151:
```suggestion
const  double K;
const  double n;
const  double shear_rate_min;
```

source/core/parameters.cc
```diff
@@ -239,6 +239,33 @@ namespace Parameters
     prm.leave_subsection();
   }
 
+  void
+  PowerLawParameters::declare_parameters(ParameterHandler &prm)
+  {
+    prm.enter_subsection("power-law");
+    {
+      prm.declare_entry("K", "1.0", Patterns::Double(), "Viscosity at rest");
```
blaisb, line 247:
```suggestion
      prm.declare_entry("K", "1.0", Patterns::Double(), "Viscosity at rest");
```
Is K really the viscosity at rest? Generally we call it the fluid consistency index : https://en.wikipedia.org/wiki/Power-law_fluid  . N is generally the flow behavior index.

blaisb (commented):
two small comments, otherwise good to go :)

source/core/rheological_model.cc
```diff
@@ -8,7 +8,11 @@ RheologicalModel<dim>::model_cast(
...
       physical_properties.fluids[0].viscosity);
-  else //  (physical_properties.non_newtonian_parameters.model ==
+  else if (physical_properties.non_newtonian_parameters.model ==
+           Parameters::NonNewtonian::Model::powerlaw)
+    return std::make_shared<PowerLaw<dim>>(
+      physical_properties.non_newtonian_parameters);
+  else // if (physical_properties.non_newtonian_parameters.model ==
        // Parameters::NonNewtonian::Model::carreau)
     return std::make_shared<Carreau<dim>>(
       physical_properties.non_newtonian_parameters);
```
oguevremont, line 18: It would make more sense to me if the Newtonian alternative was the default one, but if you prefer to have it first instead that also makes sense

oguevremont (approved):
All good! Some minor comments but the code is elegant

source/core/parameters.cc
CaroleAnneDaunais: You are right, actually K is the viscosity when n = 1. I will change both!

source/core/rheological_model.cc
CaroleAnneDaunais: That's a good point, I'll change it :)

PR #335, opened 2022-01-10
Apparent viscosity

CaroleAnneDaunais:
# Description of the problem

The Reynolds number for non Newtonian flows is presently not well defined considering the viscosity changes when the shear stresses vary. This code patch aims to better define a Reynolds number by calculating an apparent viscosity. This will be particularly usefull for flows around geometries or flows in confined spaces (i.e. in cylinders). 

# Description of the solution

The apparent viscosity is calculated this way : apparent viscosity = (Integral over the calculation domain of the local viscosity times the local shear rate magnitude) /  (Integral over the calculation domain of the local shear rate magnitude). This way, the viscosity is weighted by the shear rate, meaning that the viscosity will be driven by the highest shear rate regions. The apparent-viscosity calculation is added in the post-processing methods. 

# How Has This Been Tested?

The added test is a 1D viscosity problem where u = 2*y, and therefore shear_rate_magnitude is 2*y. For the given Carreau paramters, the viscosity equation is 1/(1+4y^2)^0.25. With a domain going from y = 0 to y = 1, the apparent viscosity can easily be integrate and gives 0.78123. 

- [x] gd_navier_stokes_2d/apparent_viscosity_carreau_gd.mpirun=2

# Future changes

- Add this new feature to the documentation


include/core/parameters.h
```diff
@@ -451,6 +454,9 @@ namespace Parameters
     // Prefix for the enstrophy output
     std::string enstrophy_output_name;
 
+    // Prefix for the enstrophy output
```
Luckabarbeau, line 457: Rename the comment

source/solvers/postprocessing_cfd.cc
```diff
@@ -423,6 +423,105 @@ calculate_kinetic_energy<3, TrilinosWrappers::MPI::BlockVector>(
...
+              integral_shear_rate += shear_rate_magnitude * fe_values.JxW(q);
+            }
+        }
+    }
+  const MPI_Comm mpi_communicator = dof_handler.get_communicator();
+  integral_viscosity_x_shear_rate =
+    Utilities::MPI::sum(integral_viscosity_x_shear_rate / domain_volume,
+                        mpi_communicator);
+  integral_shear_rate =
+    Utilities::MPI::sum(integral_shear_rate / domain_volume, mpi_communicator);
```
Luckabarbeau, line 489: You don't need to divide by the volume of the domain since it appears both on the numerator and denominator.

source/solvers/postprocessing_cfd.cc
```diff
@@ -423,6 +423,105 @@ calculate_kinetic_energy<3, TrilinosWrappers::MPI::BlockVector>(
...
+                          update_values | update_gradients |
+                            update_quadrature_points | update_JxW_values);
+
+  const FEValuesExtractors::Vector velocities(0);
+
+  const unsigned int n_q_points = quadrature_formula.size();
+
+  std::vector<Tensor<2, dim>> present_velocity_gradients(n_q_points);
+  double                      domain_volume =
+    GridTools::volume(dof_handler.get_triangulation(), mapping);
```
Luckabarbeau, line 457: Instead of using this function, I would simply integrate the JwX in the following loop as well this will eliminate a loop over the cells.

applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.mpirun=2.output
```diff
@@ -0,0 +1,13 @@
...
+   Volume of triangulation:      1
+Apparent viscosity: 0.781234 Pa.s
+
+*****************************************************************
+Steady iteration :        1/1
+*****************************************************************
+Apparent viscosity: 0.781234 Pa.s
+L2 error velocity : 5.19289e-07
+cells error_velocity error_pressure 
+   16 5.192891e-07 - 4.658475e-03 -
```
Luckabarbeau, line 13: Can we check that the apparent viscosity converges with the mesh?

Luckabarbeau (commented):
Very nice PR! Can't wait to see more results from this tool. Smalls modification here and there. I would recommend adding a mesh convergence of the apparent viscosity.

source/solvers/postprocessing_cfd.cc
CaroleAnneDaunais: right my mistake

source/solvers/postprocessing_cfd.cc
CaroleAnneDaunais: I'm not sure I understand this comment, let me fix something with your previous comment and recomment that.

applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.mpirun=2.output
CaroleAnneDaunais: I don't think there is a tool to check post-processing variable convergence with an analytical solution (@blaisb ?) but I just checked it on the side and it converges to the same rate as the velocity when refining twice.

blaisb: That would be difficult to add to the convergence table (although doable). Not sure if it is important though :).

blaisb (commented):
Good for me :)
Good job

Luckabarbeau (approved):

source/solvers/navier_stokes_base.cc
```diff
@@ -1070,6 +1070,47 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
...
+
+      this->apparent_viscosity_table.add_value(
+        "time", simulation_control->get_current_time());
+      this->apparent_viscosity_table.add_value("apparent_viscosity",
+                                               apparent_viscosity);
+      if (this->simulation_parameters.post_processing.verbosity ==
+          Parameters::Verbosity::verbose)
+        {
+          this->pcout << "Apparent viscosity: " << apparent_viscosity << " Pa.s"
+                      << std::endl;
```
blaisb, line 1093:
```suggestion
          this->pcout << "Apparent viscosity: " << apparent_viscosity << " m^2/s"
                      << std::endl;
```

PR #337, opened 2022-01-10
Create post_processing.rst

CaroleAnneDaunais:
# Description of the problem

Migration of the post-processing parameters (and addition of missing ones)

# Comments

- Description for apparent viscosity were added even though the branch is still not merged to master.


blaisb (approved):
Very good work. A next step will be to add the mathematical formulation for enstrophy and total kinetic energy, but I suggest we make that in another PR :).

PR #341, opened 2022-01-13
Add shear post-process and remove dim from rheo

blaisb:
# Description of the problem
- It is interesting to be able to visualize the shear rate in Paraview
- The Rheology classes were templated, which was not exactly necessary

# Description of the solution

- I moved the calculation of the shear rate outside of the rheology model class
- I added a shear rate post-processor
- I added a function outside to calculate the shear rate from a Tensor<2,dim>. With this, we don't need to template the rheology model.

# How Has This Been Tested?

- Existing test cover this :)

# Future changes

- None! The rheology models are robust and mature now it is getting cool.
-

### DIFF
--- a/include/core/rheological_model.h
+++ b/include/core/rheological_model.h
@@ -27,7 +27,6 @@ using namespace dealii;
  * non-newtonian viscosity on each quadrature point and the shear rate
  * magnitude.
  */
-template <int dim>
 class RheologicalModel : public PhysicalPropertyModel
 {
 public:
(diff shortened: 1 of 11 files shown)


include/core/rheological_model.h
```diff
@@ -298,5 +285,26 @@ class Carreau : public RheologicalModel<dim>
...
 };
 
+/**
+ * @brief Calculates the magnitude of the shear rate tensor given as input
+ *
+ * @param shear_rate A shear rate tensor
+ */
+template <int dim>
+inline double
+calculate_shear_rate_magnitude(const Tensor<2, dim> shear_rate)
```
CaroleAnneDaunais, line 295: Should we put the implementation in the .cc?

CaroleAnneDaunais (approved):
Very nice, this is going to be very helpful :)

include/core/rheological_model.h
CaroleAnneDaunais: got it

PR #345, opened 2022-01-16
Add the capacity for transient BC Vans & CFD-DEM

blaisb:
# Description of the problem

- Lethe supports transient boundary conditions,  but the VANS and the CFD-DEM solvers don't integrate this capability

# Description of the solution

- The feature for transient Boundary conditions are added to these solvers

# How Has This Been Tested?

- This was quickly tested on a few cases to work, since this is already tested in the basic CFD solver and it uses the same functon, I don't think it is necessary to add a test.

# Comments

- Doable because of the good work of @Luckabarbeau  in the past.

### DIFF
--- a/source/fem-dem/cfd_dem_coupling.cc
+++ b/source/fem-dem/cfd_dem_coupling.cc
@@ -19,8 +19,8 @@ template <int dim>
 bool
 check_contact_detection_method(
   unsigned int                          counter,
-  CFDDEMSimulationParameters<dim>      &param,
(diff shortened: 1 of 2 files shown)


voferreira (approved):
That's fine to me. Tested with more than one condition and it worked.

tonielgeitani (approved):

Luckabarbeau:
This look perfect to me !

Luckabarbeau (approved):
Look good to me hope it help with the unresolved cases

PR #348, opened 2022-01-17
Add the examples explicitly to ensure order

lpsaavedra:
# Description of the problem

Examples were not in the right order in the documentation.

# Description of the solution

The toctree is now numbered and the examples are added explicitly in each subsection to preserve order.

# How Has This Been Tested?

This does not require any test. The documentation compiles without any warnings.


blaisb (approved):

voferreira (approved):
Verified the sequence and it seems good to me.

Issue #315, opened 2022-01-03
Restarting overwrites Nitsche particle forces and torques


anicusan:
Hello, and thank you for developing this very high quality code.

## Problem Description

Once a simulation is restarted, the previously-computed nitsche particle forces and torques are lost; new values are overwritten instead of being appended to their files; e.g. I ran a simulation up to `t = 4 s`, saving forces acting on particle 0 in `solid_force_00.dat` and stopped - after restarting, `solid_force_00.dat` will only contain forces computed *after* `t = 4 s`.

I used the `gls_nitsche_navier_stokes_33` solver, though I guess file overwriting might be orthogonal to the specific solver employed. For completeness, I attach the `.prm` script I used:

```
#=================================================================================================
# Simulating two spheres approaching at constant velocity up to the point of contact, measuring
# the drag forces and torques acting thereon.
#
# The spheres' Reynolds number can be modified by changing the kinematic viscosity of the fluid
# (physical properties section); to fine-tune their individual Reynolds numbers change their radii
# and velocities (nitsche solid section).
#
# Author:   Andrei Leonard Nicusan <a.l.nicusan@bham.ac.uk>
# Date:     28.12.2021
# License:  LGPL
#=================================================================================================


#-------------------------------------------------------------------------------------------------
# Simulation and IO Control
#-------------------------------------------------------------------------------------------------
subsection simulation control
    set method                          = sdirk2
    set bdf startup method              = sdirk step

    set time step                       = 0.02                  # Time step
    set time end                        = 4.5                   # End time of simulation
    set adapt                           = true
    set max cfl                         = 0.5
    set adaptative time step scaling    = 1.1

    set output name                     = approaching_spheres   # Prefix for VTU outputs
    set output path                     = ./results/            # Output directory
    set output boundaries               = true                  # Output domain boundaries
    set output frequency                = 1                     # Frequency of simulation output
    set subdivision                     = 1                     # Output mesh subdivision
end


#-------------------------------------------------------------------------------------------------
# Simulation Checkpointing
#-------------------------------------------------------------------------------------------------
subsection restart
    set checkpoint                      = true                  # Enable checkpointing
    set restart                         = true                  # Start from previous checkpoint
    set filename                        = restart
    set frequency                       = 10
end


#-------------------------------------------------------------------------------------------------
# Force
#-------------------------------------------------------------------------------------------------
subsection forces
    set verbosity                       = verbose
    set calculate forces                = true
    set force name                      = force
[...]

PR #349, opened 2022-01-17
Nitsche table restart

jeannejoachim:
# Description of the problem
Fixing various problems in gls_nitsche:
1. torque tables had the same filename for multiple solids
2. when restarting a Nitsche simulation, force and torque tables were rewritten (previous values were lost)
3. maximum number of Nitsche solid was 2

# Description of the solution
1. torque filename is incremented with the number of solids
2. previous force and torque tables are loaded and used to refill the restarting simulation (see `read_checkpoint` function)
3. maximum number of Nistche solid set to 10

# How Has This Been Tested?

Tested locally on a case almost similar to the `two-bar-mixer-restart` test (adding calculations of forces and torques). Tell me if you want me to turn it into a test.

### DIFF
--- a/include/core/nitsche.h
+++ b/include/core/nitsche.h
@@ -161,7 +161,7 @@ namespace Parameters
     // Nitsche solid objects
     std::vector<std::shared_ptr<NitscheSolid<dim>>> nitsche_solids;
     unsigned int                                    number_solids;
-    static const unsigned int                       max_nitsche_solids = 2;
+    static const unsigned int                       max_nitsche_solids = 10;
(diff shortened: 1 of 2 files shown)


blaisb (approved):
Seems good to me. This would close #315

jeannejoachim:
You're right @blaisb! Sorry I missed that. It is fixed now :)

blaisb:
> You're right @blaisb! Sorry I missed that. It is fixed now :)

Waiting for CI then i'll merge

PR #357, opened 2022-01-18
Add half-step angular velocity calculation

shahabgol:
# Description of the problem

- Previously, in the half-step calculations of the velocity-verlet integration method, the half-step angular velocity wasn't calculated. In this PR we add the calculation of the half-step angular velocity to velocity-verlet integration method.

### DIFF
--- a/include/dem/explicit_euler_integrator.h
+++ b/include/dem/explicit_euler_integrator.h
@@ -54,15 +54,19 @@ public:
    * @param particle_handler The particle handler whose particle motion we wish
    * to integrate
    * @param body_force A constant volumetric body force applied to all particles
-   * @param force Force acting on particles
    * @param time_step The value of the time step used for the integration
+   * @param momentum Momentum of particles
+   * @param force Force acting on particles
+   * @param MOI A container of moment of inertia of particles
    */
   virtual void
   integrate_half_step_location(
     Particles::ParticleHandler<dim> &particle_handler,
     Tensor<1, dim> &                 body_force,
+    double                           time_step,
+    std::vector<Tensor<1, dim>> &    momentum,
     std::vector<Tensor<1, dim>> &    force,
-    double                           time_step) override;
+    std::vector<double> &            MOI) override;
 
   /**
    * Carries out integration of the motion of all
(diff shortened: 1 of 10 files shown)


include/dem/explicit_euler_integrator.h
```diff
@@ -54,15 +54,19 @@ class ExplicitEulerIntegrator : public Integrator<dim>
...
+   * @param force Force acting on particles
+   * @param MOI A container of moment of inertia of particles
    */
   virtual void
   integrate_half_step_location(
     Particles::ParticleHandler<dim> &particle_handler,
     Tensor<1, dim> &                 body_force,
+    double                           time_step,
+    std::vector<Tensor<1, dim>> &    momentum,
     std::vector<Tensor<1, dim>> &    force,
```
blaisb, line 68: Make arguments that should be const const :)

include/dem/explicit_euler_integrator.h
```diff
@@ -54,15 +54,19 @@ class ExplicitEulerIntegrator : public Integrator<dim>
    * @param particle_handler The particle handler whose particle motion we wish
    * to integrate
    * @param body_force A constant volumetric body force applied to all particles
-   * @param force Force acting on particles
    * @param time_step The value of the time step used for the integration
+   * @param momentum Momentum of particles
+   * @param force Force acting on particles
+   * @param MOI A container of moment of inertia of particles
    */
   virtual void
```
blaisb, line 62: Should this function be virtual?

blaisb (commented):
Good job. Just take the opportunity to make the const argument const :) then it is good to go

tonielgeitani (approved):

include/dem/explicit_euler_integrator.h
shahabgol: done.

include/dem/explicit_euler_integrator.h
shahabgol: yes, because it is rewritten by different integration methods.

shahabgol:
The const parameters are defined in the integrators. :)

blaisb (approved):

PR #358, opened 2022-01-18
Remove read me from all examples folders

lpsaavedra:
# Description of the problem

Every example in the examples folder had a specific README file stating the solver needed and pointing to the old wiki. 

# Description of the solution

All the specific README files are deleted and a general one pointing to the new documentation is created. 

# How Has This Been Tested?

It does not require any test.


blaisb (approved):

jeannejoachim (approved):
Good idea! :)

PR #360, opened 2022-01-19
Documentation for dynamic flow control

CaroleAnneDaunais:
# Description of the problem

I added the documentation for the dynamic flow control part.

# Comments

- I didn't know what Beta value to suggest, do we have a range to suggest?

### DIFF
--- a/doc/source/parameters/cfd/dynamic_flow_control.rst
+++ b/doc/source/parameters/cfd/dynamic_flow_control.rst
@@ -1,10 +1,7 @@
 Dynamic flow control
 ~~~~~~~~~~~~~~~~~~~~
 
-This subsection's purpose is to enable a dynamic flow control. It enables setting a volumetric flow on a specific boundary toward to normal of the wall. 
-It calculates a Beta coefficient at each time step and may be added to a source term. If the chosen wall boundary on which a volumetric flow is imposed is at the inlet, 
-the volumetric flow rate targeted must be negative since outward normal vector is used. Thereby, if the chosen wall boundary is at the outlet, 
-volumetric flow rate must be positive.
+The purpose of this subsection is to enable dynamic flow control. It is important when we want to target a volumetric flow on a specific boundary. To control the volumetric flow, the code calculates a :math:`\beta`  coefficient at each time step that is used to keep the volumetric flow value inside an acceptable threshold. If the chosen wall boundary on which a volumetric flow is imposed is an inlet, the volumetric flow rate targeted must be negative since the outward normal vector is used for the calculation. Thereby, if the chosen wall boundary is an outlet, volumetric flow rate must be positive.
 
 The default parameters are:
 
--- a/doc/source/parameters/cfd/dynamic_flow_control.rst
+++ b/doc/source/parameters/cfd/dynamic_flow_control.rst
@@ -1,0 +1,38 @@
+Dynamic flow control
+~~~~~~~~~~~~~~~~~~~~
+
+This subsection's purpose is to enable a dynamic flow control. It enables setting a volumetric flow on a specific boundary toward to normal of the wall. 
+It calculates a Beta coefficient at each time step and may be added to a source term. If the chosen wall boundary on which a volumetric flow is imposed is at inlet, 
+the volumetric flow rate targeted must be negative since outward normal vector is used. Thereby, if the chosen wall boundary is at outlet, 
+volumetric flow rate must be positive.
+
+The default parameters are:
+
+.. code-block:: text
+
+  subsection flow control
+    set enable = false
+    set volumetric flow rate = 0
+    set boundary id = 0
+    set flow direction = 0
+    set initial beta = 0
+  end
+
+* The ``enable`` parameter is set to ``true`` to enable an imposed volumetric flow rate.
+
+* The ``volumetric flow rate`` parameter specifies the flow rate (:math:`m^2/s`). 
+
+.. warning::
+
+  This value should be negative if imposed on an inlet boundary, and positive if imposed on an outlet boundary.
+
+.. note::
+
+  When there is a periodic boundary, the wall boundary should be at inlet only and if the chosen wall boundary is at inlet, flow rate must be negative and vice versa.
+
+* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled. (inlet or outlet).
+
+* The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`y` direction.
+
+* The ``initial beta`` parameter sets an initial beta coefficient that may speed up the convergence of the flow rate targeted. Some tests should be done to find a good initial beta coefficient.
+
--- a/doc/source/parameters/cfd/dynamic_flow_control.rst
+++ b/doc/source/parameters/cfd/dynamic_flow_control.rst
@@ -30,7 +30,7 @@ The default parameters are:
 
   When there is a periodic boundary, the wall boundary should be at inlet only and if the chosen wall boundary is at inlet, flow rate must be negative and vice versa.
 
-* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled. (inlet or outlet).
+* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled (inlet or outlet).
 
 * The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`y` direction.
 
--- a/doc/source/parameters/cfd/dynamic_flow_control.rst
+++ b/doc/source/parameters/cfd/dynamic_flow_control.rst
@@ -25,7 +25,7 @@ The default parameters are:
 
 .. note::
 
-  When there is a periodic boundary, the wall boundary should be at the inlet only and if the chosen wall boundary is at the inlet, flow rate must be negative and vice versa.
+  If you are simulating a problem with periodic boundary conditions at inlet and outlet walls, it is important to control the volumetric flow rate specifying the boundary id that corresponds to the inlet only (where the flow would have a negative value).
 
 * The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled (inlet or outlet).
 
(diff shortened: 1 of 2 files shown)


doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+    set enable = false
+    set volumetric flow rate = 0
+    set boundary id = 0
+    set flow direction = 0
+    set initial beta = 0
+  end
+
+* The ``enable`` parameter is set to ``true`` to enable an imposed volumetric flow rate.
+
+* The ``volumetric flow rate`` parameter specifies the flow rate (:math:`m^2/s`).
```
oguevremont, line 23:
```suggestion
* The ``volumetric flow rate`` parameter specifies the flow rate (:math:`m^{dim}/s`). 
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
+Dynamic flow control
+~~~~~~~~~~~~~~~~~~~~
+
+This subsection's purpose is to enable a dynamic flow control. It enables setting a volumetric flow on a specific boundary toward to normal of the wall. 
+It calculates a Beta coefficient at each time step and may be added to a source term. If the chosen wall boundary on which a volumetric flow is imposed is at inlet, 
+the volumetric flow rate targeted must be negative since outward normal vector is used. Thereby, if the chosen wall boundary is at outlet, 
+volumetric flow rate must be positive.
```
oguevremont, line 7:
```suggestion
This subsection's purpose is to enable a dynamic flow control. It enables setting a volumetric flow on a specific boundary toward to normal of the wall. 
It calculates a :math:`\beta` coefficient at each time step which may be added to a source term. If the chosen wall boundary on which a volumetric flow is imposed is an inlet, 
the volumetric flow rate targeted must be negative since outward normal vector is used. Thereby, if the chosen wall boundary is an outlet, 
volumetric flow rate must be positive.
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+* The ``volumetric flow rate`` parameter specifies the flow rate (:math:`m^2/s`). 
+
+.. warning::
+
+  This value should be negative if imposed on an inlet boundary, and positive if imposed on an outlet boundary.
+
+.. note::
+
+  When there is a periodic boundary, the wall boundary should be at inlet only and if the chosen wall boundary is at inlet, flow rate must be negative and vice versa.
```
oguevremont, line 31:
```suggestion
  When there is a periodic boundary, the wall boundary should be at the inlet only and if the chosen wall boundary is at an inlet, flow rate must be negative and vice versa.
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled. (inlet or outlet).
```
oguevremont, line 33:
```suggestion
* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled (inlet or outlet).
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+* The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`y` direction.
+
+* The ``initial beta`` parameter sets an initial beta coefficient that may speed up the convergence of the flow rate targeted. Some tests should be done to find a good initial beta coefficient.
```
oguevremont, line 37:
```suggestion
* The ``initial beta`` parameter sets an initial :math:`\beta` coefficient that may speed up the convergence of the flow rate targeted. Some tests should be done to find a good initial :math:`\beta` coefficient.
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+  This value should be negative if imposed on an inlet boundary, and positive if imposed on an outlet boundary.
+
+.. note::
+
+  When there is a periodic boundary, the wall boundary should be at inlet only and if the chosen wall boundary is at inlet, flow rate must be negative and vice versa.
+
+* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled. (inlet or outlet).
+
+* The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`y` direction.
```
oguevremont, line 35:
```suggestion
* The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`z` direction.
```

oguevremont (commented):
That's good! I suggested some math changes mostly, but I think it's a matter of opinion

doc/source/parameters/cfd/dynamic_flow_control.rst
CaroleAnneDaunais: I really need to use the word "at". Now would it be better to use "at the" inlet/outlet?

CaroleAnneDaunais: I agree for the beta part :)

CaroleAnneDaunais: Maybe not at exclusively but something that means the same. It could also be on

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
+Dynamic flow control
+~~~~~~~~~~~~~~~~~~~~
+
+This subsection's purpose is to enable a dynamic flow control. It enables setting a volumetric flow on a specific boundary toward to normal of the wall.
```
lpsaavedra, line 4:
```suggestion
The purpose of this subsection is to enable dynamic flow control. It is important when we want to set a specific volumetric flow on a specific boundary. 
```

doc/source/parameters/cfd/dynamic_flow_control.rst
lpsaavedra: I think I would rewrite this part a little bit like this, taking into account Olivier's comments: 

```suggestion
The purpose of this subsection is to enable dynamic flow control. It is important when we want to target a volumetric flow on a specific boundary. To control the volumetric flow, the code calculates a :math:`\beta`  coefficient at each time step that is used to keep the volumetric flow value inside an acceptable threshold. If the chosen wall boundary on which a volumetric flow is imposed is an inlet, the volumetric flow rate targeted must be negative since the outward normal vector is used for the calculation. Thereby, if the chosen wall boundary is an outlet, volumetric flow rate must be positive.
```
Somehow, I feel that it sounds better to say that the wall is indeed an outlet or inlet instead of saying that it is located at the outlet or inlet.

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+    set initial beta = 0
+  end
+
+* The ``enable`` parameter is set to ``true`` to enable an imposed volumetric flow rate.
+
+* The ``volumetric flow rate`` parameter specifies the flow rate (:math:`m^{dim}/s`). 
+
+.. warning::
+
+  This value should be negative if imposed on an inlet boundary, and positive if imposed on an outlet boundary.
```
lpsaavedra, line 27: I don't know if we should keep this, because it repeats what it is said in the first paragraph.

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. note::
+
+  When there is a periodic boundary, the wall boundary should be at inlet only and if the chosen wall boundary is at inlet, flow rate must be negative and vice versa.
```
lpsaavedra, line 31: I don't really understand this sentence. Does it mean that you should only control the flow at the inlet if there are periodic boundary conditions in the problem? Is it possible to control the flow at several boundary ids at the same time?

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -0,0 +1,38 @@
...
+
+* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled (inlet or outlet).
+
+* The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`z` direction.
+
+* The ``initial beta`` parameter sets an initial :math:`\beta` coefficient that may speed up the convergence of the flow rate targeted. Some tests should be done to find a good initial :math:`\beta` coefficient.
```
lpsaavedra, line 37: I wrote something like this for the periodic hills case, do you think it would be worth to add something similar here or should we just keep it in the specific examples? 

`.. tip:: A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will greatly speed up the convergence. `

lpsaavedra (changes requested):
Nice work, in fact I was trying to explain all this for the example I'm working on at the moment. I have only some suggestions/comments that could improve it

doc/source/parameters/cfd/dynamic_flow_control.rst
CaroleAnneDaunais: This is what I understood from the previous documentation. I think you can only control a volumetric flow on one boundary

doc/source/parameters/cfd/dynamic_flow_control.rst
CaroleAnneDaunais: that sounds great, I will add it!

doc/source/parameters/cfd/dynamic_flow_control.rst
lpsaavedra: Ah ok! So maybe we can rephrase it as this: 

If you are simulating a problem with periodic boundary conditions at inlet and outlet walls, it is important to control the volumetric flow rate specifying the boundary id that corresponds to the inlet. 

What do you think?

CaroleAnneDaunais: that looks great!

blaisb (approved):

lpsaavedra (approved):

oguevremont (approved):

PR #372, opened 2022-01-21
Very minor fix

CaroleAnneDaunais:
# Description of the problem

There was a small error in the source_term parameter section.

# Description of the solution

I fixed it

# Documentation

The documentation is already up to date with the current changes


blaisb:
Can you fix indent?
I'll also wait for CI just in case some test cases break cuz of that :)

jeannejoachim (approved):
Well spotted ^^

PR #376, opened 2022-01-22
Fix late contact detection for curved boundaries

shahabgol:
# Description of the problem

- We had several problems with the late detection of particle-wall contacts, especially in geometries with curved walls. This PR fixes this problem.

# Description of the solution

- In this PR, the particles in each cell are searched for possible collisions with boundary faces of neighbor boundary cells. To this end, we define a variable in the parameter handler (expand particle-wall contact search), which enables this feature. If enabled, a function in find_boundary_cells_information class searches for all the neighbors of all the boundary cells, which have boundary faces. The information (normal vector and point) of these boundary neighbors are added to the boundary_cells_information, so that the particles in the main boundary cell are searched for possible collisions with neighbor boundary cells. One thing to consider is that this feature should only be used in simulation with concave boundaries (for example, particles in a ball). In simulations with convex boundaries, this feature must not be used, since the particles will have collisions with the extension of boundary cells inside the simulation domain. This issue is reported as a warning in simulation with expand particle-wall contact search = true.

# How Has This Been Tested?

- An application test (packing_in_cylinder) is modified to test this feature.

# Documentation

- I will add the parameter to documentation after merging.


applications_tests/dem_3d/packing_in_cylinder.output
```diff
@@ -6,107 +6,109 @@ Starting simulation with Lethe/DEM on 1 processors
 Reading triangulation 
 
 Finished reading triangulation 
+Warning: expansion of particle-wall contact list is enabled. 
+This feature should only be activated in geometries with concanve boundaries. (For example, for particles flow inside a cylinder or sphere). In geometries with convex boundaries, this feature MUST NOT be activated.
```
blaisb, line 10: I would do it the other way around. Enable it by default and if someone disables it , it will be because they know what they are doing and Then you can print this warning.

applications_tests/dem_3d/packing_in_cylinder.prm
```diff
@@ -89,4 +89,5 @@ subsection mesh
     set grid type      	     				= cylinder
     set grid arguments       				= 0.1:0.1
     set initial refinement   				= 3
+    set expand particle-wall contact search		= true
```
blaisb, line 92: Set default value to true

source/core/parameters.cc
```diff
@@ -988,6 +988,12 @@ namespace Parameters
         Patterns::Bool(),
         "Enables checking the input grid for diamond-shaped cells.");
 
+      prm.declare_entry(
+        "expand particle-wall contact search",
+        "false",
```
blaisb, line 993:
```suggestion
        "true",
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -52,6 +55,21 @@ BoundaryCellsInformation<dim>::build(
                                                   outlet_boundaries,
                                                   check_diamond_cells,
                                                   pcout);
+
+  if (expand_particle_wall_contact_search)
+    {
+      pcout
+        << "Warning: expansion of particle-wall contact list is enabled. "
+        << std::endl
+        << "This feature should only be activated in geometries with concanve boundaries. "
```
blaisb, line 64:
```suggestion
        << "This feature should only be activated in geometries with concave boundaries. "
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -644,5 +662,89 @@ BoundaryCellsInformation<dim>::find_boundary_cells_for_floating_walls(
...
+
+                          // Add the main boundary cell with the information
+                          // (point and normal vector) of the neighbor boundary
+                          // cell to the boundary_cells_information container.
+                          // Note that since we may already have an element with
+                          // the key of face_id (key of the
+                          // boundary_cells_information map) in the
+                          // boundary_cells_information, we add the new element
+                          // with a unique key to create a unique id in the map.
+                          // This unique key is generated using Cantor pairing
```
blaisb, line 714: Holy heck, smart idea.

blaisb (commented):
Amazing work. Only alter the default behavior and I think this is good to go. Anyway, in large scale simulations particle-wall contact were never a big deal.

applications_tests/dem_3d/packing_in_cylinder.output
shahabgol: I changed the default value to true. But I think it's better to keep this warning with the current format because users could disable the feature if they have concave boundaries.

tonielgeitani (approved):

blaisb:
@shahabgol  the cfd-dem tests don't pass . Can you also disable the rolling ball case.

PR #381, opened 2022-01-24
Add model A of VANS equations

tonielgeitani:
This PR adds model A of the VANS equations to the solver. It also add a max local mass source post processing term.


source/fem-dem/parameters_cfd_dem.cc
```diff
@@ -116,6 +116,10 @@ namespace Parameters
                       "100",
                       Patterns::Integer(),
                       "dem-cfd coupling frequency");
+    prm.declare_entry("vans model",
+                      "modelB",
```
shahabgol, line 120: Did modelB show higher accuracy?

source/fem-dem/parameters_cfd_dem.cc
```diff
@@ -144,6 +148,14 @@ namespace Parameters
...
     else
       throw(std::runtime_error("Invalid drag model"));
+
+    const std::string op1 = prm.get("vans model");
+    if (op1 == "modelA")
+      vans_model = Parameters::VansModel::modelA;
+    else if (op1 == "modelB")
+      vans_model = Parameters::VansModel::modelB;
+    else
+      throw(std::runtime_error("Invalid vans model"));
```
shahabgol, line 158:
```suggestion
      throw(std::runtime_error("Invalid vans model. Valid choices are modelA and modelB."));
```

shahabgol (commented):
Very clean PR. I have two minor comments. Why the default vans model is modelB? I was expecting modelA to be the default model.

source/fem-dem/parameters_cfd_dem.cc
tonielgeitani: No, both show similar behavior. I just put it as default but we can change it. If we put model A as default than all vans and cfd-dem tests will fail because the output will change slightly. That is why I didn't do that.

shahabgol (approved):
Well done!

source/fem-dem/parameters_cfd_dem.cc
blaisb: We already did the development initially for model B, so I think that's a correct choice to keep B as default. To be honest, we no clue yet which one might be better...

include/fem-dem/parameters_cfd_dem.h
```diff
@@ -84,6 +90,7 @@ namespace Parameters
   {
     bool         grad_div;
     DragModel    drag_model;
+    VansModel    vans_model;
```
blaisb, line 93:
```suggestion
    VANSModel    vans_model;
```

include/fem-dem/parameters_cfd_dem.h
```diff
@@ -53,6 +53,12 @@ namespace Parameters
     dallavalle
   };
 
+  enum class VansModel
```
blaisb, line 56:
```suggestion
  enum class VANSModel
```

blaisb (commented):
Very few comments. Very good work. Once the grad-div PR is merged we can merge this one (once all test pass) :)

PR #383, opened 2022-01-25
Fix the definition of gamma and update tests

blaisb:
# Description of the problem

- Gamma was missing out of the grad-div correction in the BDF assembler
- The definition of gamma being constant was not a good choice

# Description of the solution

- Use a dimensionally consistent definition of gamma
- Apply gamma to the bdf scheme

# How Has This Been Tested?

- Tests results had to be changed because of this. It seems that the pressure drop is slightly altered in the packed bed case. Not significantly but still.

# Comments

- The new grad-div definition should make the model more robust.
- It could be a good idea to add a new application_test for grad-div with the step function
-


tonielgeitani (approved):
Everything is clear to me.

voferreira (approved):
Seems clear to me too.

PR #389, opened 2022-02-01
Update images in Flow around sphere example

CaroleAnneDaunais:
# Description of the problem

The images had a grey background and the mesh images were "triangulated"

# Description of the solution

I fixed the images

# Documentation

There were no major changes to make to the documentation itself


blaisb:
@CaroleAnneDaunais  Why did you remove the image with the mesh adaptation? I'm curious

PR #391, opened 2022-02-02
Introduce PhysicalPropertiesManager

blaisb:
# Description of the problem

- Physical properties model (such as rheology) are always re-created whenever we need them. This is very problematic  since we always need to carry the physical property structure everywhere (assemblers, for calculation, etc.) and we always need to "recreate" the properties everywhere we need them. It's confusing and error prone.

# Description of the solution

- Create a new class, PhysicalPropertyManager that houses all the physical properties for all the model. This class is created from the physical properties parameter. It also natively supports having different physical properties model for different fluids.
- 

# How Has This Been Tested?

- [X] physical_property_manager_01.cc

# Future changes

- The next step will be to remove the physical properties from the NavierStokesParameter and to replace them with the physical properties manager. This will give us the flexibility we finally need. Right now, the present feature is not complete, but I would rather merge it now, because the follow-up PR will be significantly larger.
The following goal will be to replace the calculation of the physical properties in each and every assembler with a calculation of the physical properties in the scratch. This will be significantly more generic and will enable  a better connection between the physics

### DIFF
--- a/applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.prm
+++ b/applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.prm
@@ -28,8 +28,8 @@ subsection physical properties
   set number of fluids = 1
   subsection fluid 0
     set non newtonian flow	= true
+   	set rheological model 	= carreau
     subsection non newtonian
-    	set model 		= carreau
     	subsection carreau
 	    	set viscosity_0		= 1.0
 	    	set viscosity_inf 	= 0
--- a/applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.prm
+++ b/applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.prm
@@ -13,18 +13,17 @@ end
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
-set number of fluids = 1
   subsection fluid 0
     set non newtonian flow	= true
-    set rheological model 		= carreau
-  subsection non newtonian
+    set rheological model 	= carreau
+    subsection non newtonian
     	subsection carreau
 	    	set viscosity_0		= 1.0
 	    	set viscosity_inf 	= 0
 	    	set lambda	      	= 1
 	    	set a	      		= 2.0
 	    	set n	      		= 0.5
-	end
+	  end
   end
 end
 end
--- a/applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.prm
+++ b/applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.prm
@@ -16,8 +16,8 @@ subsection physical properties
 set number of fluids = 1
   subsection fluid 0
     set non newtonian flow	= true
+    set rheological model 		= carreau
   subsection non newtonian
-    	set model 		= carreau
     	subsection carreau
 	    	set viscosity_0		= 1.0
 	    	set viscosity_inf 	= 0
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -187,59 +187,6 @@ namespace Parameters
   };
 
   /**
-   * @brief Fluid - Class for fluid definition
-   */
-  class Fluid
-  {
-  public:
-    Fluid()
-    {}
-
-    void
-    declare_parameters(ParameterHandler &prm, unsigned int id);
-    void
-    parse_parameters(ParameterHandler &prm, unsigned int id);
-
-    // Kinematic viscosity (nu = mu/rho) in units of L^2/s
-    double viscosity;
-    // volumetric mass density (rho) in units of kg/m^3
-    double density;
-    // specific heat capacity (cp) in J/K/kg
-    double specific_heat;
-    // thermal conductivity (k) in W/m/K
-    double thermal_conductivity;
-    // thermal expansion coefficient (alpha) in 1/K
-    double thermal_expansion;
-    // tracer diffusivity) in L^2/s
-    double tracer_diffusivity;
-
-    // Phase change parameters
-    PhaseChange phase_change_parameters;
-
-
-    enum class DensityModel
-    {
-      constant
-    } density_model;
-
-    enum class SpecificHeatModel
-    {
-      constant,
-      phase_change
-    } specific_heat_model;
-
-    enum class ThermalConductivityModel
-    {
-      constant,
-      linear
-    } thermal_conductivity_model;
-
-    // Linear thermal conductivity parameters : k = k_A0 + k_A1 * T
-    double k_A0;
-    double k_A1;
-  };
-
-  /**
    * @brief Power-law rheological model to solve for non Newtonian
    * flows.
    */
@@ -290,13 +237,6 @@ namespace Parameters
 
   struct NonNewtonian
   {
-    // Non Newtonian model
-    enum class Model
-    {
-      powerlaw,
-      carreau
-    } model;
-
     CarreauParameters  carreau_parameters;
     PowerLawParameters powerlaw_parameters;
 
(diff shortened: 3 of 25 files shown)


applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.prm
```diff
@@ -13,18 +13,17 @@ end
 # Physical Properties
 #---------------------------------------------------
 subsection physical properties
-set number of fluids = 1
```
CaroleAnneDaunais, line 16: Do you want to remove it since the default value is 1?

blaisb: yeah i'm not sure it's worth specifying if the default value is good. I prefer the "don't rewrite the parameter if the default value is used approach" since it forces us to have good and meaningful default values :)

include/core/rheological_model.h
```diff
@@ -45,7 +45,7 @@ class RheologicalModel : public PhysicalPropertyModel
    * Newtonian flow
    */
   static std::shared_ptr<RheologicalModel>
-  model_cast(const Parameters::PhysicalProperties &physical_properties);
+  model_cast(const Parameters::Fluid &fluid_properties);
```
CaroleAnneDaunais, line 48: Yes I totally agree with this architecture!

applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.prm
```diff
@@ -28,8 +28,8 @@ subsection physical properties
   set number of fluids = 1
   subsection fluid 0
     set non newtonian flow	= true
+   	set rheological model 	= carreau
```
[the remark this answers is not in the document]
jeannejoachim: Ok I got my answer in the parameter.h modifications. It's a good first implementation, but I wonder if we could make it a little more user-intuitive in the .prm

include/solvers/post_processors.h
```diff
@@ -182,7 +182,8 @@ class NonNewtonianViscosityPostprocessor : public DataPostprocessorScalar<dim>
     Parameters::PhysicalProperties p_physical_properties)
     : DataPostprocessorScalar<dim>("viscosity", update_gradients)
   {
-    rheological_model = RheologicalModel::model_cast(p_physical_properties);
+    rheological_model =
+      RheologicalModel::model_cast(p_physical_properties.fluids[0]);
```
jeannejoachim, line 186: same question as in the `navier_stokers assemblers.h`

source/core/parameters.cc
```diff
@@ -542,7 +519,46 @@ namespace Parameters
...
+      phase_change_parameters.declare_parameters(prm);
+
+      prm.declare_entry(
+        "thermal conductivity model",
+        "constant",
+        Patterns::Selection("constant|linear"),
+        "Model used for the calculation of the thermal conductivity"
+        "Choices are <constant|linear>.");
+
+      prm.declare_entry("k_A0",
```
jeannejoachim, line 553: we should think about a more explicit name in the .prm, for this parameters

source/core/rheological_model.cc
```diff
@@ -1,33 +1,26 @@
...
 std::shared_ptr<RheologicalModel>
-RheologicalModel::model_cast(
-  const Parameters::PhysicalProperties &physical_properties)
+RheologicalModel::model_cast(const Parameters::Fluid &fluid_properties)
 {
-  if (!physical_properties.fluids[0].non_newtonian_flow)
-    return std::make_shared<Newtonian>(physical_properties.fluids[0].viscosity);
-  else if (physical_properties.fluids[0].non_newtonian_parameters.model ==
-           Parameters::NonNewtonian::Model::powerlaw)
+  if (!fluid_properties.non_newtonian_flow)
```
jeannejoachim, line 6: I like that `physical_properties.fluid[0]` is now `fluid_properties`, but I see less clearly how this will apply with two-fluids simulations, with one newtonian fluid and the other non-newtonian. Could you clarify how this will be generalized in the future?

source/solvers/physical_properties_manager.cc
```diff
@@ -0,0 +1,52 @@
...
+PhysicalPropertiesManager::initialize(
+  Parameters::PhysicalProperties physical_properties)
+{
+  number_of_fluids = physical_properties.number_of_fluids;
+
+  non_newtonian_flow = false;
+  // For each fluid, declare the physical properties
+  for (unsigned int f = 0; f < number_of_fluids; ++f)
+    {
+      density.push_back(
```
jeannejoachim, line 37: we do not need also `kinematic_viscosity` and `tracer_diffusivity` in the same fashion?

jeannejoachim (approved):
I like the PhysicalPropertiesManager implementation very much! Thank you also for the test `physical_properties_manager_01` which make it a bit clearer.

I have a few questions here and there. I think I will need to see how it is applied in the solvers to really grasp the implication of this restructuring! Thank you for taking this on :)

include/solvers/simulation_parameters.h
```diff
@@ -128,7 +126,6 @@ class SimulationParameters
     forces_parameters.parse_parameters(prm);
     post_processing.parse_parameters(prm);
     flow_control.parse_parameters(prm);
-    non_newtonian.parse_parameters(prm);
```
CaroleAnneDaunais, line 131: Oops my mistake, this should have been removed a long time ago

CaroleAnneDaunais (approved):
Nice work :)

include/solvers/navier_stokes_assemblers.h
```diff
@@ -184,7 +184,8 @@ class GLSNavierStokesAssemblerNonNewtonianCore
     : simulation_control(simulation_control)
     , physical_properties(physical_properties)
   {
-    rheological_model = RheologicalModel::model_cast(physical_properties);
+    rheological_model =
+      RheologicalModel::model_cast(physical_properties.fluids[0]);
```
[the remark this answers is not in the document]
CaroleAnneDaunais: I think it is hard-coded here but will eventually not be hard-coded when added to the VOF assemblers (in navier_stokes_vof_assembler). I'm not sure if the VOF assembler uses this assembler, if yes, it will definitely have to be fixed in the future :)

source/core/rheological_model.cc
CaroleAnneDaunais: So the way I see this in when using the VOF method, you will have a fluids vector containing (two or more) fluids, each containing their parameters. Let's say the first fluid, i.e. fluids[0], has its member non_newtonian_flow set to false, and the second fluid has its member non_newtonian_flow set to true. Now instead of using physical_properties, which before contained the properties for all fluids, you now access to each fluid properties through fluid_properties. when generating the rheological_model object for the first fluid, this method will cast the rheological_model object to be a Newtonian model. For the second fluid, it's rheological_model object will be of class Carreau or PowerLaw. This way, when calculating the visocsity in the first fluid, this fluid's rheological_model object will return a constant viscosity. When calculating the viscosity in the second fluid, the returned viscosity will be calculated depending on the rheological model used. I'm not sure if I answered you question correctly, don't hesitate to reiterate

include/core/parameters.h
```diff
@@ -234,55 +260,55 @@ namespace Parameters
...
+    PhaseChange phase_change_parameters;
 
+    // Non Newtonian model parameters
+    bool non_newtonian_flow;
+    enum class RheologyModel
+    {
+      powerlaw,
+      carreau,
+      newtonian
+    } rheology_model;
```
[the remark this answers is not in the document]
lpsaavedra: This is used to declare an object of that type right after declaring the enum class. 
Using your example, that code does exactly the same as the following:
```
enum class Name {x,y,z};
Name variable_name;
```
Here we declare them to use them in parameters.cc file. The same can be done also for a class or a struct.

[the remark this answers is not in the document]
CaroleAnneDaunais: So a Fluid variable has a rheology_model member (of type RheologyModel), that can only have the given terms in curly brackets as value. The use of a enum class allows to "dynamic cast" it to check it's value or type. For example, to see if my rheology_model is of type powerlaw, I can do something like : `if (fluid_properties.rheology_model ==
           Parameters::Fluid::RheologyModel::powerlaw)`
there are also others ways to do this.

source/core/parameters.cc
CaroleAnneDaunais: I agree!

applications_tests/gd_navier_stokes_2d/apparent_viscosity_carreau_gd.prm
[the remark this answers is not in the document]
blaisb: I wanted to treat the newtonian case as a case like any other rheological model. But yeah it's not trivial. We can clearly revisit this latter and i'll be glad to.

include/solvers/navier_stokes_assemblers.h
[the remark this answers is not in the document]
blaisb: @CaroleAnneDaunais  Is spot-on. this is temporary. The goal is to migrate such that this is not hardcoded anymore and to enable multifluid simulations with different type of properties :)

source/core/parameters.cc
blaisb: Abs-So-Lu-Tely. I Just did not know how to name them. Nobody uses it for now, but we can clearly revisit that. If you have any ideas, i'll gladly do it now

source/solvers/physical_properties_manager.cc
blaisb: Kinematic viscosity is calcualted through the Carreau model. Tracer diffusivity will need to be added!

include/solvers/navier_stokes_assemblers.h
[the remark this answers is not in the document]
jeannejoachim: alright, looking forward to it :) (you can ask my review for the next PR on that also!)

source/core/rheological_model.cc
jeannejoachim: yes that clarifies it perfectly, thank you @CaroleAnneDaunais :)

CaroleAnneDaunais: no worries :)

source/core/parameters.cc
jeannejoachim: For this PR you can merge to get going, I need to give some thoughts about it. @CaroleAnneDaunais let's brainstorm that together :)

PR #392, opened 2022-02-03
Fix vof conservation

jeannejoachim:
# Description of the problem

- Equation for VOF conservation monitoring was wrong for phase 0
- Parameter name was unclear

# Description of the solution

- Fix equation
- Change parameter to `fluid monitored`

# How Has This Been Tested?

No test added, I tested it on a local example (water droplet), with theoretical fluid surface known. 

I am unsure if it should become a test, do not hesitate to tell me if it should, I can add it to this PR!

# Documentation

- [x] Parameters/Multiphysics, to update the parameter name and to make the text clearer (considering @shahabgol questions).

# Comments

- It took me a while to see what was under my nose, but at least it is fixed now!


shahabgol (approved):
Thanks for the quick response. :)

blaisb (approved):

PR #393, opened 2022-02-08
Use physical properties manager

blaisb:
# Description of the problem

- The Physical Property Manager was developed and well implemented, but it was not used yet in the assemblers and scratch

# Description of the solution

- Add the usage of the physical Property manager in 

- [x] Heat Transfer
- [x] Tracer
- [x] Fluid Dynamics
- [x] VANS

- Add the physical properties model for tracer diffusivity and thermal expansion coefficient

# How Has This Been Tested?

- If all previous tests are working, I think everything is a go
- Added a new test for linear thermal conductivity (gls_navier_stokes_2d/heat_transfer_linear_thermal_conductivity.prm) which tests a steady-state heat problem with a linear thermal conductivity (k= 1+2*T). It converges pretty well. This enabled me to fix some bugs

# Documentation

Documentation has been adapted to take into account the changes.

# Future changes

-  Integrate the locally evaluated density and viscosity for the CFD-DEM models.

### DIFF
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -37,8 +37,8 @@ enum field : int
 };
 
 inline void
-set_field_vector(const field                          id,
-                 const std::vector<double> &          data,
+set_field_vector(const field                           id,
+                 const std::vector<double> &           data,
                  std::map<field, std::vector<double>> &fields)
 {
   std::vector<double> &target = fields.at(id);
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -36,15 +36,29 @@ enum field : int
   previous_temperature
 };
 
+inline void
+set_field_vector(const field                          id,
+                 const std::vector<double> &          data,
+                 std::map<field, std::vector<double>> fields)
+{
+  std::vector<double> &target = fields.at(id);
+  size_t               sz     = target.size();
+  for (size_t i = 0; i < sz; ++i)
+    {
+      target[i] = data[i];
+    }
+}
+
 /**
  * @brief PhysicalPropertyModel. Abstract class that defines the interface for a physical property model
- * Physical property model provides an abstract interface to calculate the value
- * of a physical property or a vector of physical property value for given field
- * value. By default, the interface does not require that all (or any) fields be
- * specified. This is why a map is used to pass the dependent variable. To allow
- * for the calculation of the appropriate jacobian matrix (when that is
- * necessary) the interface also provides a jacobian function, which must
- * provide the derivative with respect to the field specified as an argument.
+ * Physical property model provides an abstract interface to calculate the
+ * value of a physical property or a vector of physical property value for
+ * given field value. By default, the interface does not require that all (or
+ * any) fields be specified. This is why a map is used to pass the dependent
+ * variable. To allow for the calculation of the appropriate jacobian matrix
+ * (when that is necessary) the interface also provides a jacobian function,
+ * which must provide the derivative with respect to the field specified as an
+ * argument.
  */
 class PhysicalPropertyModel
 {
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -73,6 +73,16 @@ public:
     model_depends_on[previous_temperature] = false;
   }
 
+  /**
+   * @brief Returns true if the PhysicalPropertyModel depends on a field, false if not.
+   */
+
+  inline bool
+  depends_on(field id)
+  {
+    return model_depends_on[id];
+  }
+
 
   /**
    * @brief value Calculates the value of a physical property.
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -37,7 +37,7 @@ enum field : int
 };
 
 inline void
-set_field_vector(const field                           id,
+set_field_vector(const field                        &   id,
                  const std::vector<double> &           data,
                  std::map<field, std::vector<double>> &fields)
 {
@@ -78,7 +78,7 @@ public:
    */
 
   inline bool
-  depends_on(field id)
+  depends_on(const field &id)
   {
     return model_depends_on[id];
   }
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -39,7 +39,7 @@ enum field : int
 inline void
 set_field_vector(const field                          id,
                  const std::vector<double> &          data,
-                 std::map<field, std::vector<double>> fields)
+                 std::map<field, std::vector<double>> &fields)
 {
   std::vector<double> &target = fields.at(id);
   size_t               sz     = target.size();
--- a/include/core/physical_property_model.h
+++ b/include/core/physical_property_model.h
@@ -37,7 +37,7 @@ enum field : int
 };
 
 inline void
-set_field_vector(const field                        &   id,
+set_field_vector(const field &                         id,
                  const std::vector<double> &           data,
                  std::map<field, std::vector<double>> &fields)
 {
--- a/include/core/thermal_expansion_model.h
+++ b/include/core/thermal_expansion_model.h
@@ -40,13 +40,13 @@ public:
 /**
  * @brief Constant thermal expansion coefficient.
  */
-class ThermalExpansionConstant : public ThermalExpansionModel
+class ConstantThermalExpansion : public ThermalExpansionModel
 {
 public:
   /**
    * @brief Default constructor
    */
-  ThermalExpansionConstant(const double p_thermal_expansion)
+  ConstantThermalExpansion(const double p_thermal_expansion)
     : thermal_expansion(p_thermal_expansion)
   {}
 
--- a/include/core/thermal_expansion_model.h
+++ b/include/core/thermal_expansion_model.h
@@ -1,0 +1,113 @@
+/* ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2019 -  by the Lethe authors
+ *
+ * This file is part of the Lethe library
+ *
+ * The Lethe library is free software; you can use it, redistribute
+ * it, and/or modify it under the terms of the GNU Lesser General
+ * Public License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ * The full text of the license can be found in the file LICENSE at
+ * the top level of the Lethe distribution.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+#ifndef lethe_thermal_expansion_model_h
+#define lethe_thermal_expansion_model_h
+
+#include <core/physical_property_model.h>
+
+/**
+ * @brief ThermalExpansionModel. Abstract class that allows to calculate the
+ * thermal expansion coefficient (with unit of inverse temperature)
+ */
+class ThermalExpansionModel : public PhysicalPropertyModel
+{
+public:
+  /**
+   * @brief Instantiates and returns a pointer to a ThermalExpansionModel object by casting it to
+   * the proper child class
+   *
+   * @param physical_properties Parameters for a single fluid
+   */
+  static std::shared_ptr<ThermalExpansionModel>
+  model_cast(const Parameters::Fluid &fluid_properties);
+};
+
+
+/**
+ * @brief Constant thermal expansion coefficient.
+ */
+class ThermalExpansionConstant : public ThermalExpansionModel
+{
+public:
+  /**
+   * @brief Default constructor
+   */
+  ThermalExpansionConstant(const double p_thermal_expansion)
+    : thermal_expansion(p_thermal_expansion)
+  {}
+
+  /**
+   * @brief value Calculates the thermale expansion
+   * @param fields_value Value of the various field on which the property may depend.
+   * @return value of the physical property calculated with the fields_value
+   */
+  double
+  value(const std::map<field, double> & /*fields_value*/) override
+  {
+    return thermal_expansion;
+  };
+
+  /**
+   * @brief vector_value Calculates the vector of thermal expansion.
+   * @param field_vectors Vectors of the fields on which the thermal expansion may depend.
+   * @param property_vector Vectors of the thermal expansion values
+   */
+  void
+  vector_value(const std::map<field, std::vector<double>> & /*field_vectors*/,
+               std::vector<double> &property_vector) override
+  {
+    std::fill(property_vector.begin(),
+              property_vector.end(),
+              thermal_expansion);
+  }
+
+  /**
+   * @brief jacobian Calculates the jacobian (the partial derivative) of the thermal expansion with respect to a field
+   * @param field_values Value of the various fields on which the property may depend.
+   * @param id Indicator of the field with respect to which the jacobian
+   * should be calculated.
+   * @return value of the partial derivative of the thermal expansion with respect to the field.
+   */
+
+  double
+  jacobian(const std::map<field, double> & /*field_values*/,
+           field /*id*/) override
+  {
+    return 0;
+  };
+
+  /**
+   * @brief vector_jacobian Calculates the derivative of the thermal expansion with respect to a field.
+   * @param field_vectors Vector for the values of the fields used to evaluate the property.
+   * @param id Identifier of the field with respect to which a derivative should be calculated.
+   * @param jacobian vector of the value of the derivative of the thermal expansion with respect to the field id.
+   */
+
+  void
+  vector_jacobian(
+    const std::map<field, std::vector<double>> & /*field_vectors*/,
+    const field /*id*/,
+    std::vector<double> &jacobian_vector) override
+  {
+    std::fill(jacobian_vector.begin(), jacobian_vector.end(), 0);
+  };
+
+private:
+  const double thermal_expansion;
+};
+
+#endif
--- a/include/solvers/navier_stokes_scratch_data.h
+++ b/include/solvers/navier_stokes_scratch_data.h
@@ -15,7 +15,12 @@
  */
 
 #include <core/bdf.h>
+#include <core/density_model.h>
 #include <core/parameters.h>
+#include <core/physical_property_model.h>
+#include <core/rheological_model.h>
+
+#include <solvers/physical_properties_manager.h>
 
 #include <dem/dem.h>
 #include <dem/dem_properties.h>
@@ -32,6 +37,7 @@
 #include <deal.II/numerics/vector_tools.h>
 
 #include <deal.II/particles/particle_handler.h>
+
 
 
 #ifndef lethe_navier_stokes_scratch_data_h
@@ -79,11 +85,13 @@ public:
    * @param mapping The mapping of the domain in which the Navier-Stokes equations are solved
    *
    */
-  NavierStokesScratchData(const FESystem<dim> &      fe,
-                          const Quadrature<dim> &    quadrature,
-                          const Mapping<dim> &       mapping,
+  NavierStokesScratchData(PhysicalPropertiesManager &properties_manager,
+                          const FESystem<dim>       &fe,
+                          const Quadrature<dim>     &quadrature,
+                          const Mapping<dim>        &mapping,
                           const Quadrature<dim - 1> &face_quadrature)
-    : fe_values(mapping,
+    : properties_manager(properties_manager)
+    , fe_values(mapping,
                 fe,
                 quadrature,
                 update_values | update_quadrature_points | update_JxW_values |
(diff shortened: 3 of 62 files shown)


include/core/physical_property_model.h
```diff
@@ -36,15 +36,29 @@ enum field : int
   previous_temperature
 };
 
+inline void
+set_field_vector(const field                          id,
+                 const std::vector<double> &          data,
+                 std::map<field, std::vector<double>> fields)
+{
+  std::vector<double> &target = fields.at(id);
```
CaroleAnneDaunais, line 44: Does this mean that the given fields parameter already contains an element for each possible id (i.e. shear_rate, temperature and previous_temperature)? What I mean is, when this function will be called, is the fields parameters something that looks like this:
`{
{shear_rate, []},
{temperature, []},
{previous_temperature, []},
}` ?
If yes, why would it already contain all 3 fields? If not, the at() method will not work if the field is not already existing. The [ ] accessor on the other had will get the proper field if already existing, but will create it if it does not exist. We'll discuss that tomorrow :)

include/core/physical_property_model.h
```diff
@@ -36,15 +36,29 @@ enum field : int
...
+  size_t               sz     = target.size();
+  for (size_t i = 0; i < sz; ++i)
+    {
+      target[i] = data[i];
```
CaroleAnneDaunais, line 48: The std::vector class has it's own copy constructor. It could look something like:
`std::vector<double> target(fields[id]); `

include/core/physical_property_model.h
CaroleAnneDaunais: I think I better understand now that the specified id has to be in the fields.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -14,12 +14,11 @@ HeatTransferAssemblerCore<dim>::assemble_matrix(
 {
   // Gather physical properties in case of mono fluids simulations (to be
   // modified by cell in case of multiple fluids simulations)
-  double density       = this->physical_properties.fluids[0].density;
-  double specific_heat = this->physical_properties.fluids[0].specific_heat;
-  double thermal_conductivity =
-    this->physical_properties.fluids[0].thermal_conductivity;
-  double rho_cp = density * specific_heat;
-  double alpha  = thermal_conductivity / rho_cp;
+  const std::vector<double> &density       = scratch_data.density;
```
CaroleAnneDaunais, line 17: Goodbye simplification to the Navier-Stokes equations

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -47,29 +46,8 @@ HeatTransferAssemblerCore<dim>::assemble_matrix(
   // assembling local matrix and right hand side
   for (unsigned int q = 0; q < n_q_points; ++q)
     {
-      if (this->multiphysics_parameters.VOF)
```
CaroleAnneDaunais, line 50: Will this be taken care of in the VOF assembler?

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -79,6 +90,114 @@ HeatTransferScratchData<dim>::enable_vof(const FiniteElement<dim> &fe,
...
+          thermal_conductivity[q] =
+            calculate_point_property(this->phase_values[q],
+                                     this->thermal_conductivity_0[q],
+                                     this->thermal_conductivity_1[q]);
+
+          viscosity[q] = calculate_point_property(this->phase_values[q],
+                                                  this->viscosity_0[q],
+                                                  this->viscosity_1[q]);
+        }
+    }
```
CaroleAnneDaunais, line 200: Love this

include/core/physical_property_model.h
blaisb: Yeah but here we don't want to allocate a new vector right, just to copy the data into it.We want to avoid dynamic allocation of vectors when assembling the matrices.

include/core/physical_property_model.h
blaisb: It's a very good point. The field vector has to have that memory pre-allocated, otherwise it won't work. I faced a challenge here. I wanted to have a flexible structure, but then I ended up not wanting to do dynamic memory allocation in the scratch reinit, because that's the whole point of the scratch. So my best solution was to allocate the memory from the get_go for the scratch, but only fill the fields that we would need.

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -59,16 +64,24 @@ HeatTransferScratchData<dim>::allocate()
...
+    std::pair<field, std::vector<double>>(field::temperature, n_q_points));
+  fields.insert(
+    std::pair<field, std::vector<double>>(field::previous_temperature,
+                                          n_q_points));
+  fields.insert(
+    std::pair<field, std::vector<double>>(field::shear_rate, n_q_points));
   specific_heat        = std::vector<double>(n_q_points);
   density              = std::vector<double>(n_q_points);
   thermal_conductivity = std::vector<double>(n_q_points);
+  viscosity            = std::vector<double>(n_q_points);
```
CaroleAnneDaunais, line 77: What to these lines do that's different from lines 36 to 39? I think I'm a bit confused

CaroleAnneDaunais: Awesome I'll make the changes for Navier-stokes :)

blaisb:
@CaroleAnneDaunais  I think everything is migrated. However, I am not very satisfied with how density is handled right now. It's making me unsure if we should make it as a property.
I propose we review the solver as-is now, as this PR is already getting our of hand. once it's merged, I'll continue the cleanup. I'll start a github project page to keep note of what needs to be cleaned.

blaisb:
@CaroleAnneDaunais  I added defensive assertions in the locations where the non-constant properties are not acceptable yet. Future work will be to extend these things, but I'd rather move onward with this PR now instead of making it crazy large. Right now everything works  and things are, I believe, simpler :).

source/solvers/tracer_scratch_data.cc
```diff
@@ -50,5 +54,39 @@ TracerScratchData<dim>::allocate()
...
+    {
+      // In this case, only viscosity is the required property
+      const auto diffusivity_model =
+        properties_manager.get_tracer_diffusivity();
+      diffusivity_model->vector_value(fields, tracer_diffusivity);
+    }
+  else // properties_manager.get_number_of_fluids() == 2
+    {
+      // In this case,  we need both density and viscosity
+      const auto diffusivity_model_0 = properties_manager.get_density(0);
```
jeannejoachim, line 72: Why did you choose to create two variables, rather than a vector containing them? (thus having `diffusivity_model[0]` or something like that)

include/core/thermal_expansion_model.h
```diff
@@ -0,0 +1,113 @@
...
+  ThermalExpansionConstant(const double p_thermal_expansion)
+    : thermal_expansion(p_thermal_expansion)
+  {}
+
+  /**
+   * @brief value Calculates the thermale expansion
+   * @param fields_value Value of the various field on which the property may depend.
+   * @return value of the physical property calculated with the fields_value
+   */
+  double
```
jeannejoachim, line 58: why not `inline` here?

include/core/thermal_expansion_model.h
```diff
@@ -0,0 +1,113 @@
...
+  value(const std::map<field, double> & /*fields_value*/) override
```
jeannejoachim, line 59: why `/*fields_value*/` is commented? Does this mean that right now nothing is parsed, but it should be the `fields_values` eventually? (so it's a WIP)

include/core/thermal_expansion_model.h
```diff
@@ -0,0 +1,113 @@
...
+ * ---------------------------------------------------------------------
+ */
+
+#ifndef lethe_thermal_expansion_model_h
+#define lethe_thermal_expansion_model_h
+
+#include <core/physical_property_model.h>
+
+/**
+ * @brief ThermalExpansionModel. Abstract class that allows to calculate the
```
jeannejoachim, line 23: So it's abstract, just as `AuxiliaryPhysics`? If yes, should not the methods be `virtual`?

include/solvers/physical_properties_manager.h
```diff
@@ -23,10 +23,29 @@
...
 
 using namespace dealii;
 
 
+DeclException1(
+  RequiresConstantDensity,
+  std::string,
+  "The following assembler or post-processing utility "
+    << arg1
+    << " that you are trying to use does not support the use of a non-constant density model. Modifications to Lethe are required to take this into account.");
```
jeannejoachim, line 37: why is this line to long? It should have been cut by the clang formatting, no?

include/solvers/physical_properties_manager.h
```diff
@@ -23,10 +23,29 @@
...
+
+
+DeclException1(
+  RequiresConstantViscosity,
+  std::string,
+  "The following assembler or post-processing utility "
+    << arg1
+    << " that you are trying to use does not support the use of a non-constant viscosity model. Modifications to Lethe are required to take this into account.");
```
jeannejoachim, line 45: same as above

jeannejoachim (approved):
What a PR! I have few questions here and there. Overall I think the code really gained a lot of clarity, well done for this new structure!

source/solvers/gd_navier_stokes.cc
```diff
@@ -766,17 +765,25 @@ GDNavierStokesSolver<dim>::set_initial_condition_fd(
...
     {
       this->set_nodal_values();
-      double viscosity =
-        this->simulation_parameters.physical_properties.fluids[0].viscosity;
-      this->simulation_parameters.physical_properties.fluids[0].viscosity =
-        this->simulation_parameters.initial_condition->viscosity;
+      std::shared_ptr<RheologicalModel> original_viscosity_model =
+        this->simulation_parameters.physical_properties_manager.get_rheology();
+
+      std::shared_ptr<Newtonian> temporary_rheology =
```
shahabgol, line 771: maybe add a comment to explain what temporary_rheology is doing?

source/solvers/gls_navier_stokes.cc
```diff
@@ -730,17 +723,25 @@ GLSNavierStokesSolver<dim>::set_initial_condition_fd(
...
     {
       this->set_nodal_values();
-      double viscosity =
-        this->simulation_parameters.physical_properties.fluids[0].viscosity;
-      this->simulation_parameters.physical_properties.fluids[0].viscosity =
-        this->simulation_parameters.initial_condition->viscosity;
+      std::shared_ptr<RheologicalModel> original_viscosity_model =
+        this->simulation_parameters.physical_properties_manager.get_rheology();
+
+      std::shared_ptr<Newtonian> temporary_rheology =
```
shahabgol, line 729: the same comment here.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -373,30 +309,14 @@ HeatTransferAssemblerBDF<dim>::assemble_rhs(
...
-
-          specific_heat = calculate_point_property(
-            scratch_data.phase_values[q],
-            this->physical_properties.fluids[0].specific_heat,
-            this->physical_properties.fluids[1].specific_heat);
-
-          // Useful definitions
-          rho_cp = density * specific_heat;
-        }
+      double rho_cp = density[q] * specific_heat[q];
```
shahabgol, line 315:
```suggestion
      const double rho_cp = density[q] * specific_heat[q];
```

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -569,28 +491,7 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
-
-          viscosity = calculate_point_property(
-            scratch_data.phase_values[q],
-            this->physical_properties.fluids[0].viscosity,
-            this->physical_properties.fluids[1].viscosity);
-
-          // Useful definitions
-          dynamic_viscosity = viscosity * density;
-        }
+      double dynamic_viscosity = viscosity[q] * density[q];
```
shahabgol, line 494:
```suggestion
      const double dynamic_viscosity = viscosity[q] * density[q];
```

include/solvers/physical_properties_manager.h
blaisb: Good question. I'll manually break it into multiple lines.

source/solvers/tracer_scratch_data.cc
blaisb: I don't have any good reason for that. I guess that was mostly laziness? I could make a routine for the property_manager that returns the vector of density model straight away and that could be a good way to work. I'll fix that.

source/solvers/tracer_scratch_data.cc
```diff
@@ -50,5 +54,39 @@ TracerScratchData<dim>::allocate()
...
+{
+  // Case where you have one fluid
+  if (properties_manager.get_number_of_fluids() == 1)
+    {
+      // In this case, only viscosity is the required property
+      const auto diffusivity_model =
+        properties_manager.get_tracer_diffusivity();
+      diffusivity_model->vector_value(fields, tracer_diffusivity);
+    }
+  else // properties_manager.get_number_of_fluids() == 2
```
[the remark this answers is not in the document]
blaisb: Agreed, defensive programming is good programming. I'll add that.

source/fem-dem/vans_assemblers.cc
```diff
@@ -59,7 +59,7 @@ GLSVansAssemblerCoreModelB<dim>::assemble_matrix(
       const double u_mag = std::max(velocity.norm(), 1e-12);
 
       // Grad-div weight factor used to be constant 0.1
-      const double gamma = viscosity + u_mag * h;
+      const double gamma = viscosity[q] + u_mag * h;
```
shahabgol, line 62: store viscosity[q] as a local variable?

source/fem-dem/vans_assemblers.cc
```diff
@@ -234,12 +234,12 @@ GLSVansAssemblerCoreModelB<dim>::assemble_rhs(
         this->simulation_control->get_assembly_method() ==
             Parameters::SimulationControl::TimeSteppingMethod::steady ?
           1. / std::sqrt(std::pow(2. * u_mag / h, 2) +
-                         9 * std::pow(4 * viscosity / (h * h), 2)) :
+                         9 * std::pow(4 * viscosity[q] / (h * h), 2)) :
```
shahabgol, line 237: same comment.

source/fem-dem/vans_assemblers.cc
```diff
@@ -374,9 +374,9 @@ GLSVansAssemblerCoreModelA<dim>::assemble_matrix(
...
```
shahabgol, line 377: same comment.

source/fem-dem/vans_assemblers.cc
```diff
@@ -539,9 +540,9 @@ GLSVansAssemblerCoreModelA<dim>::assemble_rhs(
...
```
shahabgol, line 543: same comment.

include/core/physical_property_model.h
```diff
@@ -36,15 +36,29 @@ enum field : int
   previous_temperature
 };
 
+inline void
```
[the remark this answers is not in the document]
blaisb: Inlining getters is generally a good idea. However, if the function has inheritance (e..g class B derives from class A) inlining will not occur because of the need to go through a vtable. So the TLDR would be : If the function is used by the class or the class has no inheritance tree, inline functions in the .h if they are short, otherwise dont :).

source/fem-dem/vans_assemblers.cc
```diff
@@ -924,7 +962,7 @@ GLSVansAssemblerDallavalle<dim>::calculate_particle_fluid_interactions(
       // Particle's Reynolds number
       double re = 1e-1 + relative_velocity.norm() *
                            particle_properties[DEM::PropertiesIndex::dp] /
-                           physical_properties.fluids[0].viscosity;
+                           viscosity;
```
shahabgol, line 965:
```suggestion
                           (viscosity + DBL_MIN);
```

include/core/thermal_expansion_model.h
blaisb: I don't think the virtual keyword here is necessary. Since the class does not implement the value, vector_value and etc. of the base class, I don't think it could be instantiated (although I might be wrong ) :).

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -211,14 +211,14 @@ GLSNavierStokesAssemblerCore<dim>::assemble_rhs(
         this->simulation_control->get_assembly_method() ==
             Parameters::SimulationControl::TimeSteppingMethod::steady ?
           1. / std::sqrt(std::pow(2. * u_mag / h, 2) +
-                         9 * std::pow(4 * viscosity / (h * h), 2)) :
+                         9 * std::pow(4 * viscosity[q] / (h * h), 2)) :
```
shahabgol, line 214: same comment.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -523,20 +511,17 @@ GLSNavierStokesAssemblerNonNewtonianCore<dim>::assemble_rhs(
...
         this->simulation_control->get_assembly_method() ==
             Parameters::SimulationControl::TimeSteppingMethod::steady ?
-          1. /
-            std::sqrt(std::pow(2. * u_mag / h, 2) +
-                      9 * std::pow(4 * non_newtonian_viscosity / (h * h), 2)) :
-          1. /
-            std::sqrt(std::pow(sdt, 2) + std::pow(2. * u_mag / h, 2) +
-                      9 * std::pow(4 * non_newtonian_viscosity / (h * h), 2));
+          1. / std::sqrt(std::pow(2. * u_mag / h, 2) +
+                         9 * std::pow(4 * viscosity[q] / (h * h), 2)) :
```
shahabgol, line 515: same comment.

source/fem-dem/vans_assemblers.cc
blaisb: The viscosity is always defined positive, why add a DBL_MIN? I don't think the model makes much sense with viscosity = 0 ?

include/core/thermal_expansion_model.h
jeannejoachim: interesting, I did not know this usage!

source/fem-dem/vans_assemblers.cc
shahabgol: In the 3d simulation I'm doing to find the pressure difference between inside and outside of a bubble, the viscosity of the phases are equal to zero (so that the NS simulations can be solved analytically). But if you think it's unnecessary, we can ignore it.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -720,12 +734,28 @@ class NavierStokesScratchData
...
-  /**
-   * @brief enable_hessian Enables the collection of the hesian tensor when it's a non Newtonian flow
+  /** @brief Calculates the physical properties. This function calculates the physical properties
+   * that may be required by the fluid dynamics problem. Namely the kinematic
+   * viscosity and, when required, the density.
+   *
    */
-
   void
-  enable_hessian();
```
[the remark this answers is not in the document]
blaisb: It is not necessary anymore because the scratch can decide if the hessians are required from the rheologicla model.

source/solvers/multiphysics_interface.cc
```diff
@@ -6,7 +6,7 @@
 
 template <int dim>
 MultiphysicsInterface<dim>::MultiphysicsInterface(
-  const SimulationParameters<dim> &                            nsparam,
+  SimulationParameters<dim> &                                  nsparam,
```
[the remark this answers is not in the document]
blaisb: I managed to revert this back and make it const. It is back to being const now, like it should have been all along.

source/solvers/tracer_scratch_data.cc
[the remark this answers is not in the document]
blaisb: Made  it a switch. Cleaner.

source/fem-dem/vans_assemblers.cc
blaisb: Ohh ok, I did not think about it this way. I'll add this then.

source/solvers/gd_navier_stokes.cc
blaisb: Agreed :)!. Done

blaisb:
I applied everyone's comment. Thank you for the good reviews :). I'll merge tomorrow once CI has passed.

blaisb:
Documentation has been finalized this morning :).

PR #394, opened 2022-02-08
Refactor dem 2d

shahabgol:
# Description of the problem

- Previously, the force and torque tensors in two-dimensional DEM solver (dem_2d) had x and y elements. This caused issued in the dem_2d, as the rotation of particles in the third dimension was neglected. We need to address this issue before working on the resolved CFD-DEM, since resolved CFD-DEM will use dem_2d.

# Description of the solution

- All the force, torque and gravitational acceleration vectors have been changed to three-dimensional vectors. Therefore, all the calculations are performed in three-dimensions. Only in the integration stage, we neglect the force and torques in the z direction.

# Future changes

- Modified dem_2d solver will be used in the resolved CFD-DEM solver.


Luckabarbeau (approved):
Very nice PR this will simplify some manipulation in 2D. Nothing special to say. Just check if it's normal that the value of the test change that much. I guess it is because you changed the test itself?

shahabgol:
> Very nice PR this will simplify some manipulation in 2D. Nothing special to say. Just check if it's normal that the value of the test change that much. I guess it is because you changed the test itself?

@Luckabarbeau The results of the two-dimensional tests change because they didn't have any torques before.

blaisb:
A lot of tests don't pass in debug mode @shahabgol

shahabgol:
> 

There was a bug in the acceleration vector. It was still <1, dim> instead of <1, 3>. It is fixed now.

tonielgeitani (approved):
It looks good on my side.

blaisb (approved):
Very good work. @shahabgol  Before I merge, can you verify that performance is not affected by this change?

shahabgol:
> Very good work. @shahabgol Before I merge, can you verify that performance is not affected by this change?

I've added a benchmark for tracking the simulation times of the dem_3d solver.

blaisb:
@shahabgol  Can you add your processor information in the readme file you added?
This way we could have multiple type of processors and have the timing for different systems as a function of time for our machines.
After that, ready to merge

shahabgol:
> 

Added to both readme and simulation time file.

blaisb (approved):

PR #396, opened 2022-02-17
Vof peeling

jeannejoachim:
# Description of the problem

- Peeling/wetting was not implemented in VOF solver

# Description of the solution

- create BC section for VOF
- implement peeling/wetting basic methods

# How Has This Been Tested?

No test for now, will be added when the solver is more mature. I tested it locally on the "glissomètre" (dough on slopped plane) that we used during our experimental campaign.

# Documentation

For now I did not put it in the documentation, as it is really immature so I do not really want anyone to use it. Will be in the next PR with improved peeling/wetting!

# Future changes

- `none` BC could be generalized
- improving peeling/wetting : this is a basic implementation of the methods needed
- give a physical meaning to the pressure threshold (link it to the friction coefficient?)
- I really want to clarify the VOF solver: VOF or CFD parameters have different suffix throughout the code and I want to make it uniform. This will be my next PR!

# Comments

I thought I had to rebase but git says it is able to automatically merge. Do not hesitate to tell me if you think a rebase is still necessary!


include/solvers/vof.h
```diff
@@ -422,6 +417,37 @@ class VolumeOfFluid
...
+  template <typename VectorType>
+  void
+  apply_peeling_wetting(const unsigned int i_bc,
+                        const VectorType & current_solution_cfd);
+
+  /**
+   * @brief Change cell phase, small method called to avoid code repetition and reduce sloppy
+   * error likelihood in apply_peeling_wetting.
+   *
+   * @param type a string stating the needed change ("w" or "p")
```
blaisb, line 436: Why use a string to indicate a status? It would be more appropriate to use an enum class. String compares are quite expensive (and prone to mistakes).

source/solvers/vof.cc
```diff
@@ -1082,5 +1116,219 @@ VolumeOfFluid<dim>::assemble_mass_matrix_diagonal(
...
+  for (const auto &cell_vof : dof_handler.active_cell_iterators())
+    {
+      if (cell_vof->is_locally_owned() && cell_vof->at_boundary())
+        {
+          for (const auto face : cell_vof->face_indices())
+            {
+              if (cell_vof->face(face)->at_boundary() &&
+                  cell_vof->face(face)->boundary_id() == boundary_id)
+                {
+                  // Equivalent dof_handler for CFD
```
blaisb, line 1178:
```suggestion
                  // Get fluid dynamics active cell iterator
```

source/solvers/vof.cc
```diff
@@ -1082,5 +1116,219 @@ VolumeOfFluid<dim>::assemble_mass_matrix_diagonal(
...
+                              // increment cell count
+                              nb_q_wet++;
+
+                              // if enough quadrature points meet the condition
+                              if (nb_q_wet > n_q_points / 2)
+                                {
+                                  change_cell_phase("w",
+                                                    1,
+                                                    solution_pw,
+                                                    dof_indices_vof);
```
blaisb, line 1223: Here using a string is a bad idea. Let's move to an enum class sytem.

source/solvers/vof.cc
```diff
@@ -1082,5 +1116,219 @@ VolumeOfFluid<dim>::assemble_mass_matrix_diagonal(
...
+                  // Local index (see deal.II step 4)
+                  cell_vof->get_dof_indices(dof_indices_vof);
+
+                  for (unsigned int q = 0; q < n_q_points; q++)
+                    {
+                      if (pressure_values[q] > wetting_threshold)
+                        {
+                          // wetting of lowest density fluid
+                          if ((density_fluid_1 > density_fluid_0) &&
+                              (phase_values[q] < 0.5))
```
blaisb, line 1212: It would be less confusing if the less dense and the dense fluids where pre-established and these ifs in the quadrature point logic were removed.

source/solvers/vof.cc
```diff
@@ -1082,5 +1116,219 @@ VolumeOfFluid<dim>::assemble_mass_matrix_diagonal(
...
+                                  change_cell_phase("w",
+                                                    1,
+                                                    solution_pw,
+                                                    dof_indices_vof);
+                                }
+                            }
+                          else if ((density_fluid_0 > density_fluid_1) &&
+                                   (phase_values[q] > 0.5))
+                            {
+                              // increment cell count
```
blaisb, line 1229: This is a quadrature point count no? Why cell count?

include/core/boundary_conditions.h
```diff
@@ -678,6 +681,162 @@ namespace BoundaryConditions
     prm.leave_subsection();
   }
 
+  /**
+   * @brief This class manages the boundary conditions for VOF solver
+   * It introduces the boundary functions and declares the boundary conditions
+   * coherently. (TODO update)
+   * The members ... contain double used for bc calculation :
+   *  - if bc type is "..." (... condition), "..." is the
+   * double passed to the deal.ii ConstantFunction
```
blaisb, line 690: I don't understand this part?

blaisb (commented):
Some slight work remains to be done. I think it would be appropriate to add a test (can be a lump of liquid falling or something) so that the feature is always reproducable.

include/core/boundary_conditions.h
```diff
@@ -678,6 +681,162 @@ namespace BoundaryConditions
...
+   * @brief Declare the boundary conditions default parameters
+   * Calls declareDefaultEntry method for each boundary (max 14 boundaries)
+   *
+   * @param prm A parameter handler which is currently used to parse the simulation information
+   */
+  template <int dim>
+  void
+  VOFBoundaryConditions<dim>::declare_parameters(ParameterHandler &prm)
+  {
+    this->max_size = 14;
```
shahabgol, line 755: maybe rename this variable to max_bundary_size?

source/solvers/vof.cc
```diff
@@ -241,6 +241,8 @@ void
 VolumeOfFluid<dim>::attach_solution_to_output(DataOut<dim> &data_out)
 {
   data_out.add_data_vector(dof_handler, present_solution, "phase");
+  // Peeling/wetting output
+  data_out.add_data_vector(dof_handler, marker_pw, "marker_pw");
```
shahabgol, line 245: don't we need an if to output the peeling/wetting data?

source/solvers/vof.cc
```diff
@@ -455,6 +457,35 @@ template <int dim>
 void
 VolumeOfFluid<dim>::modify_solution()
 {
+  // Peeling/wetting
```
shahabgol, line 460: the same here. I think a parameter in the parameter handler and an if is required to enable the calculations.

source/solvers/vof.cc
```diff
@@ -1082,5 +1116,219 @@ VolumeOfFluid<dim>::assemble_mass_matrix_diagonal(
...
+
+  const FEValuesExtractors::Scalar pressure(dim);
+  std::vector<double>              pressure_values(n_q_points);
+  std::vector<double>              phase_values(n_q_points);
+
+  unsigned int boundary_id =
+    this->simulation_parameters.boundary_conditions.id[i_bc];
+
+  // Useful definitions for readability
+  const unsigned int density_fluid_0 =
```
shahabgol, line 1159: why are the types of these values unsigned int?

include/solvers/vof.h
jeannejoachim: you are very right, at the time I implemented it I had no better idea. I will look into that!

source/solvers/vof.cc
jeannejoachim: yes you are right, I will correct that

include/core/boundary_conditions.h
jeannejoachim: thank you for spotting that I forgot to update that :)

source/solvers/vof.cc
jeannejoachim: I am not sure that I understand you: do you suggest that "fluid 1" would always be set to be the densest one? In this case we would have to ask the user to follow this when writting the prm file, it did not seem appropriate

shahabgol (commented):
Clean PR. I have a few comments. Besides, I think adding a simple test would be a good idea :)

source/solvers/vof.cc
blaisb: no I meant like... instead of having all of these if in the loop, you could define a variable called "critical phase value" or something and pre-define it.

jeannejoachim: I did that at first, but it occurred to me that it did not change much the readability, because I would trade this if condition `if ((density_fluid_1 > density_fluid_0) && (phase_values[q] < 0.5))` for `if ((wetting_phase_value == 1) && (phase_values[q] < 0.5))`, which is not a lot clearer. Besides, I would have an if loop prior to that to define the `wetting_phase_value` and `peeling_phase_value`, so it seemed not a good trade. Maybe we could talk briefly about it to clarify that?

include/core/boundary_conditions.h
jeannejoachim: I agree that it will be clearer, though it is defined in the same fashion than the `NSBoundaryConditions`, `HTBoundaryConditions` and `TracerBoundaryConditions`. Do you suggest that I change this parameter name everywhere?

source/solvers/vof.cc
jeannejoachim: absolutely! Well spotted, thank you

source/solvers/vof.cc
jeannejoachim: it should absolutely not! Thank you for spotting this sloppy copy/paste ^^"

include/core/boundary_conditions.h
shahabgol: It's your decision, maybe we can do it in another PR. :)

shahabgol:
@jeannejoachim  Is the PR ready for the final review?

jeannejoachim:
> @jeannejoachim Is the PR ready for the final review?

No not at all! I rebased on master and now I should deal with the new physical property manager, I still have some work to do. Thank you for asking, I will notify you when this is the case :)

jeannejoachim:
@shahabgol and @blaisb it is ready to be reviewed again! Thank you for your feedback, it improved this PR a lot.

I also merged (twice) with master to resolve conflicts and use the new `property_manager`. Please tell me what you think about it!

The only thing missing from your suggestions is the test. I have it (water droplet falling on a plane), but it would require a more mature peeling/wetting mechanism for the results to be convincing. I could still add the test and its output (number of cells wet/peeled) but it would still need to be rewritten in future versions. I would rather merge what is already here and then work on a version 2 than continue on the same PR, but do not hesitate to tell me if you think otherwise!

blaisb (approved):
Good job! Once CI passes, it's good to go for me :)

PR #398, opened 2022-02-18
Fix timeout syntax for the new deal.II version

blaisb:
# Description of the problem

- The new deal.II 10 CMAKE change broke the test timeout sequence in Debug

# Description of the solution

- Fixes this issue by  changing the cmake syntax for the test timeout

This will be merged when https://github.com/dealii/dealii/pull/13413#pullrequestreview-886762709 is merged.


applications_tests/dem_3d/CMakeLists.txt
```diff
@@ -14,11 +14,19 @@ string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_LOWER_BUILD_TYPE)
...
        	set_tests_properties(dem_3d/packing_in_ball.mpirun=2.debug PROPERTIES TIMEOUT 3000) 
-       	set_tests_properties(dem_3d/packing_in_cylinder.debug PROPERTIES TIMEOUT 3000) 
-       	set_tests_properties(dem_3d/packing_in_ball_dynamic_contact.debug PROPERTIES TIMEOUT 3000)
-       	set_tests_properties(dem_3d/sliding_in_box.debug PROPERTIES TIMEOUT 3000)
+
+    else()
+	    set_tests_properties(dem_3d/packing_in_box.mpirun=1.threads=0.debug PROPERTIES TIMEOUT 3000)
+      	set_tests_properties(dem_3d/packing_in_box.mpirun=2.threads=0.debug PROPERTIES TIMEOUT 3000) 
+       	set_tests_properties(dem_3d/packing_in_ball.mpirun=1.threads=0.debug PROPERTIES TIMEOUT 3000) 
+       	set_tests_properties(dem_3d/packing_in_ball.mpirun=2.threads=0.debug PROPERTIES TIMEOUT 3000)
```
tamiko, line 30: @blaisb Yes, I have changed the target name.
It might not have been the best decision to unconditionally add the `.threads=0` (it simply saves another set of if clauses to decide whether `.mpirun=N` or `.threads=N` has to be present). So i this change is inconvenient we can certainly try to address it in upstream deal.II.

blaisb: @tamiko  I am cool with both solutions (either way). I will adapt myself to the library. Just tell me which way you chose and I'll live with it :).

PR #400, opened 2022-02-18
Clean VOF solver

jeannejoachim:
Clean and homogeneize to make it clear what comes from the vof vs. cfd

# Description of the problem
In vof, it was not obvious to tell what came from the vof solving vs. cfd solving
- parameters coming from fluid had different suffix: none, `_fluid`, `_navier_stokes`
- parameters coming from vof had different suffix/prefix: none, `_fs` (old free surface)

# Description of the solution
Clean and homogenize the code
- when in the VOF files, added `this->` for every vof attributes for extra clarity
- change parameter names where it felt clearer
- minor homogenization in `heat_transfer` and `tracer` physics
- removed old/dead parameters

# How Has This Been Tested?

No tests, and if all tests run in debug we should be good to go. Should I change the names for the vof tests (called `free_surface` for now)?

# Documentation

None, variable names were changed in the code, not the prm.

# Future changes

None except another solver needs clarification as well.

# Comments

I will update the peeling-wetting branch so that the parameters are already in the same fashion.

### DIFF
--- a/include/solvers/heat_transfer_scratch_data.h
+++ b/include/solvers/heat_transfer_scratch_data.h
@@ -90,7 +90,7 @@ public:
                           const FiniteElement<dim> &      fe_ht,
                           const Quadrature<dim> &         quadrature,
                           const Mapping<dim> &            mapping,
-                          const FiniteElement<dim> &      fe_navier_stokes,
+                          const FiniteElement<dim> &      fe_fd,
                           const Quadrature<dim - 1> &     face_quadrature)
     : properties_manager(properties_manager)
     , fe_values_T(mapping,
@@ -98,10 +98,7 @@ public:
                   quadrature,
                   update_values | update_quadrature_points | update_JxW_values |
                     update_gradients | update_hessians)
-    , fe_values_navier_stokes(mapping,
-                              fe_navier_stokes,
-                              quadrature,
-                              update_values | update_gradients)
+    , fe_values_fd(mapping, fe_fd, quadrature, update_values | update_gradients)
     , fe_face_values_ht(mapping,
                         fe_ht,
                         face_quadrature,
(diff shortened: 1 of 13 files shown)


include/solvers/heat_transfer_scratch_data.h
```diff
@@ -427,9 +424,9 @@ class HeatTransferScratchData
    */
   FEValuesExtractors::Vector velocities;
   // This FEValues must mandatorily be instantiated for the velocity
-  FEValues<dim>               fe_values_navier_stokes;
-  std::vector<Tensor<1, dim>> velocity_values;
-  std::vector<Tensor<2, dim>> velocity_gradient_values;
+  FEValues<dim>               fe_values_fd;
+  std::vector<Tensor<1, dim>> velocity_values_fd;
```
CaroleAnneDaunais, line 428: I think that since the velocity_values vector of tensor is independant of the number of fluids, maybe we should keep it as velocity_values

include/solvers/heat_transfer_scratch_data.h
```diff
@@ -427,9 +424,9 @@ class HeatTransferScratchData
...
+  std::vector<Tensor<2, dim>> velocity_gradient_values_fd;
```
CaroleAnneDaunais, line 429: Same here

include/solvers/heat_transfer_scratch_data.h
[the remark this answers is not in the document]
jeannejoachim: yes @CaroleAnneDaunais said the same, I was not sure myself about that but thought that maybe it would be clearer. As you both say it is not, I will revert this :)

jeannejoachim:
> I think it's a good idea to use vof instead of fs, it's much clearer! Can you enlighten me on why we use two different fe_values when using vof? That has nothing to do with this PR, I'm only curious. I made some comments on changes I think should be dismissed but overall as long as tests pass this makes sense to me.

This is an excellent question! Let me clarify how the VOF solver works.

The location of the interface between the two fluids are initially given in the prm file. Then, at each time step, it moves accordingly with the velocity inside both fluids domain. On the VOF solver, we use an advection equation to move the interface. The fluid velocity needed is contained in `fe_values_fd`, see for instance in the `vof_scratch_data`: `fe_values_fd[velocities_fd].get_function_values(current_solution, velocity_values);`

That is why the solver needs two fe_values: the one coming from the fluid dynamics (FD), to get the fluid velocity, and the one that will have the computation of the phase values (0 in one fluid, 1 in the other, 0.5 considered the interface location) from the VOF solver. This occurs also for the Heat Transfer and the Tracer physics, so I wanted to make extra clear what comes from the FD and what is computed by the current auxiliary physic.

Do not hesitate if you have any question about that :) If you feel the documentation for the VOF is not clear enough to grab that (in the code or the website), do not hesitate to tell me.

CaroleAnneDaunais (approved):

blaisb (approved):

PR #401, opened 2022-02-20
Add full jacobian for phase change

blaisb:
# Description of the problem

- The jacobian matrix for the specific heat was not completely implemented. This reduced the speed of convergence

# Description of the solution

- Implement the full jacobian matrix

# How Has This Been Tested?

- Results of currently tests are not changed. No need to add a new test.

# Comments

- This was much easier than anticipated...

### DIFF
--- a/source/solvers/heat_transfer_assemblers.cc
+++ b/source/solvers/heat_transfer_assemblers.cc
@@ -204,6 +204,8 @@ HeatTransferAssemblerBDF<dim>::assemble_matrix(
 
   const std::vector<double> &density       = scratch_data.density;
   const std::vector<double> &specific_heat = scratch_data.specific_heat;
+  const std::vector<double> &grad_specific_heat_temperature =
+    scratch_data.grad_specific_heat_temperature;
 
   // Copy data elements
   auto &strong_residual = copy_data.strong_residual;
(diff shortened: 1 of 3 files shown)


source/solvers/heat_transfer_assemblers.cc
```diff
@@ -204,6 +204,8 @@ HeatTransferAssemblerBDF<dim>::assemble_matrix(
 
   const std::vector<double> &density       = scratch_data.density;
   const std::vector<double> &specific_heat = scratch_data.specific_heat;
+  const std::vector<double> &grad_specific_heat_temperature =
```
shahabgol, line 207: I don't understand the physical meaning of this parameter.

shahabgol (commented):
Clean PR, I just don't understand the meaning of grad_specific_heat_temperature. Maybe you can change the variable name or add a comment to make it clear. :)

source/solvers/heat_transfer_assemblers.cc
blaisb: It is the gradient of the specific heat with respect to the temperature ( so dCp / dT). I'll add a comment to make this clearer.

shahabgol (approved):
Awesome :)

blaisb:
> Awesome :)

I added a bit more information as a comment to make the thing more clear :). Once CI passes i'll merge.

PR #408, opened 2022-02-27
Small assembly optimization for Navier-Stokes

blaisb:
# Description of the problem

- During the assembly of the NS equations, some variables were copied when the reference could have been used

# Description of the solution

- Use const reference instead of creating new tensors

# How Has This Been Tested?

- All tests results were unchanged

# Comments

- This has a 3-4% influence on speed, not that much but still worth it.


Luckabarbeau (approved):
This is a nice update that would reduce the instantiation of variables. Everything seems good to me

PR #412, opened 2022-03-01
change how the pf force and drag are applied

tonielgeitani:
# Description of the problem

- Model A and B of the VANS gave different results in liquid fluidized beds
- The forces applied on the particles and fluid for both models were modified in order to make the models mathematically equivalent. 
 -Problem with restarting simulation from checkpoints. DEM was exploding

# Description of the solution

- The momentum exchange coefficient of Model B was take to be that of Model A / void fraction on both phases.
- The pf interactions for model A on the fluid side where taken to be only drag while the undisturbed flow forces where applied on the fluid side for model B only.
- The koch and hill drag model was added. All drag models were calculated for model B and then multiplied by the void fraction to obtain those for model A.
- - Fix the problem for restarting simulations by initially setting the checkpoing_step to true if it is a restart time step.
- Fixed a typo with the pressure force where there was a - sign that was misplaced.

# How Has This Been Tested?

- All tests were rerun where the values of the pressure drop, mass conservation, and particle location were slightly modified.
Only the 2D VANS tests were unchanged while all other tests were changed based on this modification.

# Documentation

- No applicable documentation was changed

# Future changes

- 

# Comments

-


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -686,10 +686,10 @@ class NavierStokesScratchData
     fe_values_local_particles.reinit(cell);
 
     fe_values_local_particles[velocities].get_function_laplacians(
-      current_solution, fluid_velocity_laplacian_at_particle_location);
+      previous_solution, fluid_velocity_laplacian_at_particle_location);
 
     fe_values_local_particles[pressure].get_function_gradients(
-      current_solution, fluid_pressure_gradients_at_particle_location);
+      previous_solution, fluid_pressure_gradients_at_particle_location);
```
blaisb, line 692: Why use previous solutions instead of current one here?
Is it because it does not converge if you use current one?

blaisb (commented):
Very good, I just have one question related to the forces. See my comment

include/solvers/navier_stokes_scratch_data.h
tonielgeitani: Yes, when using model A, this the shear force is not applied and thus there is no problem. However, for model B, it stagnates at a residue of 10^-5 if I use the current solution. Using the previous solution allows it to converge quickly.

blaisb: Ok I understand, we have to understand that this will generate a small difference in the results between A and B :). Good. I'll proceed and merge then!

PR #414, opened 2022-03-02
Fix ht vof

jeannejoachim:
# Description of the problem

Launching a simulation with both VOF and heat transfer resulted in seg-fault error.

# Description of the solution

- Fix the heat transfer solver issue: VOF scratch_data was never gathered
- Add a test with both VOF and heat transfer, to assess that similar problem will be catched in the future

# How Has This Been Tested?

Test added : gls_free-surface_heat-transfer_hydrostat

The dumbest test you can think of: free surface that is supposed to stay static, and heat transfer mechanism called for no heat transfer mechanism happening (homogeneous temperature at the start). Still it guarantees that both physics can be called in the same simulation!

# Documentation

Not directly linked to this PR, but I saw that there are things missing in the `initial_condition` part of the documentation (namely initial condition on the temperature). I will open another PR for that.

# Comments

While I was at it, I cleaned what I saw in the code.

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/gls_free-surface_heat-transfer_hydrostat.output
+++ b/applications_tests/gls_navier_stokes_2d/gls_free-surface_heat-transfer_hydrostat.output
@@ -1,0 +1,39 @@
+Running on 1 MPI rank(s)...
+   Number of active cells:       1024
+   Number of degrees of freedom: 3267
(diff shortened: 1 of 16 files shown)


blaisb (approved):
Nice catch. Good job, this is much cleaner this way.
No comments on my side.

PR #428, opened 2022-03-10
Non newtonian jacobian assembler

CaroleAnneDaunais:
# Description of the problem

Previously, all non-Newtonian simulations converged, but the jacobian matrix was ill-posed since it didn't take into account the jacobian of the viscosity (since it changes with velocity). 

# Description of the solution

A small addition was made to the jacobian matrix.

# How Has This Been Tested?

There were no added tests, but these two tests were slightly changed

- applications_tests/gls_navier_stokes_2d/mms2d_carreau_1st-order_gls.output
- applications_tests/gls_navier_stokes_2d/mms2d_carreau_2nd-order_gls.output

# Documentation

No documentation changes

### DIFF
--- a/source/solvers/navier_stokes_assemblers.cc
+++ b/source/solvers/navier_stokes_assemblers.cc
@@ -416,10 +416,10 @@ GLSNavierStokesAssemblerNonNewtonianCore<dim>::assemble_matrix(
               double local_matrix_ij =
                 viscosity *
                   scalar_product(grad_phi_u_j_non_newtonian, grad_phi_u_i) +
-                scratch_data.grad_viscosity_shear_rate[q] /
+                0.5 * scratch_data.grad_viscosity_shear_rate[q] /
                   shear_rate_magnitude *
                   scalar_product(grad_phi_u_j_non_newtonian, shear_rate) *
-                  scalar_product(velocity_gradient, grad_phi_u_i) +
+                  scalar_product(shear_rate, grad_phi_u_i) +
                 velocity_gradient_x_phi_u_j[j] * 0.5 * phi_u_i +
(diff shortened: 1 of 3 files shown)


source/solvers/navier_stokes_assemblers.cc
```diff
@@ -384,7 +384,8 @@ GLSNavierStokesAssemblerNonNewtonianCore<dim>::assemble_matrix(
           velocity_gradient_x_phi_u_j[j] = velocity_gradient * phi_u_j;
         }
 
-
+      shear_rate_magnitude =
+        shear_rate_magnitude > 1e-3 ? shear_rate_magnitude : 1e-3;
```
Luckabarbeau, line 388: I am not sure that hard coding this 1e-3 is the best way. Maybe define a parameter or use an already existing parameter.

Luckabarbeau (commented):
Small comment. Very nice update this will speed up a lot the computation

source/solvers/navier_stokes_assemblers.cc
blaisb: That's a good point, I don't know what would be the best solution for this though. If it's another parameter in the physical properties it means we will have to percolate it everywhere and it will be a bit of a pain. The other solution could be that all rheological model could provide a "minimum shear" parameter and every rheological model could provide it. This should work well.

blaisb: What do you think @CaroleAnneDaunais ?

blaisb: The more I think about it, the more it is best we leave this automatic. Maybe we can merge it as is for now and look into this later.

blaisb (approved):
PR is good for me :)!

PR #429, opened 2022-03-11
correct precision overwrite

Luckabarbeau:
# Description of the problem

- A line of code was overwriting the output precision

# Description of the solution

- remove the line

# How Has This Been Tested?

- modify the tests that were affected.


blaisb (approved):
Will merge once CI passes.

PR #431, opened 2022-03-15
Fix compilation warnings

blaisb:
# Description of the problem

- Previous patch had introduced some compilation warnings

# Description of the solution

- Fix them. They were mostly unused variables

# How Has This Been Tested?

- If all tests pass this is ok


Luckabarbeau (approved):
Everything looks good. This also cleans the code a bit by removing some useless variables.

PR #434, opened 2022-03-19
Add create output directory function

shahabgol:
# Description of the problem

- This PR creates the output folder if it does not exist in the simulation directory.

### DIFF
--- a/applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
+++ b/applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
@@ -10,7 +10,7 @@ subsection simulation control
   set time step               = 0.1    # Time step
     set number mesh adapt     = 0       # If steady, nb mesh adaptation
   set time end                = 1.00      # End time of simulation
-  set output path             =  # Path to where the output results are made
+  set output path             = ./ # Path to where the output results are made
   set output name             = tgv # Prefix for VTU outputs
   set output frequency        = 100       # Frequency of simulation output
   set subdivision             = 1       # Mesh subdivision when outputend
--- a/applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
+++ b/applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
@@ -8,11 +8,10 @@
 subsection simulation control
   set method                  = bdf1
   set time step               = 0.1    # Time step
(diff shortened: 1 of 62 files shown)


blaisb:
Seems like the CI is complaining friend :)!

applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
```diff
@@ -10,7 +10,7 @@ subsection simulation control
   set time step               = 0.1    # Time step
     set number mesh adapt     = 0       # If steady, nb mesh adaptation
   set time end                = 1.00      # End time of simulation
-  set output path             =  # Path to where the output results are made
+  set output path             = ./ # Path to where the output results are made
```
blaisb, line 13: It would be better to put the default value of output path to ./ and just remove this line from the prm file!
At the same time, i would take the opportunity to remove the output of these tests by putting output frequency to 0.

source/solvers/navier_stokes_base.cc
```diff
@@ -117,6 +119,18 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
...
+  std::string output_dir_name =
+    simulation_parameters.simulation_control.output_folder;
+  struct stat buffer;
+
+  // If output directory does not exist, create it
+  if (stat(output_dir_name.c_str(), &buffer) != 0 &&
+      Utilities::MPI::this_mpi_process(this->mpi_communicator) == 0)
+    {
+      create_output_folder(output_dir_name);
+    }
```
blaisb, line 132: I would put the this->mpi_process ==0 in a first if so that only proc 0 checks if the folder exists. at 1000 cores otherwise you will have 1000 processes checking the same folder.

blaisb (commented):
2 small changes and this will be good if the CI passes :)!

applications_tests/gd_navier_stokes_2d/taylor-green-vortex_gd_bdf1.prm
shahabgol: the default value of the output path is already ./ so I removed all the set output path lines from the parameter handler files. The second comment is also applied.

source/solvers/navier_stokes_base.cc
shahabgol: done.

blaisb (approved):
Awesome. Actually this should make the CI faster now that we don,t write any files...

PR #435, opened 2022-03-19
Cfd dem b work

tonielgeitani:
# Description of the problem

-The earlier formulation of gamma for the grad div stabilization was not sufficient in certain cases to converge the cfd-dem solver. The solution was either stagnating or crashing.

# Description of the solution

-Change the grad-div formulation as well as allowing for either expect or implicit SUPG and grad div stabilizations based on user input.

# How Has This Been Tested?

-For two spouted and fluidized bed test cases, the solution as wells convergence were evaluated between the two gamma formulation.

# Documentation

- the CFD-DEM-dem.rst file was modified to add the two newly added parameters.


doc/source/parameters/unresolved_cfd-dem/cfd_dem.rst
```diff
@@ -17,6 +17,8 @@ This subsection includes parameters related to multiphase flow simulations using
       set coupling frequency = 200
       set inlet boundary id = 2
       set outlet boundary id = 3
+      set implicit stabilization = false
```
blaisb, line 20: I think it's better to have the default here at true. And if we want more robustness, we can change it to false :). Let's keep the previous behavior as is and when we voluntarily face challenge, we put this to false :).

blaisb:
@tonielgeitani I just changed the default value so that implicit stabilization=true is the default. Otherwise, very nice PR. Will merge once the CI finishes

PR #436, opened 2022-03-23
Non newtonian ramp n initial condition

CaroleAnneDaunais:
# Description of the problem

It is currently difficult to solve solutions for highly rheological flows, even when using an initial newtonian condition at Re = 1. The dynamic of the flow progresses logarithmically, and therefore switching from a newtonian solution to a highly non-newtonian solution gives a hard time to the linear solver. 

# Description of the solution

The solution consists in ramping on the n value which dictates the level of rheology of the flow. Likewise, a ramp on viscosity was added that can also be used for Newtonian simulations.

# How Has This Been Tested?

A test which uses both the viscosity ramp and the n ramp was added using a power-law model

- applications_tests/gls_sharp_navier_stokes_3d/sphere_power_law_ramp

# Documentation

The cfd/initial_conditions.rst file was updated

# Future changes

A Lethe example on non-newtonian flow around a sphere should be added shortly.


blaisb:
I think you need to fix the timeout time for your test in debug mode :)!

doc/source/parameters/cfd/initial_conditions.rst
```diff
@@ -30,3 +43,22 @@ It is often necessary to set-up complex initial conditions when simulating trans
 
 .. note::
    The ``Function expression`` can be used to establish an even more complex free surface initial geometry. For example, one can create a circle of fluid : ``if ( (x^2+y^2)<=(r)^2 ,1,0)``
+
+
+* The ``subsection ramp`` holds the parameters to operate a ramp on either or both the viscosity and the ``n`` parameter in rheological models (see :doc:`physical_properties` for nor information on this parameter). When ramping on the viscosity value,
```
blaisb, line 48:
```suggestion
* The ``subsection ramp`` holds the parameters to operate a ramp on either or both the viscosity and the ``n`` parameter in rheological models (see :doc:`physical_properties` for more information on this parameter). When ramping on the viscosity value, 
```

blaisb (approved):
Very good. Just fix CI and address the single comment and this will be good to go :)

blaisb (approved):

PR #442, opened 2022-03-30
Fix VANS equations so that Model A and Model B become equivalent

voferreira:
# Description of the problem

- The VANS models A and B were not equivalent in the unresolved CFD-DEM code;
- It was mainly related to a difference in the Drag model since the Beta of model A was equivalent to the Beta of model B times the voidfraction;
- Additionally, the Beta B should be multiplied by epsilon in all drag models.

# Description of the solution

- Made the two models equivalent by using the same drag model for both;
- All changes were done in the file 'vans_assemblers.cc'.

# How Has This Been Tested?

- Comparing the terminal velocity of a sedimenting single particle with estimations;
- Comparing the pressure drop from the simulations of a fixed bed with the one calculated by Ergun's equation;
- Comparing the pressure drop from the simulations of a liquid fluidized bed with the one calculated from the mass of particles;
- Comparing the porosity (expansion) of the liquid fluidized bed with experimental results.

# Documentation

- 

# Future changes

- Include a section about the VANS equations in the theory guide.

# Comments

-


blaisb (approved):

tonielgeitani (approved):
Good job

PR #446, opened 2022-04-08
Laser heat source

shahabgol:
# Description of the problem

- This PR adds the laser heat source to the heat transfer solver.

# Description of the solution

- Exponentially decaying model is used to simulate the laser heat source: "Liu, S., Zhu, H., Peng, G., Yin, J. and Zeng, X., 2018. Microstructure prediction of selective laser melting AlSi10Mg using finite element analysis. Materials & Design, 142, pp.319-328."

# How Has This Been Tested?

- An application test is added (heat_transfer_laser), but it needs a post-processing method to report the maximum temperature, and radiation boundary condition, which will be added in separate PRs.

# Future changes

- Maximum temperature post-processing
- Radiation boundary condition

# Comments

- Documentation will be updated after finishing all these three PRs.

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
+++ b/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
@@ -8,31 +8,31 @@ Running on 1 MPI rank(s)...
 Transient iteration : 1        Time : 0.01     Time step : 0.01     CFL : 0       
 *****************************************************************
 L2 error velocity : 0
-L2 error temperature : 0.164285
+L2 error temperature : 150.231
 
 *****************************************************************
 Transient iteration : 2        Time : 0.02     Time step : 0.01     CFL : 0       
 *****************************************************************
 L2 error velocity : 0
-L2 error temperature : 0.275693
+L2 error temperature : 278.378
 
 *****************************************************************
 Transient iteration : 3        Time : 0.03     Time step : 0.01     CFL : 0       
 *****************************************************************
 L2 error velocity : 0
-L2 error temperature : 0.344329
+L2 error temperature : 378.601
 
 *****************************************************************
 Transient iteration : 4        Time : 0.04     Time step : 0.01     CFL : 0       
 *****************************************************************
 L2 error velocity : 0
-L2 error temperature : 0.200529
+L2 error temperature : 287.622
 
 *****************************************************************
 Transient iteration : 5        Time : 0.05     Time step : 0.01     CFL : 0       
 *****************************************************************
 L2 error velocity : 0
-L2 error temperature : 0.148557
+L2 error temperature : 245.927
  time  error_velocity 
 0.0100 0.000000e+00   
 0.0200 0.000000e+00   
@@ -40,8 +40,8 @@ L2 error temperature : 0.148557
 0.0400 0.000000e+00   
 0.0500 0.000000e+00   
 cells error_temperature 
-1024  1.642847e-01      
-1024  2.756934e-01      
-1024  3.443293e-01      
-1024  2.005288e-01      
-1024  1.485571e-01      
+1024  1.502310e+02      
+1024  2.783781e+02      
+1024  3.786006e+02      
+1024  2.876222e+02      
+1024  2.459270e+02      
--- a/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
+++ b/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
@@ -1,0 +1,47 @@
+Running on 1 MPI rank(s)...
+   Number of active cells:       1024
+   Number of degrees of freedom: 3267
+   Volume of triangulation:      0.01
+   Number of thermal degrees of freedom: 1089
+
+*****************************************************************
+Transient iteration : 1        Time : 0.01     Time step : 0.01     CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error temperature : 0.164285
+
+*****************************************************************
+Transient iteration : 2        Time : 0.02     Time step : 0.01     CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error temperature : 0.275693
+
+*****************************************************************
+Transient iteration : 3        Time : 0.03     Time step : 0.01     CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error temperature : 0.344329
+
+*****************************************************************
+Transient iteration : 4        Time : 0.04     Time step : 0.01     CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error temperature : 0.200529
+
+*****************************************************************
+Transient iteration : 5        Time : 0.05     Time step : 0.01     CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error temperature : 0.148557
+ time  error_velocity 
+0.0100 0.000000e+00   
+0.0200 0.000000e+00   
+0.0300 0.000000e+00   
+0.0400 0.000000e+00   
+0.0500 0.000000e+00   
+cells error_temperature 
+1024  1.642847e-01      
+1024  2.756934e-01      
+1024  3.443293e-01      
+1024  2.005288e-01      
+1024  1.485571e-01      
--- a/doc/source/parameters/cfd/laser_heat_source.rst
+++ b/doc/source/parameters/cfd/laser_heat_source.rst
@@ -37,7 +37,7 @@ If a laser heat source is present in a simulation, it can be added in this secti
 * The ``beam orientation`` parameter shows the orientation of the laser beam. For instance, if a laser beam is emitted perpendicular on a plane in x-y coordinates, the orientation of the laser beam will be in the z direction.
 
 .. note:: 
-    In two-dimensional simulations, the laser beam orientation is always in the perpendicular direction to the simulation domain.
+    In two-dimensional simulations, the laser beam orientation cannot be in the z direction.
 
 
 * In the ``path`` subsection, the laser scanning path is defined using a ``Function expression``.
--- a/doc/source/parameters/cfd/laser_heat_source.rst
+++ b/doc/source/parameters/cfd/laser_heat_source.rst
@@ -24,7 +24,7 @@ If a laser heat source is present in a simulation, it can be added in this secti
 
 * Laser ``concentration factor`` parameter indicates the definition of the beam radius. In almost all the articles, it is assumed equal to 2.0.
 
-* The ``power`` parameter sets the power of the laser in W.
+* The ``power`` parameter sets the power of the laser [:math:`ML^2T^{-3}`].
 
 * The ``absorptivity`` parameter is defined as the fraction of the amount of incident radiation that is absorbed by the surface, and it is measured using diffuse reﬂectance spectroscopy (DRS). Generally, a constant value in the range of 0.3-0.8 (for welding processes with titanium) are used in the literature. However, recent studies show that it varies with powder particle size distribution, and the angle of incidence that changes due to the dynamic melt pool surface [1].
 
--- a/doc/source/parameters/cfd/laser_heat_source.rst
+++ b/doc/source/parameters/cfd/laser_heat_source.rst
@@ -1,0 +1,50 @@
+Laser heat source
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If a laser heat source is present in a simulation, it can be added in this section. The default parameters are:
+
+.. code-block:: text
+
+   subsection laser parameters
+	set enable                 = false
+	set concentration factor   = 2.0
+	set power                  = 100.0
+	set absorptivity           = 0.5
+	set penentration depth     = 0.0
+	set beam radius            = 0.0
+	set start time             = 0.0
+	set end time               = 1.0
+	set beam orientation       = z
+	subsection path
+	  set Function expression  =  0.0; 0.0
+	end
+   end
+
+* The ``enable`` parameter is set to true if the problem has a laser heat source term and enables its calculation.
+
+* Laser ``concentration factor`` parameter indicates the definition of the beam radius. In almost all the articles, it is assumed equal to 2.0.
+
+* The ``power`` parameter sets the power of the laser in W.
+
+* The ``absorptivity`` parameter is defined as the fraction of the amount of incident radiation that is absorbed by the surface, and it is measured using diffuse reﬂectance spectroscopy (DRS). Generally, a constant value in the range of 0.3-0.8 (for welding processes with titanium) are used in the literature. However, recent studies show that it varies with powder particle size distribution, and the angle of incidence that changes due to the dynamic melt pool surface [1].
+
+* The ``penentration depth`` parameter determines the penetration depth of the laser in the simulation domain in the direction of emission.
+
+* The ``beam radius`` parameter defines the radius of the laser beam.
+
+* The ``start time`` and ``end time`` parameters define the operation time window of the laser.
+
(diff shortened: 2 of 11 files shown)


source/core/parameters.cc
```diff
@@ -813,6 +813,114 @@ namespace Parameters
...
+      prm.declare_entry("start time",
+                        "0.0",
+                        Patterns::Double(),
+                        "Start time of laser");
+      prm.declare_entry("end time",
+                        "0.0",
+                        Patterns::Double(),
+                        "End time of laser");
+
+      prm.declare_entry("beam direction",
```
blaisb, line 866: I think "orientation" would be a better choice of word than direction here.

source/core/parameters.cc
```diff
@@ -2019,6 +2127,8 @@ namespace Parameters
     prm.leave_subsection();
   }
 
+  template class Laser<2>;
```
blaisb, line 2130: Only things that comes to mind for me here is : "pew pew"

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -523,3 +522,118 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+      if constexpr (dim == 3)
+        {
+          laser_location[2] =
+            laser_path.value(laser_location,
+                             laser_parameters->beam_direction_coordinate);
+          laser_location_in_depth = laser_location[2];
+        }
+
+      // Get laser location on the operation surface
+      Point<2> laser_location_on_surface;
```
blaisb, line 576: why is this always a point<2>. Should it not be a Point<dim-1> ???

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -523,3 +522,118 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+          // Get quadrate point location
+          Point<2> quadrature_point_on_surface;
+          quadrature_point_on_surface[0] = scratch_data.quadrature_points[q][0];
+          quadrature_point_on_surface[1] = scratch_data.quadrature_points[q][1];
+
+          // Get quadrate point depth
+          double laser_quadrate_point_distance_in_depth = 0.0;
+          if constexpr (dim == 3)
+            {
+              const double quadrate_point_depth =
```
blaisb, line 599:
```suggestion
              const double quadrature_point_depth =
```

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -523,3 +522,118 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+      auto &strong_residual = copy_data.strong_residual;
+      auto &local_rhs       = copy_data.local_rhs;
+
+      // assembling right hand side
+      for (unsigned int q = 0; q < n_q_points; ++q)
+        {
+          // Get quadrate point location
+          Point<2> quadrature_point_on_surface;
+          quadrature_point_on_surface[0] = scratch_data.quadrature_points[q][0];
+          quadrature_point_on_surface[1] = scratch_data.quadrature_points[q][1];
```
blaisb, line 593: I am not sure how this work in 2D? Should it not be that  the second variable is the depth in 2D? Also, shouldnt the depth depend on the direction / orientation of the laser? This is a bit unclear to me right now.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -523,3 +522,118 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+
+          // Get quadrate point depth
+          double laser_quadrate_point_distance_in_depth = 0.0;
+          if constexpr (dim == 3)
+            {
+              const double quadrate_point_depth =
+                scratch_data.quadrature_points[q][2];
```
blaisb, line 600: Shouldnt the depth direction depend on the laser direction?

blaisb (changes requested):
Good work, but i think there are some dimensionality confusion. Like if you think the model should not make sense in 2D or it should be different, you can code two different assemblers (one for 2D and one for 3D) if it makes things easier. But right now I am under the impression that it is not general? 

In addition, can you add the model formulation in the .rst for the documentation so that the parameters are explained and the model equations are given? :)

include/core/parameters.h
```diff
@@ -456,6 +455,44 @@ namespace Parameters
...
+    enum class BeamDirection
+    {
+      x,
+      y,
+      z,
+    } beam_direction;
+
+    unsigned int beam_direction_coordinate;
+    unsigned int perpendicular_plane_coordinate_one;
+    unsigned int perpendicular_plane_coordinate_two;
```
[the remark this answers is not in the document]
shahabgol: added.

include/core/parameters.h
```diff
@@ -456,6 +455,44 @@ namespace Parameters
...
+  class Laser
+  {
+  public:
+    // Laser parameters
+    bool   activate_laser;
+    double concentration_factor;
+    double laser_power;
+    double laser_absorptivity;
+    double penentration_depth;
+    double beam_radius;
```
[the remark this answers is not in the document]
shahabgol: added.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -523,3 +522,118 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+  const double beam_radius          = laser_parameters->beam_radius;
+  const double current_time = this->simulation_control->get_current_time();
+
+  if (current_time >= laser_start_time && current_time <= laser_end_time)
+    {
+      // Get laser path
+      Function<dim> &laser_path = *(laser_parameters->laser_path);
+      laser_path.set_time(current_time);
+
+      // Get laser lcoation
```
CaroleAnneDaunais, line 557:
```suggestion
      // Get laser location
```

source/core/parameters.cc
shahabgol: changed to orientation

include/core/parameters.h
CaroleAnneDaunais, line 483: I'm not sure I understood what are perpendicular place coordinates. Are they two points in the plane? Also, where does this place originate?

source/solvers/heat_transfer_assemblers.cc
shahabgol: True, it was my mistake. It is corrected now.

source/solvers/heat_transfer_assemblers.cc
shahabgol: corrected.

source/core/parameters.cc
```diff
@@ -813,6 +813,114 @@ namespace Parameters
...
+        }
+      else if (op == "z")
+        {
+          beam_direction                     = BeamDirection::z;
+          perpendicular_plane_coordinate_one = 0;
+          perpendicular_plane_coordinate_two = 1;
+          beam_direction_coordinate          = 2;
+        }
+    }
+    prm.leave_subsection();
```
CaroleAnneDaunais, line 921: I think I now understand what are the coordinates here. Maybe they should be called perpendicular_plane_direction_one and _two?

include/core/parameters.h
shahabgol: These auxiliary variables are used to store the coordinates of a perpendicular plane to the direction of emission. For instance, if the laser direction of emission is in the z coordinates (2), then perpendicular_plane_coordinate_one will be equal to 0 [x direction] and perpendicular_plane_coordinate_two will be equal to 1 [y direction]. I have added more descriptions to make these variables more clear.

source/core/parameters.cc
shahabgol: As Bruno suggested to use coordinate instead of direction for the laser beam emission, I think it would be better to keep these names.

shahabgol:
> Good work, but i think there are some dimensionality confusion. Like if you think the model should not make sense in 2D or it should be different, you can code two different assemblers (one for 2D and one for 3D) if it makes things easier. But right now I am under the impression that it is not general?
> 
> In addition, can you add the model formulation in the .rst for the documentation so that the parameters are explained and the model equations are given? :)

All the comments are addressed/answered. The documentation is also updated.

source/solvers/heat_transfer_assemblers.cc
blaisb: > For the calculation of the laser heat source term, we need the radial distance between the laser focal point (in a 2d plane perpendicular to the laser beam direction) and the quadrature points, and axial distance (in the emission direction) between the laser focal point and the quadrature points, separately. Hence, we store the laser position on the surface as a Point<2> to calculate of the radial distance, and laser position in the emission direction (laser_location_in_depth as a double) to calculate the axial distance.

Yeah but this does not work in a 2D simulation. In this case the point will be a 1D point and not a 2D point.
You need to make sure it's  consistent...

shahabgol: It is revised according to our discussion :)

applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
```diff
@@ -0,0 +1,47 @@
...
+0.0100 0.000000e+00   
+0.0200 0.000000e+00   
+0.0300 0.000000e+00   
+0.0400 0.000000e+00   
+0.0500 0.000000e+00   
+cells error_temperature 
+1024  1.502310e+02      
+1024  2.783781e+02      
+1024  3.786006e+02      
+1024  2.876222e+02
```
blaisb, line 46: I guess this is temporary until you have the post-processing for max temperature right?

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -0,0 +1,50 @@
...
+
+* The ``start time`` and ``end time`` parameters define the operation time window of the laser.
+
+* The ``beam orientation`` parameter shows the orientation of the laser beam. For instance, if a laser beam is emitted perpendicular on a plane in x-y coordinates, the orientation of the laser beam will be in the z direction.
+
+.. note:: 
+    In two-dimensional simulations, the laser beam orientation cannot be in the z direction.
+
+
+* In the ``path`` subsection, the laser scanning path is defined using a ``Function expression``.
```
blaisb, line 43: Can you add the equation for the laser heat source?

source/core/parameters.cc
```diff
@@ -813,6 +819,114 @@ namespace Parameters
...
+      prm.declare_entry("concentration factor",
+                        "2.0",
+                        Patterns::Double(),
+                        "Concentration factor");
+      prm.declare_entry("power", "100.0", Patterns::Double(), "Laser power");
+      prm.declare_entry("absorptivity",
+                        "0.5",
+                        Patterns::Double(),
+                        "Laser absorptivity");
+      prm.declare_entry("penentration depth",
```
blaisb, line 838:
```suggestion
      prm.declare_entry("penetration depth",
```

source/core/parameters.cc
```diff
@@ -813,6 +819,114 @@ namespace Parameters
...
+                        "0.0",
+                        Patterns::Double(),
+                        "Penentration depth");
```
blaisb, line 841:
```suggestion
                        "Penetration depth");
```

source/core/parameters.cc
```diff
@@ -813,6 +819,114 @@ namespace Parameters
...
+  void
+  Laser<dim>::parse_parameters(ParameterHandler &prm)
+  {
+    prm.enter_subsection("laser parameters");
+    {
+      activate_laser       = prm.get_bool("enable");
+      concentration_factor = prm.get_double("concentration factor");
+      laser_power          = prm.get_double("power");
+      laser_absorptivity   = prm.get_double("absorptivity");
+      penentration_depth   = prm.get_double("penentration depth");
```
blaisb, line 885:
```suggestion
      penetration_depth   = prm.get_double("penetration depth");
```

blaisb (approved):
Very good. Just 2-3 last comments to address and this is good to go :)
It would be good to add the maximal temperature and average temperature post-processing utilities in a next short PR to modify the example. Otherwise right now the test is not very conclusive.

applications_tests/gls_navier_stokes_2d/heat_transfer_laser.output
shahabgol: Right, I will modify the output of this test after adding the post-processing.

doc/source/parameters/cfd/laser_heat_source.rst
shahabgol: Added.

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -0,0 +1,59 @@
...
+	subsection path
+	  set Function expression  =  0.0; 0.0
+	end
+   end
+
+* The ``enable`` parameter is set to true if the problem has a laser heat source term and enables its calculation.
+
+* Laser ``concentration factor`` parameter indicates the definition of the beam radius. In almost all the articles, it is assumed equal to 2.0.
+
+* The ``power`` parameter sets the power of the laser in W.
```
Luckabarbeau, line 27: I don't think the problem will be in W if the distance unit used is not in meters. maybe give a more general form as L^2MT^-3

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -0,0 +1,59 @@
...
+* The ``penetration depth`` parameter determines the penetration depth of the laser in the simulation domain in the direction of emission.
+
+* The ``beam radius`` parameter defines the radius of the laser beam.
+
+* The ``start time`` and ``end time`` parameters define the operation time window of the laser.
+
+* The ``beam orientation`` parameter shows the orientation of the laser beam. For instance, if a laser beam is emitted perpendicular on a plane in x-y coordinates, the orientation of the laser beam will be in the z direction.
+
+.. note:: 
+    In two-dimensional simulations, the laser beam orientation cannot be in the z direction.
```
Luckabarbeau, line 40: Would it be more interesting to generalize the orientation along a vector in space? That vector could even be a function of time similar to the position. By default, if you are in 2D the laser orientation will be in the z-direction no?
I think this would have a nice application for modeling 5 axis laser cutter or welder.

include/core/parameters.h
```diff
@@ -456,6 +455,79 @@ namespace Parameters
...
+    enum class BeamOrientation
+    {
+      x,
+      y,
+      z,
+    } beam_orientation;
+
+    // beam_orientation_coordinate parameter stores the integer (x = 0, y = 1,
+    // z =2) value of the beam_orientation parameter
+    unsigned int beam_orientation_coordinate;
```
Luckabarbeau, line 509: See comments above for the orientation of the laser

source/core/parameters.cc
```diff
@@ -813,6 +819,114 @@ namespace Parameters
...
+          if constexpr (dim == 3)
+            {
+              beam_orientation                   = BeamOrientation::z;
+              perpendicular_plane_coordinate_one = 0;
+              perpendicular_plane_coordinate_two = 1;
+              beam_orientation_coordinate        = 2;
+            }
+          else if constexpr (dim == 2)
+            Assert(dim == 2, TwoDimensionalLaserError(dim));
+        }
```
Luckabarbeau, line 925: See comments above

Luckabarbeau (commented):
Very nice work! The only comment I have is about how general the laser orientation is. I think generalizing the orientation of the laser has multiple applications.

doc/source/parameters/cfd/laser_heat_source.rst
shahabgol: In our current project (additive manufacturing), the laser orientation does not change during the operation. I will add changing orientation in another PR if it is necessary.

CaroleAnneDaunais (approved):

PR #448, opened 2022-04-10
Refactoring stabilization

blaisb:
# Description of the problem

- In the current version of Lethe, the solver hard-codes the stabilization mechanism. The GD solver uses a grad-div stabilization and LBB stable elements while the GLS solver uses PSPG/SUPG stabilization. This PR introduces a mechanism which allows  more freedom in that respect

# Description of the solution

- Introduce a stabilization parameter and introduce also a real GLS stabilization. The default stabilization is still the old PSPG/SUPG, but the full GLS can also be used. It seems to be pretty good for order 2 and +. Not sure, but still. 

# How Has This Been Tested?

- I added a unit test for this.

# Documentation

- [x] Still needs to be done

# Future changes

- As some projects move forward, we may introduce additional (or different) stabilized solvers. With this PR,  it is easy to do.

# Comments

One issue I have found is that the PSPG/SUPG and GLS are currently not coherent with the block solver, that is they just don't converge. I don't know why. Clearly a rework of the block solver will be in order this summer. I am also not very happy about the code duplication for the GLS and the PSPG/SUPG solvers, but I felt like this was pretty much the only place where they could be seperated.

I also need to do more work on assembly, maybe again this summer.

### DIFF
--- a/source/solvers/navier_stokes_assemblers.cc
+++ b/source/solvers/navier_stokes_assemblers.cc
@@ -10,14 +10,14 @@
 template <int dim>
 void
 PSPGSUPGNavierStokesAssemblerCore<dim>::assemble_matrix(
-  NavierStokesScratchData<dim>         &scratch_data,
+  NavierStokesScratchData<dim> &        scratch_data,
   StabilizedMethodsTensorCopyData<dim> &copy_data)
 {
   // Scheme and physical properties
   const std::vector<double> &viscosity_vector = scratch_data.viscosity;
 
   // Loop and quadrature informations
-  const auto        &JxW_vec    = scratch_data.JxW;
+  const auto &       JxW_vec    = scratch_data.JxW;
   const unsigned int n_q_points = scratch_data.n_q_points;
   const unsigned int n_dofs     = scratch_data.n_dofs;
   const double       h          = scratch_data.cell_size;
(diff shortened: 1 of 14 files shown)


blaisb:
> Very nice PR. I don't see any specific modification to bring. On a more general note, would it be interesting to put the stabilization terms in their own assembler and avoid duplication of code for the Navier-Stokes equations in each assembler?

Yeah, I would love to do that eventually. Just not sure exactly on how to do it. Maybe templace the core assembler with the stabilization type? Right now the idea is that we could have really messed up stabilization that require different shenanigans, hence the duplication. I'm not a fan of it...

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -65,10 +66,8 @@ GLSNavierStokesAssemblerCore<dim>::assemble_matrix(
       const double tau =
         this->simulation_control->get_assembly_method() ==
             Parameters::SimulationControl::TimeSteppingMethod::steady ?
-          1. / std::sqrt(std::pow(2. * u_mag / h, 2) +
-                         9 * std::pow(4 * viscosity / (h * h), 2)) :
-          1. / std::sqrt(std::pow(sdt, 2) + std::pow(2. * u_mag / h, 2) +
-                         9 * std::pow(4 * viscosity / (h * h), 2));
+          calculate_navier_stokes_tau_steady(u_mag, viscosity, h) :
+          calculate_navier_stokes_tau_transient(u_mag, viscosity, h, sdt);
```
CaroleAnneDaunais, line 70: clean

CaroleAnneDaunais (approved):
Looks good to me :)

blaisb:
I added the documentation for the stabilization :)!. Thanks for the reviews folks

PR #450, opened 2022-04-12
Radiation heat transfer

shahabgol:
# Description of the problem

- This PR modifies the convection heat transfer boundary condition to convection-radiation. This new BC is a combination of convection and radiation. If the Stefan-Boltzmann constant (with default value = 0) is set to 0, only the convection component is considered, whereas if the convection coefficient, h (with default value = 0), is set to 0, only the radiation component is considered. Otherwise, both the convection and radiation are significant on the boundary.

- It also adds "calculate temperature range" option to post-processing. By enabling "calculate temperature range", the solver reports the minimum and maximum temperature in the simulation domain.

# How Has This Been Tested?

- heat_transfer_laser application test is improved to test this new BC.

# Documentation

- boundary_conditions_multiphysics is updated.

### DIFF
--- a/include/solvers/postprocessing_cfd.h
+++ b/include/solvers/postprocessing_cfd.h
@@ -273,6 +273,23 @@ calculate_flow_rate(const DoFHandler<dim> &    dof_handler,
                     const Mapping<dim> &       mapping);
 
 
+/**
+ * @brief Calculates the minimum and maximum temperature in the simulation domain
+ *
+ * @param dof_handler. The argument used to get finite elements
+ *
+ * @param face_quadrature_formula The face quadrature formula for the calculation
+ *
+ * @param mapping The mapping of the simulation
+ */
+template <int dim, typename VectorType>
+std::pair<double, double>
+calculate_min_max_temperature(const DoFHandler<dim> &dof_handler_ht,
+                              const VectorType &     evaluation_point,
+                              const Quadrature<dim> &quadrature_formula,
+                              const Mapping<dim> &   mapping);
+
+
 
 #  define lethe_postprocessing_cfd_h
 
--- a/include/solvers/postprocessing_cfd.h
+++ b/include/solvers/postprocessing_cfd.h
@@ -276,11 +276,13 @@ calculate_flow_rate(const DoFHandler<dim> &    dof_handler,
 /**
  * @brief Calculates the minimum and maximum temperature in the simulation domain
  *
- * @param dof_handler. The argument used to get finite elements
+ * @param dof_handler_ht. The dof handler associated with the heat transfer problem
  *
- * @param face_quadrature_formula The face quadrature formula for the calculation
+ * @param evaluation_point Temperature solution
  *
- * @param mapping The mapping of the simulation
+ * @param quadrature_formula The quadrature formula for the calculation
+ *
+ * @param mapping
  */
 template <int dim, typename VectorType>
 std::pair<double, double>
--- a/source/solvers/heat_transfer.cc
+++ b/source/solvers/heat_transfer.cc
@@ -468,7 +468,7 @@ HeatTransfer<dim>::postprocess(bool first_iteration)
         }
     }
 
-  // Minimum, maximum and average temperature
+  // Minimum and maximum temperature
   if (simulation_parameters.post_processing.calculate_min_max_temperature)
     {
       std::pair<double, double> min_max_temperature =
--- a/source/solvers/heat_transfer.cc
+++ b/source/solvers/heat_transfer.cc
@@ -6,6 +6,7 @@
 #include <solvers/heat_transfer.h>
 #include <solvers/heat_transfer_assemblers.h>
 #include <solvers/heat_transfer_scratch_data.h>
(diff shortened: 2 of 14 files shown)


include/solvers/postprocessing_cfd.h
```diff
@@ -273,6 +273,23 @@ calculate_flow_rate(const DoFHandler<dim> &    dof_handler,
                     const Mapping<dim> &       mapping);
 
 
+/**
+ * @brief Calculates the minimum and maximum temperature in the simulation domain
+ *
+ * @param dof_handler. The argument used to get finite elements
```
blaisb, line 279:
```suggestion
 * @param dof_handler. The dof handler associated with the heat transfer problem.
```

include/solvers/postprocessing_cfd.h
```diff
@@ -273,6 +273,23 @@ calculate_flow_rate(const DoFHandler<dim> &    dof_handler,
...
+ *
+ * @param face_quadrature_formula The face quadrature formula for the calculation
```
blaisb, line 281: This is not a face quadrature right? please review these comments. I think there is some copy paste errors there :)!

source/solvers/postprocessing_cfd.cc
```diff
@@ -1040,3 +1040,81 @@ calculate_flow_rate(const DoFHandler<3> &                     dof_handler,
...
+  const FESystem<dim, dim> fe = dof_handler_ht.get_fe();
+
+  const unsigned int  n_q_points = quadrature_formula.size();
+  std::vector<double> local_temperature_values(n_q_points);
+
+  FEValues<dim> fe_values(mapping,
+                          fe,
+                          quadrature_formula,
+                          update_values | update_quadrature_points |
+                            update_JxW_values);
```
blaisb, line 1060: Why update the quadrature points and the JxW values, you only use the values anyway? Might as well. update as minimally as possible.

blaisb (commented):
Very good work. Do you plan on adding the capability to output this to a text file instead of just into the log? Or is that planned for a follow-up PR?

include/solvers/postprocessing_cfd.h
shahabgol: True :) corrected.

shahabgol:
> Very good work. Do you plan on adding the capability to output this to a text file instead of just into the log? Or is that planned for a follow-up PR?

Is it necessary to write minimum and maximum temperatures in a text file? I think we will only need it for testing. The rest of comments have been addressed.

blaisb:
> > Very good work. Do you plan on adding the capability to output this to a text file instead of just into the log? Or is that planned for a follow-up PR?
> 
> Is it necessary to write minimum and maximum temperatures in a text file? I think we will only need it for testing. The rest of comments have been addressed.

Not necessary at all, you are right

blaisb:
Could you add a quick unit test with an analytical solution for pure radiation BC (like a couette with a radiation BC on one side orsomething) just to make sure the code work as intended :)

shahabgol:
> When CI passes, this seems very good to me! Only one question, would we also like a Conduction boundary condition?

Conduction BC is significant if we have two solids (with high k) in contact or the convection coefficient (h) is small, otherwise it is negligible.

source/solvers/heat_transfer.cc
```diff
@@ -466,6 +467,24 @@ HeatTransfer<dim>::postprocess(bool first_iteration)
                       << std::endl;
         }
     }
+
+  // Minimum, maximum and average temperature
```
oguevremont, line 471:
```suggestion
  // Minimum and maximum temperature
```
```suggestion
  // Minimum and maximum temperature
```

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -422,18 +422,27 @@ HeatTransferAssemblerRobinBC<dim>::assemble_rhs(
...
+  const double Stefan_Boltzmann_constant =
+    this->boundary_conditions_ht.Stefan_Boltzmann_constant;
+
+  // Robin boundary condition, loop on faces (Newton's cooling law +
+  // Stefan-Boltzmann law) Convection-radiation BC is a combination of
+  // convection and radiation. If the Stefan-Boltzmann constant (with default
+  // value = 0) is set to 0, only the convection component is considered,
+  // whereas if the convection coefficient, h (with default value = 0), is set
+  // to 0, only the radiation component is considered. Otherwise, both the
+  // convection and radiation are significant on the boundary implementation
```
oguevremont, line 435: Should it be explicitly said what happens when both are zero?

oguevremont (commented):
Good work!

blaisb:
@shahabgol it seems a lot of tests are breaking in debug mode?

source/solvers/heat_transfer_assemblers.cc
shahabgol: if both are 0, then using the convection-radiation BC is meaningless. I think it's not necessary to mention it.

oguevremont (approved):

PR #451, opened 2022-04-13
Changed test to be in 2D

CaroleAnneDaunais:
# Description of the problem

The applications_tests/gls_sharp_navier_stokes_3d/sphere_power_law_ramp example was very unstable in the CI. 

# Description of the solution

I removed a dimension, hoping this works

# How Has This Been Tested?
The test it now:
- applications_tests/gls_sharp_navier_stokes_2d/sphere_power_law_ramp


blaisb (approved):
Seems good, but you kept the prm for the 3d case, is there a reason for that?

CaroleAnneDaunais:
> Seems good, but you kept the prm for the 3d case, is there a reason for that?

No reason, my mistake!

PR #460, opened 2022-04-22
correct bug

Luckabarbeau:
# Description of the problem

- Some lines of code could lead to incorrect reload of particle state

# Description of the solution

- Correct the lines that could lead to errors


blaisb (changes requested):
Do you think you can add a test that reproduce the bug you had and fixed with this?

blaisb (approved):

PR #469, opened 2022-05-01
Update cfd dem example

blaisb:
# Description of the problem

- The square fluidized bed example has some sub-optimal parameters for the DEM that made loading the particles far too long

# Description of the solution

- Optimized the parameters. Loading the  30k particle takes about 20-25 minutes now.

# How Has This Been Tested?

- I have rerun the example completely on my machine at home and at work

# Documentation

- I updated the corresponding documentation

# Future changes

- We will need more CFD-DEM examples :)


shahabgol (approved):
It is very clear. I have no comment.

PR #470, opened 2022-05-03
Fix heat transfer bugs

shahabgol:
# Description of the problem

- This PR fixes two bugs in the heat transfer solver:
1, source term is added to the strong residual vector
2, laser term is added to the local matrix of the Robin boundary condition.

# How Has This Been Tested?

- The results of one application test (on one and two CPUs) are updated.

### DIFF
--- a/source/solvers/heat_transfer_assemblers.cc
+++ b/source/solvers/heat_transfer_assemblers.cc
@@ -165,7 +165,7 @@ HeatTransferAssemblerCore<dim>::assemble_rhs(
 
       // Calculate the strong residual for GLS stabilization
       strong_residual_vec[q] += rho_cp * velocity * temperature_gradient -
-                                thermal_conductivity[q] * temperature_laplacian;
+                                thermal_conductivity[q] * temperature_laplacian + scratch_data.source[q];
 
       for (unsigned int i = 0; i < n_dofs; ++i)
         {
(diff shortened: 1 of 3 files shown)


source/solvers/heat_transfer_assemblers.cc
```diff
@@ -164,8 +164,10 @@ HeatTransferAssemblerCore<dim>::assemble_rhs(
                       9 * std::pow(4 * alpha / (h * h), 2));
 
       // Calculate the strong residual for GLS stabilization
-      strong_residual_vec[q] += rho_cp * velocity * temperature_gradient -
-                                thermal_conductivity[q] * temperature_laplacian;
+      strong_residual_vec[q] +=
+        rho_cp * velocity * temperature_gradient -
+        thermal_conductivity[q] * temperature_laplacian +
+        scratch_data.source[q];
```
blaisb, line 170: I think there is a sign error here. It should be - and not +.
The equation is rho_cp * u * grad(T) - k * laplacian T - source = 0
This is also what is coded in the RHS
k * (grad T) * grad(test) + rho_cp*u*grad(T) - source 

Only the laplacian changes sign due to integration by part.

blaisb (commented):
Just a slight mistake to address.

source/solvers/heat_transfer_assemblers.cc
shahabgol: It is corrected. The errors also decreased in the failed application test. :)

shahabgol:
@lpsaavedra Could you please review this PR and merge it if it's ok?

lpsaavedra (approved):

PR #474, opened 2022-05-09
Correct_thread_bug

Luckabarbeau:
# Description of the problem

- The force evaluation was wrong when forcing the code to use one thread

# Description of the solution

- Find a patch by forcing the use of high order mapping on the surface of the particles

# Comments

- We still do not fully understand what was happening exactly. But a patch that fix the problems was added
- All the tests that use force in q1q1 have changed slightly since we always use high order mapping on the sphere


blaisb (approved):
Very good work!

PR #479, opened 2022-05-19
Melting example

shahabgol:
# Description of the problem

- I was planning to add a melting example in this PR, but I noticed there are some missing elements that should be added/addressed before adding the example. This PR adds the phase_change thermal expansion coefficient model.

# How Has This Been Tested?

- thermal_expansion_phase_change

# Future changes

- As mentioned, melting example will be added.


source/solvers/heat_transfer_assemblers.cc
```diff
@@ -209,6 +209,7 @@ HeatTransferAssemblerBDF<dim>::assemble_matrix(
     scratch_data.grad_specific_heat_temperature;
 
   // Copy data elements
+  // ?? do we need strong residual here??
```
tonielgeitani, line 212: I don't know if you mean't to keep this comment or if you want to delete it

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -355,6 +356,7 @@ HeatTransferAssemblerBDF<dim>::assemble_rhs(
 
               if (GGLS)
                 {
+                  // ?? why rho_cp * rho_cp
```
tonielgeitani, line 359: same here

tonielgeitani (approved):

source/core/thermal_expansion_model.cc
```diff
@@ -19,6 +19,11 @@
...
-  return std::make_shared<ConstantThermalExpansion>(
-    fluid_properties.thermal_expansion);
+  if (fluid_properties.thermal_expansion_model ==
+      Parameters::Fluid::ThermalExpansionModel::phase_change)
+    return std::make_shared<ThermalExpansionPhaseChange>(
+      fluid_properties.phase_change_parameters);
+  else
+    return std::make_shared<ConstantThermalExpansion>(
+      fluid_properties.thermal_expansion);
 }
```
acdaigneault, line 29: Only curiosity : 
Why keeping _ThermalExpansionModel_ class instead of having only _ConstantThermalExpansion_ and _ThermalExpansionPhaseChange_ with inheritance of _PhysicalPropretyModel_ instead of _ThermalExpansionModel._
And having the model_cast condition in the physical_properties_manager.cc file line 79?

acdaigneault (approved):

source/core/thermal_expansion_model.cc
shahabgol: It's a good idea that will make the code more clear. but I used the method that is currently implemented for all the properties. maybe we will change it to a base class and derived models structure in the future.

source/solvers/heat_transfer_assemblers.cc
shahabgol: I kept these comments to ask them from Bruno. I will remove them after that.

shahabgol:
I am going to merge this PR if you have no more comments.

blaisb:
@shahabgol if a test is broken and you need to disable it, you need to open an issue so we can keep track of it.

include/core/thermal_expansion_model.h
```diff
@@ -18,6 +18,7 @@
 #define lethe_thermal_expansion_model_h
 
 #include <core/physical_property_model.h>
+#include <core/specific_heat_model.h>
```
blaisb, line 21: I would move the phase change stuff out of the specific heat model, there is no reason why thermal expansion should depend on specific heat.

include/core/thermal_expansion_model.h
```diff
@@ -110,4 +111,122 @@ class ConstantThermalExpansion : public ThermalExpansionModel
...
+public:
+  /**
+   * @brief Default constructor
+   */
+  ThermalExpansionPhaseChange(const Parameters::PhaseChange phase_change_params)
+    : p_phase_change_params(phase_change_params)
+    , thermal_expansion_s(phase_change_params.thermal_expansion_s)
+    , thermal_expansion_l(phase_change_params.thermal_expansion_l)
+    , T_solidus(phase_change_params.T_solidus)
+    , T_liquidus(phase_change_params.T_liquidus)
```
blaisb, line 128: Why store a copy of all parameters + the phase change params? It's like a weird duplication of parameters. I would just store either what is only needed OR the phase change params overall, but not two versions of the same info.

include/core/thermal_expansion_model.h
```diff
@@ -110,4 +111,122 @@ class ConstantThermalExpansion : public ThermalExpansionModel
...
+              calculate_liquid_fraction(T[i], p_phase_change_params);
+
+            property_vector[i] = thermal_expansion_l * l_frac +
+                                 thermal_expansion_s * (1. - l_frac);
+          }
+      }
+  };
+
+  /**
+   * @brief jacobian Calcualtes the jacobian (the partial derivative) of the thermal expansion with respect to a field
```
blaisb, line 193:
```suggestion
   * @brief jacobian Calculates the jacobian (the partial derivative) of the thermal expansion with respect to a field
```

source/core/thermal_expansion_model.cc
blaisb: The idea is that eventually Thermal Expansion, Specific Heat, etc. might have more information in common or might need additional sanity checks that can only be provided if we know what type of properties they are. The idea is to have some sort of interface where at one point you know that aproperty is viscosity for example, otherwise properties will be moved around and could be used in weird fashion (e.g. specific heat passed when it should be a thermal conductivity). I thought it made the interface saner this way.

source/solvers/heat_transfer_assemblers.cc
blaisb: You can look at the following paper for more info on GGLS
https://doi.org/10.1016/S0045-7825(02)00242-6

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1712,16 +1712,16 @@ BuoyancyAssembly<dim>::assemble_rhs(
       // Current temperature values
       double current_temperature = scratch_data.temperature_values[q];
 
-      strong_residual[q] += -force * thermal_expansion * current_temperature;
+      strong_residual[q] += force * thermal_expansion * current_temperature;
```
blaisb, line 1715: I don't think this is correct. What are the equations you are solving?
u . grad(u) = -grad(p) + mu lap(u) + beta * "force" * (T-T0) no?
In this case, force should have a minus sign in front of it. 
I suggest you take the opportunity to update the documentation to make that clear.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1712,16 +1712,16 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
 
       // Assembly of the right-hand side
       for (unsigned int i = 0; i < n_dofs; ++i)
         {
           const auto phi_u_i = scratch_data.phi_u[q][i];
 
           // Laplacian on the velocity terms
-          local_rhs(i) +=
-            -force * thermal_expansion * current_temperature * phi_u_i * JxW;
+          local_rhs(i) -=
```
blaisb, line 1723: same comment as above.
Clearly, if you are not sure about the signs, take the time to add the information to the theory guide and it will help you sort it out. it's there for that.

tests/core/thermal_expansion_phase_change.cc
```diff
@@ -0,0 +1,87 @@
+/**
+ * @brief Tests the thermal_conductivity_phase_change model.
+ * During a phase change the physical properties (including conductivity) of the
```
blaisb, line 3:
```suggestion
 * @brief Tests the thermal_expansion_phase_change model.
 * During a phase change the physical properties (including thermal expansion) of the
```

Update the rest of the text also. It only talks about conductivity.

blaisb (changes requested):
Still some important elements to address. Please take the time to update the theory guide and add a section related to buoyany driven flow, otherwise you will keep being confused by teh forcing term due to buoyancy :)!

blaisb:
@shahabgol  I opened an issue for the test, it broke on my machine at home also. See #483

include/core/thermal_expansion_model.h
shahabgol: I use the function calculate_liquid_fraction which is defined in the specific heat model. Should I move this function to utilities?

include/core/thermal_expansion_model.h
shahabgol: Corrected.

source/solvers/navier_stokes_assemblers.cc
shahabgol: The equation is: u . grad(u) = -grad(p) + mu lap(u) - beta * "force" * (T-T0)

include/core/thermal_expansion_model.h
blaisb: If you can yeah. You could even define a phase_change.h file in core that is used to house these functions, that would be even cleaner.

blaisb (approved):
Please just move the phase change "functions" outside of the specific heat and to a phase_change.h file and this is ready for merge :)!

shahabgol:
> Please just move the phase change "functions" outside of the specific heat and to a phase_change.h file and this is ready for merge :)!

Done. :)

blaisb:
@shahabgol can you rebase and I will be able to merge? You still have a disabled test on in this branch

shahabgol:
> @shahabgol can you rebase and I will be able to merge? You still have a disabled test on in this branch

The test is enabled and the branch is rebased.

PR #480, opened 2022-05-24
Add DEM mesh documentation

shahabgol:
# Documentation

- This PR updates the mesh subsection in the DEM solver's documentation.


doc/source/parameters/dem/mesh.rst
```diff
@@ -1,3 +1,30 @@
...
-The mesh subsection of DEM simulations is identical to the `CFD <https://lethe-cfd.github.io/lethe/parameters/cfd/mesh.html>`_ in Lethe.
\ No newline at end of file
+The mesh subsection of DEM simulations is almost identical to the `CFD <https://lethe-cfd.github.io/lethe/parameters/cfd/mesh.html>`_ in Lethe. There are two additional parameters mainly used in DEM and CFD-DEM simulations. These parameters are ``check diamond cells`` and ``expand particle-wall contact search``.
+
+.. code-block:: text
+
+ subsection mesh
+    set check diamond cells		           = true
+
+    set expand particle-wall contact search = true
```
blaisb, line 10: The default value of these parameters is false no?

doc/source/parameters/dem/mesh.rst
```diff
@@ -1,3 +1,30 @@
...
+
+* ``check diamond cells`` parameter enables searching for diamond-shaped boundary cells and adding them to particle-wall contact search cells. The following image shows a diamond-shaped boundary cell colored in red.
+
+.. image:: images/diamond_cell.png
+    :alt: Schematic
+    :align: center
+    :width: 400
+
+
+* ``expand particle-wall contact search`` enables adding the boundary neighbor cells of boundary cells to the particle-wall contact search list. This feature should only be activated in geometries with concave boundaries. (For example, for particles flow inside a cylinder or sphere). The following image shows the boundary neighbor cells (colored in teal) of the read boundary cell.
```
blaisb, line 21: Add an explanation as to why this is set to true only for concave geometry (e.g. that otherwise it leads to unphysical contacts with walls that don't exist). You could also add an issue regarding this and maybe we could implement a version that is more intelligent in the future. E.g. it would add the extension only if the normal vector between the walls is at an angle below 90 degrees for example.

doc/source/examples/dem/rotating-drum/rotating-drum.rst
```diff
@@ -28,15 +28,16 @@ Parameter file
 Mesh
 ~~~~~
 
-In this example, we choose a ``cylinder`` grid type to create a cylinder. Grid arguments are the radius and half-length, respectively. Therefore, the specified grid arguments create a cylinder with a diameter of 0.24 m and a length of 0.36 m. The grid is refined 4 times to reach the desired cell size to particle diameter ratio (see packing in ball example for more details).
+In this example, we choose a ``cylinder`` grid type to create a cylinder. Grid arguments are the radius and half-length, respectively. Therefore, the specified grid arguments create a cylinder with a diameter of 0.24 m and a length of 0.36 m. The grid is refined 4 times to reach the desired cell size to particle diameter ratio (see packing in ball example for more details). The ``expand particle-wall contact search`` is used in concave geometries to enable particle-wall contact search with boundary faces of neighbor cells for particles located in each boundary cell.
```
blaisb, line 31: I think this explanation is a bit too short, Can you give more details as to why this is needed? It's important to be pedagological, this is not an easy option to understand.

blaisb (changes requested):
Great work, but I think the explanations need to be a bit more pedagological in the examples

doc/source/parameters/dem/mesh.rst
shahabgol: done. I've added it to my work list in the hackathon.

doc/source/examples/dem/rotating-drum/rotating-drum.rst
shahabgol: done.

blaisb (approved):

PR #481, opened 2022-05-24
Moving grid dim spacedim

emilebergeron:
Small PR to prepare for floating triangulations in the DEM solver. It is not implemented in the solver in this PR as it is still a work in progress. 

The changes mostly affect the moving grid class to allow it to function in dim/spacedim rather than dim/dim. A class and parser is also added for the moving triangulation. 

Since no functionality is added, there are no new tests yet, but the changes should not affect existing code. Tests will be added along with the solver integration in the other PR.


shahabgol:
I guess this PR is not compatible with dealii 9.3 at the moment, right?

emilebergeron:
> I guess this PR is not compatible with dealii 9.3 at the moment, right?

I'm working on fixing this, the deal.ii API changed in v10.0 so I need to check for the version before calling a specific function

source/dem/floating_grid.cc
```diff
@@ -0,0 +1,64 @@
+//
+// Created by Emile Bergeron on 2022-01-30.
+//
```
blaisb, line 3: We try to avoid keeping authors name in the files now :)

source/dem/floating_grid.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <deal.II/grid/grid_out.h>
+
+template <int dim, int spacedim>
+FloatingGrid<dim, spacedim>::FloatingGrid(
+  const DEMSolverParameters<spacedim> &dem_parameters,
+  const ConditionalOStream &           pcout,
+  const double &                       dem_time_step)
+  : gridMotion(dem_parameters, dem_time_step)
+{
+  if (dem_parameters.floating_grid.mesh.file_name != "none")
```
blaisb, line 19: It would be better to have a number of floating grids parameter which is set to 0, this will pave the support for multiple one instead of using a string for that.

blaisb (changes requested):
Good first PR, one or two small things to fix and fixing 9.3 compatibility and this will be good to go.

source/dem/floating_grid.cc
emilebergeron: I changed the code so that the FloatingGrid class only takes a subsection of the parameters as an arugument, which should make this change easy (only need to change the parser, more or less). I will implement this in the next PR as I add the code to the DEM solver.

blaisb (approved):

PR #488, opened 2022-06-06
Increase scale of resolved cfd dem

voferreira:
# Description of the problem

- Resolved CFD-DEM did not scale well with the number of particles in terms of performance.

# Description of the solution

- Implementation of a Verlet list for the contact search.

# How Has This Been Tested?

- ctests
- Case with 1000 particles

# Documentation

- Parameters resolved CFD-DEM.

# Future changes

- Further implementations on the search for the cut cells will be implemented with the same goal.


oguevremont (approved):
Good job, I have nothing to add

doc/source/parameters/resolved_cfd-dem/resolved_cfd-dem.rst
```diff
@@ -109,6 +110,8 @@ To sharpen the immersed boundary of each particles, a layer of cells around the
 
 * The ``alpha`` parameter is the relaxation parameter used when solving the dynamics equation of the particle.
 
+* The ``contact search radius factor`` parameter is used to create the list of potential contacting particles. If ``contact search radius factor`` > 1, two given particles with respective radius :math:`R_1` and :math:`R_2` are in potential contact if the distance between them is < :math:`(R_1 + R_2) * factor`. Alternatively, if ``contact search radius factor`` < or = 1, all particles are considered in potential contact with each other. If no value is provided this parameter is set to 0.
```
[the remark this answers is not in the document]
blaisb: Why set it to zero, I don't understand...

include/fem-dem/ib_particles_dem.h
```diff
@@ -237,12 +243,15 @@ class IBParticlesDEM
   std::shared_ptr<ParticleParticleContactForce<dim>>
     particle_particle_contact_force_object;
 
+  std::vector<std::set<int>> particles_contact_candidates;
```
blaisb, line 246: Shouldnt it be an unsigned int?

include/fem-dem/ib_particles_dem.h
```diff
@@ -237,12 +243,15 @@ class IBParticlesDEM
...
+
   std::shared_ptr<ParticleWallContactForce<dim>>
     particle_wall_contact_force_object;
 
   // Empty parameters to initilize particle_wall_contact_force_object
   double triangulation_cell_diameter = 0.0;
 
+
```
blaisb, line 254:
```suggestion
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -67,6 +67,69 @@ IBParticlesDEM<dim>::update_particles(
...
+{
+  particles_contact_candidates.resize(dem_particles.size());
+
+  double radius_factor = parameters->contact_search_radius_factor;
+
+  // If contact_search_radius_factor < 1, all particles are taken into
+  // account in the contact search. Otherwise, do it for the particles inside
+  // the region of contact search. By default, radius_factor is zero, so all
+  // particles are taken into account.
+  if (radius_factor <= 1)
```
blaisb, line 82: I don't like these weird behaviors based on a double. If you want all particles to take into account all particles, but a very large value of the radius factor. There is no reason why <0 would lead to all particles taking into account all particles. It's like changing the meaning of the parameter.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -67,6 +67,69 @@ IBParticlesDEM<dim>::update_particles(
...
+
+                  particles_contact_candidates[particle_one.particle_id].insert(
+                    particle_two.particle_id);
+                }
+            }
+        }
+    }
+
+
+  else
```
blaisb, line 104: This whole if, else can be completely removed by just setting a large value of the radius factor if you want everyone to see evryone

source/fem-dem/ib_particles_dem.cc
```diff
@@ -77,8 +140,14 @@ IBParticlesDEM<dim>::calculate_pp_contact_force(
...
     {
-      for (auto &particle_two : dem_particles)
+      for (auto particle_contact_candidates_id =
+             particles_contact_candidates[particle_one.id].begin();
+           particle_contact_candidates_id !=
+           particles_contact_candidates[particle_one.id].end();
+           ++particle_contact_candidates_id)
         {
+          auto  particle_contact_id = *particle_contact_candidates_id;
+          auto &particle_two        = dem_particles[particle_contact_id];
```
blaisb, line 150: Why is one a reference and the other one a pointer?
If you are not modifying the particle, use const auto & instead of auto &

source/fem-dem/ib_particles_dem.cc
```diff
@@ -184,8 +253,14 @@ IBParticlesDEM<dim>::calculate_pp_lubrication_force(
...
```
blaisb, line 263: Same comment as above,.

blaisb (changes requested):
Good idea. There are some things to address though regarding coding style. Especially this if/else that is not necessary and this default value of 0 + the meaning changing when it is lower than 1

doc/source/parameters/resolved_cfd-dem/resolved_cfd-dem.rst
```diff
@@ -21,6 +21,7 @@ This subsection contains the parameters related to the resolved CFD-DEM around p
 		set particle nonlinear tolerance            = 1e-6
 		set DEM coupling frequency                  = 1000
 		set alpha                                   = 1
+		set contact search radius factor            = 0
```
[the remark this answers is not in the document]
Luckabarbeau: I would also force it to be bigger than 1 and throw an error otherwise.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3001,14 +3001,14 @@ GLSSharpNavierStokesSolver<dim>::solve()
                                                    this->dof_handler,
                                                    *this->face_quadrature,
                                                    *this->mapping);
+          ib_dem.update_contact_candidates();
```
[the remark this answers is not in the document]
Luckabarbeau: I agree I would add this. Very good suggestion.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -67,6 +67,35 @@ IBParticlesDEM<dim>::update_particles(
...
+            {
+              const Point<dim> particle_one_location = particle_one.position;
+              const Point<dim> particle_two_location = particle_two.position;
+
+              if ((particle_one_location - particle_two_location).norm() <
+                  (particle_one.radius + particle_two.radius) * radius_factor)
+                {
+                  particles_contact_candidates[particle_one.particle_id].insert(
+                    particle_two.particle_id);
+                }
```
oguevremont, line 92:
```suggestion
          if (particle_one.particle_id < particle_two.particle_id)
            {
              const Point<dim> particle_one_location = particle_one.position;
              const Point<dim> particle_two_location = particle_two.position;

              if ((particle_one_location - particle_two_location).norm() <
                  (particle_one.radius + particle_two.radius) * radius_factor)
                {
                  particles_contact_candidates[particle_one.particle_id].insert(
                    particle_two.particle_id);
                  particles_contact_candidates[particle_two.particle_id].insert(
                    particle_one.particle_id);
                }
```

Luckabarbeau: This shouldn't be implemented as it is redundant information since only particle one will do the calculation for both particles. The original lines of codes are more appropriate and reduce memory usage. Maybe add a comment to explain that only particle 1 needs to have the particle two in its contact candidate

oguevremont (approved):

doc/source/parameters/resolved_cfd-dem/resolved_cfd-dem.rst
```diff
@@ -109,6 +110,16 @@ To sharpen the immersed boundary of each particles, a layer of cells around the
 
 * The ``alpha`` parameter is the relaxation parameter used when solving the dynamics equation of the particle.
 
+* The ``contact search radius factor`` parameter is used to create the list of potential contacting particles. Two given particles with respective radii :math:`R_1` and :math:`R_2` are in potential contact if the distance between them is < :math:`(R_1 + R_2) * factor`. If no value is provided this parameter is set to 3.
```
blaisb, line 113:
```suggestion
* The ``contact search radius factor`` parameter is used to create the list of potential contacting particles. Two given particles with respective radii :math:`R_1` and :math:`R_2` are in potential contact if the distance between them is < :math:`(R_1 + R_2) * factor`. The default value of this parameter is set to 3.
```

blaisb (approved):
A tiny comment and this is good to go

PR #491, opened 2022-06-07
Add laser beam direction to laser heat source term

shahabgol:
# Description of the problem

- Previously we did not have any parameters to define the laser beam direction of emission. I noticed it while working in VOF meltpool example. In this example, both air and the solid were heated by the laser as there were no parameter to limit the laser heat source to its direction of emission. In this PR, I have modified the laser heat source parameters to take the emission direction into account. Corresponding application test and example are updated, as well as the documentation.

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.prm
+++ b/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.prm
@@ -32,7 +32,7 @@ subsection laser parameters
 	set beam radius 		= 0.002
 	set start time 			= 0
 	set end time 			= 0.05
-	set beam orientation 		= y negative
+	set beam orientation 		= y-
 	subsection path
 	    set Function expression 	=  3 * t; 0.1
 	end
--- a/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.prm
+++ b/applications_tests/gls_navier_stokes_2d/heat_transfer_laser.prm
@@ -32,7 +32,7 @@ subsection laser parameters
 	set beam radius 		= 0.002
 	set start time 			= 0
 	set end time 			= 0.05
-	set beam orientation 		= y
+	set beam orientation 		= y negative
 	subsection path
 	    set Function expression 	=  3 * t; 0.1
 	end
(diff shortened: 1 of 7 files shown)


applications_tests/gls_navier_stokes_2d/heat_transfer_laser.prm
```diff
@@ -32,7 +32,7 @@ subsection laser parameters
 	set beam radius 		= 0.002
 	set start time 			= 0
 	set end time 			= 0.05
-	set beam orientation 		= y
+	set beam orientation 		= y negative
```
blaisb, line 35: Maybe it would be better to use y+ and y- ?

blaisb (changes requested):
Some nomenclature to review I believe, but the rest is all great

lpsaavedra (approved):
I don't have any comment, everything seems good to me :)

shahabgol:
Your comments have been applied. :)

source/core/parameters.cc
```diff
@@ -844,27 +844,61 @@ namespace Parameters
...
+          if constexpr (dim == 3)
+            {
+              beam_direction                     = 1;
+              beam_orientation                   = BeamOrientation::positive_z;
+              perpendicular_plane_coordinate_one = 0;
+              perpendicular_plane_coordinate_two = 1;
+              beam_orientation_coordinate        = 2;
+            }
+          else if constexpr (dim == 2)
+            Assert(dim == 2, TwoDimensionalLaserError(dim));
```
hepap, line 894: Maybe it would be interesting if it was working in 2D for geometries in the xz or yz planes.

hepap (approved):
I have no other comments than the one on 2D geometries in the xz and yz planes for the laser orientation :)

source/core/parameters.cc
shahabgol: Good question. Since the current laser model is 3d (it has penetration depth as a parameter), in 2d simulations we need to have a laser beam compatible with the grid, and we can't have beam emissions perpendicular to the 2d plane.

blaisb (approved):
Can you just change the enum for the parameter to y+ and Y- nomenclature. Then I can merge.

shahabgol:
> Can you just change the enum for the parameter to y+ and Y- nomenclature. Then I can merge.

Done :)

PR #492, opened 2022-06-07
Add none BC to heat_transfer

jeannejoachim:
# Description of the problem

- heat transfer BC were by default "temperature", so there was no mean to have nothing happening if say the heat transfer BC were not the first ordered ones

# Description of the solution

- add a "none" BC, by default, similarly to what have been done for the VOF BC

# How Has This Been Tested?

Tested with a modification of the `warming-up-a-viscous-fluid` (domain turned clockwise), which was not possible before this BC existed. This has been added to the example bonuses.

# Documentation

- example `warming-up-a-viscous-fluid`

- parameter: multiphysics boundary condition

# Future changes

@blaisb tell me if you would like the same `none` BC for the tracer (already implemented for VOF).

### DIFF
--- a/doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
+++ b/doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
@@ -331,25 +331,17 @@ Several adjustments have to be made in the `.prm` to turn the domain clockwise, 
 * in ``subsection mesh``: ``set grid arguments = 0, 0 : 1, 0.5 : true``
 * in ``subsection analytical solution``, ``subsection temperature``: 
    ``set Function expression = Tw+(((rho*nu)*v*v)/(2*K))*(1-(y/B)*(y/B))``
-* and most importantly, the ``boundary conditions`` subsection should state explicitly ``outlet`` for the fluid boundary conditions, as the ``number`` of boundary conditions should be adapted to use the bottom and top wall (see the `deal.II documentation on hyper_rectangle grid generator <https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a56019d263ae45708302d5d7599f0d458>`_ for further details):
+* and most importantly, the ``id`` of ``boundary conditions`` should be adapted to use the bottom and top wall (see the `deal.II documentation on hyper_rectangle grid generator <https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a56019d263ae45708302d5d7599f0d458>`_ for further details):
 
 .. code-block:: text
 
 	subsection boundary conditions
-	  set number                  = 4
+	  set number                  = 2
 	    subsection bc 0
-	    set id = 0
-		set type              = outlet
-	    end
-	    subsection bc 1
-	    set id = 1
-		set type              = outlet
-	    end
-	    subsection bc 2
 	    set id = 2
 		set type              = noslip
 	    end
-	    subsection bc 3
+	    subsection bc 1
 	    set id = 3
 		set type              = function
 		subsection u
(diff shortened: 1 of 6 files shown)


blaisb:
It was actually doable before (you just needed to specify like 3 boundary conditions on surface 0, 2 and 3 for example), but that's just nitpicking. Its a good idea to be more explicit.

doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
```diff
@@ -322,6 +322,64 @@ After the fluid has been heated up by the right plate, the temperature is really
...
+	subsection boundary conditions
+	  set number                  = 4
+	    subsection bc 0
+	    set id = 0
+		set type              = outlet
+	    end
+	    subsection bc 1
+	    set id = 1
+		set type              = outlet
+	    end
```
blaisb, line 347: I'm not sure I understand, here you don't have only outlets no? one is an inlet and one is an outlet? Be careful  because the outlet boundary condition penalizes inflow, it's not like a nothing boundary condition, it's different depending on the value of beta and it's not meant to be used as an inlet

doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
```diff
@@ -322,6 +322,64 @@ After the fluid has been heated up by the right plate, the temperature is really
...
+	    subsection bc 3
+	    set id = 3
+		set type              = temperature
+		set value             = 80
+	    end
+	end
+
+.. tip::
+	Heat transfer boundary conditions are of ``type = none`` by default, so this is the type implicitly used for ``subsection bc 0`` and ``subsection bc 1``.
```
blaisb, line 380: It's would be better to have the type=none be replaced by something like zero flux or 'natural' because that's what is being done when you don't do anything. I like no flux or zero flux, but I am open to alternatives.

doc/source/parameters/cfd/boundary_conditions_multiphysics.rst
```diff
@@ -36,8 +36,9 @@ The default parameters of each are shown:
 
 * ``id`` is the number associated with the boundary condition. By default, Lethe assumes that the id is equivalent to the number of the bc.
 
-* ``type``: This is the type of boundary condition been imposed. At the moment, choices are:
-    * ``temperature`` (Dirichlet BC), to impose a given temperature ``value`` at the boundary.
+* ``type``: type of boundary condition being imposed. At the moment, choices are:
+    * ``none`` so that there is no heat transfer boundary condition (nothing happens),
```
blaisb, line 40: I would replace that by no flux or a similar thing, see above comment. It's a bit easier to understand, because it's not that nothing happens, it's that no flux is assembled at the boundary.

include/core/boundary_conditions.h
```diff
@@ -35,6 +35,8 @@ namespace BoundaryConditions
 {
   enum class BoundaryType
   {
+    // common
+    none,
```
blaisb, line 39:
```suggestion
    zero_flux,
```

blaisb (changes requested):
Some nomenclature to fix, but otherwise it's good

doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
jeannejoachim: I have a preference for `noflux`, I will change that

include/core/boundary_conditions.h
jeannejoachim: I use `none` for the VOF BC, so I will not remove it. I will just create an additional `noflux` :)

doc/source/examples/multiphysics/warming-up-a-viscous-fluid/warming-up-a-viscous-fluid.rst
blaisb: You can specify a nothing BC, you just need to specify the id of the bc.
For example, if you a nothing BC on the id 0, you can have something like.
subsection bc 0
 set id =1
 set type ...
end
subsection bc 1
 set id = 2
 set type ...
end

It is not mandatory that the bc "number" matches the boundary id.  In all cases, it would be better to make the horizontal example either with a nothing bc or without specifying an outlet (specifying nothing) because this "inlet" behavior of the outlet is not desired, it's because the default beta value is small.

jeannejoachim: Oh ok I got it! Sorry, you are right, I will change that right away

tests/solvers/restart_01.cc
```diff
@@ -164,7 +164,7 @@ test()
   NSparam.restart_parameters.frequency  = 1;
   NSparam.non_linear_solver.verbosity   = Parameters::Verbosity::quiet;
   NSparam.linear_solver.verbosity       = Parameters::Verbosity::quiet;
-  NSparam.boundary_conditions.createDefaultNoSlip();
+  NSparam.boundary_conditions.createNoSlip();
```
blaisb, line 167: Is NoSlip the boundary condition created by default or is it a none now? If it is still no Slip, then that's ok. Just to verify.

blaisb (changes requested):
One last comment i want to be sure off, then ready for a merge.

tests/solvers/restart_01.cc
jeannejoachim: It is `none`! In fact this `createDefaultNoSlip` is only used in this test, still I had to keep the method so that the test is not broken. I renamed it `createNoSlip` to clarify that this is not the case by default anymore :)

blaisb: Ohh it's just in a test. I see.
Ready fo rmergsaurus rex then.

blaisb (approved):

PR #493, opened 2022-06-08
Update rheological physical properties documentation

CaroleAnneDaunais:
# Description of the problem

The rheological models lacked some explanation

# Description of the solution

- Adding images for both the Carreau and power-law model
- Added a note saying that the Carreau model should not be used for shear-thickening flows
- Rearranged the models and added them to the list of contents


blaisb (approved):
Very good, nothing to add. Ready to merge once CI passess

PR #498, opened 2022-06-15
Add density ratio term to the surface tension force calculations

shahabgol:
# Description of the problem

- A bug is fixed in the surface tension force calculations. The density ratio term is added to the surface tension force calculations.

# How Has This Been Tested?

- gls_static_bubble_surface_tension_force is changed to gls_static_droplet_surface_tension_force, because it shows a droplet, not the bubble; and its pressure calculations are revised.

# Documentation

- The documentation is updated (rising bubble example).


applications_tests/gls_navier_stokes_2d/gls_static_droplet_surface_tension_force.prm
```diff
@@ -51,7 +51,7 @@ subsection analytical solution
   set verbosity = verbose
     subsection uvwp
     	set Function constants = sigma=0.023, R=0.022
-    	set Function expression = 0; 0; if ((x-0.04) * (x-0.04) + (y-0.04) * (y-0.04) < 0.02 * 0.02 , 2 * sigma / R , 0)
+    	set Function expression = 0; 0; if ((x-0.04) * (x-0.04) + (y-0.04) * (y-0.04) < 0.02 * 0.02 , sigma / R , 0)
```
blaisb, line 54: There was an error in the analytical solution? Have you looked into convergence?

blaisb (approved):
Very good, seems like the new results are also much better for the sphere. I think it is time you start writing a theory section for the VOF model :).
Quick question, have you tried running the bubble with the new mass conservation routine Jeanne implemented?

shahabgol:
> Very good, seems like the new results are also much better for the sphere. I think it is time you start writing a theory section for the VOF model :). Quick question, have you tried running the bubble with the new mass conservation routine Jeanne implemented?

I have prepared another example (square bubble reshaping into circle) which uses the mass conservation routine. I will add it in a future PR.

tonielgeitani (approved):

doc/source/examples/multiphysics/rising-bubble-VOF/rising-bubble-VOF.rst
```diff
@@ -111,9 +111,9 @@ Volume of Fluid (VOF)
 In Lethe, the surface tension force (:math:`{\bf{F_{\sigma}}}`) is calculated using the following equation [1, 2]:
 
     .. math::
-        {\bf{F_{\sigma}}} = 2 \sigma k {\bf{\psi}}
+        {\bf{F_{\sigma}}} = 2 \frac{\rho}{\rho_0 + \rho_1} \sigma k {\bf{\psi}}
 
-where :math:`\sigma`, :math:`k` and :math:`\bf{\psi}` denote surface tension coefficient, filtered curvature and filtered phase fraction gradient.
+where :math:`\sigma`, :math:`k` and :math:`\bf{\psi}` denote surface tension coefficient, filtered curvature and filtered phase fraction gradient. :math:`\rho`, :math:`\rho_1`, and :math:`\rho_2` are density of the flow, density of phase 0, and density of phase 1, respectively.
```
hepap, line 116:
```suggestion
where :math:`\sigma`, :math:`k` and :math:`\bf{\psi}` denote respectively the surface tension coefficient, the filtered curvature and the filtered phase fraction gradient. :math:`\rho`, :math:`\rho_1`, and :math:`\rho_2` are the density of the flow, the density of phase 0, and the density of phase 1, respectively.
```

hepap (approved):

PR #507, opened 2022-06-24
Add two performance tests for DEM

blaisb:
# Description of the problem

- The current performance test we have for DEM is good, but it starts with initial overlaps which make it not very good to test scalability because the dynamics is unpredictable

# Description of the solution

- Add a new short performance test that I have been using for a year or so

# Comments

- The performance of the code is very good. It is actually as good as it as ever been...
-


shahabgol (approved):
Very good PR. I'll use these benchmarks.

PR #510, opened 2022-06-29
Fix vof ht

jeannejoachim:
# Description of the problem

Heat transfer did not consider multiple fluid parameters when used with VOF

# Description of the solution

- Solve the bug (silly parameter naming)
- Add test with analytical solution

# How Has This Been Tested?

Test with analytical solution: two fluids with different `k` and between two walls at fixed temperature.

# Future changes

I could see that the documentation pages on `initial condition` and `analytical solution` could use another check: `temperature` subsection is missing on `initial condition`, a few typos, and `analytical solution` could be clarified. 

Also I think it would be clearer to rename the `subsection VOF` in `initial condition` to `subsection phase`, so that it is coherent with `temperature` and `subsection phase` in the `analytical solution`. 

I will do that in another documentation PR as it is not directly related.

# Comments

I changed the `output` for the marangoni effect test with @shahabgol approval. We are not sure if it is better now, it will be addressed by Shahab in his next example.

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
+++ b/applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
@@ -16,23 +16,23 @@ Kinetic energy : 2.06454e-05
 *****************************************************************
 Transient iteration : 2        Time : 0.002    Time step : 0.001    CFL : 0.00936298
 *****************************************************************
-Enstrophy  : 91.7191
-Kinetic energy : 0.001354
+Enstrophy  : 50.4139
+Kinetic energy : 0.000761582
 
 *****************************************************************
-Transient iteration : 3        Time : 0.003    Time step : 0.001    CFL : 0.160143
+Transient iteration : 3        Time : 0.003    Time step : 0.001    CFL : 0.122855
 *****************************************************************
-Enstrophy  : 291.448
-Kinetic energy : 0.00504483
+Enstrophy  : 189.088
+Kinetic energy : 0.00325828
 
 *****************************************************************
-Transient iteration : 4        Time : 0.004    Time step : 0.001    CFL : 0.279504
+Transient iteration : 4        Time : 0.004    Time step : 0.001    CFL : 0.232129
 *****************************************************************
-Enstrophy  : 507.019
-Kinetic energy : 0.00983761
+Enstrophy  : 371.442
+Kinetic energy : 0.0071153
 
 *****************************************************************
-Transient iteration : 5        Time : 0.005    Time step : 0.001    CFL : 0.35994 
+Transient iteration : 5        Time : 0.005    Time step : 0.001    CFL : 0.317114
 *****************************************************************
-Enstrophy  : 711.921
-Kinetic energy : 0.014945
+Enstrophy  : 566.914
+Kinetic energy : 0.0117275
--- a/applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls.output
+++ b/applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls.output
@@ -1,0 +1,59 @@
+Running on 1 MPI rank(s)...
+   Number of active cells:       1024
+   Number of degrees of freedom: 3267
+   Volume of triangulation:      0.2
+   Number of VOF degrees of freedom: 1089
+   Number of thermal degrees of freedom: 1089
+
+*****************************************************************
+Transient iteration : 1        Time : 0.2      Time step : 0.2      CFL : 0       
+*****************************************************************
+L2 error velocity : 0
+L2 error phase : 0.195496
+L2 error temperature : 0.413192
+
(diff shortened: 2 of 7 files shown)


applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
```diff
@@ -16,23 +16,23 @@ Kinetic energy : 2.06454e-05
 *****************************************************************
 Transient iteration : 2        Time : 0.002    Time step : 0.001    CFL : 0.00936298
 *****************************************************************
-Enstrophy  : 91.7191
-Kinetic energy : 0.001354
+Enstrophy  : 50.4139
+Kinetic energy : 0.000761582
```
blaisb, line 20: Do the results still make sense @shahabgol ?

applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls.output
```diff
@@ -0,0 +1,59 @@
...
+0.4000 0.000000e+00   0.0000         
+0.6000 0.000000e+00   0.0000         
+0.8000 0.000000e+00   0.0000         
+1.0000 0.000000e+00   0.0000         
+ time  error_phase  
+0.2000 1.954964e-01 
+0.4000 1.954506e-01 
+0.6000 1.956370e-01 
+0.8000 1.956360e-01 
+1.0000 1.956358e-01
```
blaisb, line 53: Isnt that a bit high for phase error? Is it because of numerical diffusion?

blaisb (approved):
Nothing really, but some comments are food for thought. This is good to merge once @shahabgol  has checked my first question.

blaisb:
> # Future changes
> I could see that the documentation pages on `initial condition` and `analytical solution` could use another check: `temperature` subsection is missing on `initial condition`, a few typos, and `analytical solution` could be clarified.
> 
> Also I think it would be clearer to rename the `subsection VOF` in `initial condition` to `subsection phase`, so that it is coherent with `temperature` and `subsection phase` in the `analytical solution`.


I would rather we rename everything "phase" to VOF. VOF means volume of fluid anyway, so it's a more coherent nomenclature than phase.

applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls.output
jeannejoachim: Yes it is due to diffusion and my mesh is quite coarse. It is lowered when I augment the mesh refinement so I would not bother too much about it

jeannejoachim:
> > # Future changes
> > I could see that the documentation pages on `initial condition` and `analytical solution` could use another check: `temperature` subsection is missing on `initial condition`, a few typos, and `analytical solution` could be clarified.
> > Also I think it would be clearer to rename the `subsection VOF` in `initial condition` to `subsection phase`, so that it is coherent with `temperature` and `subsection phase` in the `analytical solution`.
> 
> I would rather we rename everything "phase" to VOF. VOF means volume of fluid anyway, so it's a more coherent nomenclature than phase.

I would advise against it especially in the `.prm`. For a user, in the `subsection initial conditions` for instance, it is much more simple to know what `subsection phase` means intuitively (because we set the initial value of a quantity, which is the phase), than `subsection vof` (setting an initial value for a method's name does not mean anything imho). I think it is clearer to make a difference between the solvers's name (`vof` and `heat transfer`), as enabled in the `multiphysics subsection`, and the quantity of interest that is really set, like the `temperature` and the `phase`.

jeannejoachim:
> > > > # Future changes
> > > > I could see that the documentation pages on `initial condition` and `analytical solution` could use another check: `temperature` subsection is missing on `initial condition`, a few typos, and `analytical solution` could be clarified.
> > > > Also I think it would be clearer to rename the `subsection VOF` in `initial condition` to `subsection phase`, so that it is coherent with `temperature` and `subsection phase` in the `analytical solution`.
> > > 
> > > 
> > > I would rather we rename everything "phase" to VOF. VOF means volume of fluid anyway, so it's a more coherent nomenclature than phase.
> > 
> > 
> > I would advise against it especially in the `.prm`. For a user, in the `subsection initial conditions` for instance, it is much more simple to know what `subsection phase` means intuitively (because we set the initial value of a quantity, which is the phase), than `subsection vof` (setting an initial value for a method's name does not mean anything imho). I think it is clearer to make a difference between the solvers's name (`vof` and `heat transfer`), as enabled in the `multiphysics subsection`, and the quantity of interest that is really set, like the `temperature` and the `phase`.
> 
> I think we will have to agree to disagree here. VOF meaning volume of fluid has plenty of meaning. In addition, if we have a second type of multiphase flow method (e.g. a level-set approach), things will get overly messy. I'd rather more everything towards vof than phase.

For other readers: in the end we agreed on renaming `phase` in the code and prm to `vof indicator` :)

blaisb:
Good to merge I think :)!

Issue #483, opened 2022-05-29
gls_free-surface_drop-wetting_inverted  gls_free-surface_drop-peeling are broken


blaisb:
The tests break on my machine at home and break regularly on the CI
(see https://github.com/lethe-cfd/lethe/runs/6641090352?check_suite_focus=true)

Until the test is fixed and made stable, I have disabled it.
@jeannejoachim I think you were the one that wrote it? There seems to be something highly non-reproducible with that test indicating either a bug with the feature or a non-robust test.

I suggest running them through valgrind to see if they are sane

jeannejoachim:
Hi Bruno, I had trouble to setup these tests to begin with, so I narrowed them a lot. It always passes on my machine now and it passed multiple time on the CI during my last PR review so I thought I had found a stable way of testing these mechanisms. Sorry that it was not, you were right to disable it and we can talk about is this week!

PR #516, opened 2022-07-06
Fix tests vof pw

jeannejoachim:
# Description of the problem

- two tests using the peeling/wetting mechanisms where broken on the CI

# Description of the solution

- found the problem with valgrind: the `n_q_point` where not at the correct value
- changed and made this more explicit

# How Has This Been Tested?

- 2 tests for pw retested on my machine and with valgrind. Hope the CI passes now!

### DIFF
--- a/source/core/parameters.cc
+++ b/source/core/parameters.cc
@@ -1987,7 +1987,7 @@ namespace Parameters
         prm.set("Function expression", "0; 0; 0");
       prm.leave_subsection();
 
-      unsigned int max_ib_particles = 1000;
+      unsigned int max_ib_particles = 10;
       particles.resize(max_ib_particles);
       for (unsigned int i = 0; i < max_ib_particles; ++i)
         {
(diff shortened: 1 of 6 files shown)


jeannejoachim:
This fixes the issue #483

source/core/parameters.cc
```diff
@@ -1987,7 +1987,7 @@ namespace Parameters
         prm.set("Function expression", "0; 0; 0");
       prm.leave_subsection();
 
-      unsigned int max_ib_particles = 1000;
+      unsigned int max_ib_particles = 10;
```
hepap, line 1990: Is the `max_ib_particles = 10` still right?

hepap (approved):
Looks good to me! Vive Valgrind! :1st_place_medal:

source/core/parameters.cc
jeannejoachim: yes this is the value on master now!

lpsaavedra (approved):
Everything looks good to me!

blaisb (approved):

PR #518, opened 2022-07-06
Output BC at first iteration only

jeannejoachim:
# Description of the problem

Output BC was outputted at every iteration. That was not necessary.

# Description of the solution

Output BC only at the first iteration.

# How Has This Been Tested?

Locally on a prm I have.

### DIFF
--- a/source/fem-dem/gls_sharp_navier_stokes.cc
+++ b/source/fem-dem/gls_sharp_navier_stokes.cc
@@ -883,7 +883,7 @@ GLSSharpNavierStokesSolver<dim>::postprocess_fd(bool firstIter)
 {
   if (this->simulation_control->is_output_iteration())
     {
-      this->write_output_results(this->present_solution, firstIter);
+      this->write_output_results(this->present_solution);
     }
 
   bool enable =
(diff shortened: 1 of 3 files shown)


source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -883,7 +883,7 @@ GLSSharpNavierStokesSolver<dim>::postprocess_fd(bool firstIter)
 {
   if (this->simulation_control->is_output_iteration())
     {
-      this->write_output_results(this->present_solution);
+      this->write_output_results(this->present_solution, firstIter);
```
blaisb, line 886: You can get from the simulation control if this is the first iteration, since the function already uses the simulation control, why not use it directly?
See : https://github.com/lethe-cfd/lethe/blob/master/include/core/simulation_control.h#L186

blaisb (changes requested):
Just a small thing to fix

source/fem-dem/gls_sharp_navier_stokes.cc
jeannejoachim: Oh with the `is_at_start()`? I did not know, good suggestion I will change that

jeannejoachim: in the end I used `get_step_number() == 0` because `is_at_start()` is true for both iteration `0` and `1`

blaisb: Makes sense

voferreira (approved):
All perfect. Nice one :)

jeannejoachim:
> Not yet rebased.

It is not necessary here. I only rebase when there are conflicts that prevent direct merging! :angel:

PR #527, opened 2022-07-11
Vof ht fix dissipation

jeannejoachim:
# Description of the problem

Viscous dissipation in the air, with physical values chosen for numerical stability, lead to the air heating up the fluid of interest

# Description of the solution

- Add a parameter to choose where to apply viscous dissipation in the case of VOF
- Note that I preferred to leave the possibility to have viscous dissipation in both fluids, if they are close in viscosity
- Add a specialized `ViscousDissipationVOF` constructor

# How Has This Been Tested?

- Test added: air advected at the top edge, which does not lead to viscous dissipation so does not heat up the denser bottom fluid, as intended

# Documentation

- [x] Update heat transfer documentation page

# Comments

- On the test, the error on the velocity and pressure is big because I did not give any analytical solution for that. I intend to change soon the analytic output to make it more adaptable, so that pressure and velocity are not outputted by default.
- Already rebased


source/solvers/heat_transfer.cc
```diff
@@ -64,9 +64,19 @@ HeatTransfer<dim>::setup_assemblers()
...
+            std::make_shared<HeatTransferAssemblerViscousDissipation<dim>>(
+              this->simulation_control));
+        }
+      else
+        {
+          // Call for the specialized assembler
+          this->assemblers.push_back(
+            std::make_shared<HeatTransferAssemblerViscousDissipationVOF<dim>>(
+              this->simulation_control));
+        }
```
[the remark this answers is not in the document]
jeannejoachim: Yes I agree for readability, but for performances I always try to put the most probable case as the first `if` statement. This considered, do you still wish that I change the order? (I would be very ok with favoring readability also)

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -209,6 +212,23 @@ HeatTransferScratchData<dim>::calculate_physical_properties()
                 this->phase_values[q],
                 this->grad_specific_heat_temperature_0[q],
                 this->grad_specific_heat_temperature_1[q]);
+
+              // Coefficient used to neglect viscous dissipation in the fluid
+              // which is more than 10 times less dense than the other fluid
+              if (this->viscosity_1[q] > 10. * this->viscosity_0[q])
```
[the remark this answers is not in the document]
jeannejoachim: Oups sorry forgot to update that. At first I compared the density but then it seemed more logical to compare the viscosity, so I changed the code. I will update the comment!

source/solvers/heat_transfer.cc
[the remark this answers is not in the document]
blaisb: It won't change performance. This place has no impact on performance at all because choosing the assembler is done once per iteration and these are cheap objects to create.

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -209,6 +212,23 @@ HeatTransferScratchData<dim>::calculate_physical_properties()
                 this->phase_values[q],
                 this->grad_specific_heat_temperature_0[q],
                 this->grad_specific_heat_temperature_1[q]);
+
+              // Coefficient used to neglect viscous dissipation in the fluid
+              // which is more than 10 times less viscous than the other fluid
+              if (this->viscosity_1[q] > 10. * this->viscosity_0[q])
```
blaisb, line 218: Why do you want to neglect the viscous dissipation automatically in the fluid which is less viscous? Especially because here viscosity is the kinematic viscosity. For example, if I have a liquid with rho=1000 and mu =1 and air with rho =1 and mu = 0.01, I know that my air will have negligible viscous dissipation when compared to my liquid (liquid will be 100x larger). Yet this comparison will give you air as the viscous dissipater. I would do the comparison on  the dynamic viscosity and not the viscosity. 

In this case, I  think it would be more relevant to just specify from a user parameter on which phase viscous dissipation is applied (if you want to enable for both). The threshold for 10 seems a bit arbitrary. Might as well specify the fluid indicator with a default value of 1 and let it be this way :).

blaisb (changes requested):

source/solvers/heat_transfer_scratch_data.cc
jeannejoachim: Yes the user-defined option was also on my mind, I did not want to add yet another parameter, but I agree that it will be really less hacky than hard-coding a threshold. So I will change that!

doc/source/parameters/cfd/multiphysics.rst
```diff
@@ -22,6 +22,13 @@ This subsection defines the multiphysics interface of Lethe and enables the solu
 
   When ``set heat transfer = true``, these optional parameters can be used:
    * ``viscous dissipation``: controls if the viscous dissipation is taken into account in the heat transfer equation.
+
+   .. important::
+
+	When both ``viscous dissipation`` and ``VOF`` are ``true``, ``viscous dissipation`` will be neglected in the fluid with a ``kinematic viscosity`` (in :doc:`physical_properties`): :math:`\nu_\text{less viscous} < 10 \times \nu_\text{more viscous}`.
+
+	This mechanism is particularly useful when one fluid is the air: for numerical stability, the ``kinematic viscosity`` for the air is usually increased, but we do not want to have viscous dissipation in the air, which would heat it in a non-physical way.
```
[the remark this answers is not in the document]
voferreira: I'm not sure if that is what you meant, but I think it is clearer this way.

voferreira (commented):
Just a few comments, but it looks good for me! Regarding the parameter, you could have two ways to address this: either specify the ids of the phases you want to apply the viscous dissipation or apply a threshold specified in a second parameter. I like the first option better.

voferreira (changes requested):
I'll double-check it when you implement the parameter  :)

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -104,6 +104,8 @@ HeatTransferScratchData<dim>::enable_vof(const FiniteElement<dim> &fe,
   viscosity_1                      = std::vector<double>(n_q_points);
   grad_specific_heat_temperature_1 = std::vector<double>(n_q_points);
 
+  viscous_dissipation_indicator =
+    Parameters::DissipationIndicator::fluid1; // TODO take from parameters
```
jeannejoachim, line 108: @blaisb Would you have an insight about how you wish we access this parameter?

I have considered: 
- creating a getter method (in multiphysics_interface or in vof?), 
- modifying the `HeatTransferAssemblerViscousDissipationVOF` specialized method (but then I have to add an optional parameter to the overriden methods in `HeatTransferAssemblerBase`)
- leaving it as is (dissipation in `fluid1` only), knowing that if we need to apply it differently a portion of the code already exists

I would be happy to have your feedback on that :)

source/core/parameters_multiphysics.cc
```diff
@@ -116,6 +122,17 @@ Parameters::VOF::parse_parameters(ParameterHandler &prm)
...
+    // DissipationIndicator
+    const std::string op = prm.get("viscous dissipation application");
+    if (op == "fluid 1")
+      viscous_dissipation_indicator = Parameters::DissipationIndicator::fluid1;
+    else if (op == "fluid 0")
+      viscous_dissipation_indicator = Parameters::DissipationIndicator::fluid0;
+    else if (op == "both")
+      viscous_dissipation_indicator = Parameters::DissipationIndicator::both;
+    else
+      throw(std::runtime_error("Invalid viscous dissipation application"));
```
voferreira, line 134:
```suggestion
      throw(std::runtime_error("Invalid viscous dissipation application"));
```
```suggestion
      throw(std::runtime_error("Invalid viscous dissipation application. Options are 'fluid 0', 'fluid 1' or 'both'."));
```

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -551,7 +549,115 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+          viscous_dissipation_coefficient = phase_value_q;
+        }
+      else if (this->viscous_dissipation_indicator ==
+               Parameters::DissipationIndicator::fluid0)
+        {
+          // if phase = 1, no viscous dissipation
+          // if phase = 0, maximum viscous dissipation
+          viscous_dissipation_coefficient = 1. - phase_value_q;
+        }
+      else
```
voferreira, line 611: I know you have the throw on the parameter, but I would use "else if" here just in case.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -551,7 +549,115 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+  StabilizedMethodsCopyData &   copy_data)
+{
+  const unsigned int n_q_points = scratch_data.n_q_points;
+  const unsigned int n_dofs     = scratch_data.n_dofs;
+
+  const std::vector<double> &density   = scratch_data.density;
+  const std::vector<double> &viscosity = scratch_data.viscosity;
+
+  double viscous_dissipation_coefficient(0.);
+  //  const std::vector<double> &viscous_dissipation_coefficient =
```
voferreira: I think you forgot to erase this.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -551,7 +549,115 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+  // Time steps and inverse time steps which is used for stabilization
+  // constant
+  std::vector<double> time_steps_vector =
+    this->simulation_control->get_time_steps_vector();
+
+  // Copy data elements
+  auto &local_rhs = copy_data.local_rhs;
+
+
+  //  // Coefficient used to consider viscous dissipation in
```
voferreira: This part, as well.

voferreira (changes requested):
Very nice and clean. Maybe the only thing I would change is the name of the parameter to "viscous dissipative fluid", because I think it is more towards the options you have.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -551,7 +549,95 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+      else if (this->viscous_dissipation_indicator ==
+               Parameters::DissipationIndicator::both)
+        {
+          // maximum viscous dissipation everywhere
+          viscous_dissipation_coefficient = 1.;
+        }
+      else
+        {
+          throw(
+            std::runtime_error("Invalid viscous dissipation application. "
```
voferreira, line 620:
```suggestion
            std::runtime_error("Invalid viscous dissipative fluid. "
```

include/solvers/heat_transfer_assemblers.h
```diff
@@ -230,6 +230,45 @@ class HeatTransferAssemblerViscousDissipation
...
+   * @param scratch_data (see base class)
+   * @param copy_data (see base class)
+   */
+  virtual void
+  assemble_rhs(HeatTransferScratchData<dim> &scratch_data,
+               StabilizedMethodsCopyData &   copy_data) override;
+
+
+protected:
+  Parameters::DissipationIndicator viscous_dissipation_indicator;
```
voferreira, line 270: I think it would be clearer if the parameter had the same name as the variable, but it is really up to you. I feel like if someone else is modifying it, the person may want to do a ctrl+f and find everything about it.

source/core/parameters_multiphysics.cc
```diff
@@ -116,6 +122,18 @@ Parameters::VOF::parse_parameters(ParameterHandler &prm)
     peeling_wetting.parse_parameters(prm);
     surface_tension_force.parse_parameters(prm);
 
+    // DissipationIndicator
+    const std::string op = prm.get("viscous dissipative fluid");
```
voferreira, line 126: I'd use the full name of the parameter here, but not a big deal.

voferreira (approved):
The very last few comments. I think after addressing it, it is good to go :)

include/solvers/heat_transfer_assemblers.h
jeannejoachim: Yes you are right! It was a previous name of the parameter, I will change that

source/core/parameters_multiphysics.cc
jeannejoachim: I prefer to leave it with an "op" which is just used for the transition between the input string and the real parameter. It's the same machanism as for the "verbosity" parameter!

jeannejoachim:
Thanks to your suggestion @voferreira I have been able to see that it would be cleaner to generalize the use of this new parameter type (rename ``Parameters::FluidIndicator``). It went a bit over this PR scope though, but I thought I would do so while I was at it! I also renamed the older "free surface" tests with "vof" for homogeneity

voferreira:
Just checked it. It is better this way for sure. Consistent parameters make it easier for the user 👍

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -551,7 +549,94 @@ HeatTransferAssemblerViscousDissipation<dim>::assemble_rhs(
...
+        {
+          // maximum viscous dissipation everywhere
+          viscous_dissipation_coefficient = 1.;
+        }
+      else
+        {
+          throw(
+            std::runtime_error("Invalid viscous dissipative fluid. "
+                               "Options are 'fluid 0', 'fluid 1' or 'both'."));
+        }
```
blaisb, line 621: This cannot happen, because viscous_dissipative_fluid is an enum,but more defensive programming is always good.

blaisb (approved):
Good from my point of view! great work!

PR #528, opened 2022-07-12
Remove warnings in DEM and RPT

blaisb:
# Description of the problem

- Warnings to be cleaned inDEM and RPT
# Description of the solution

- Warnings were cleaned.

### DIFF
--- a/include/rpt/rpt_utilities.h
+++ b/include/rpt/rpt_utilities.h
@@ -41,7 +41,7 @@ using namespace dealii;
  */
 template <int dim>
 void
-attach_grid_to_triangulation(Triangulation<dim>              &triangulation,
+attach_grid_to_triangulation(Triangulation<dim> &             triangulation,
                              const Parameters::RPTParameters &parameters,
                              const unsigned int               n_refinement)
 {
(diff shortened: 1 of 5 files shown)


shahabgol (approved):

include/rpt/rpt_utilities.h
```diff
@@ -51,7 +51,8 @@ attach_grid_to_triangulation(Triangulation<dim> &             triangulation,
                                      5,
                                      parameters.reactor_radius,
                                      parameters.reactor_height / 2.);
-  GridTools::rotate(M_PI_2, 1, triangulation);
+  Tensor<1, dim> axis({0, 1, 0});
```
tonielgeitani, line 54: I don't understand why here you defined axis (0,1,0) while the other files you defined it as axis(0,0,0) and the changed axis[rotation_axis] to 1. What is the difference between the two cases?

tonielgeitani (approved):

include/rpt/rpt_utilities.h
blaisb: In this case this rotation is hardcoded to be in this directoin. In the other one it's with a user paraemter. This is a code we need to make more general in the future :)

PR #531, opened 2022-07-14
Optimize generate cut cells

voferreira:
# Description of the problem

- Cut cell mapping was not optimized, since all cells searched for all particles.

# Description of the solution

- Map the potential cut/inside particles cells in cascade;
- Propagate the cascade along the cells' levels;
- Add a parameter to control which cut cells mapping function is called;

# How Has This Been Tested?

- List and describe the test(s) for the added code (if there is no test for the pull request, explain why (i.e. the test already exists but was changed))

- Optimized static stokes case: used to verify if the results of the optimized function are the same as the regular cut cells function.

# Future changes

- Making it work for other shapes would be necessary. Currently, the optimized cut cells mapping works only for spherical particles;
- Eventually, both load_cells_from_file and checkpoint write and read functions will need to be fixed to use with other particle shapes as well.

# Comment

- [x] Fix check_point test
- [x] Fix debug tests


blaisb:
@voferreira Can you please rebase one lat time so I can see the warnings checkup?

voferreira:
@blaisb Done!

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -303,14 +306,40 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+   *cut, and the particle id of the particle that cut it.
+   */
+  void
+  generate_cut_cells_map();
+
   /**
    * @brief
    *This function create a map (cut_cells_map) that indicates if a cell is cut,
    *and the particle id of the particle that cut it.
+   *The algorithm was optimized to reduce the cost of this step.
```
Luckabarbeau, line 330: I would specify that it only applied to spheres which is why we need also the old function.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3169,20 +3441,22 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
           particles[p_i].orientation[2] = restart_data["orientation_z"][p_i];
           // Initialize shape of particles according to the type parameter in
           // the file
-          if (restart_data["type"][p_i] == 0)
+          if (restart_data["type"][p_i] == Shape<dim>::ShapeType::sphere)
```
Luckabarbeau, line 3444: Nice !

Luckabarbeau (approved):
Very nice work! this should help a lot when simulating a large number of particles.

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -303,14 +306,40 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
    */
   void
-  generate_cut_cells_map();
+  optimized_generate_cut_cells_map();
 
+  /**
+   * @brief
+   *This function abstracts the generation of cell candidates that possibly are
```
oguevremont, line 337:
```suggestion
   *This function abstracts the generation of cell candidates that possibly are cut
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -148,6 +166,252 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
 }
 
 
+template <int dim>
+void
+GLSSharpNavierStokesSolver<dim>::optimized_generate_cut_cells_map()
+{
+  TimerOutput::Scope t(this->computing_timer, "optimized_cut_cells_mapping");
+  MappingQ1<dim>     mapping;
+  unsigned int       max_childs = GeometryInfo<dim>::max_children_per_cell;
```
oguevremont, line 175: In a few places there is 'max_childs', which I would replace by 'max_children'

oguevremont (approved):
Very nice PR! Good job!

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1362,61 +1628,61 @@ GLSSharpNavierStokesSolver<dim>::integrate_particles()
           // domain. If this is the case, the particle won't be updated.
           IBParticle<dim> save_particle_state         = particles[p];
           bool            save_particle_state_is_used = false;
-          try
-            {
-              // Define the correction vector.
-              Tensor<1, 3> dv = residual_velocity * 1 / d_residual_dv;
+          // Define the correction vector.
+          Tensor<1, 3> dv = residual_velocity * 1 / d_residual_dv;
```
blaisb, line 1632: Are you sure that d_residual_dv is always going to be non-zero?

blaisb (changes requested):
Some minor changes. Very very good work.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -148,6 +166,252 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
+    {
+      if (cell->is_locally_owned() || cell->is_ghost())
+        {
+          cut_cells_map[cell]    = {false, 0};
+          cells_inside_map[cell] = {false, 0};
+        }
+    }
+
+  // Loop over particles
+  for (int p = particles.size() - 1; p >= 0; --p)
```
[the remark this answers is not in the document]
voferreira: To guarantee that the particle in the map is the one with the lowest ID. For instance, if you have two particles cutting the same cell, it guarantees that the lowest ID particle is associated with the cell in the map.

source/fem-dem/gls_sharp_navier_stokes.cc
voferreira: I added DBL_MIN to it.

source/fem-dem/gls_sharp_navier_stokes.cc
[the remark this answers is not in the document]
blaisb: Add a comment to explain that exactly :)

source/core/lethegridtools.cc
```diff
@@ -48,43 +48,54 @@ LetheGridTools::find_cell_around_point_with_tree(
...
+              const double dist =
+                GeometryInfo<dim>::distance_to_unit_cell(p_cell);
 
-          if (dist < 1e-12)
-            {
-              // cell on lvl 0 found
-              cell_on_level_0_found = true;
-              best_cell_iter        = cell;
-              break;
+              if (dist < 1e-12)
```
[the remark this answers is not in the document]
voferreira: What do you suggest for the distance? I did not change this condition in this PR so I did not think about it before, but I think we could standardize it by putting a factor on the size of the cell, maybe. I am just not sure if it is necessary. What do you think @Luckabarbeau ?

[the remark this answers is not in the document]
blaisb: I would at least maybe open an issue. It can be done outside of this PR, but I feel this might eventually bite us back.

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -497,6 +532,8 @@ Return a bool that describes  if a cell contains a specific point
    */
 
 private:
+  bool all_spheres = true;
```
blaisb, line 535:
```suggestion
  bool all_spheres;
```

You fixed the default value in the constructor, so you don't need to put a value here. It's not a good idea to have the values in the .h.

blaisb (changes requested):

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -148,6 +166,252 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
+        {
+          // Assign vertex to manifold superpoint
+          manifold_points[i] = face_iter->vertex(i);
+        }
+
+      // Create array of points on face
+      auto surrounding_face_points =
+        make_array_view(manifold_points.begin(), manifold_points.end());
+
+      // Project particles' position on face
```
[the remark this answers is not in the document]
voferreira: What about now?

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -148,6 +166,252 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
+    {
+      cell_is_inside = true;
+      cell_is_cut    = true;
+      return {cell_is_inside, cell_is_cut};
+    }
+
+  // Initialize superpoint of manifold
+  std::vector<Point<dim>> manifold_points(
+    GeometryInfo<dim - 1>::vertices_per_cell);
+
```
voferreira, line 365:
```suggestion

  // The last check consists of projecting the particle's position on the cells'
  // face and checking whether the projected points are within the particle. If
  // one of the projected points is inside the particle, either the cell is
  // inside or is cut by the particles. If this is true, the next
  // level will be tested again by this function.

```

source/core/lethegridtools.cc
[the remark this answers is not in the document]
voferreira: Just talked to Lucka about it. This distance is dimension-independent.

source/fem-dem/gls_sharp_navier_stokes.cc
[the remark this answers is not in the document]
blaisb: Looks good to me.

blaisb (approved):
Ready for merge once CI passes.

PR #535, opened 2022-07-21
Implement time varying source term + Update doc

jeannejoachim:
# Description of the problem

Source term for cfd did not vary in time

# Description of the solution

Add the feature for both gls and gd solvers. It was already implemented in other solvers.

# How Has This Been Tested?

Locally

# Documentation

Updated and refactored the Source Term documentation page


blaisb (approved):
Good for me. Nice catch, this really was a missing thing.

PR #537, opened 2022-07-22
Vof v3

jeannejoachim:
# Description of the problem

peeling/wetting mechanism was not robust, so it did not work well with adaptative mesh.

# Description of the solution

Change pw mechanism :
 - use of pressure gradient
 - add method to calculate the pressure average of the monitored fluid
 - simplify with a globalized approach: from 5 parameters to 1 and update tests

Other changes:
 - change multiphysics order of resolution so that VOF is solved last
 - remove error_quadrature parameter (change for cell_quadrature) and update tests accordingly

# How Has This Been Tested?

- [x] Bouncing drop : tested locally, too complex to be made a test, but can be an example in the future
- [x] Dam break with noslip condition
- [x] Drop with adaptative mesh
- [x] Update VOF tests

# Documentation

- [x] VOF

# Future changes

- [x] Rebase on master

# Comments


doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -146,36 +142,30 @@ The default values of the VOF parameters are given in the text box below.
...
-      * if the ``fluid 0`` has a higher density than ``fluid 1``, and ``set wetting phase distance = 0.1``, the wetting can occur where the phase value is above :math:`= 0.6`.
-
-    * ``diffusivity``: value of the diffusivity (diffusion coefficient) in the transport equation of the phase fraction. Default value is 0 to have pure advection. This can be used to :ref:`improve wetting`.
+    * ``diffusivity``: value of the diffusivity (diffusion coefficient) in the transport equation of the phase fraction. Default value is ``0`` to have pure advection. Increase ``diffusivity`` to :ref:`improve wetting`.
 
 .. warning::
 
-  As peeling/wetting mechanisms result in fluid creation and disparition, is it highly advised to monitor the mass conservation of the fluid of interest (``subsection mass conservation``) and to change the type of sharpening threshold to adaptative (``subsection sharpening``).
+  As peeling/wetting mechanisms result in fluid creation and removal, is it highly advised to monitor the mass conservation of the fluid of interest (``subsection mass conservation``) and to change the type of sharpening threshold to adaptative (``subsection sharpening``).
```
shahabgol, line 168:
```suggestion
  As peeling/wetting mechanisms result in fluid generation and loss, is it highly advised to monitor the mass conservation of the fluid of interest (``subsection mass conservation``) and to change the type of sharpening threshold to adaptative (``subsection sharpening``).
```

include/solvers/heat_transfer.h
```diff
@@ -75,7 +75,6 @@ class HeatTransfer : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
         cell_quadrature = std::make_shared<QGaussSimplex<dim>>(fe->degree + 1);
         face_quadrature =
           std::make_shared<QGaussSimplex<dim - 1>>(fe->degree + 1);
-        error_quadrature = std::make_shared<QGaussSimplex<dim>>(fe->degree + 2);
```
shahabgol, line 78: Why do you remove the error_quadrature?

source/solvers/vof.cc
```diff
@@ -409,6 +409,105 @@ VolumeOfFluid<dim>::calculate_volume_and_mass(
     Utilities::MPI::sum(this->mass_monitored, mpi_communicator);
 }
 
+template <int dim>
+template <typename VectorType>
+void
+VolumeOfFluid<dim>::calculate_pressure_monitored_avg(
```
shahabgol, line 415: Isn't find_monitored_fluid_avg_pressure a better name?

source/solvers/vof.cc
```diff
@@ -409,6 +409,105 @@ VolumeOfFluid<dim>::calculate_volume_and_mass(
...
+
+          for (unsigned int q = 0; q < n_q_points; q++)
+            {
+              // Gather minimum and maximum pressure on the phase of interest
+              if ((monitored_fluid == Parameters::FluidIndicator::fluid0 &&
+                   local_phase_values[q] < 0.5) ||
+                  (monitored_fluid == Parameters::FluidIndicator::fluid1 &&
+                   local_phase_values[q] > 0.5))
+                {
+                  this->pressure_monitored_avg += local_pressure_values[q];
```
shahabgol, line 471: I saw you have defined this parameter in the header file, but I didn't see where you initialize it.

source/solvers/vof.cc
```diff
@@ -2062,21 +2168,21 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
                         {
-                          if (pressure_gradients[q][d] < peeling_grad_p)
-                            {
-                              nb_pressure_grad_meet_peel_condition += 1;
-                            }
+                          double pressure_gradient_q_d =
+                            pressure_gradients[q][d] *
+                            fe_face_values_fd.normal_vector(q)[d];
+
+                          if (pressure_gradient_q_d < -1e-5)
```
shahabgol, line 2182: If you want to use a small negative value, I think using -DBL_MIN is a better choice. If you need this 1e-5 value, I think it's better to keep it as a variable as use the variable elsewhere.

source/solvers/vof.cc
```diff
@@ -2062,21 +2168,21 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
+                            nb_pressure_grad_meet_peel_condition += 1;
+                          else if (pressure_gradient_q_d > 1e-5)
```
shahabgol, line 2184: The same here.

source/solvers/vof.cc
```diff
@@ -2095,20 +2201,49 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
                 round(phase_lighter_fluid_q / n_q_points_face);
               double phase_values_cell =
                 phase_values_q / static_cast<double>(n_q_points_face);
+              double pressure_values_cell =
+                pressure_values_q / static_cast<double>(n_q_points_face);
+
+              // Check peeling condition on the pressure gradient
+              if (pressure_values_cell < this->pressure_monitored_avg and
```
shahabgol, line 2208:
```suggestion
              if (pressure_values_cell < this->pressure_monitored_avg &&
```

source/solvers/vof.cc
```diff
@@ -2095,20 +2201,49 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
+                  nb_pressure_grad_meet_peel_condition > n_q_points_face / 2)
```
shahabgol, line 2209:
```suggestion
                  nb_pressure_grad_meet_peel_condition > n_q_points_face * 0.5)
```

source/solvers/vof.cc
```diff
@@ -2095,20 +2201,49 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
+                {
+                  // Peel the higher density fluid
+                  // conditions phase_values_cell < 1 or phase_values_cell
+                  // > 0 prevent from treating values outside of the range
+                  // [0,1]
+                  if (phase_denser_fluid_cell == 1 and phase_values_cell > 0.5)
+                    {
+                      // Progressive phase value change
+                      double new_phase =
+                        std::max(phase_values_cell - 0.25,
```
shahabgol, line 2219: What is this 0.25?

source/solvers/vof.cc
```diff
@@ -2095,20 +2201,49 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
+
+              // Check peeling condition on the pressure gradient
+              if (pressure_values_cell < this->pressure_monitored_avg and
+                  nb_pressure_grad_meet_peel_condition > n_q_points_face / 2)
+                {
+                  // Peel the higher density fluid
+                  // conditions phase_values_cell < 1 or phase_values_cell
+                  // > 0 prevent from treating values outside of the range
+                  // [0,1]
+                  if (phase_denser_fluid_cell == 1 and phase_values_cell > 0.5)
```
shahabgol, line 2215:
```suggestion
                  if (phase_denser_fluid_cell == 1 && phase_values_cell > 0.5)
```

source/solvers/vof.cc
```diff
@@ -2095,20 +2201,49 @@ VolumeOfFluid<dim>::apply_peeling_wetting(const unsigned int i_bc,
...
-              if (pressure_values_q > wetting_p_value)
+                      change_cell_phase(PhaseChange::peeling,
+                                        new_phase,
+                                        solution_pw,
+                                        dof_indices_vof);
+                    }
+                }
+              else if (pressure_values_cell > this->pressure_monitored_avg and
+                       nb_pressure_grad_meet_wet_condition >
+                         n_q_points_face / 2)
```
shahabgol, line 2242:
```suggestion
                         n_q_points_face * 0.5)
```

shahabgol (commented):
Very good work, I only have a few comments/questions.

source/solvers/vof.cc
jeannejoachim: I do not have particular attachment to `&&` vs. `and`, but I want it to be consistent, at least in this solver. On other parts of the code, Bruno asked me to change to `and` so I changed it here also. Do you want me to revert back to `&&` everywhere?

source/solvers/vof.cc
jeannejoachim: Oh yeah good one!

source/solvers/vof.cc
jeannejoachim: this number is a bit arbitrary, I use it to change progressively the phase value rather than go from `1` to `0` directly. Would you prefer it to be a parameter?

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -146,36 +142,30 @@ The default values of the VOF parameters are given in the text box below.
...
+    The cell is then filled with the lower density fluid by changing its phase value progressively.
 
-  * Wetting of the lower density fluid occurs where those conditions are met: 
+  * Wetting occurs in a cell where those conditions are met: 
 
     * the cell is in the domain of the lower density fluid,
-    * the cell pressure value is above ``wetting pressure value``, and
-    * the distance (on the phase value) to the interface is above ``wetting phase distance``.
+    * the cell pressure value is above the average pressure on the ``monitored fluid`` (``fluid 1`` by default, see ``subsection mass conservation``), and
+    * the pressure gradient is positive for more than half of the quadrature points.
```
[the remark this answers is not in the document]
jeannejoachim: I do not have a reference, this is something thought and tried by us, to have a more general approach than what we saw in the literature. I hope we will have an article as a reference in the future :)

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -146,36 +142,30 @@ The default values of the VOF parameters are given in the text box below.
...
           -number of peeled cells: 1
 
-  * Peeling of the higher density fluid occurs where those conditions are met:
+  * Peeling occurs in a cell where those conditions are met:
 
     * the cell is in the domain of the higher density fluid,
-    * the cell pressure value is below ``peeling pressure value``, and
-    * more than half of the quadrature points in the cell have a pressure gradient below ``peeling pressure gradient``.
+    * the cell pressure value is below the average pressure on the ``monitored fluid`` (``fluid 1`` by default, see ``subsection mass conservation``), and
+    * the pressure gradient is negative for more than half of the quadrature points.
```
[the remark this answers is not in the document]
jeannejoachim: For both these conditions, it is more something that we try on, out of physical sense and reflections. None of the free surface methods I have seen in the literature were general enough and robust enough for us to use on an industrial case, so we came up with this (it's the 3rd version of the peeling/wetting mechanism), and it gives satisfactory results for now

source/solvers/vof.cc
jeannejoachim: It is in the same function, just before the loop on `cell_vof`. Do you think it should be elsewhere?

applications_tests/gls_navier_stokes_2d/gls_vof_wetting_mesh-adapt.prm
```diff
@@ -0,0 +1,186 @@
...
+# Simulation and IO Control
+#---------------------------------------------------
+subsection simulation control
+  set method                  = bdf1
+  set number mesh adapt       = 0
+  set time end                = 0.1
+  set time step               = 0.02
+  set output name             = drop_wetting_adapt_VOF
+  set output frequency        = 0
+
```
blaisb, line 13:
```suggestion
```

applications_tests/gls_navier_stokes_2d/gls_vof_wetting_mesh-adapt.prm
```diff
@@ -0,0 +1,186 @@
...
+subsection boundary conditions VOF
+  set number                  = 4
+    subsection bc 2 #lower wall
+	set id = 2
+        set type              = peeling/wetting
+    end
+    subsection bc 3 #upper wall
+	set id = 3
+        set type              = peeling/wetting
+    end
```
blaisb, line 136: This is weird. If you only have two active bcs you should set number  = 2 and not 4... this does not follow the regular syntax. We only specify the BCs we need.

doc/source/parameters/cfd/volume_of_fluid.rst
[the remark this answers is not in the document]
blaisb: I would add a bit of explanation that explains that this is an heuristic. Just to give more context to the user.

include/solvers/vof.h
```diff
@@ -547,6 +565,7 @@ class VolumeOfFluid
   calculate_mass_deviation(const Parameters::FluidIndicator monitored_fluid,
                            const double sharpening_threshold);
 
+
```
blaisb, line 568:
```suggestion
```

include/solvers/vof.h
```diff
@@ -719,6 +737,10 @@ class VolumeOfFluid
   double       mass_first_iteration;
   double       sharpening_threshold;
 
+  double pressure_min;
+  double pressure_max;
+  double pressure_monitored_avg;
```
blaisb, line 742: Maybe put this into a small struct. PressureMonitoring or something similar. i would rather have as little floating doubles lying around.

source/solvers/vof.cc
blaisb: well I would give it an initial value in the constructor of the class just to be sure, but still that's ok.
This brings the question though. Does this value really needs to be a member of the class vof or could it not just be a return argument of the function? It's best not to store things as class members when it can be avoided. If you just use it somewhere else, better return it and not alter the state.

source/solvers/vof.cc
```diff
@@ -409,6 +409,105 @@ VolumeOfFluid<dim>::calculate_volume_and_mass(
...
+                   local_phase_values[q] > 0.5))
+                {
+                  this->pressure_monitored_avg += local_pressure_values[q];
+                  nb_values++;
+                }
+            }
+        }
+    }
+
+  this->pressure_monitored_avg =
```
blaisb, line 478: Why not return it and not store it as a class member?

source/solvers/vof.cc
blaisb: let's revert back to &&. "and" is not portable whereas && is.

source/solvers/vof.cc
blaisb: Let's not add parameter. This mechanism is still very project specific anyway...

blaisb (changes requested):
Just a few comments there and there. I would avoid making the pressure things as class members and instead would use return arguments for the functions.

applications_tests/gls_navier_stokes_2d/gls_vof_wetting_mesh-adapt.prm
jeannejoachim: Yes you are right, this was from previous testing on other BC as well, I will change that

source/solvers/vof.cc
jeannejoachim: You are right, it will be cleaner to just return it, I will change that

source/solvers/vof_assemblers.cc
```diff
@@ -166,7 +168,9 @@ VOFAssemblerCore<dim>::assemble_rhs(VOFScratchData<dim> &      scratch_data,
   const auto method = this->simulation_control->get_assembly_method();
 
   // Add a small diffusivity, used in the context of the wetting mechanism
-  const double diffusivity = this->vof_parameters.peeling_wetting.diffusivity;
+  double diffusivity = 0.;
+  if (this->vof_parameters.peeling_wetting.enable)
+    diffusivity = this->vof_parameters.peeling_wetting.diffusivity;
```
blaisb, line 173: Are you sure you want to enable diffusivity only for peeling and wetting? it could be useful for other stuff. Then you could put the default value to 0 and it would have no effect

blaisb (changes requested):
1-2 minor things to address. Can you rebase also one last time, 2 tests were added with VOF and the results of them might change with this PR. You can replace their output with the new ones

jeannejoachim:
@blaisb I rebased again on master and changed the new test output! If CI passes this should be ready to merge

blaisb:
Will merge as soon as CI passes

applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat_mesh-adapt.output
```diff
@@ -28,36 +28,36 @@ Transient iteration : 3        Time : 0.06     Time step : 0.02     CFL : 2.1861
...
-0.0200 3.071207e-05   0.0682         
-0.0400 4.143914e-06   0.0668         
-0.0600 3.312462e-06   0.0669         
-0.0800 2.763695e-06   0.0669         
-0.1000 3.568241e-06   0.0669         
+0.0200 3.071207e-05   0.9287         
+0.0400 4.143914e-06   0.9285         
+0.0600 3.312462e-06   0.9283         
+0.0800 3.571638e-06   0.9284         
+0.1000 2.763693e-06   0.9283
```
blaisb, line 57: There is something wrong here. The pressure analytical solution has been redefined in the new version of Lethe so that it gives the right solution. i would watch that out here. You can compare with the regular hydrostat prm file for the right analytical solution.

blaisb (changes requested):
Can you double check the analytical solution for the hydrostat problem with mesh adaptation? it seems like the prm you are using reverted back to the old analytical solution. The new one is the correct one.

blaisb (approved):

PR #540, opened 2022-07-28
Fix help text of some CMake options

paulapatience:
* CMakeLists.txt (PROTOTYPES, MARCH_NATIVE): Fix help text.


blaisb (approved):
:)!
Great catch

PR #544, opened 2022-08-01
use the contact search radius

Luckabarbeau:
# Description of the problem

- the contact search radius for particle-particle contact and particle-wall contact was not the same

# Description of the solution

- use the same factor

# How Has This Been Tested?

- no test needed


shahabgol (approved):

blaisb (approved):
Good catch my man. Just fix clang format and good to go

PR #548, opened 2022-08-03
Corrected 2 bug that were affecting the convergence in large case

Luckabarbeau:
# Description of the problem

-The pressure reference point was imposed inside the particle even when we were not solving NS inside the particle. This affected the non-linear convergence significantly for cases where a large number of particles are in motion.
-An error in the update candidate criteria was preventing the update of the contact candidates.

# Description of the solution

- We no longer impose a reference pressure point inside the particle when the NS equation is not solved inside the particle.
- We corrected the criteria for the update of the contact candidates.

# How Has This Been Tested?

- Multiple tests are affected by the removal of the reference pressure inside the particle. This change is due to the change of the solution inside the particle that affects the non-linear convergence of the solution outside the particle. All perturbations are minimal and smaller than the non-linear tolerance of the solver.
- The test that used the reference pressure point of the particle to define a pressure reference point outside of the particle was modified to remove this strategy.


applications_tests/gls_sharp_navier_stokes_3d/moving_stokes.prm
```diff
@@ -133,7 +133,8 @@ set time dependent =true
     end
     subsection bc 5
 	set id = 5
-        set type              = function
+        set type              = function weak
```
CaroleAnneDaunais, line 136: Can you tell me why you made that change? Same question for the other .prm

CaroleAnneDaunais (approved):
Neat, good catch!

applications_tests/gls_sharp_navier_stokes_3d/moving_stokes.prm
Luckabarbeau: This case has Dirichlet boundary conditions everywhere. This means that there is no reference pressure. we were using the pressure reference point of the particle outside the particle to define a reference pressure. We cannot do this anymore (and it was a bad use of the pressure reference point). Instead, we now use the weak Dirichlet boundary conditions which still act as a pressure reference point.

oguevremont (approved):
Very nice!

blaisb (approved):
Good job, ready for merger

applications_tests/gls_sharp_navier_stokes_3d/moving_stokes.prm
blaisb: Seems like the weak dirichlet BC are getting more and more useful...

PR #550, opened 2022-08-04
Add invitation to cite deal.II when citing Lethe

blaisb:
# Description of the problem

- Our referencing Lethe section discusses how to cite Lethe, but it omits to encourage the citation of the deal.II library. Without deal.II this code does not exist, so it should be natural that every citation of Lethe comes with a citation of deal.II

# Description of the solution

- Added an invitation to cite deal.II as well as the motivation for it. Including a link to the "citing deal.II" page"

# Comments

- I think it's important to remain grateful to all these things we are inheriting from deal.II and using without even really thinking about it :).

### DIFF
--- a/doc/source/referencing.rst
+++ b/doc/source/referencing.rst
@@ -2,7 +2,7 @@
 Referencing Lethe
 ###################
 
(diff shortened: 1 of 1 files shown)


CaroleAnneDaunais (approved):
Good point Bruno, will do in my article :)

lpsaavedra (approved):

PR #552, opened 2022-08-05
Enhancements vans

tonielgeitani:
# Description of the problem

-New void fraction scheme was not reflecting good values at boundaries.
-Some drag models had issues with their units
-Added a feature to print particle velocity at every time step for the sedimentation test case and output it in a file.

# Description of the solution

- Removed the need for special calculation at the boundaries for the new void fraction scheme and added the quadrature weights in the calculation of the void fraction. This way every quadrature point will have a different contribution to the cell void fraction depending on its relative weight.
- Fixed the units for the drag models and used formulations that are presented in article for the momentum exchange coefficient

# How Has This Been Tested?

- For the void fraction it has been tested by comparing different void fraction schemes for the spouted bed and finding that treating the boundaries indifferently led to the most realistic void fraction scheme and mass conservation.
- For the drag models, the spouted have been simulated with these models and the first bubble formed has been examined. Previously, the bubble was very small and incomparable to the article experiment and simulations. Now its form is completely comparable.

# Documentation

-N/A

# Future changes

-N/A

# Comments

-N/A

### DIFF
--- a/include/fem-dem/cfd_dem_coupling.h
+++ b/include/fem-dem/cfd_dem_coupling.h
@@ -319,6 +319,7 @@ private:
   DoFHandler<dim> background_dh;
   PVDHandler      grid_pvdhandler;
   PVDHandler      particles_pvdhandler;
+  TableHandler    particle_velocity_table;
 
   DEMSolverParameters<dim>      dem_parameters;
   double                        dem_time_step;
--- a/include/fem-dem/cfd_dem_coupling.h
+++ b/include/fem-dem/cfd_dem_coupling.h
@@ -319,7 +319,6 @@ private:
   DoFHandler<dim> background_dh;
   PVDHandler      grid_pvdhandler;
   PVDHandler      particles_pvdhandler;
-  TableHandler    particle_velocity_table;
 
   DEMSolverParameters<dim>      dem_parameters;
   double                        dem_time_step;
--- a/source/fem-dem/cfd_dem_coupling.cc
+++ b/source/fem-dem/cfd_dem_coupling.cc
@@ -1267,22 +1267,7 @@ CFDDEMSolver<dim>::print_particles_summary()
         << "--------------------------------------------------------------------------"
         << "--------------------------------------------------------------------------"
         << std::endl;
-
-      particle_velocity_table.add_value(
-        "time", this->simulation_control->get_current_time());
-      particle_velocity_table.add_value(
-        "particle-velocity", particle_properties[DEM::PropertiesIndex::v_y]);
     }
-
-  {
-    std::string filename =
-      this->simulation_parameters.simulation_control.output_folder +
-      "particle_velocity" + ".dat";
-    std::ofstream output(filename.c_str());
-    particle_velocity_table.set_precision("time", 12);
-    particle_velocity_table.set_precision("particle-velocity", 12);
-    particle_velocity_table.write_text(output);
-  }
 }
 
 template <int dim>
--- a/source/fem-dem/cfd_dem_coupling.cc
+++ b/source/fem-dem/cfd_dem_coupling.cc
@@ -1267,7 +1267,22 @@ CFDDEMSolver<dim>::print_particles_summary()
         << "--------------------------------------------------------------------------"
         << "--------------------------------------------------------------------------"
         << std::endl;
+
+      particle_velocity_table.add_value(
+        "time", this->simulation_control->get_current_time());
+      particle_velocity_table.add_value(
+        "particle-velocity", particle_properties[DEM::PropertiesIndex::v_y]);
     }
+
+  {
+    std::string filename =
+      this->simulation_parameters.simulation_control.output_folder +
+      "particle_velocity" + ".dat";
+    std::ofstream output(filename.c_str());
+    particle_velocity_table.set_precision("time", 12);
+    particle_velocity_table.set_precision("particle-velocity", 12);
+    particle_velocity_table.write_text(output);
+  }
 }
 
 template <int dim>
--- a/source/fem-dem/gls_vans.cc
+++ b/source/fem-dem/gls_vans.cc
@@ -565,77 +565,50 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
                     R_sphere =
                       (pow(3 * active_neighbors[n]->measure() / (4 * M_PI),
                            1. / 3.));
-                  if (((!cell->at_boundary() || !cell->has_boundary_lines()) &&
-                       (!active_neighbors[n]->at_boundary() ||
-                        !active_neighbors[n]->has_boundary_lines())) ||
-                      ((cell->at_boundary() || cell->has_boundary_lines()) &&
-                       (!active_neighbors[n]->at_boundary() ||
-                        !active_neighbors[n]->has_boundary_lines())))
-                    {
-                      // Loop over quadrature points
-                      for (unsigned int k = 0; k < n_q_points; ++k)
-                        {
-                          // Distance between particle and quadrature
-                          // point
-                          double neighbor_distance =
-                            particle.get_location().distance(
-                              neighbor_quadrature_point_location[n][k]);
+                  {
+                    // Loop over quadrature points
+                    for (unsigned int k = 0; k < n_q_points; ++k)
+                      {
+                        // Distance between particle and quadrature
+                        // point
+                        double neighbor_distance =
+                          particle.get_location().distance(
+                            neighbor_quadrature_point_location[n][k]);
 
-                          // Particle completely in reference sphere
-                          if (neighbor_distance <= (R_sphere - r_particle))
-                            {
+                        // Particle completely in reference sphere
+                        if (neighbor_distance <= (R_sphere - r_particle))
+                          {
+                            particle_properties
+                              [DEM::PropertiesIndex::volumetric_contribution] +=
+                              M_PI *
+                              pow(particle_properties[DEM::PropertiesIndex::dp],
+                                  dim) /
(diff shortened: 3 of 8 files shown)


source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1267,7 +1267,22 @@ CFDDEMSolver<dim>::print_particles_summary()
         << "--------------------------------------------------------------------------"
         << "--------------------------------------------------------------------------"
         << std::endl;
+
+      particle_velocity_table.add_value(
```
voferreira, line 1271: I would use a separate member and use a parameter to do this. I think it would be cleaner this way :)

voferreira (changes requested):
Nice catch on the Drag models! Good job on cleaning the void fraction part as well. I still want to double-check the formulation and I must finish it tomorrow.

include/fem-dem/cfd_dem_coupling.h
```diff
@@ -319,6 +319,7 @@ class CFDDEMSolver : public GLSVANSSolver<dim>
   DoFHandler<dim> background_dh;
   PVDHandler      grid_pvdhandler;
   PVDHandler      particles_pvdhandler;
+  TableHandler    particle_velocity_table;
```
blaisb, line 322: ifyou are going to print all the particle velocity, will not print everything related to the particles? Shahab has already made a post-processor for this. it's called print xyz or something like this and it is already used in the DEM. No need to make a new one :)

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1267,7 +1267,22 @@ CFDDEMSolver<dim>::print_particles_summary()
...
     }
+
+  {
+    std::string filename =
+      this->simulation_parameters.simulation_control.output_folder +
+      "particle_velocity" + ".dat";
+    std::ofstream output(filename.c_str());
+    particle_velocity_table.set_precision("time", 12);
+    particle_velocity_table.set_precision("particle-velocity", 12);
+    particle_velocity_table.write_text(output);
```
blaisb, line 1284: Ohh I see, you wanted to write it to a file?  Then I guess that's ok, but I would add more than the y component of the velocity.
I am not sure about adding another parameter. I'm wondering what is the use of this file other than for a few comparison? Because you can get the particles velocity using the pvd and the vtu files, so what's the point of having them in a text file?

source/fem-dem/gls_vans.cc
```diff
@@ -565,77 +565,50 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
-                          // Particle completely in cell (PCM)
-                          particle_properties
-                            [DEM::PropertiesIndex::volumetric_contribution] +=
-                            M_PI *
-                            pow(particle_properties[DEM::PropertiesIndex::dp],
-                                dim) /
-                            (2.0 * dim);
-                        }
-                    }
+                            else if (dim == 3)
```
blaisb, line 603: Wow this is much cleaner now.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1093,21 +1093,21 @@ GLSVansAssemblerKochHill<dim>::calculate_particle_fluid_interactions(
...
+                8.16 * pow(1 - cell_void_fraction, 3));
         }
       else if ((1 - cell_void_fraction) >= 0.4)
         {
           f0 = 10 * (1 - cell_void_fraction) / pow(cell_void_fraction, 3);
         }
 
       double f3 = 0.0673 + 0.212 * (1 - cell_void_fraction) +
-                  0.0232 / pow(cell_void_fraction, 5);
+                  0.0232 * pow(cell_void_fraction, -5);
```
blaisb, line 1104: why this change?

source/fem-dem/vans_assemblers.cc
```diff
@@ -1310,8 +1306,6 @@ GLSVansAssemblerGidaspow<dim>::calculate_particle_fluid_interactions(
 
       particle_number += 1;
     }
-
-  beta_drag = beta_drag / scratch_data.cell_volume;
```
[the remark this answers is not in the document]
blaisb: I'm asking myself the same question here.

blaisb (changes requested):
First batch of comments, but I will need to relook at the drag expressions. Do you think you could put the reference for the formulas in the documentation so at least we all agree we are reading from the same source? :)

source/fem-dem/cfd_dem_coupling.cc
voferreira: I agree. There is no reason to have it in a separate file. If you want, I can help you to do it in the postprocessing using pyvista.

source/fem-dem/vans_assemblers.cc
voferreira: I think I would keep the division here, just because it is easier to compare with the article.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1273,35 +1272,32 @@ GLSVansAssemblerGidaspow<dim>::calculate_particle_fluid_interactions(
           C_d = 0.44;
         }
 
-      if (cell_void_fraction >= 0.8)
+      if (cell_void_fraction > 0.8)
```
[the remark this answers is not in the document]
tonielgeitani: Yes this is a typo. I removed it by mistake.

source/fem-dem/vans_assemblers.cc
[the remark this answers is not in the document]
tonielgeitani: The term of the beta drag in all the drag models should be 1/s to be equal to the units in the VANS equation. Before the momentum_transfer_coefficient unit was m^3/s so we had to divide by the element volume to get 1/s. In the Gidaspow model, the momentum_transfer_coefficient unit is already 1/s . So dividing beta by the element volume would result in the beta unit to be 1/(m^3.s) and this was leading to a very very small initial bubble in the fluidized bed. It is just a matter of units

source/fem-dem/vans_assemblers.cc
```diff
@@ -1156,8 +1156,7 @@ GLSVansAssemblerBeetstra<dim>::calculate_particle_fluid_interactions(
   Assert(
     !scratch_data.properties_manager.is_non_newtonian(),
     RequiresConstantViscosity(
-      "GLSVansAssemblerBeetstra<dim>::calculate_particle_fluid_interactions"));
-  const double viscosity = scratch_data.properties_manager.viscosity_scale;
+      "GLSVansAssemblerDiFelice<dim>::calculate_particle_fluid_interactions"));
```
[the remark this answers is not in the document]
tonielgeitani: Done

source/fem-dem/cfd_dem_coupling.cc
tonielgeitani: True I only had it for the comparison of the sedimentation velocity of different void fraction models. This can be removed entirely. its useless except in this case. I will remove it then.

include/fem-dem/cfd_dem_coupling.h
tonielgeitani: I will remove it. It was just for one single case

source/fem-dem/cfd_dem_coupling.cc
tonielgeitani: I will remove it. It was only needed for one comparison case

source/fem-dem/vans_assemblers.cc
tonielgeitani: Done

tonielgeitani:
I removed the table and file creation for particle velocity as it was only needed for a single comparison case.  These are the equations for the drag:  [Beestra](https://www.sciencedirect.com/science/article/pii/S0301932220305346 ) eq. 7 and 8. They reference the main Beestra paper but they write the drag for in a way that can be written as F=Beta(u-v). All the other equations remained the same as the references from the documentation of the parameters of Unresolved CFD-DEM.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1179,26 +1178,27 @@ GLSVansAssemblerBeetstra<dim>::calculate_particle_fluid_interactions(
...
-                     -(1 + 4 * (1 - cell_void_fraction)) * 0.5));
-
-      drag_force = normalized_drag_force * 3 * M_PI * viscosity * density *
-                   particle_properties[DEM::PropertiesIndex::dp] *
-                   superficial_velocity;
-
-      beta_drag +=
-        drag_force.norm() /
-        (density * (relative_velocity[particle_number].norm()) + 1e-15);
+      // Beetstra drag coefficient
```
voferreira, line 1181: Since you are reframing the drag model, the comment in the .h file needs to be changed as well. I would add the reference in the .h and put a note on this at the documentation.

source/fem-dem/vans_assemblers.cc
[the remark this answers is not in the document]
blaisb: Ohh I see, so the original correlation does not lead to the same unit by definition? Interesting. It's true that the unit of the Navier-Stokes equation we solve are Velocity / time so L^1 / T^2. Since beta is multiplied by the velocity, it needs to have units of seconds inverse.

blaisb (approved):
Good explanations, everything makes sense to me

source/fem-dem/vans_assemblers.cc
tonielgeitani: Done

voferreira:
Nice catch with this PR, @tonielgeitani ! All is good to me now.

voferreira (approved):

PR #557, opened 2022-08-08
PBC in DEM - Periodic mapping & Particles displacement

acdaigneault:
# Description of the problem

- Current DEM solver doesn't support periodic boundary conditions (PBC)
- Prior having smaller domain that could use a PBC feature (as pneumatic transport of particles) over a large domain, implementing this feature could be useful
- This feature will have multiple PR (2 or 3) for every step of implementation, this one is related to parameters, periodic cells mapping and particles displacement 

# Description of the solution

- Add boundary ids exportation for DEM for debug helping as it is already possible for CFD 
- Add parameters for the periodic boundaries in the subsection DEM boundary conditions :
    -  Modify `type` parameter to allow `periodic` option
    -  `boundary id` is already available for outlet BC type, the same parameter is also use for periodic boundaries to get id of one periodic wall (sometime referred as outlet in code)
    - Add `periodic id` for the id of the other periodic wall
    - Add `periodic direction` to set the perpendicular axis of the periodic walls
- Modify the `read_mesh` to allow the periodicity in the triangulation (same as done for CFD)
- Create new class `PeriodicBoundariesManipulator` for all periodic boundary operations since data structure is not the same as other classes (personal choice over implementing various functions in other classes)
- `PeriodicBoundariesManipulator` currently allows : 
    -  Periodic cells mapping : map as [key : cell id at outlet, value : <cell at outlet info object, cell at periodic info object>]
    Info of cells are store as pairs of `boundary_cells_info_struct`, similar as implemented in `BoundaryCellsInformation` class. Also, info is stored for locally owned cells with no duplicate pair.
    - Particle displacement : The position of every particle that passes through a periodic wall (that goes outside the domain) is modified to move it to the periodic cell after the integration step. The ParticleHandler doesn't support automated displacement.
- Add an application test with 2 threads with 10 particles to check possible issues with multi core. 
   

# How Has This Been Tested?

- No unit test yet, will be done after the full implementation of the PBC in DEM
- Has been tested in debug, with multi core (had some issues for not all thread are used situation, but is fixed now)

# Documentation

- No documentation yet, will be done after the full implementation of the PBC in DEM

# Future changes

- Addition of similar particle-particle detection as board search and fine search 
- Make particles interact in periodic cells as with cell neighbors   
- Addition of documentation in the wiki

# Comments

- I first planned to implement this feature for the Hooper case, so the branch is related to the example (which was a bad idea)
- Sorry about the squashed commits and merges from master, still improving my Git skills.


blaisb:
@acdaigneault Can you add an application_test? It can be really simple (like a particle falling through the domain and coming back up). this is more important than having an example.

source/dem/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ SET(TARGET "lethe-dem")
 
 FILE(GLOB_RECURSE TARGET_SRC *.cc)
 FILE(GLOB_RECURSE TARGET_INC ${CMAKE_SOURCE_DIR}/include/dem/*.h)
-SET(TARGET_SRC ${TARGET_SRC}  ${TARGET_INC}) 
+SET(TARGET_SRC ${TARGET_SRC} ${TARGET_INC})
```
blaisb, line 18: ?

source/dem/dem.cc
```diff
@@ -16,6 +16,7 @@
  *
  * Author: Bruno Blais, Shahab Golshan, Polytechnique Montreal, 2019-
```
blaisb, line 17:
```suggestion
```

source/dem/dem.cc
```diff
@@ -321,6 +323,12 @@ DEMSolver<dim>::load_balance()
           triangulation, solids[i_solid]->get_solid_triangulation());
     }
 
+  if (parameters.boundary_conditions.BC_type ==
+      Parameters::Lagrangian::BCDEM::BoundaryType::periodic)
+    {
+      periodic_boundaries_object.map_periodic_cells(triangulation);
+    }
+
```
blaisb, line 331: Here this will have to be replaced with a loop over boundaryconditions no? if that's the case,I would put this small verification into some sort of utility.

source/dem/dem.cc
```diff
@@ -864,6 +882,17 @@ DEMSolver<dim>::solve()
     cell_neighbors_object.find_full_cell_neighbors(triangulation,
                                                    cells_total_neighbor_list);
 
+  if (parameters.boundary_conditions.BC_type ==
+      Parameters::Lagrangian::BCDEM::BoundaryType::periodic)
```
blaisb, line 886: Again this won't work if bc_0 is not periodic right?

source/dem/floating_grid.cc
```diff
@@ -17,11 +17,17 @@ FloatingGrid<dim, spacedim>::FloatingGrid(
 {
   if (floating_grid_parameters.mesh.file_name != "none")
```
blaisb, line 18: I don't think this floating grid is actually used anywhere now or am I wrong? The current floating grids are stored into the Rigid solid object in core...This whole thing should be deleted, am I wrong @shahabgol ??

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -0,0 +1,194 @@
...
+          // cell the particle is coming from and will execute displacement
+          // with the locally owned cell.
+          if (cell->is_locally_owned())
+            {
+              typename Particles::ParticleHandler<dim>::particle_iterator_range
+                         particles_in_cell = particle_handler.particles_in_cell(cell);
+              const bool particles_exist_in_main_cell =
+                !particles_in_cell.empty();
+
+              if (particles_exist_in_main_cell)
```
blaisb, line 165: Why not instead request the number of particles in the cell? this would improve readability instead of looking at the iterator range.

source/dem/read_mesh.cc
```diff
@@ -2,11 +2,12 @@
...
-          const bool &                  restart,
-          const ConditionalOStream &    pcout,
-          Triangulation<dim, spacedim> &triangulation,
-          double &                      triangulation_cell_diameter)
+read_mesh(const Parameters::Mesh &       mesh_params,
+          const bool &                   restart,
+          const ConditionalOStream &     pcout,
+          Triangulation<dim, spacedim> & triangulation,
+          double &                       triangulation_cell_diameter,
+          Parameters::Lagrangian::BCDEM &bc_params)
```
blaisb, line 10: I kinda hate the fact that this routine is seperate from the grid one in the core of the librarie. It would be worth it to eventually look into reconciling the two.

blaisb (changes requested):
Just a few changes there and there. I think the whole floating grid this is deprecated and should be completely removed.
Right now does the current implementation support having multiple walls with periodic bc or just wall 0? I would finalize the feature so that it actually supports all wall for now, and then this could be merged.

examples/dem/periodic/periodic_box.prm
```diff
@@ -0,0 +1,100 @@
+# Listing of Parameters
+# ---------------------
+# --------------------------------------------------
```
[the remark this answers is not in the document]
acdaigneault: Application test is already commit :)

include/core/parameters_lagrangian.h
```diff
@@ -326,6 +327,11 @@ namespace Parameters
       // Rotational axes of rotating boundaries
       std::unordered_map<unsigned int, Tensor<1, 3>> boundary_rotational_vector;
 
+      // Periodic boundary IDs
+      std::vector<unsigned int> periodic_boundaries;
+      std::vector<unsigned int> periodic_direction;
```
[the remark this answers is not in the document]
acdaigneault: I did the same as outlet_boundaries, will check that and modify other parameters that should use this type

source/dem/periodic_boundaries_manipulator.cc
acdaigneault: Ok, will have to change it in another place also!

source/dem/read_mesh.cc
acdaigneault: I agree, I will work on that

source/dem/dem.cc
acdaigneault: So far parameters are stored in vectors to allow multiple periodic walls, but only parameters in [0] are used, which makes only one periodic wall available. I should modify the PR to implement it

shahabgol (approved):
Very good work. I have a few minor comments, but generally this is a very good PR.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -0,0 +1,194 @@
...
+{}
+
+template <int dim>
+void
+PeriodicBoundariesManipulator<dim>::get_boundary_info(
+  typename Triangulation<dim>::cell_iterator cell,
+  unsigned int                               face_id,
+  boundary_cells_info_struct<dim> &          boundary_information)
+{
+  // Initialize a simple quadrature for on the system. This will be used to
```
tonielgeitani, line 27:
```suggestion
  // Initialize a simple quadrature for the system. This will be used to
```

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -0,0 +1,194 @@
...
+
+      // Calculate distance between particle position and the face of
+      // boundary cell d = n•(pt_particle - pt_face)
+      double distance_with_face =
+        scalar_product(particle_position - cell_1.point_on_face,
+                       cell_1.normal_vector);
+
+
+      // If distance >= 0, particle is outside of cell (or on face).
+      // If so, particle location is modified to get moved in the periodic cell.
```
tonielgeitani, line 119:
```suggestion
      // If so, particle location is modified to get moved into the periodic cell.
```
I am not sure what do you mean here. Are you moving the particle from one cell into the periodic cell? If so this need to be into

tonielgeitani (approved):

applications_tests/dem_3d/periodic_boundary_box.prm
```diff
@@ -0,0 +1,104 @@
...
+    set enable 						 = true
+end
+
+# --------------------------------------------------
+# Model parameters
+#---------------------------------------------------
+subsection model parameters
+  set contact detection method 		   		         = dynamic
+  set dynamic contact search size coefficient	 	 = 0.9
+  set load balance method				             = frequent
```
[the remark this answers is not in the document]
acdaigneault: The application test is in parallel since I got some issues with that, especially when not all threads contains cells after load-balancing.

source/dem/periodic_boundaries_manipulator.cc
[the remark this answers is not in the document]
acdaigneault: I know, but the data structure is not the same. I had to make a choice between implementation of a new class dealing with periodic matter or implementation of new functions/modification with boolean for PBC cases in current classes. 
The main issue is that data structure for periodic cells have to keep info of the periodic cell. I tried to do it, but it gives addition of many new containers in the DEM class. Also, I had to pass the BC_type as argument to allow specific operation related to boundary cells.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -0,0 +1,194 @@
...
+                    periodic_cell = cell->periodic_neighbor(face_id);
+                  get_boundary_info(periodic_cell,
+                                    cell->periodic_neighbor_face_no(face_id),
+                                    periodic_boundary_information);
+
+                  // Store both cell information in map with cell id at
+                  // outlet as key
+                  periodic_boundary_cells_information.insert(
+                    {boundary_information.cell->global_active_cell_index(),
+                     std::make_pair(boundary_information,
```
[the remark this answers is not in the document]
acdaigneault: Because I need to keep position of the periodic cell to allow particle displacement. And will need it also in future for contact detection.

applications_tests/dem_3d/periodic_boundary_box.prm
```diff
@@ -0,0 +1,104 @@
+# Listing of Parameters
+# ---------------------
+# --------------------------------------------------
+# Simulation and IO Control
+#---------------------------------------------------
+subsection simulation control
+  set time step                 		 = 1e-4
```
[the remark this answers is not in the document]
acdaigneault: Didn't pay attention, thanks. It says 1e-5 is too small, I will put 1.5e-5

applications_tests/dem_3d/periodic_boundary_box.prm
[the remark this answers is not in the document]
acdaigneault: After all, I removed the load balancing because it seems to make the application test crash (particles may be owned by a random processor)

include/core/manifolds.h
```diff
@@ -97,6 +97,40 @@ class BoundaryPostprocessor : public DataPostprocessorScalar<dim>
   {
     auto current_cell = input_data.template get_cell<dim>();
```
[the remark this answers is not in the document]
acdaigneault: I actually added the virtual function `evaluate_scalar_field` because only `evaluate_vector_field` was implemented in Lethe. 
The  `evaluate_vector_field` function do not work for the dummy solution vector related to background_dof in DEM.
I did not add a new loop, just copy pasted the `evaluate_vector_field` and modified `DataPostprocessorInputs::Vector<dim> &input_data` to `DataPostprocessorInputs::Scalar<dim> &input_data` argument.

source/dem/read_mesh.cc
acdaigneault: ... will open an issue ;)

source/dem/periodic_boundaries_manipulator.cc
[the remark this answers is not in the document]
shahabgol: I see, so if you need, the information of the boundary cell and the periodic cell, why don't you add a new member (periodic cell) to boundary_cell_info_struct, or define a new struct for periodic boundaries? By doing this, you don't have to build two structures, fill them and extract information from two structures.

applications_tests/dem_3d/periodic_boundary_box.prm
```diff
@@ -0,0 +1,104 @@
...
+# Boundary conditions
+#---------------------------------------------------
+subsection DEM boundary conditions
+   set number of boundary conditions                     = 1
+
+   subsection boundary condition 0
+       # Choices are fixed_wall|outlet|rotational|translational|periodic
+       set type                                        = periodic
+       set boundary id                                 = 2
+       set periodic id                                 = 3
```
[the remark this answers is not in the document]
shahabgol: what about "periodic boundary id"? I leave this decision to you. If you think periodic id is better, keep it. :)

source/dem/periodic_boundaries_manipulator.cc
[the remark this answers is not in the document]
acdaigneault: I think it's a good idea and it could makes collision detection implementation easier in the future.
I commited my changes :)

blaisb (approved):

PR #559, opened 2022-08-09
Fix contact bug

Luckabarbeau:
# Description of the problem

Multiple bugs were present in the code

- Only particles with index 0 could contact other particles due to a typo.
- Particles could only contact one wall at a time which causes problems for particles in the corner
- The contact history was stored on a point-on-face basis instead of on a face basis which meant that the contact history was lost after a mesh refinement. 
- Particle contact with non-zero boundary conditions was leading to conflict between the value of two Dirichlet boundary conditions.
- Optimised cut cell mapping is leading to instability of the code for some cases where some cells that should be cut are not cut.
- Optimised cut cell mapping is sometimes slower than the original cut cell mapping if a significant portion of the volume of the domain is occupied by particles.

# Description of the solution

- Corrected the typo that was preventing particle contact calculation
- Allows particles to contact multiple walls at the same time.
- Stores the contact history based on the face indices instead than on a point-on-wall basis.
- Add a check that prevents the application of the sharp_IB BC if the DOF has already another boundary condition applied. Also, we added support for the Floating wall implemented in the DEM section to prevent particle wall contact with the inlet.
- Remove temporarily the optimized cut cell mapping and only use the standard version. (The fix of this function will require more time and research and will be part of a future update.) 

# How Has This Been Tested?

- Most sharp IB tests were affected by the removal of the optimized cut-cel mapping. All changes are due to changes in the cut-cell mapping.
- This update has also been used in the simulation of a small fluidized bed with 300 particles with significantly more stability than the current version of the code.

# Future changes

- Correct the optimized cut cell mapping and define the range of applicability of this approach.


source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
+      std::map<unsigned int, BoundaryCellsInfo> best_cells;
       // For each particle loop over the point and normal identified as
       // potential contact candidate.
       for (auto &boundary_cell_iter : boundary_cells[particle.particle_id])
         {
           // find the best candidate (the closest point).
           double dist =
             (boundary_cell_iter.point_on_boundary - particle.position).norm();
-          if (dist < best_dist)
+          if (dist < best_dist[boundary_cell_iter.boundary_index].value)
```
blaisb, line 413: Ok this is like such a hack. You need to comment this routine a lot better than it is commented right now.... You are using the default constructor of your struct to check if this thing already exist or not because value will always call the default constructor if it does not find an element. This is kinda nasty... You need to document it a lot better than that.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
     parameters->wall_restitution_coefficient;
 
+  struct DefaultDBL_MAX
+  {
+    double value = DBL_MAX;
+  };
+  struct DefaultUINT_MAX
+  {
+    int value = UINT_MAX;
+  };
```
blaisb, line 392: I would make it a tiny class instead with a default constructor instead of a struct with a value assigned. Also define it outside of the routine.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
   double wall_restitution_coefficient =
     parameters->wall_restitution_coefficient;
 
+  struct DefaultDBL_MAX
```
blaisb, line 385: You need to add comments as to why you are doing this. Otherwise we only understand 20 lines below why the heck this exists...

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
+              best_cells[boundary_cell_iter.boundary_index] =
+                boundary_cell_iter;
             }
           boundary_index += 1;
         }
+
+      // start the indexing of the floating wall at 1000000 so it doesn't
+
+      // add all the floating wall has contact candidates. Their indices start
+      // from 1000000 to avoid conflict with other boundary condition.
```
blaisb, line 427:
```suggestion
      // We start the indexing of the floating wall at 1000000 so it does not
      // add all the floating wall as contact candidates. Their indices start
      // from lowest_floating_wall_index to avoid conflict with other boundary conditions.
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
-      unsigned int best_index     = UINT_MAX;
+      unsigned int                              boundary_index = 0;
+      std::map<unsigned int, DefaultDBL_MAX>    best_dist;
+      std::map<unsigned int, DefaultUINT_MAX>   best_indices;
+      std::map<unsigned int, BoundaryCellsInfo> best_cells;
       // For each particle loop over the point and normal identified as
       // potential contact candidate.
       for (auto &boundary_cell_iter : boundary_cells[particle.particle_id])
         {
           // find the best candidate (the closest point).
```
blaisb, line 410:
```suggestion
          // Find the best candidate (which is the closest point).
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
   for (auto &particle : dem_particles)
     {
-      unsigned int boundary_index = 0;
-      double       best_dist      = DBL_MAX;
-      unsigned int best_index     = UINT_MAX;
+      unsigned int                              boundary_index = 0;
+      std::map<unsigned int, DefaultDBL_MAX>    best_dist;
+      std::map<unsigned int, DefaultUINT_MAX>   best_indices;
+      std::map<unsigned int, BoundaryCellsInfo> best_cells;
       // For each particle loop over the point and normal identified as
```
blaisb, line 406: This loop is not over particles, so why say for each particle?

source/fem-dem/ib_particles_dem.cc
```diff
@@ -520,31 +573,71 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+  };
+  struct DefaultUINT_MAX
+  {
+    int value = UINT_MAX;
+  };
+  enum lowest_floating_wall_indices
+  {
+    lowest_floating_wall_indices = 1000000
+  };
+
```
blaisb, line 588: Yeah clearly if you find yourself defining the struct twice, you need to do it cleanly in the .h...

source/fem-dem/ib_particles_dem.cc
```diff
@@ -520,31 +573,71 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
             }
           boundary_index += 1;
         }
-      // Do the particle wall lubrication calculation with the best candidate.
-      if (boundary_cells[particle.particle_id].size() > 0)
+
+      // start the indexing of the floating wall at 1M so it doesn't
+
+      // add all the floating wall has contact candidate their indices start
+      // from the largest indices of the face found in the candidate.
```
blaisb, line 617: See changes to the same comment as before.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -520,31 +573,71 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+  {
+    int value = UINT_MAX;
+  };
+  enum lowest_floating_wall_indices
+  {
+    lowest_floating_wall_indices = 1000000
+  };
+
   // Loop over the particles
   for (auto &particle : dem_particles)
```
blaisb, line 590: I feel like there are repetitions there, are you sure you can't refactor some of the stuff in the previous routine to work also with this one?

blaisb (changes requested):
Some refactoring and cleaning to do.
Please try to be more verbose when commenting your stuff, otherwise it is very hard to understand the intent of what you are trying to achieve. Comments inside the code serve an important purpose.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -40,9 +40,11 @@
 // Constructor for class GLSNavierStokesSolver
 template <int dim>
 GLSSharpNavierStokesSolver<dim>::GLSSharpNavierStokesSolver(
-  SimulationParameters<dim> &p_nsparam)
-  : GLSNavierStokesSolver<dim>(p_nsparam)
+  CFDDEMSimulationParameters<dim> &p_nsparam)
+  : GLSNavierStokesSolver<dim>(p_nsparam.cfd_parameters)
+  , cfd_dem_parameters(p_nsparam)
```
CaroleAnneDaunais, line 45: it's a bit redundant. Why the two attributes?

CaroleAnneDaunais (commented):
Bruno made good comments, I only had one more.

applications_tests/gls_sharp_navier_stokes_2d/couette2d_gls.output
```diff
@@ -7,26 +7,26 @@ Running on 1 MPI rank(s)...
...
-cells  error_velocity     error_pressure   
- 1024 2.178861e-03    - 0.000000e+00     - 
- 4096 6.279473e-04 1.79 5.395080e-09  -inf 
-16384 2.033314e-04 1.63 3.395293e-08 -2.65 
+div u : 4.49643e-09
+L2 error velocity : 0.00030358 L2 error pressure: 3.80696e-08
+cells  error_velocity    error_pressure   
+ 1024 4.526187e-03    - 0.000000e+00    - 
+ 4096 1.193397e-03 1.92 4.016706e-08 -inf 
+16384 3.035804e-04 1.97 3.806957e-08 0.08
```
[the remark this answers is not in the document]
Luckabarbeau: I do too

source/fem-dem/ib_particles_dem.cc
```diff
@@ -372,31 +382,72 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
             }
           boundary_index += 1;
         }
+
+      // start the indexing of the floating wall at 1000000 so it doesn't
+
+      // add all the floating wall has contact candidates. Their indices start
+      // from 1000000 to avoid conflict with other boundary condition.
+      for (unsigned int i = 0;
+           i < floating_walls_parameters->points_on_walls.size();
```
shahabgol, line 429: Does it iterate through all the floating walls? It's a little unclear that you use the size of points_on_walls

shahabgol (approved):
Very good work, I only have 2-3 minor comments.

source/fem-dem/ib_particles_dem.cc
Luckabarbeau: I move all the definitions in the .h, so they share all the definitions of the struct and of the enum.

source/fem-dem/ib_particles_dem.cc
Luckabarbeau: I move the definition of the struct in the .h.I kept the struct instead of the class since it is in line with other similar use of struct in the code. Also, the value is used outside of the struct and, as such, is public.

source/fem-dem/gls_sharp_navier_stokes.cc
Luckabarbeau: The parameter of the GLSNavierStokesSolver only takes the CFD part of the parameter and the cfd_dem_parameters need the full set of the parameter to be initialized.

source/fem-dem/ib_particles_dem.cc
Luckabarbeau: hopefully the new comments make it more clear

CaroleAnneDaunais (approved):

applications/gls_sharp_navier_stokes_2d/gls_sharp_navier_stokes_2d.cc
```diff
@@ -32,8 +32,8 @@ main(int argc, char *argv[])
       Utilities::MPI::MPI_InitFinalize mpi_initialization(
         argc, argv, numbers::invalid_unsigned_int);
 
-      ParameterHandler        prm;
-      SimulationParameters<2> NSparam;
+      ParameterHandler              prm;
+      CFDDEMSimulationParameters<2> NSparam;
```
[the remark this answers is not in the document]
Luckabarbeau: They are exactly the same object

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3106,18 +3124,19 @@ GLSSharpNavierStokesSolver<dim>::write_checkpoint()
...
   if (Utilities::MPI::this_mpi_process(this->mpi_communicator) == 0)
     {
+      std::string prefix =
+        this->simulation_parameters.simulation_control.output_folder +
+        this->simulation_parameters.restart_parameters.filename;
+
+      TableHandler particles_information_table;
+      std::string  filename =
+        this->simulation_parameters.simulation_control.output_folder + prefix +
+        ".ib_particles";
```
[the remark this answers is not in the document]
Luckabarbeau: It is used to write the restart file

blaisb (approved):
Good for me, whille merge once CI finishes

PR #563, opened 2022-08-12
Overhaul build system

paulapatience:
This PR overhauls the build system, updating it to use modern CMake and simplifying it in the process.

- [ ] Compiles on Manjaro
- [x] Compiles on Mac OS ARM architecture
- [ ] Compiles on Niagara
- [x] Compiles on Beluga
- [ ] Compiles on Narval


blaisb:
@paulapatience next time, please make the PR from a branch that is a branch from the lethe repository instead of a fork, this makes testing on multiple machine easier (instead of having to reclone and redo everything anew). This PR will require compilation on a few machines to test if everything is ok.

source/core/CMakeLists.txt
```diff
@@ -1,28 +1,73 @@
+add_library(lethe-core
+  # Sources
```
blaisb, line 2: I am not sure I am a fan of this. It means that anyone adding a new .cc and new .h will need to manually modify the CMakeList right? Why use this instead of the previous glob mechanism? It was a bit cleaner. Keep in mind that most of our developer have no CMake knowledge, so that becomes a new thing to teach everyone that will be hard to document (where to document it)

blaisb (changes requested):
Very good change. Much cleaner CMakeLists in the end.
The only thing I am not a fan off is the manual specification of the files to be included and the source file. It means that everybody that wants to add a file will need to manually go into the CMakeLists file. This will be difficult to document and it will be another hurdle that prevents people from developing. Can't we keep the recursive glob mechanism instead?

paulapatience:
> @paulapatience next time, please make the PR from a branch that is a branch from the lethe repository instead of a fork

Sorry about that. I just need you to invite me to the repository, as I don't currently have write access.

CMakeLists.txt
```diff
@@ -1,114 +1,144 @@
...
+# TARGET_SUPPORTS_SHARED_LIBS beforehand may be necessary in some
+# environments.
+set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
+
+# Assume that deal.II_DIR won't be provided on the command line in favor
+# of DEAL_II_DIR and can thus be used to detect whether CMake was
+# already run.
+if(DEFINED deal.II_DIR)
+  set(_had_deal_ii TRUE)
+endif()
```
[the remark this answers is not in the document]
paulapatience: This was added in commit https://github.com/lethe-cfd/lethe/pull/563/commits/d53052892fd59a3e014a44bebaffc018246047f4, the message of which explains the rationale: it's not CMake that is prevented from rerunning (because of course that would be quite incorrect), rather only the status messages related to deal.II features are prevented from printing, because when you rerun CMake on the same build directory, you most often will have already come to terms with the deal.II features which are available.

source/core/CMakeLists.txt
paulapatience: I suspected you might have ambiguous feelings towards this change, which is why I tried to explain the issue in the commit message (see https://github.com/lethe-cfd/lethe/pull/563/commits/488590c9731dd3272722ca1cc650a96e679a6010).

To expand upon the description in the commit message, the problem with file(GLOB) is that the build system (i.e., the generated Makefiles, Ninja files, etc.) are given the list of globbed files, but when you add or remove a source file, this list is not updated when rerunning CMake because it doesn't know that it has to update the list of files in the generated build system because no CMakeLists.txt files were changed. CMake basically assumes that the generated build system is still up to date.

The only way to inform the build system about the new files is to remove the build directory -- which causes you to lose all built artifacts, thus requiring a full rebuild -- and rerun CMake. In other words, every time you pull changes from someone, you have to either check if they added or removed files, or systematically delete the build directory.

The only recommended way to use file(GLOB) for source files is with CONFIGURE_DEPENDS, which was introduced in CMake 3.12. However, you had mentioned to me that maintaining compatibility with Ubuntu 18.04 was important, and that version of Ubuntu ships with CMake 3.10.2.

The [official CMake documentation](https://cmake.org/cmake/help/latest/command/file.html#filesystem) itself discourages using file(GLOB), and maintains that even using CONFIGURE_DEPENDS is suboptimal.

The two (correct) options you have are:

* Require CMake 3.12, use CONFIGURE_DEPENDS, and require users of Ubuntu 18.04 to update their version of CMake (which is apparently quite simple to do); or

* Require contributors to update the source files listed in the CMakeLists.txt files when adding or removing files.

I have (and as I mentioned, so does CMake) a preference for the latter option -- if you are worried about contributors forgetting to update the list, we could add a pre-commit hook to check if the lists of source files in the CMakeLists.txt files are up to date and print a warning if not -- but either one would be better than the current behavior.

blaisb: I understand the rationale and it does make sense to me. The only challenge I see is more from an education purpose. Other than you, I am the only one that has done the CMake management in the past. As such, it will be a tricky thing to explain to any new developer that if they add a file, for their code to take the file into account, they need to edit the CMakeList. Can you maybe add a section on that in the contributing section of the documentation? With this, I'd be confident in moving forward with the change you suggest. Seems healthier. I was never a fan of the Glob approach (because it leads to the problems you mentioned and I encountered all of those in the past). I just want to make sure that this won't hinder everyone. CMake is not something most of Lethe developer know how to use or write.

blaisb:
> > @paulapatience next time, please make the PR from a branch that is a branch from the lethe repository instead of a fork
> 
> Sorry about that. I just need you to invite me to the repository, as I don't currently have write access.

No biggie. Added you

blaisb:
@paulapatience  Can you rebase with master? There were some missing includes that prevented compilation on the latest manjaro rolling release. Once this is rebased, I can finish testing on all platforms.

CMakeLists.txt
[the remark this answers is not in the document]
blaisb: I see. Sorry, I did not look at the individual commit message. They are kinda hard to get to from the github interface which I use to review the PR.

blaisb (approved):

PR #565, opened 2022-08-15
Organize dem data structure

shahabgol:
# Description of the problem

- Data structures used in the dem solver were complicated and difficult to read.

# Description of the solution

- Store complex dem data structures in data_containers.h and use them in dem and unresolved cfd-dem solvers.

### DIFF
--- a/include/core/data_containers.h
+++ b/include/core/data_containers.h
@@ -16,6 +16,7 @@
  *
  */
 
+#include <dem/boundary_cells_info_struct.h>
 #include <dem/particle_particle_contact_info_struct.h>
 #include <dem/particle_point_line_contact_info_struct.h>
 #include <dem/particle_wall_contact_info_struct.h>
@@ -178,6 +179,10 @@ namespace dem_data_containers
 
     typedef std::map<unsigned int, std::map<types::boundary_id, Tensor<1, 3>>>
       vector_on_boundary;
+
+    typedef std::map<types::global_cell_index,
+                     periodic_boundaries_cells_info_struct<dim>>
+      periodic_boundaries_cells_info;
   };
 
 } // namespace dem_data_containers
--- a/include/core/data_containers.h
+++ b/include/core/data_containers.h
@@ -16,6 +16,10 @@
  *
  */
 
+#include <dem/particle_particle_contact_info_struct.h>
+#include <dem/particle_point_line_contact_info_struct.h>
+#include <dem/particle_wall_contact_info_struct.h>
+
 #include <deal.II/base/tensor.h>
 
 #include <deal.II/distributed/tria.h>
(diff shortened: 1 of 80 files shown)


source/core/lethe_grid_tools.cc
```diff
@@ -11,8 +11,6 @@
 #include <deal.II/grid/manifold_lib.h>
 
 #include <cmath>
-#include <unordered_map>
-
```
blaisb, line 15: Why remove this include? Unordered map are used in the lethe_grid_tools. This was required for ARM compilation.

include/core/data_containers.h
```diff
@@ -90,5 +94,91 @@ namespace dem_data_containers
...
+
+    typedef std::unordered_map<
+      types::global_cell_index,
+      std::vector<typename Triangulation<dim>::active_cell_iterator>>
+      cells_total_neighbor_list;
+
+    typedef std::map<types::boundary_id, std::pair<Tensor<1, 3>, Point<3>>>
+      boundary_points_and_normal_vectors;
+
+    typedef std::map<unsigned int, std::map<types::boundary_id, Tensor<1, 3>>>
```
blaisb, line 179: I would pre-declare these structures in the .cc file so that they are existing for dim=2 and dim=3.

include/dem/particle_particle_broad_search.h
```diff
@@ -78,18 +74,14 @@ class ParticleParticleBroadSearch
...
-    std::unordered_map<types::particle_index,
-                       std::vector<types::particle_index>>
-      &local_contact_pair_candidates,
-    std::unordered_map<types::particle_index,
-                       std::vector<types::particle_index>>
-      &ghost_contact_pair_candidates);
+    const typename dem_data_containers::dem_data_structures<
+      dim>::cells_neighbor_list *cells_local_neighbor_list,
+    const typename dem_data_containers::dem_data_structures<
+      dim>::cells_neighbor_list *cells_ghost_neighbor_list,
```
blaisb, line 80: Why are these pointers when the other ones were references?

blaisb (changes requested):
First mini batch of comments, will continue later

include/dem/particle_point_line_fine_search.h
```diff
@@ -84,14 +84,12 @@ class ParticlePointLineFineSearch
...
+    dim>::particle_point_line_contact_info
   particle_line_fine_search(
-    const std::unordered_map<
-      types::particle_index,
-      std::tuple<Particles::ParticleIterator<dim>, Point<dim>, Point<dim>>>
-      &           particle_line_contact_candidates,
-    const double &neighborhood_threshold);
+    const typename dem_data_containers::dem_data_structures<
+      dim>::particle_line_candidates &particle_line_contact_candidates,
+    const double &                    neighborhood_threshold);
```
blaisb, line 92: Does this need to be a reference? if it's just a double I would pass it by value.

blaisb (changes requested):
Just small tiny changes. God this is clean. We should have done this from the beginning...

acdaigneault (approved):
I really like the idea of DEM data structures! My only comment is about the names of some data structures. I think they are a bit too specific about usage and/or not explicit enough about the kind of data structure (key type & value type). 

You may also add the `periodic_boundaries_cells_information` type from `periodic_boundaries_manipulator.h`.

source/core/lethe_grid_tools.cc
shahabgol: We don't have any unordered_maps in the lethe_grid_tools anymore. is it necessary?

blaisb: if we don't then it's cool. just make sure all includes are there.

include/core/data_containers.h
```diff
@@ -90,5 +94,91 @@ namespace dem_data_containers
...
+      vector_on_boundary;
+  };
+
```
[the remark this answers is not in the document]
blaisb: I think for the struct this make sense, I like the current typedef though

include/core/data_containers.h
shahabgol: Do you mean declaring the structure template (like the way I did?)

include/core/data_containers.h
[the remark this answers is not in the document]
blaisb: I'm not sure I agree with this regarding the typedef. The proposed typedef actually indicate what the containers are used to do instead of what they contain. I rather like this.

include/dem/particle_particle_broad_search.h
shahabgol: Corrected.

include/core/data_containers.h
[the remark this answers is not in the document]
acdaigneault: Ok I understand

include/dem/particle_point_line_fine_search.h
shahabgol: Corrected.

include/dem/update_particle_wall_contact_container.h
```diff
@@ -20,8 +20,6 @@
 
 #include <dem/particle_wall_contact_info_struct.h>
```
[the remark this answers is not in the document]
shahabgol: Removed.

source/dem/localize_contacts.cc
```diff
@@ -255,96 +227,48 @@ localize_contacts(
...
-  std::unordered_map<
-    types::particle_index,
-    std::unordered_map<types::boundary_id, Particles::ParticleIterator<3>>>
-    &pfw_contact_candidates,
-  std::vector<std::map<
-    typename Triangulation<2, 3>::active_cell_iterator,
-    std::unordered_map<types::particle_index, Particles::ParticleIterator<3>>,
-    dem_data_containers::cut_cell_comparison<3>>>
+  typename dem_data_containers::dem_data_structures<
+    3>::particle_floating_wall_candidates &pfw_contact_candidates,
```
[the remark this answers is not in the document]
shahabgol: corrected.

include/core/data_containers.h
```diff
@@ -90,5 +94,91 @@ namespace dem_data_containers
...
+      particle_floating_wall_candidates;
+
+    typedef std::unordered_map<
+      types::particle_index,
+      std::unordered_map<types::boundary_id,
+                         std::tuple<Particles::ParticleIterator<dim>,
+                                    Tensor<1, dim>,
+                                    Point<dim>,
+                                    types::boundary_id,
+                                    types::global_cell_index>>>
```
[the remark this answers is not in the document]
shahabgol: This PR is getting very big. I will apply this comment in another PR.

blaisb (approved):
All good for me

PR #566, opened 2022-08-15
Nitsche ib for heat

jeannejoachim:
# Description of the solution

Add heat transfer to the Nitsche IB formulation, so that the Nitsche solid can now have a given temperature. Dirichlet condition only for now.

# How Has This Been Tested?

- @Luckabarbeau cases
- Rotating Nitsche bodies

# Documentation

- [x] Nitsche documentation

# Future changes

- [x] Rebase and solve conflicts
- [x] Add test with Nitsche static object
- [x] Add test with Nitsche moving object
- [x] Return warning if using `gls_nitsche` solver but `number_of_solids = 0` in prm
- [x] Return warning if using `gls_navier_stokes` solver but `number_of_solids > 0` in prm

# Comments

@Luckabarbeau do not hesitate to change this description as you see fit!

- Generalize mesh adaptation in the next PR! (branch `Kelly_estimator_multivariables`)


blaisb:
@jeannejoachim can you please rebase? This branch is now quite old.

jeannejoachim:
> @jeannejoachim can you please rebase? This branch is now quite old.

Yes sure! It is in the tasks in Future Changes. Do not bother reviewing for now :)

jeannejoachim:
It should be properly rebased on master! I followed the classical squash-rebase and solved conflicts manually, not sure why some commits on master are considered as commits to this branch (Shahab and Audrey's namely), sorry for the files to review that are added by that :/

jeannejoachim (commented):
I reviewed all the files and pushed some modifications. The code and documentations can be reviewed! Please disregard any `dem` and `cfd-dem` files, the modifications only came from the rebase with master. All it needs now IMO is a test, I will add that.

blaisb:
> I reviewed all the files and pushed some modifications. The code and documentations can be reviewed! Please disregard any `dem` and `cfd-dem` files, the modifications only came from the rebase with master. All it needs now IMO is a test, I will add that.

I think you made a mistake when rebasing. I can't review the PR with 110 changes. Some of them are actually reverting some changes that were done yesterday and before... Can you backtrack and re-rebase to ensure that you only have modifications on file you actually modified? It's too dangerous to review this PR with 110 files changed.. It should not be the case

jeannejoachim:
Ok now it should be good to review! Sorry for the mess ^^"

blaisb (changes requested):
First batch of comments, review not finished yet

include/solvers/multiphysics_interface.h
```diff
@@ -540,6 +540,46 @@ class MultiphysicsInterface
     return reynolds_stress_solutions;
   }
 
+  /**
+   * @brief Request the solid objects. Used when a heat boundary condition
+   * is applied on a solid, through Nitsche immersed boundary method.
```
blaisb, line 545:
```suggestion
   * @brief Request the solid objects. Used an auxiliary physics
   * needs to apply a boundary condition on a solid through Nitsche immersed boundary method.
```

include/solvers/multiphysics_interface.h
```diff
@@ -540,6 +540,46 @@ class MultiphysicsInterface
...
+  TrilinosWrappers::MPI::Vector *
+  get_rhs(PhysicsID physics_id)
+  {
+    AssertThrow((std::find(active_physics.begin(),
+                           active_physics.end(),
+                           physics_id) != active_physics.end()),
+                ExcInternalError());
+    return physics_rhs[physics_id];
+  }
+
```
blaisb, line 582: Do you really need those? Why can't the heat transfer solver itself apply the restriction. Passing RHS and matrices around seems off...

include/solvers/multiphysics_interface.h
```diff
@@ -628,6 +668,19 @@ class MultiphysicsInterface
...
+   *
+   * @param physics_id The physics of the DOF handler being requested
+   *
+   * @param dof_handler The dof handler for which the reference is stored
+   */
+  void
+  set_solid(std::vector<std::shared_ptr<SolidBase<dim, dim>>> *solids_input)
+  {
+    solids = solids_input;
+  }
```
blaisb, line 682: Comments and briefs need to be changed here

include/solvers/multiphysics_interface.h
```diff
@@ -684,7 +737,39 @@ class MultiphysicsInterface
...
+   */
+  void
+  set_rhs(PhysicsID physics_id, TrilinosWrappers::MPI::Vector *rhs_vector)
+  {
+    AssertThrow((std::find(active_physics.begin(),
+                           active_physics.end(),
+                           physics_id) != active_physics.end()),
+                ExcInternalError());
+    physics_rhs[physics_id] = rhs_vector;
+  }
```
blaisb, line 772: I don't understand why you need those?

include/solvers/multiphysics_interface.h
```diff
@@ -764,15 +849,25 @@ class MultiphysicsInterface
...
+          estimated_error_per_cell);
       }
-    for (auto &iphys : block_physics)
+    else
       {
-        iphys.second->compute_kelly(estimated_error_per_cell);
+        for (auto &iphys : physics)
+          {
+            iphys.second->compute_kelly(estimated_error_per_cell);
+          }
```
blaisb, line 866: Can you explain what brought upon this change? I don't understand why this mechanism is needed?
If you want to combine the error per cell of the temperature and the velocity, you can already do it with the Vector<float> that is calculated here?

include/solvers/multiphysics_interface.h
```diff
@@ -809,7 +904,7 @@ class MultiphysicsInterface
       }
   };
 
-
+  bool constraing_solver;
```
blaisb, line 907: What is this parameter and why is it public?

include/solvers/multiphysics_interface.h
```diff
@@ -819,6 +914,9 @@ class MultiphysicsInterface
   // Data structure to store all physics which were enabled
   std::vector<PhysicsID> active_physics;
 
+  // Fluid solver is a constraing solver. (GLS_NITSCHE ,GLS_SHARP)
+
+
```
blaisb, line 919: What? contraing? like constrained?

include/solvers/multiphysics_interface.h
```diff
@@ -840,6 +938,14 @@ class MultiphysicsInterface
 
   std::map<PhysicsID, DoFHandler<dim> *> physics_dof_handler;
 
+
+  std::map<PhysicsID, TrilinosWrappers::SparseMatrix *> physics_matrix;
+  std::map<PhysicsID, TrilinosWrappers::MPI::Vector *>  physics_rhs;
```
blaisb, line 943: Very unclear to me why these would need to be passed around...

source/core/parameters.cc
```diff
@@ -1599,6 +1600,8 @@ namespace Parameters
         variable = Variable::phase;
       if (vop == "temperature")
         variable = Variable::temperature;
+      if (vop == "velocity and temperature")
+        variable = Variable::velocity_temperature;
```
blaisb, line 1604: Maybe here it would be better to move to a mask approach where we would have like 00000 means false false false false false and then 10001 for example would mean true false false false true. This would be a lot more flexible to allow multiple adaptation criteria. This mechanism proposed here won't be maintainable, every new combination of variable means a new enum variable and don't get me started if we need 3...

source/solvers/heat_transfer.cc
```diff
@@ -40,6 +43,142 @@ void
...
+            {
+              local_matrix = 0;
+              local_rhs    = 0;
+
+#if (DEAL_II_VERSION_MAJOR < 10 && DEAL_II_VERSION_MINOR < 4)
+              const auto &cell =
+                particle->get_surrounding_cell(*this->triangulation);
+#else
+              const auto &cell = particle->get_surrounding_cell();
+#endif
```
blaisb, line 109: We don,t support 9.3 anymore, this shouldnt be needed.
```suggestion
              const auto &cell = particle->get_surrounding_cell();

```

source/solvers/heat_transfer.cc
```diff
@@ -40,6 +43,142 @@ void
...
+              auto &system_rhs = this->system_rhs;
+              constraints_used.distribute_local_to_global(local_matrix,
+                                                          local_rhs,
+                                                          heat_dof_indices,
+                                                          this->system_matrix,
+                                                          system_rhs);
+              particle = pic.end();
+            }
+          this->system_matrix.compress(VectorOperation::add);
+          this->system_rhs.compress(VectorOperation::add);
```
blaisb, line 179: This is all fine and dandy, but then why did you need to add the matrix and the rhs to the multiphysics interface?

include/solvers/multiphysics_interface.h
```diff
@@ -540,6 +540,46 @@ class MultiphysicsInterface
...
   }
 
+  /**
+   * @brief Request the solid objects. Used when a heat boundary condition
+   * is applied on a solid, through Nitsche immersed boundary method.
+   */
+  std::vector<std::shared_ptr<SolidBase<dim, dim>>> *
+  get_solids()
+  {
+    return solids;
```
blaisb, line 550: I would add an assertion here based on the number of solids or something like this for defensive programming.

blaisb (changes requested):
Some work left to do before this is mergeable, lots of cleaning mostly, there are lot of leftover things that are not used.

source/core/parameters.cc
jeannejoachim: I really like the idea of having a flexible approach! We can talk about how to integrate it well, also with the user input on the prm. Maybe this would be better in another PR? (and I remove the mechanism on velocity_temperature in this one?)

include/core/solid_base.h
```diff
@@ -194,6 +197,12 @@ class SolidBase
   void
   rotate_grid(double angle, int axis);
 
+  /**
+   * @brief prints the positions of the particles
+   */
+  void
+  update_temperature_time(double time);
```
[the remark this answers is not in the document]
Luckabarbeau: We prefer this design has it is more uniform with other modification to the state of the solid

include/solvers/multiphysics_interface.h
Luckabarbeau: This has been removed since they are not used anymore

include/solvers/multiphysics_interface.h
Luckabarbeau: same has above

include/solvers/multiphysics_interface.h
Luckabarbeau: this is also not used anymore.

include/solvers/multiphysics_interface.h
Luckabarbeau: This has been removes it was use in an other architecture that wasn't working

include/solvers/multiphysics_interface.h
Luckabarbeau: This has been removes it was use in an other architecture that wasn't working

include/solvers/multiphysics_interface.h
Luckabarbeau: same has above

include/solvers/multiphysics_interface.h
Luckabarbeau: This was used in another architecture but it's not used anymore.

source/solvers/heat_transfer.cc
Luckabarbeau: They are no longer used; the original code added the Nitsche restriction in the Nitsche solver directly. But it was very complex for nothing I changed the architecture.

include/solvers/multiphysics_interface.h
jeannejoachim: I tried to implement it as you suggested, but it was just never triggered. This is because `get_solids` in the `multiphysics_interface` is called only by the `heat_transfer` solver, in the method `assemble_nitsche_restriction`, which is itself is called only if `number_solids > 0`, in the method `assemble_matrix_and_rhs` (still in the `heat_transfer.cc`). I will specify it in the description of this method because it took me a while to pinpoint that! 

We do not have `set nitsche = true` anywhere (unlike for subphysics), that could be used to assert that if we use NIB, we have `number of solids` getter than 0. I think that a cleaner way to do so would be to assert that if the solver `gls_nitsche_navier_stokes` is called, a solid should be defined (and likewise, if another solver is called but `number of solids` is not 0 in the prm, a warning is thrown). Maybe by asserting it in the solvers directly? I do not know if this would have a negative impact for testing/debugging. What do you think?

source/core/parameters.cc
jeannejoachim: I talked with Lucka and he needs to have the velocity+temperature still available, so I will implement the generalized approach directly in this PR in the end :) to be coming soon!

applications_tests/gls_nitsche_navier_stokes_22/nitsche-heating_moving.prm
```diff
@@ -0,0 +1,130 @@
...
+end
+
+#---------------------------------------------------
+# Initial condition
+#---------------------------------------------------
+subsection initial conditions
+    set type = nodal
+    subsection temperature
+            set Function expression = 0
+    end
```
CaroleAnneDaunais, line 28: Aren't these the default values?

include/solvers/heat_transfer.h
```diff
@@ -127,6 +127,12 @@ class HeatTransfer : public AuxiliaryPhysics<dim, TrilinosWrappers::MPI::Vector>
   void
   assemble_rhs();
 
+  /**
+   * @brief Call for the assembly of the matrix and the right-hand side of the Nitsche restriction for the heat transfert equation.
+   */
+  void
+  assemble_nitsche_restriction(bool assemble_matrix);
```
CaroleAnneDaunais, line 134: From the header I get that it's only for the heat restriction, I think a better name for this function would be `assemble_nitsche_heat_restriction()`

CaroleAnneDaunais (commented):
Nice PR! :) I will approve once 2 last tasks are checked

applications_tests/gls_nitsche_navier_stokes_22/nitsche-heating_moving.prm
jeannejoachim: Yes it is! But I thought it would be clearer to specify it, as we are monitoring the temperature evolution.

include/solvers/multiphysics_interface.h
blaisb: Well if this can't be called if there are no solids, then it won't happen, but I would still add an assertion here even if it never happens. Better be more defensive than not enough.
It's all right to write assertions which may never be triggered. The more we have, the better it is. Assertions have zero cost in release mode
:)

source/core/parameters.cc
blaisb: Okidoke, don't hestiate to talk to me for the implementation. I Have a very good idea for the architecture.

jeannejoachim: In the end it was becoming too big so I removed it and opening another branch for that! Lucka is ok with his case being broken for a few days. I cleaned this PR :)

applications_tests/gls_nitsche_navier_stokes_22/nitsche-heating_moving.prm
CaroleAnneDaunais: Fair enough!

CaroleAnneDaunais (approved):

applications_tests/gls_nitsche_navier_stokes_22/nitsche-heating_static.prm
```diff
@@ -0,0 +1,129 @@
+# Listing of Parameters
```
blaisb, line 1: Underscore in file name

doc/source/parameters/cfd/initial_conditions.rst
```diff
@@ -19,6 +19,10 @@ It is often necessary to set-up complex initial conditions when simulating trans
      set Function expression = if (x<0.5 & y<1, 1, 0)
    end
 
+   subsection temperature
```
blaisb, line 22: Good catch!

include/core/parameters.h
```diff
@@ -863,6 +863,13 @@ namespace Parameters
       fraction
     } fractionType;
 
+    // Booleans for mesh adaptation
+    bool refine_on_velocity;
+    bool refine_on_temperature;
+    bool refine_on_pressure;
+    bool refine_on_phase;
+    bool refine_on_tracer;
```
blaisb, line 871: Are those actually used?

include/core/solid_objects_parameters.h
```diff
@@ -238,9 +264,12 @@ namespace Parameters
...
       number_solids = prm.get_integer("number of solids");
-      for (unsigned int i_solid = 0; i_solid < number_solids; ++i_solid)
+      nitsche_solids.resize(number_solids);
+
+      if (number_solids > 0)
         {
-          nitsche_solids[i_solid]->parse_parameters(prm, i_solid);
+          for (unsigned int i_solid = 0; i_solid < number_solids; ++i_solid)
+            nitsche_solids[i_solid]->parse_parameters(prm, i_solid);
         }
```
blaisb, line 273: Why do you need an if around the for? The for loop won't do anything is the number of solid is 0.

include/solvers/gd_navier_stokes.h
```diff
@@ -33,10 +33,16 @@
 #include <deal.II/lac/trilinos_precondition.h>
 #include <deal.II/lac/trilinos_solver.h>
 
-
-
 using namespace dealii;
 
+DeclException3(SolidWarning,
```
blaisb, line 38: You can declare the exception once somewhere in core for example. This will prevent you from having to redeclare it for every solver. You can make a  core/exceptions.h if you want even.

include/solvers/multiphysics_interface.h
```diff
@@ -43,6 +44,12 @@
 
 using namespace dealii;
 
+DeclException1(
```
blaisb, line 47: I  would move the exception declaration to a seperate file.

doc/source/parameters/cfd/mesh_adaptation_control.rst
```diff
@@ -45,6 +45,7 @@ This subsection controls the mesh adaptation method, with default values given b
 	* pressure
 	* phase (for multiphase flows)
 	* temperature
+	* velocity and temperature
```
blaisb, line 48: This has been removed no?

blaisb (changes requested):
One or two tiny things but this is ready for merge afterwards.
I was just wondering why there are still these bools for adaptation and why there is still an explanation for temperature and velocityadaptation?
I would try to move the declaration of the exceptions to more general files to prevent repeating them for every solver. You can create a core/exceptions or something similar to that. Otherwise, very good work.

include/core/parameters.h
jeannejoachim: no more! Good catch thank you

jeannejoachim:
@blaisb I am not sure why there are build errors with deal.ii master and v9.4.0 (everytime at the stage of `lethe-dem`), as I did not touch these files. Would you know what is happening?

jeannejoachim:
> > @blaisb I am not sure why there are build errors with deal.ii master and v9.4.0 (everytime at the stage of `lethe-dem`), as I did not touch these files. Would you know what is happening?
> 
> You have a missing include somewhere. The compilation error is this one:
> 
> Screenshot_20220906_072847

thank you! I thought I put it everywhere needed

blaisb (approved):

PR #572, opened 2022-08-23
Vof example

shahabgol:
# Description of the problem

- Fixes two bugs in the VOF solver
- Adds a limit force parameter to enable limiting the source term in VOF solver


source/core/parameters_multiphysics.cc
```diff
@@ -114,6 +114,11 @@ Parameters::VOF::declare_parameters(ParameterHandler &prm)
 {
   prm.enter_subsection("VOF");
   {
+    prm.declare_entry("limit force",
```
jeannejoachim, line 117: For what I remember, I used a parameter a while ago (`apply_gravity_on_fluid0` or something), but we decided with @blaisb that it should always be `true` so we removed the parameter. Do you have a case where you want it to be false?

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -89,7 +90,9 @@ The default values of the VOF parameters are given in the text box below.
 
 	Applying viscous dissipation in one of the fluids instead of both is particularly useful when one of the fluids is air. For numerical stability, the ``kinematic viscosity`` of the air is usually increased. However, but we do not want to have viscous dissipation in the air, because it would result in an unrealistic increase in its temperature.
 
-* ``subsection interface sharpening``: defines parameters to counter numerical diffusion of the VOF method and to avoid the interface between the two fluids becoming more and more blurry after each time step.
+  * ``limit force`` enables limiting the source term (for instance gravity force).
```
jeannejoachim, line 93:
```suggestion
  * ``limit force``: defines if the ``source term`` (doc:`../source_term`, classically the gravity force) should be applied to the non-conservative fluid. This parameter is used if ``conservative fluid`` is not ``both`` in ``subsection mass conservation``.     
```

source/core/parameters_multiphysics.cc
jeannejoachim: like, is there a case where you would like to skip mass conservation on a fluid, and still apply the source term on it?

applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
```diff
@@ -10,29 +10,29 @@ Kinetic energy : 0
...
+Kinetic energy : 0.00760271
 
 *****************************************************************
-Transient iteration : 5        Time : 0.005    Time step : 0.001    CFL : 0.317114
+Transient iteration : 5        Time : 0.005    Time step : 0.001    CFL : 0.31828 
 *****************************************************************
-Enstrophy  : 566.914
-Kinetic energy : 0.0117275
+Enstrophy  : 585.52
+Kinetic energy : 0.0124625
```
jeannejoachim, line 38: is it for better or worse? Beware that I have a PR on VOF opened, which changes VOF test results, so the last to be merged should rebase and update the test output (race is on! :smile: ) to account for both changes

source/solvers/gd_navier_stokes.cc
```diff
@@ -60,6 +60,13 @@ GDNavierStokesSolver<dim>::setup_assemblers()
 {
   this->assemblers.clear();
 
+  // Buoyant force
+  if (this->simulation_parameters.multiphysics.buoyancy_force)
```
jeannejoachim, line 64: so you want to use the `buoyancy_force` even if VOF is not used?

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -102,10 +102,12 @@ NavierStokesScratchData<dim>::enable_vof(const FiniteElement<dim> &fe,
...
-  density_0   = std::vector<double>(n_q_points);
-  density_1   = std::vector<double>(n_q_points);
-  viscosity_0 = std::vector<double>(n_q_points);
-  viscosity_1 = std::vector<double>(n_q_points);
+  density_0           = std::vector<double>(n_q_points);
+  density_1           = std::vector<double>(n_q_points);
+  viscosity_0         = std::vector<double>(n_q_points);
+  viscosity_1         = std::vector<double>(n_q_points);
+  thermal_expansion_0 = std::vector<double>(n_q_points);
+  thermal_expansion_1 = std::vector<double>(n_q_points);
```
jeannejoachim, line 110: it's not used in this PR, right? It will be in a next PR?

source/core/parameters_multiphysics.cc
jeannejoachim: oh, or do you want to limit it no matter the density ratio?

jeannejoachim (commented):
I am not sure I understand the need for this new parameter (it existed before and was removed), could you clarify that for me (and maybe in the documentation also, as a tip)?

source/core/parameters_multiphysics.cc
blaisb: The idea is that the behavior of not applying the force in one of the fluid is not valid for multiple cases.
I would be for the idea of removing the parameter all together and applying force everywhere.

doc/source/parameters/cfd/volume_of_fluid.rst
blaisb: Let's leave it this way for now, but I am really, really starting to hate this multiplication of parameters. I would rather remove it alltogether and apply the force everywhere. The VOF models already has a bit too many shenigans right now.

source/solvers/gd_navier_stokes.cc
blaisb: Buoyancy force should exists even when VOF is not used. Buoyancy and VOF are not related.

source/solvers/navier_stokes_scratch_data.cc
blaisb: It is used right now...

source/solvers/navier_stokes_vof_assemblers.cc
```diff
@@ -31,16 +31,16 @@ GLSNavierStokesVOFAssemblerCore<dim>::assemble_matrix(
...
-
   Assert(
     scratch_data.properties_manager.density_is_constant(),
     RequiresConstantDensity(
       "GLSVansAssemblerDiFelice<dim>::calculate_particle_fluid_interactions"));
 
+  // Limit force application : not applied if cell density is below
+  // density_ratio of the maximum density (e.g. when one of the fluids is air)
+  const double density_ratio = 2;
+  double       phase_cutoff  = 0;
```
blaisb, line 42: I don't like this. I think it's just better to remove this force limitation mechanism right now. This is not a good idea. Let's just apply the force everywhere and be done with it... This is going to be removed ASAP otherwise and it's not a good idea. Force is force, it applies to both fluid, let's be done with it.

blaisb (changes requested):
Why is this PR called VOF examples? There are not VOF examples.
Can you just remove the force limitation parameter. This is not a good idea. We should aim at reducing the number of parameters, not adding new ones. Let's streamline the code and remove this thresholding for the force. It was a bad idea (my bad idea) from the get go. If it changes the example results, so be it, it will be much cleaner this way.

source/core/parameters_multiphysics.cc
shahabgol: This parameter and limiting force in the VOF solver are removed.

doc/source/parameters/cfd/volume_of_fluid.rst
shahabgol: It is removed.

applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
shahabgol: I can't say if it's better or worse, but I've seen removing the "limit force" results in better agreement with experimental results in the rayleigh-taylor instability example.

source/solvers/navier_stokes_vof_assemblers.cc
shahabgol: Limiting the source term is removed from the VOF assemblers.

shahabgol:
Your comments have been applied.
An example (rayleigh taylor instability) is added.

examples/multiphysics/rayleigh_taylor_instability/rayleigh_taylor_instability.prm
```diff
@@ -0,0 +1,165 @@
...
+# --------------------------------------------------
+# Linear Solver Control
+#---------------------------------------------------
+subsection linear solver
+  set verbosity					= verbose
+  set method                  			= gmres
+  set max iters               			= 10000
+  set relative residual                      	= 1e-8
+  set minimum residual                       	= 1e-10
+  set ilu preconditioner fill 			= 0
```
jeannejoachim, line 162: remove unnecessary parameters (from `ilu preconditioner fill` to the end of this subsection)

source/solvers/navier_stokes_vof_assemblers.cc
jeannejoachim: so the `density_ratio` and `phase_cutoff` are removed as well? It does not appear as such on git

jeannejoachim (changes requested):
Very cool additions for the example! I have comments here and there, but mainly I think you should clean the prm for the example to keep only non-default values, and explain why they are non-default (namely in the solver controls)

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
```diff
@@ -0,0 +1,252 @@
...
+      set fraction type           	= fraction
+      set max refinement level    	= 7
+      set min refinement level    	= 5
+      set frequency               	= 1
+      set fraction refinement     	= 0.9
+      set fraction coarsening     	= 0.001
+    end
+
+
+In the ``VOF`` subsection, we enable ``interface sharpening`` to reconstruct the interface and keep it sharp during the simulation. Note that here we use the ``constant`` method for interface sharpening. We can also use ``adaptative`` type for interface sharpening, but the ``mass conservation`` results show that choosing a ``constant`` method does not affect the mass conservation significantly.
```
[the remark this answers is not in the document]
blaisb: Yeah I would really add another small section to this example where you do the same thing with adaptive mass conservation. It would be the ideal place to show it.

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
```diff
@@ -0,0 +1,252 @@
...
+In the following figure, we compare the simulation results with that of Garoosi and Hooman [2].
+
+
+.. image:: images/comparison.png
+    :alt: Schematic
+    :align: center
+    :width: 400
+
+
+The following figure shows the mass of fluid 1 throughout the simulation.
```
blaisb, line 238: Maybe you can choose a better y axis scale? It's a bit too zoomed out.

source/solvers/navier_stokes_vof_assemblers.cc
blaisb: Yeah they should be removed no?

blaisb (changes requested):
Micro changes left, nothing big

blaisb:
Could also be good to add a comment in the example regarding the periodic bcs.

blaisb:
@shahabgol CI breaking is not your fault. Deal.II change. I'll need to fix that tonight

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
```diff
@@ -0,0 +1,252 @@
...
+
+which result in Reynolds and Atwood numbers equal to
+    .. math::
+        Re = \frac{\rho_h H \sqrt{H \bf{g} }}{\mu_h} = 256
+
+    .. math::
+        At = \frac{\rho_r - 1}{\rho_r + 1} = 0.5
+
+
+A perturbed interface defined as :math:`2H + 0.1 \cos{2 \pi x} / H` separates the fluids. At the top and bottom boundaries, no-slip boundary condition is applied, while on the left and right walls, periodic boundary condition is used. The temporal evolution of the interface is compared with the simulations of Garoosi and Hooman [2] at dimensionless times (:math:`t = t \sqrt{\bf{g} / H}`) of 1.5, 2.5, 3.5, 4.0 and 4.5.
```
lpsaavedra, line 55:
```suggestion
A perturbed interface defined as :math:`2H + 0.1 \cos{2 \pi x} / H` separates the fluids. At the top and bottom boundaries, a no-slip boundary condition is applied, while on the left and right walls, a periodic boundary condition is used. The temporal evolution of the interface is compared with the simulations of Garoosi and Hooman [2] at dimensionless times (:math:`t = t \sqrt{\bf{g} / H}`) of 1.5, 2.5, 3.5, 4.0 and 4.5.
```

blaisb:
I think this example also shows some of the limitation of the output control time. when the time step becomes super small, the linear solver becomes very innefficient. I'll have to double check that. :)

lpsaavedra (commented):
Great work, I love the Rayleigh-Taylor instability and this is a very cool example! I don't have much to say...

blaisb:
One last thing, I tested with the adaptive mass conservation. It works super well. So this could be just a slight appendix at the end to demonstrate this other possibility. There no issues and the sharpening threshold alternates between 0.498 and 0502....

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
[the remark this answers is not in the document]
shahabgol: Added.

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
```diff
@@ -0,0 +1,252 @@
...
+        set enable      		= true
+        set threshold		= 0.5
+        set interface sharpness	= 1.5
+        set frequency		= 25
+        set type        		= constant
+      end
+     
+      subsection mass conservation
+        set monitoring          	= true
+        set monitored fluid     	= fluid 1
```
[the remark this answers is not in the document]
shahabgol: I think it's better to keep them, since it is the first example that uses adaptive refinement.

examples/multiphysics/rayleigh_taylor_instability/rayleigh_taylor_instability.prm
shahabgol: Bruno modified this subsection.

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
shahabgol: Done.

applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat.output
```diff
@@ -13,19 +13,19 @@ L2 error phase : 0.00369306
...
-0.0600 4.051310e-05   0.0236         
+0.0200 3.071207e-05   0.9287         
+0.0400 3.079621e-05   0.9288         
+0.0600 3.079248e-05   0.9288         
  time  error_phase  
 0.0200 3.693057e-03 
-0.0400 3.693720e-03 
-0.0600 3.694369e-03 
+0.0400 3.693377e-03 
+0.0600 3.693698e-03
```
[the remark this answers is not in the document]
shahabgol: Done.

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
shahabgol: Applied.

shahabgol:
Thank you all for your constructive comments. 

- I have added two separate prm files for constant and adaptive interface sharpening.
- I have noticed this problem with the output control time. But I could not find the reason.
- The density ratio and phase cutoff calculations cannot be removed yet, since they are used to determine whether continuity equation should be solved or not (solve_continuity parameter). If you believe these parameters should also be removed, I think it's better to remove them in another PR, since the current PR already changes the results of many tests.
- A section is added to the documentation file to address the comment on periodic BC.

Thank you again for taking the time to review my PR. :)

blaisb:
Awesome. this is ready for merge for me. I will fix the issues with the repo myself.

applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat_mesh-adapt.output
```diff
@@ -17,47 +17,47 @@ Transient iteration : 2        Time : 0.04     Time step : 0.02     CFL : 8.8806
...
-0.0200 3.071207e-05   0.0541         
-0.0400 5.079275e-05   0.0189         
-0.0600 3.311897e-06   0.0198         
-0.0800 2.763905e-06   0.0198         
-0.1000 2.400827e-06   0.0198         
+0.0200 3.071207e-05   0.9287         
+0.0400 4.143914e-06   0.9285         
+0.0600 3.312462e-06   0.9283         
+0.0800 3.571638e-06   0.9284         
+0.1000 2.763693e-06   0.9283
```
blaisb, line 57: I think your analytical solution for pressure is still wrong...

applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat.prm
```diff
@@ -72,8 +72,8 @@ subsection analytical solution
   set enable     	= true
   set verbosity 	= verbose
     subsection uvwp
-	set Function constants = rho=2., g=9.81, h=0.4
-        set Function expression =  0 ; 0 ; if (y<h, rho*g*(h-y), 0)
+	set Function constants = rho1=2., g=9.81, h=0.4, rho2=0.5,l=1
+        set Function expression =  0 ; 0 ; if (y<h, rho2*g*(l-h) + rho1*g*(h-y), rho2*g*(l-y))
```
blaisb, line 76:
```suggestion
        set Function expression =  0 ; 0 ; if (y<h, rho2*g*(l-h) + rho1*g*(h-y), rho2*g*(l-h))
```
You also need to define it for y>h because now it's not zero.

blaisb (approved):
Can you just fix the analytical solution for pressure and we should be good. The output seems to be wrong.

doc/source/examples/multiphysics/rayleigh-taylor-instability/rayleigh-taylor-instability.rst
```diff
@@ -0,0 +1,310 @@
...
+Parameter file
+--------------
+
+Time integration is handled by a 1st order backward differentiation scheme 
+`(bdf1)`, for a :math:`0.75` s simulation time with an initial 
+time step of :math:`0.0002` second.
+
+.. note::   
+    This example uses an adaptive time-stepping method, where the 
+    time-step is modified during the simulation to keep the maximum value of the CFL condition below a given threshold (0.5 here). Using ``output control = time``, and ``output time = 0.005`` the simulation results are written every 0.005 s.
```
lpsaavedra, line 70:
```suggestion

.. note::   
    This example uses an adaptive time-stepping method, where the 
    time-step is modified during the simulation to keep the maximum value of the CFL condition below a given threshold (0.9 here). Using ``output control = time``, and ``output time = 0.005`` the simulation results are written every 0.005 s.
```

lpsaavedra (approved):

PR #575, opened 2022-08-29
Add initial adaptive refinement

shahabgol:
# Description of the problem

- Choosing a large number of mesh initial refinement makes the code very slow in the first step (before adaptive refinement can make the grids larger in unnecessary regions).

# Description of the solution

- Adds initial (pre-solve) adaptive mesh refinement.

### DIFF
--- a/source/core/parameters.cc
+++ b/source/core/parameters.cc
@@ -1531,6 +1531,11 @@ namespace Parameters
   {
     prm.enter_subsection("mesh adaptation");
     {
+      prm.declare_entry("initial refinement steps",
+                        "0",
+                        Patterns::Integer(),
+                        "Number of pre-solve adaptive mesh refinement steps");
+
       prm.declare_entry("type",
                         "none",
                         Patterns::Selection("none|uniform|kelly"),
@@ -1582,6 +1587,8 @@ namespace Parameters
   {
     prm.enter_subsection("mesh adaptation");
     {
+      initial_refinement = prm.get_integer("initial refinement steps");
+
       const std::string op = prm.get("type");
       if (op == "none")
         type = Type::none;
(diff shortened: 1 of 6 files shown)


tonielgeitani (approved):

source/core/parameters.cc
```diff
@@ -1531,6 +1531,11 @@ namespace Parameters
   {
     prm.enter_subsection("mesh adaptation");
     {
+      prm.declare_entry("initial refinement steps",
```
Luckabarbeau, line 1534: Number of initial refinement step would be better I think

Luckabarbeau (commented):
very nice and short Pr i would also implement this in all the other solver

source/solvers/gd_navier_stokes.cc
```diff
@@ -1215,7 +1215,17 @@ GDNavierStokesSolver<dim>::solve()
...
+        {
+          for (unsigned int i = 0;
+               i <
+               this->simulation_parameters.mesh_adaptation.initial_refinement;
+               i++)
+            NavierStokesBase<dim,
+                             TrilinosWrappers::MPI::BlockVector,
+                             std::vector<IndexSet>>::refine_mesh();
+
+          this->iterate();
```
blaisb, line 1227: This is a good idea really, but I suggest you move it to the initial condition. This way, for example, you could re-apply the initial condition between every mesh refinement. In addition, this way every solver would have this capability automatically.
I really suggest moving it to initial conditions in the NavierStokesBase and to reapply the initial condition every time.

blaisb (changes requested):
One major suggestion + don't forget to add documentation to .rst file :)

source/core/parameters.cc
shahabgol: I think a little bit long for a parameter name

blaisb (approved):

PR #577, opened 2022-08-30
fix timer scope vans

tonielgeitani:
# Description of the problem

-Timer Output was not adding up to the total value.

# Description of the solution

The setup preconditioner was called inside the scope of the assemble matrix. It had to be called outside of it.

# How Has This Been Tested?

- No tests needed.

# Documentation

- N/A

# Future changes

-N/A

# Comments

- N/A


shahabgol (approved):
I couldn't find the difference, but the changed lines seem good to me.

blaisb (approved):

PR #582, opened 2022-09-07
Doc win install

jeannejoachim:
The existing documentation may be difficult to follow for typical windows users (as my industrial partner) + we found with @voferreira that some changes were necessary to install deal.ii with WSL. This PR addressed this concern.

Another small fix: I am always on black mode and the Lethe logo did not display well, so I simply added a white background.

# Comments

I would advise you to ``make html`` in a local repository to make your review easier and better see the changes!


doc/source/installation/windows_wsl.rst
```diff
@@ -0,0 +1,254 @@
...
+	+--------+------------------------------------------------+-----------------------------------------------+
+
+	* save and close 
+
+6. |linux_shell| Still in the candi subfolder, run candi installation:
+	* for a 8Gb RAM computer:
+
+	.. code-block:: text
+
+		./candi.sh -j 1
```
voferreira, line 152:
```suggestion
		./candi.sh -j1
```

doc/source/installation/windows_wsl.rst
```diff
@@ -0,0 +1,254 @@
...
+
+	* for a 16Gb (and more) RAM computer:
+
+	.. code-block:: text
+
+		./candi.sh -j 4
```
voferreira, line 158:
```suggestion
		./candi.sh -j4
```

doc/source/installation/windows_wsl.rst
```diff
@@ -0,0 +1,254 @@
...
+
+.. tip::
+
+	Candi will print messages asking you if you installed the dependency. Hit ``Enter`` two times to validate and the installation will launch. If new lines are written in the console, this means the installation is correctly going on. The installation will take from 1 to 3 hours depending on your hardware.
```
voferreira, line 162:
```suggestion
	Candi will print messages asking you if you installed the dependency. Hit ``Enter`` two times to validate and the installation will launch. If new lines are written in the console, this means the installation is going on correctly. The installation will take from 1 to 3 hours depending on your hardware.
```

doc/source/installation/windows_wsl.rst
```diff
@@ -0,0 +1,254 @@
...
+
+	If the installation is stuck (no change on the console for a few minutes), hitting ``Ctrl+C`` can unstuck it.
```
voferreira, line 164: I think hitting enter is safer and works the same way.

doc/source/installation/windows_wsl.rst
```diff
@@ -0,0 +1,254 @@
...
+	cd build
+	cmake ../git -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=../inst/
+
+4. |linux_shell| Compile lethe:
+
+.. code-block:: text
+
+	make -j$numprocs
+
+Where ``$numprocs`` corresponds to the number of processors used for the compilation. As a rule of thumb, this can be equal to the number of physical cores on your computer minus 1, so for a 6 physical cores computer, the command is: ``make -j5``
```
voferreira, line 216: This same comment couldn't be up there before line 147?

voferreira (commented):
This is very nice, Jeanne! Very good job! There are some comments here and there, but it is almost ready to go.

doc/source/installation/windows_wsl.rst
jeannejoachim: I honestly don't know... I have 6 physical proc (12 logical) and I always compile Lethe on 5 (following Bruno's advice), but I never tried to compile candi on more than 4 proc, as stated for a "16Gb RAM computer" in the documentation ^^"

jeannejoachim: That's why in the candi part I only said "-j1" or "-j4" depending on the RAM. Here I state that it depends on the number of physical cores. But if you have further knowledge/advice about that I would be happy to change this information :)

voferreira: I think it is better to ask @blaisb here. I always use the number of physical procs on my machine, but I have 32 gb, so I don't know if it would crash with it. I think the recommendation should be the same both for candi and lethe compilation.

voferreira (approved):
Great job, Jeanne! Apart from the uniformization of the -j recommendation, I have nothing else to add.

doc/source/installation/windows_wsl.rst
blaisb: It's a tricky question I guess. It depends on the amount of ram you have on your machine. Anything above 16 gig I would always use number of physical core - 1 (or even number of physical cores, on my machine i do -j16 for example).
At 16 gig, I don't think this creates an issue. I thought I had seen in the past sometimes that Trilinos struggles with 16 Gb of ram, but I have never been able to reproduce this lately.
At 8gig, it's really important not to use more than 2 cores or so, otherwise it can just stall because of an out of ram error.

So I guess I would say : Number of physical cores -1 if you have 16GB and above. Otherwise, limit yourself to 2 at most

blaisb (approved):
Just homogeneize the proc thing and this is good to go. It's very clear. I like this.

jeannejoachim:
Good to go :) thanks a ton for the fast reviews! :smile_cat:

PR #583, opened 2022-09-07
fix erros in gidaspow and beetstra models

tonielgeitani:
# Description of the problem

- Errors in the description of the drag for Beetstra and Gidaspow.

# Description of the solution

- Gidaspaw model calculates the drag per cell. We thus don't have to su mthe beta over all particles as it was already summed - Return to the normalized form of the force, but evaluate the individual particle force based on the formulation in the article

# How Has This Been Tested?

- N/A

# Documentation

- N/A

# Future changes

- N/A

# Comments

- N/A

### DIFF
--- a/include/solvers/navier_stokes_scratch_data.h
+++ b/include/solvers/navier_stokes_scratch_data.h
@@ -739,13 +739,17 @@ public:
 
     // Relative velocity and particle Reynolds
     {
-      unsigned int particle_no = 0;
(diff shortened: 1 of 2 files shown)


voferreira (approved):
It is good to go, I think. In my case, it was close to the previous results I had with Rong. I like the way you put the average relative velocity :)

blaisb (approved):
I'm happy with those changes. Good job @tonielgeitani

PR #585, opened 2022-09-09
Parameters for multivariable mesh adaptation

jeannejoachim:
# Description of the problem

For now, only one parameter can be parsed for mesh adaptation

# Description of the solution

Create parameters to be used for mesh adaptation with multiple variables. Multiple parameters can be parsed separated by a comma, for instance:

<pre><code>
   set types = kelly,kelly
   set variables = velocity,temperature
   set fractions coarsening = 0.2,0.5
   set fractions refinement = 0.3,0.4
</code></pre>

This created a map `mesh_adaptation.variables`, with :
- `first` being the adaptation variable (like `Variable::velocity`), and 
- `second` a structure `MultipleAdaptationParameters`, with its element `type`, `coarsening_fraction` and `refinement_fraction`

For now the parameters name has been put in plural (`variables etc`) to prevent any modification of the existing code (with `variable`) but this can be renamed when the mechanism is implemented.

# How Has This Been Tested?

Locally on a prm file. All configurations and assertions work as expected.

# Documentation

None for now, will be updated when implemented.

# Future changes

For now, the mono-variable mesh adaptation is the only implemented, so these parameters are not used. This will be implemented in the next PR about this issue.

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -31,11 +31,12 @@
 #ifndef lethe_parameters_h
 #define lethe_parameters_h
 
+#include <core/ib_particle.h>
+#include <core/multiphysics.h>
+
 #include <deal.II/base/conditional_ostream.h>
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/base/parsed_function.h>
-
-#include <core/ib_particle.h>
 
 
 using namespace dealii;
@@ -831,6 +832,18 @@ namespace Parameters
   };
 
   /**
+   * @brief Mesh Adaptation Parameters that can differ for each refinement variable
+   */
+  struct MultipleAdaptationParameters
+  {
+    // Coarsening fraction
+    double coarsening_fraction;
+
+    // Refinement fraction
+    double refinement_fraction;
+  };
+
+  /**
    * @brief MeshAdaption - Parameters that control dynamic mesh adaptation.
    * Dynamic mesh adaptation in Lethe is very flexible and can be both local
    * and global.
@@ -853,9 +866,14 @@ namespace Parameters
       velocity,
       pressure,
       phase,
-      temperature,
-      velocity_temperature,
+      temperature
     } variable;
+
+    // Map containing the refinement variables
+    std::map<Variable, MultipleAdaptationParameters> variables;
+    // declaration for parsing variables
+    Variable                     vars;
+    MultipleAdaptationParameters var_adaptation_param;
 
     // Decision factor for Kelly refinement (number or fraction)
     enum class FractionType
@@ -875,12 +893,6 @@ namespace Parameters
 
     // Refinement after frequency iter
     unsigned int frequency;
-
-    // Refinement fraction
-    double refinement_fraction;
-
-    // Coarsening fraction
-    double coarsening_fraction;
 
     static void
     declare_parameters(ParameterHandler &prm);
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -31,11 +31,11 @@
 #ifndef lethe_parameters_h
 #define lethe_parameters_h
 
-#include <core/ib_particle.h>
-
 #include <deal.II/base/conditional_ostream.h>
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/base/parsed_function.h>
+
+#include <core/ib_particle.h>
 
 
 using namespace dealii;
@@ -853,7 +853,8 @@ namespace Parameters
       velocity,
       pressure,
       phase,
-      temperature
+      temperature,
+      velocity_temperature,
     } variable;
 
     // Decision factor for Kelly refinement (number or fraction)
--- a/source/core/parameters.cc
+++ b/source/core/parameters.cc
@@ -1,6 +1,6 @@
-#include <deal.II/base/exceptions.h>
-
 #include <core/parameters.h>
+
+#include <deal.II/base/exceptions.h>
 
 DeclException2(
   PhaseChangeIntervalError,
@@ -21,6 +21,14 @@ DeclException1(
   unsigned int,
   << "Laser beam orientation in : " << arg1
   << "-dimensional simulations cannot be defined in the z direction");
+
+DeclException3(MultipleAdaptationSizeError,
+               std::string,
+               unsigned int,
+               unsigned int,
+               << "Error in 'mesh adaptation' : number of '" << arg1 << "' ("
+               << arg2 << ") does not correspond to the number of 'variables' ("
+               << arg3 << ")");
 
 namespace Parameters
 {
@@ -1526,6 +1534,7 @@ namespace Parameters
     }
     prm.leave_subsection();
   }
+
   void
   MeshAdaptation::declare_parameters(ParameterHandler &prm)
   {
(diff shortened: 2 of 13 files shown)


source/core/parameters.cc
```diff
@@ -1548,6 +1558,39 @@ namespace Parameters
                           "velocity|pressure|phase|temperature"),
                         "Variable for kelly estimation"
                         "Choices are <velocity|pressure|phase|temperature>.");
+      // end WIP
+      // WIP declare multivariable parameters
+      prm.declare_entry(
+        "types",
```
Luckabarbeau, line 1564: There will be only one type of refinement and not multiple. For a given variable you have kelly, none, uniform. If you have a uniform for one variable you will have the equivalent of a uniform for all variables. If you have none for a given variable you will just not have this variable in the list of variables. I would keep the original type of refinement.

include/core/parameters.h
```diff
@@ -830,6 +833,26 @@ namespace Parameters
     parse_parameters(ParameterHandler &prm);
   };
 
+  /**
+   * @brief Mesh Adaptation Parameters that can differ for each refinement variable
+   */
+  struct MultipleAdaptationParameters
+  {
+    // Type of mesh adaptation
+    enum class Type
```
Luckabarbeau, line 842: See comments bellow

source/core/parameters.cc
```diff
@@ -1607,6 +1653,79 @@ namespace Parameters
...
+          else if (var_vec[i] == "phase")
+            vars = Variable::phase;
+          else if (var_vec[i] == "temperature")
+            vars = Variable::temperature;
+          else
+            {
+              std::cerr
+                << "  WARNING skipping unknown variable in 'subsection mesh adaptation':"
+                << std::endl
+                << var_vec[i] << std::endl;
```
Luckabarbeau, line 1698: I would just throw here instead of giving a warning.

source/core/parameters.cc
```diff
@@ -1607,6 +1653,79 @@ namespace Parameters
...
+          else if (type_vec[i] == "uniform")
+            var_adaptation_param.type =
+              MultipleAdaptationParameters::Type::uniform;
+          else if (type_vec[i] == "kelly")
+            var_adaptation_param.type =
+              MultipleAdaptationParameters::Type::kelly;
+          else
+            {
+              std::cerr
+                << "  WARNING unknown type in 'subsection mesh adaptation':"
```
Luckabarbeau, line 1713: same as above

Luckabarbeau (commented):
Good job this will be a very nice feature once it's done. The only comment is about the list of types of refinement, see comments related to this.

source/core/parameters.cc
jeannejoachim: and well, I still need to implement the "uniform" type, for it to be usable also :)

blaisb: I would rather keep one type (kelly, uniform, none) for all variables. Otherwise it will get confusing. I understand what you mean about turning on or off, but this will be better accomplished by commenting / uncommenting the 3-4 lines that defines the variables and their parameter instead.

source/core/parameters.cc
blaisb: Yup, throwing is a better option here.

include/core/parameters.h
```diff
@@ -31,12 +31,15 @@
...
-
 #include <deal.II/base/conditional_ostream.h>
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/base/parsed_function.h>
 
+//#include <deal.II/lac/vector.h>
+
+#include <core/ib_particle.h>
+#include <core/multiphysics.h>
+
```
blaisb, line 42: Can you remove commented include?
Also, Lethe includes should be above the deal.II includes generally. The automatic indentation should take care of that.

include/core/parameters.h
```diff
@@ -840,13 +863,15 @@ namespace Parameters
     // Initial adaptive refinement
     unsigned int initial_refinement;
 
+    // WIP remove when multivariable is implemented
```
blaisb, line 866:
```suggestion
```

include/core/parameters.h
```diff
@@ -840,13 +863,15 @@ namespace Parameters
...
     // Type of mesh adaptation
     enum class Type
     {
       none,
       uniform,
       kelly
     } type;
+    // end WIP
```
blaisb, line 874:
```suggestion
```

include/core/parameters.h
```diff
@@ -856,6 +881,20 @@ namespace Parameters
       temperature
     } variable;
 
+    // WIP Map containing the refinement variables
+    std::map<Variable, MultipleAdaptationParameters> variables;
+    // declaration for parsing variables
+    Variable                     vars;
+    MultipleAdaptationParameters var_adaptation_param;
```
blaisb, line 888: I know this is something you want to remove eventually, but you could actually remove it right now by already using the key of the map (assuming there is only one). This way generallisation will be quasi-done.

include/core/parameters.h
```diff
@@ -856,6 +881,20 @@ namespace Parameters
...
+    // end WIP
+
+    // WIP remove when multivariable is implemented
+    // Coarsening fraction
+    double coarsening_fraction;
+    // Refinement fraction
+    double refinement_fraction;
+    // end WIP
```
blaisb, line 896: This could already be removed by using the values of the struct in the first element of the map.

include/core/parameters.h
```diff
@@ -875,16 +914,28 @@ namespace Parameters
...
-
     static void
     declare_parameters(ParameterHandler &prm);
     void
     parse_parameters(ParameterHandler &prm);
+
+    // @brief Splitting method, used to fill the map 'variables'
+    // @input variables, string with mesh adaptation variables separated by a
+    // comma
+    std::vector<std::string>
```
blaisb, line 925: You don't need a function for this. There is already one in deal.II
https://github.com/lethe-cfd/lethe/blob/master/source/core/parameters_lagrangian.cc#L362

source/core/parameters.cc
```diff
@@ -1548,6 +1558,39 @@ namespace Parameters
                           "velocity|pressure|phase|temperature"),
                         "Variable for kelly estimation"
                         "Choices are <velocity|pressure|phase|temperature>.");
+      // end WIP
+      // WIP declare multivariable parameters
+      prm.declare_entry(
+        "types",
+        "none",
+        Patterns::Anything(),
```
blaisb, line 1566: It is not a good idea to use Patterns::Anything. There is a deal.II pattern for this, we already use it:
https://github.com/lethe-cfd/lethe/blob/master/source/core/parameters_lagrangian.cc#L313

source/core/parameters.cc
```diff
@@ -1607,6 +1653,79 @@ namespace Parameters
         variable = Variable::phase;
       if (vop == "temperature")
         variable = Variable::temperature;
+      // end WIP
+
+      // WIP Create map of refinement variables
+      const std::string        var_op    = prm.get("variables");
+      std::vector<std::string> var_vec   = split_input_variables(var_op);
+      const std::string        type_op   = prm.get("types");
+      std::vector<std::string> type_vec  = split_input_variables(type_op);
```
blaisb, line 1662: Use dealII version of the split function

source/core/parameters.cc
```diff
@@ -1617,8 +1736,11 @@ namespace Parameters
       maximum_refinement_level = prm.get_integer("max refinement level");
       minimum_refinement_level = prm.get_integer("min refinement level");
       frequency                = prm.get_integer("frequency");
-      coarsening_fraction      = prm.get_double("fraction coarsening");
-      refinement_fraction      = prm.get_double("fraction refinement");
+
+      // WIP remove when multivariable is implemented
+      coarsening_fraction = prm.get_double("fraction coarsening");
+      refinement_fraction = prm.get_double("fraction refinement");
+      // end WIP
```
blaisb, line 1743: This can already be removed here

blaisb (changes requested):
Some changes to be done:
- Remove the previous variables, you can already start using the map by "assuming" is only has one element in it
- Use the deal.II functions to split the list
- Use the deal.II format for lists instead of using an Anything, that's a dangerous approach because it won't sanity check.
-

source/core/parameters.cc
jeannejoachim: I was searching for that, thank you! Just a note that these `list x`, `y` and `z` are missing from the documentation (https://lethe-cfd.github.io/lethe/parameters/dem/insertion_info.html)

source/solvers/navier_stokes_base.cc
```diff
@@ -865,52 +865,58 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_kelly()
...
+          ivar.second.coarsening_fraction,
+          this->simulation_parameters.mesh_adaptation.maximum_number_elements);
+
+      else if (this->simulation_parameters.mesh_adaptation.fractionType ==
+               Parameters::MeshAdaptation::FractionType::fraction)
+        parallel::distributed::GridRefinement::
+          refine_and_coarsen_fixed_fraction(tria,
+                                            estimated_error_per_cell,
+                                            ivar.second.refinement_fraction,
+                                            ivar.second.coarsening_fraction);
```
blaisb, line 918: This will eventually need to be moved outside but it's ok for now. It will be done shortly anyway in another PR.

blaisb (approved):
Good to go. Merging

PR #590, opened 2022-09-16
Update name of section of the location of the examples in doc

lpsaavedra:
# Description of the problem

All the examples in the documentation had a section called  "Location of the examples"  with the exception of three cases.

# Description of the solution

The section was renamed for all the examples to"Location of the files used in this example".

# Comments

We still need to agree on the capitalization of the titles in the documentation... as all the rpt examples and two examples in the incompressible flow examples (the 3D mixer and ribbon mixer) have all the titles capitalized. We could agree on something and correct it in this PR).


doc/source/examples/dem/granular-mixer/granular-mixer.rst
```diff
@@ -12,8 +12,8 @@ Features
 - Bidispersed particles (same size and properties, but different types)
 
 
-Location of the examples
-------------------------
+Location of the files used in this example
+--------------------------------------------
 ``/examples/dem/3d-granular-mixer/granular-mixer.prm``
```
blaisb, line 17: I think we should remove the prm right and just point to the folder itself no?

blaisb (commented):
Do we want to give the file path total, or just the path to the folder which should be enough? I feel like just giving the path would be enough no?

lpsaavedra:
The titles were updated according to our talk!

PR #591, opened 2022-09-17
Mesh adapt multivar

jeannejoachim:
Implement generalized multiple variable mesh adaptation.

# How Has This Been Tested?

Test added with mesh adapt on phase and temperature.

# Documentation

Documentation updated (+ corrected a warning in an example).

### DIFF
--- a/applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls_multivar_adapt.output
+++ b/applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls_multivar_adapt.output
@@ -15,7 +15,7 @@ Transient iteration : 1        Time : 0.2      Time step : 0.2      CFL : 0
 *****************************************************************
 L2 error velocity : 0
 L2 error temperature : 0.417811
-L2 error phase : 0.195496
+L2 error phase : 0.00208462
 
 *****************************************************************
 Transient iteration : 2        Time : 0.4      Time step : 0.2      CFL : 0       
@@ -27,7 +27,7 @@ Transient iteration : 2        Time : 0.4      Time step : 0.2      CFL : 0
    Number of VOF degrees of freedom: 775
 L2 error velocity : 0
 L2 error temperature : 0.0955313
-L2 error phase : 0.195753
+L2 error phase : 0.00208364
 
 *****************************************************************
 Transient iteration : 3        Time : 0.6      Time step : 0.2      CFL : 0       
@@ -39,7 +39,7 @@ Transient iteration : 3        Time : 0.6      Time step : 0.2      CFL : 0
    Number of VOF degrees of freedom: 873
 L2 error velocity : 0
 L2 error temperature : 0.0239307
-L2 error phase : 0.195743
+L2 error phase : 0.00208379
 
 *****************************************************************
 Transient iteration : 4        Time : 0.8      Time step : 0.2      CFL : 0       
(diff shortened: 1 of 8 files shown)


applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls_multivar_adapt.output
```diff
@@ -0,0 +1,84 @@
...
+688   9.553132e-02      
+784   2.393069e-02      
+784   5.409094e-03      
+736   1.090529e-03      
+ time  error_phase  
+0.2000 1.954961e-01 
+0.4000 1.957534e-01 
+0.6000 1.957427e-01 
+0.8000 1.957411e-01 
+1.0000 1.957406e-01
```
blaisb, line 84: Is there a reason why the error is so big here?

source/solvers/navier_stokes_base.cc
```diff
@@ -865,6 +865,14 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_kelly()
   const FEValuesExtractors::Scalar pressure(dim);
   auto &                           present_solution = this->present_solution;
 
+  // Global flags
+  // Their dimension is consistent with the dimension returned by
+  // save_refine_flags(), in order to be able to use load_refine_flags()
+  std::vector<bool> global_refine_flags(dim * tria.n_active_cells(), false);
+  std::vector<bool> global_coarsen_flags(dim * tria.n_active_cells(), false);
+
+  bool first_variable(true);
```
blaisb, line 874: What you could also do is not give a size to the vector and use their size to check if it's been filled or no, this would save you the use of another variable, but it's maybe more readable in the current fashion.

blaisb (approved):

applications_tests/gls_navier_stokes_2d/gls_vof_ht_heated-walls_multivar_adapt.prm
```diff
@@ -0,0 +1,189 @@
...
+  set verbosity  = verbose
+  subsection temperature
+    set Function constants = k1=2, k0=0.1, T1=5, T0=1, L=0.5
+    #temperature for fluid 1 (x<=0) : T = k0*(T0-T1)/(L*(k1+k0)) * x + T1+(T0-T1)*k0/(k1+k0)
+    #temperature for fluid 0 (x>0) : T = k1*(T0-T1)/(L*(k1+k0)) * x + T1+(T0-T1)*k0/(k1+k0)
+    set Function expression = if(x<=0, k0*(T0-T1)/(L*(k1+k0)) * x + T1+(T0-T1)*k0/(k1+k0), k1*(T0-T1)/(L*(k1+k0)) * x + T1+(T0-T1)*k0/(k1+k0) )
+  end  
+  subsection phase
+    set Function expression = if(x<=0,0,1)
+  end
```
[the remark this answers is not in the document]
jeannejoachim: Oh wow good catch! I will modify the original test on which I based this one also

source/solvers/navier_stokes_base.cc
jeannejoachim: Yes I tried not using another variable at first but I liked that using a simple boolean made this extra clear

blaisb: Cool with me

PR #592, opened 2022-09-19
Add rbf sharp

oguevremont:
# Description of the problem

- Simulations based on arbitrary shapes described by grids had to be done using the Nitsche solver;
- The high precision of the Sharp solver couldn't be used for complex shapes.
 
# Description of the solution

- RBF (Radial basis function) objects can now be used to describe complex objects with a clear definition, which allows the use of the Sharp solver;
- These objects are described in a file of 6 columns containing, for each node: weight, support radius, basis function, x, y, z. Lethe can now read these files and consider those parameters to solve the flow around complex objects;
- The RBF text files are generated using a script that will be imminently added to the public domain. This script will have to be run before the Lethe executable. 

# How Has This Been Tested?

- [gls_sharp_navier_stokes_3d/flow_around_rbf] This tests the flow around a complex object containing one hole, one cone tip and several sharp edges;
- There is no 2D test since the RBF generation script only works in 3D for the moment.

# Documentation

- The Resolved CFD-DEM parameters page was updated with the new geometry type.

# Future changes

- The RBF generation script will be made public, and the way to use it will be explained;
- One example using the RBF generation script and the Sharp solver will be made.


include/core/shape.h
```diff
@@ -531,4 +533,139 @@ class CompositeShape : public Shape<dim>
...
+
+  std::shared_ptr<Shape<dim>>
+  static_copy() const override;
+
+  double
+  displaced_volume(const double fluid_density) override;
+
+  double
+  // Inspired by Optimad Bitpit. https://github.com/optimad/bitpit
+  evaluate_basis_function(const unsigned int basis_function_id,
```
paulapatience, line 602: Use the scoped enum type, not a raw `unsigned int`.

include/core/shape.h
```diff
@@ -531,4 +533,139 @@ class CompositeShape : public Shape<dim>
...
+  double
+  c2c1(double) const;
+  double
+  c0c2(double) const;
+  double
+  c1c2(double) const;
+  double
+  c2c2(double) const;
+
+  std::pair<std::string, int>
```
paulapatience, line 632: Here also, why not `enum ShapeType` rather than `int`?

include/core/shape.h
```diff
@@ -531,4 +533,139 @@ class CompositeShape : public Shape<dim>
...
+    GAUSS95    = 4,
+    GAUSS99    = 5,
+    C1C0       = 6,
+    C2C0       = 7,
+    C0C1       = 8,
+    C1C1       = 9,
+    C2C1       = 10,
+    C0C2       = 11,
+    C1C2       = 12,
+    C2C2       = 13,
```
paulapatience, line 659: Enums are numbered from 0 by default, so providing these values is optional.

source/core/shape.cc
```diff
@@ -495,6 +495,353 @@ CompositeShape<dim>::displaced_volume(const double fluid_density)
...
+}
+
+/*!
+ * Wendland C2 function
+ * @param[in] dist distance normalized with respect to support radius
+ * @return rbf value
+ */
+template <int dim>
+double
+RBFShape<dim>::wendlandc2(double dist) const
```
paulapatience, line 556: The usage of const by-value parameters is inconsistent. Here `dist` isn't modified but is also not declared const, whereas various other functions do have const by-value parameters.

paulapatience (changes requested):

applications_tests/gls_sharp_navier_stokes_3d/RBF_test_shape.txt
```diff
@@ -0,0 +1,4097 @@
+support_radius;basis_function;xnode;ynode;znode;weight
```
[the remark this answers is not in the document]
oguevremont: That's the way I output the data from bitpit, but it shouldn't be a problem to output in CSV instead. I'm not used to TSV, but let me check.

include/core/shape.h
```diff
@@ -531,4 +533,139 @@ class CompositeShape : public Shape<dim>
...
+{
+public:
+  /**
+   * @param support_radius the scaling of the reach of the nodes
+   * @param basis_function the basis function that was used to parametrize the RBF object
+   * @param weight the weighting associated to each node for the sum operation
+   * @param nodes the center of each basis function
+   */
+  RBFShape<dim>(const std::vector<double>         support_radius,
+                const std::vector<unsigned int>   basis_function,
```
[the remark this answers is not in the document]
blaisb: Is the underlying enum an enum class? I'm always afraid of using int and enum interchangeably, If it's a vector of the enum, then let it be a vector of the enum and not a vector of unsigned int.

include/core/shape.h
blaisb: Also, since you take the time specify them as enum class, always use the enum class itself and never integer values.

source/core/ib_particle.cc
```diff
@@ -215,6 +215,41 @@ IBParticle<dim>::initialize_shape(const std::string         type,
                                                  position,
                                                  orientation);
     }
+  else if (type == "rbf")
+    {
+      if constexpr (dim == 3)
```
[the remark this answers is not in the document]
blaisb: Also why would this be only supported for dim==3?

source/core/shape.cc
```diff
@@ -495,6 +495,353 @@ CompositeShape<dim>::displaced_volume(const double fluid_density)
...
+  return copy;
+}
+
+template <int dim>
+double
+RBFShape<dim>::displaced_volume(const double fluid_density)
+{
+  std::cout
+    << "Warning: For a RBF shape, the real volume will be lower than output."
+    << std::endl;
```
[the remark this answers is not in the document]
blaisb: @paulapatience is right, but this is why we never write to std::cout. If you write to std::cout in parallel you just get a cluster fuck. Either throw if this is an error or use the parallel terminal writing capabilities.

source/core/shape.cc
blaisb: Should be const here.

source/core/shape.cc
```diff
@@ -495,6 +495,353 @@ CompositeShape<dim>::displaced_volume(const double fluid_density)
...
+}
+
+/*!
+ * Linear function
+ * @param[in] dist distance normalized with respect to support radius
+ * @return rbf value
+ */
+template <int dim>
+double
+RBFShape<dim>::linear(double dist) const
```
blaisb, line 578:
```suggestion
RBFShape<dim>::linear(const double dist) const
```

source/core/shape.cc
```diff
@@ -495,6 +495,353 @@ CompositeShape<dim>::displaced_volume(const double fluid_density)
...
+}
+
+/*!
+ * Non compact gaussian function with 0.1 value at dist equal to 1
+ * @param[in] dist distance normalized with respect to support radius
+ * @return rbf value
+ */
+template <int dim>
+double
+RBFShape<dim>::gauss90(double dist) const
```
blaisb, line 590:
```suggestion
RBFShape<dim>::gauss90(const double dist) const
```

source/core/shape.cc
```diff
@@ -495,6 +495,353 @@ CompositeShape<dim>::displaced_volume(const double fluid_density)
...
+    {
+      return (1. - dist);
+    }
+}
+
+/*!
+ * Non compact gaussian function with 0.1 value at dist equal to 1
+ * @param[in] dist distance normalized with respect to support radius
+ * @return rbf value
+ */
```
blaisb, line 587: Fix the brief, this is copy pasted....

blaisb (commented):
Just some comments there and there before I fully review.
- Please review all of your briefs
- Check all function parameters. If they don't need to be modified make them const. This helps people understand your code afterwards.

include/core/shape.h
oguevremont: Good to know! And yes I'll make the change

source/core/ib_particle.cc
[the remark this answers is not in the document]
oguevremont: The bitpit library doesn't allow 2D for all their objects, and that's the case for one that is used in the RBF generation script. At the moment the script can't produce 2D objects, but in theory this limitation wouldn't interfere Lethe. I thought we shouldn't allow using 2D RBF shapes since we can't produce them, but I can allow them if you think it would be pertinent.

include/core/shape.h
```diff
@@ -531,4 +533,138 @@ class CompositeShape : public Shape<dim>
...
+  double
+  displaced_volume(const double fluid_density) override;
+
+  double
+  // Inspired by Optimad Bitpit. https://github.com/optimad/bitpit
+  evaluate_basis_function(const unsigned int basis_function_id,
+                          const double       distance) const;
+
+  double
+  wendlandc2(double) const;
```
hepap, line 605: Following the review comments in shape.cc

```suggestion
  wendlandc2(const double) const;
```

source/core/shape.cc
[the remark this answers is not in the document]
oguevremont: I tried with parallel output, but I am not convinced yet by the structure. 
I could put the warning as a comment, or throw an error since displaced_volume() is only used to integrate particle motion, and we can only do it when particles are spherical.

applications_tests/gls_sharp_navier_stokes_3d/RBF_test_shape.txt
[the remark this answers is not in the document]
oguevremont: I changed to CSV in bitpit and Lethe

source/core/parameters.cc
```diff
@@ -2200,8 +2199,65 @@ namespace Parameters
...
-            Utilities::string_to_double(shape_arguments_str_list);
+          std::vector<double> shape_arguments;
+          if (shape_type == "rbf")
+            {
+              if constexpr (dim == 3)
+                {
+                  std::string shape_name = shape_arguments_str_list[0];
+                  std::map<std::string, std::vector<double>> rbf_data;
+                  fill_vectors_from_file(rbf_data, shape_name, ",");
+                  size_t number_of_nodes = rbf_data["weight"].size();
```
blaisb, line 2210: Add comments to explain waht you are doing.

blaisb: Nice usage of the function Lucka wrote though. Very very cool.

source/core/parameters.cc
```diff
@@ -2200,8 +2199,65 @@ namespace Parameters
...
+
+                  for (size_t n_i = 0; n_i < number_of_nodes; n_i++)
+                    {
+                      weights[n_i]         = rbf_data["weight"][n_i];
+                      support_radii[n_i]   = rbf_data["support_radius"][n_i];
+                      basis_functions[n_i] = rbf_data["basis_function"][n_i];
+                      nodes_x[n_i]         = rbf_data["xnode"][n_i];
+                      nodes_y[n_i]         = rbf_data["ynode"][n_i];
+                      nodes_z[n_i]         = rbf_data["znode"][n_i];
+                    }
```
blaisb, line 2234: Is this intermediary generation of a vector necessary? I Mean it's not a big deal, but at least please comment what you are doing.

source/core/parameters.cc
```diff
@@ -2200,8 +2199,65 @@ namespace Parameters
...
             Utilities::split_string_list(shape_arguments_str, ";"));
-          std::vector<double> shape_arguments =
-            Utilities::string_to_double(shape_arguments_str_list);
+          std::vector<double> shape_arguments;
+          if (shape_type == "rbf")
+            {
+              if constexpr (dim == 3)
+                {
+                  std::string shape_name = shape_arguments_str_list[0];
+                  std::map<std::string, std::vector<double>> rbf_data;
```
blaisb, line 2208: I don't understand. When you declare a default entry, why are you filling from a file like this? It does not make sense no?
I feel like this bunch of code is repeated lower when you actually do the parsing?

source/core/shape.cc
```diff
@@ -171,7 +172,8 @@ Sphere<dim>::gradient(const Point<dim> &evaluation_point,
 
 template <int dim>
 double
-Sphere<dim>::displaced_volume(const double fluid_density)
+Sphere<dim>::displaced_volume(const double fluid_density,
+                              std::shared_ptr<ConditionalOStream> /*pcout*/)
```
blaisb, line 176: Why are you passing the pcout there?

source/core/shape.cc
```diff
@@ -392,22 +400,28 @@ template <int dim>
...
-       "output."
-    << std::endl;
+CutHollowSphere<dim>::displaced_volume(
+  const double /*fluid_density*/,
+  std::shared_ptr<ConditionalOStream> pcout)
+{
+  if (pcout)
+    *pcout
+      << "Warning: For a cut hollow sphere, the real volume will be lower than "
+      << "output." << std::endl;
```
blaisb, line 421: What is meant by this here? I don't think the displaced volume part is a good place to print stuff... Only print outside...

source/core/shape.cc
```diff
@@ -481,20 +501,222 @@ CompositeShape<dim>::static_copy() const
...
+template <int dim>
+double
+RBFShape<dim>::value(const Point<dim> &evaluation_point,
+                     const unsigned int /*component*/) const
+{
+  Point<dim> centered_point = this->align_and_center(evaluation_point);
+
+  // The computation and addition of the bounding box distance is necessary
+  // since the RBF nodes may not cover the whole simulation domain. In that
+  // case, it is assumed that the distance from the RBF object is approximately
```
blaisb, line 530: You should explain that also in the constructor.

source/core/shape.cc
```diff
@@ -481,20 +501,222 @@ CompositeShape<dim>::static_copy() const
...
+
+template <int dim>
+double
+RBFShape<dim>::displaced_volume(const double fluid_density,
+                                std::shared_ptr<ConditionalOStream> pcout)
+{
+  if (pcout)
+    *pcout
+      << "Warning: For a RBF shape, the real volume will be lower than output."
+      << std::endl;
```
blaisb, line 569: I don't think you need these warnings here.

source/core/shape.cc
```diff
@@ -481,20 +501,222 @@ CompositeShape<dim>::static_copy() const
...
+template <int dim>
+double
+RBFShape<dim>::gauss99(const double distance) const
+{
+  double eps = std::pow(-1.0 * std::log(0.01), 0.5);
+  return std::exp(-1.0 * std::pow(distance * eps, 2.0));
+}
+
+template <int dim>
+double
```
blaisb, line 613: Why not implement those in the .h and inline them? It would be really really faster.

source/core/shape.cc
```diff
@@ -481,20 +501,222 @@ CompositeShape<dim>::static_copy() const
...
+  return distance > 1.0 ?
+           0.0 :
+           (1.0 - 10.0 * std::pow(distance, 3.0) +
+            15.0 * std::pow(distance, 4.0) - 6.0 * std::pow(distance, 5.0));
+}
+
+template <int dim>
+double
+RBFShape<dim>::evaluate_basis_function(const RBFBasisFunction basis_function,
+                                       const double           distance) const
```
blaisb, line 685: Would it have been possible instead of using a switch, to template the kernel type in the shape ? 
Does not need to be done now, but keep this in mind.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3606,6 +3607,62 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
...
+
+              for (size_t n_i = 0; n_i < number_of_nodes; n_i++)
+                {
+                  weight[n_i]         = shape->weights[n_i];
+                  support_radius[n_i] = shape->support_radii[n_i];
+                  basis_function[n_i] =
+                    static_cast<double>(shape->basis_functions[n_i]);
+                  nodes_x[n_i] = shape->nodes[n_i][0];
+                  nodes_y[n_i] = shape->nodes[n_i][1];
+                  nodes_z[n_i] = shape->nodes[n_i][2];
```
blaisb, line 3643: This code is copy pasted everywhere. Please refactor it in a function if you have teh same code twice.

blaisb (changes requested):
Great work. Some changes are necessary. Please remove the pcout from the volume calculation for exemple, but very good nonetheless.

applications_tests/gls_sharp_navier_stokes_3d/flow_around_rbf.prm
```diff
@@ -0,0 +1,107 @@
...
+  set refine mesh inside radius factor        = 0.7
+  set refine mesh outside radius factor       = 1.5
+  set initial refinement                      = 1
+  set integrate motion                        = false
+  set assemble Navier-Stokes inside particles = false
+  subsection particle info 0
+    set type            = rbf
+    set shape arguments = rbf_test_shape.csv
+    subsection position
+      set Function expression = 0.01;0.01;0.01
```
Luckabarbeau, line 70: I would also include a rotation in the test

applications_tests/gls_sharp_navier_stokes_3d/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ file(COPY check_point.triangulation.info DESTINATION "${CMAKE_CURRENT_BINARY_DIR
 file(COPY check_point.triangulation_fixed.data DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/check_point.${_build_type}")
 file(COPY ib_force.00.dat DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/check_point.${_build_type}")
 file(COPY particles DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/load_particles_from_file_test.${_build_type}")
+file(COPY rbf_test_shape.csv DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/flow_around_rbf.${_build_type}")
```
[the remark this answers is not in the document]
oguevremont: I'll change it to .input with space separation

include/core/shape.h
```diff
@@ -104,9 +109,12 @@ class Shape : public AutoDerivativeFunction<dim>
    * Return the volume displaced by the solid
    *
    * @param fluid_density The density of the fluid that is displaced
+   * @param pcout An optional output stream to display warnings regarding volume
+   * calculation
    */
   virtual double
-  displaced_volume(const double fluid_density) = 0;
+  displaced_volume(const double                        fluid_density,
```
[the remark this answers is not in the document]
oguevremont: It's to print warnings when the displaced volume formula called is inexact, but since this function is only called when we integrate the particle motion, and we only do this with spheres, I can safely remove the warnings.

include/core/shape.h
```diff
@@ -531,4 +555,217 @@ class CompositeShape : public Shape<dim>
...
+class RBFShape : public Shape<dim>
+{
+public:
+  /**
+   * Class taken from Optimad Bitpit. https://github.com/optimad/bitpit
+   * @enum RBFBasisFunction
+   * @ingroup RBF
+   * @brief Enum class defining types of RBF kernel functions that could be used in bitpit::RBF class
+   */
+  enum class RBFBasisFunction : int
```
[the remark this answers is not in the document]
oguevremont: Yes it needs to be an int, because the basis function type is written as a number when it comes out of bitpit. If we changed the data type it wouldn't be possible to load the data using fill_vectors_from_file().

include/core/shape.h
```diff
@@ -531,4 +555,217 @@ class CompositeShape : public Shape<dim>
...
+   * @param distance distance to the node normalized by the support radius
+   */
+  double
+  c2c2(const double distance) const;
+
+  std::pair<std::string, enum Shape<dim>::ShapeType>
+  get_shape_name() override
+  {
+    return std::make_pair("rbf_shape", Shape<dim>::ShapeType::rbf_shape);
+  }
```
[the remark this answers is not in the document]
oguevremont: This function was used in a bit of code in GLSSharpNavierStokesSolver<dim>::check_whether_all_particles_are_sphere(). Its call was commented, so I can most likely just delete get_shape_name().

include/core/shape.h
```diff
@@ -531,4 +555,217 @@ class CompositeShape : public Shape<dim>
...
+              low_bounding_point[d] = nodes[i][d];
+            if (high_bounding_point[d] < nodes[i][d])
+              high_bounding_point[d] = nodes[i][d];
+          }
+        bounding_box_center[d] =
+          0.5 * (low_bounding_point[d] + high_bounding_point[d]);
+        half_lengths[d] =
+          0.5 * (high_bounding_point[d] - low_bounding_point[d]);
+      }
+    bounding_box = std::make_shared<Rectangle<dim>>(half_lengths,
```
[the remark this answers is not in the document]
oguevremont: Yes, I added an explanation above

applications_tests/gls_sharp_navier_stokes_3d/flow_around_rbf.prm
```diff
@@ -0,0 +1,107 @@
...
+    set type = slip
+  end
+  subsection bc 2
+    set id   = 3
+    set type = slip
+  end
+end
+
+subsection particles
+  set number of particles                     = 1
```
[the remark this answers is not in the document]
oguevremont: Done

source/core/shape.cc
oguevremont: I'll keep this in mind, but first I'll check if it would be useful to have different kernel types for different nodes. That is probably not useful though, just an idea.

applications_tests/gls_sharp_navier_stokes_3d/rbf_test_shape.csv
```diff
@@ -0,0 +1,4097 @@
+support_radius,basis_function,xnode,ynode,znode,weight
```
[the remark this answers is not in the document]
oguevremont: I removed some detail. We'll still get the idea and see whether the RBF shapes work with the solver. It now takes 13s in Debug.

source/core/parameters.cc
oguevremont: I removed the code duplication and added the default case

applications_tests/gls_sharp_navier_stokes_3d/flow_around_rbf.prm
```diff
@@ -0,0 +1,115 @@
...
+  set type                 = kelly
+  set variable             = pressure
+  set fraction type        = number
+  set max number elements  = 200000
+  set max refinement level = 7
+  set min refinement level = 2
+  set frequency            = 1
+  set fraction refinement  = 0.3
+  set fraction coarsening  = 0.1
+end
```
blaisb, line 97:
```suggestion
```
You don't have mesh adaptation so you can remove this part....

include/core/shape.h
```diff
@@ -25,8 +25,11 @@
 #endif
 
 
+#include <deal.II/base/conditional_ostream.h>
```
blaisb, line 28: Do you still need this include?

include/core/shape.h
```diff
@@ -510,25 +596,361 @@ class CompositeShape : public Shape<dim>
...
+   * radial basis function has a location and properties that are used in the
+   * sum.
+   * @param shape_arguments the concatenated vector of all shape arguments for
+   * an RBF in the order: weights, support_radii, basis_functions, nodes_x,
+   * nodes_y, nodes_z
+   * @param position the location of the RBF shape
+   * @param orientation the orientation of the shape in relation to each main
+   * axis
+   */
+  RBFShape<dim>(const std::vector<double> &shape_arguments,
```
blaisb, line 707: I would implement the constructor in the .cc file

blaisb (changes requested):
Some minor changes left. Nothing big

include/core/shape.h
oguevremont: Do you mean the 2nd constructor only or both of them?

blaisb: I'd say both

blaisb (approved):

blaisb:
Just needs @hepap and we are good for a merge I think

source/core/shape.cc
```diff
@@ -484,17 +487,166 @@ double
...
+                                                  bounding_box_center,
+                                                  Tensor<1, 3>());
+}
+
+template <int dim>
+double
+RBFShape<dim>::evaluate_basis_function(const RBFBasisFunction basis_function,
+                                       const double           distance) const
+{
+  switch (basis_function)
```
hepap, line 617: If you ever have only one basis function type for your entire object (if it becomes a requirement of your rbf geometry), this switch case could be removed for optimization. :smile:

hepap (approved):
Looks good to me! Only 3 minor comments :smile:

source/core/shape.cc
oguevremont: Good suggestion, thanks!

include/core/shape.h
```diff
@@ -194,27 +190,41 @@ class Sphere : public Shape<dim>
...
   displaced_volume(const double fluid_density) override;
 
-
-  std::pair<std::string, int>
-  get_shape_name() override
-  {
-    return std::make_pair("sphere", Shape<dim>::ShapeType::sphere);
-  };
-
   void
```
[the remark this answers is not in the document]
oguevremont: No, for getters and setters I don't think it's necessary since the function name is very clear, but I could add it if it's important ^^

source/core/parameters.cc
```diff
@@ -2200,8 +2199,55 @@ namespace Parameters
...
+                                         rbf_data["basis_function"].end());
+                  shape_arguments.insert(shape_arguments.end(),
+                                         rbf_data["xnode"].begin(),
+                                         rbf_data["xnode"].end());
+                  shape_arguments.insert(shape_arguments.end(),
+                                         rbf_data["ynode"].begin(),
+                                         rbf_data["ynode"].end());
+                  shape_arguments.insert(shape_arguments.end(),
+                                         rbf_data["znode"].begin(),
+                                         rbf_data["znode"].end());
```
paulapatience, line 2243: Why `*node` here but `nodes_*` in the input file?

paulapatience (approved):

PR #598, opened 2022-09-26
Multiphysics interface previous solutions

blaisb:
# Description of the problem

- The code use to only be able to gather one previous solution from the auxiliary physics instead of the entire list of vector. This lead to breakdown of BDF schemes when physics A needed the previous values of physics B

# Description of the solution

- Fixed this by allowing the passing of the entire previous solutions vector
- Also cleaned the multiphysics interface at the same time.

# How Has This Been Tested?

- All previous tests pass so this means this is correct.


Luckabarbeau (approved):
I would only check if other solvers like Nitsche or Vans required these modifications as well. Otherwise, everything looks good to me! Nice job!

PR #599, opened 2022-09-26
Transient function-imposed velocity field

blaisb:
# Description of the problem

- For some test cases, it is desirable to run the auxiliary physics solver with an ever changing velocity field that is not coming from the Navier-Stokes equations

# Description of the solution

- Enables to changes the velocity field as a function of time by recalling the initial conditions
- Fix some naming issues with the VOF solver.

### DIFF
--- a/source/solvers/navier_stokes_base.cc
+++ b/source/solvers/navier_stokes_base.cc
@@ -572,8 +572,8 @@ NavierStokesBase<dim, VectorType, DofsType>::iterate()
 {
   auto &present_solution = this->present_solution;
 
-  // If the fluid dynamics is not to be solved, update initial condition
-  // and move on
+  // If the fluid dynamics is not to be solved, but rather specified
+  // via an initial condition. Update condition and move on.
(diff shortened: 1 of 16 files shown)


source/solvers/navier_stokes_base.cc
```diff
@@ -571,9 +571,27 @@ void
 NavierStokesBase<dim, VectorType, DofsType>::iterate()
 {
   auto &present_solution = this->present_solution;
-  if (simulation_control->get_assembly_method() ==
-        Parameters::SimulationControl::TimeSteppingMethod::sdirk22 &&
-      simulation_parameters.multiphysics.fluid_dynamics)
+
+  // If the fluid dynamics is not to be solved, update initial condition
+  // and move on
```
lpsaavedra, line 576:
```suggestion
  // If the fluid dynamics is not to be solved, but rather specified
  // via an initial condition. Update condition and move on.
```

lpsaavedra (approved):
Seems good to me!

PR #603, opened 2022-10-03
Fix a bug in the definition of stabilization constant for VOF simulations

blaisb:
# Description of the problem

- The stabilization constant was not defined correctly for VOF simulations. There was a density term missing on the transient term
- The Dam-Break example was using a high fill level for the ILU preconditioner for no reason

# Description of the solution

- I decided to use the small inlined stabilization functions we have in the stabilization.h file. This way all the calculations are the same for everyone
- Reduce fill level of the dam-break example to 0.

# How Has This Been Tested?

- I reran the examples (nothing changed, but they are slightly more stable from a linear solver perspective)
- The outputs of some tests have changed very very slightly, they have been updated.

# Comments

- High-order time integration is still sketchy at best for the VOF solvers. This will be fixed in another PR where we change the order of the solvers.

### DIFF
--- a/include/solvers/stabilization.h
+++ b/include/solvers/stabilization.h
@@ -30,14 +30,19 @@ using namespace dealii;
  * @param viscosity Kinematic viscosity
  *
  * @param h Cell size. Should be calculated using the diameter of a sphere of equal volume to that of the cell
+ *
+ * @param density Density of the fluid, assumed to be 1 by default.
+ *
  */
 inline double
 calculate_navier_stokes_gls_tau_steady(const double u_mag,
                                        const double viscosity,
-                                       const double h)
+                                       const double h,
+                                       const double density = 1)
 {
-  return 1. / std::sqrt(Utilities::fixed_power<2>(2. * u_mag / h) +
-                        9 * Utilities::fixed_power<2>(4 * viscosity / (h * h)));
+  return 1. / std::sqrt(Utilities::fixed_power<2>(2. * density * u_mag / h) +
+                        9 * Utilities::fixed_power<2>(4 * density * viscosity /
+                                                      (h * h)));
 }
 
 /**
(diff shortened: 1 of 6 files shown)


include/solvers/stabilization.h
```diff
@@ -30,14 +30,19 @@ using namespace dealii;
  * @param viscosity Kinematic viscosity
  *
  * @param h Cell size. Should be calculated using the diameter of a sphere of equal volume to that of the cell
+ *
+ * @param density Density of the fluid, assumed to be 1 by default.
```
jeannejoachim, line 34: I like that this implementation is general! Maybe it would be good to ass a little explanation on why we set the density on 1 by default (eg because we sometimes work with NS equations on rho, and sometimes we need to compute the density, for VOF or rheological properties)?

source/solvers/navier_stokes_vof_assemblers.cc
```diff
@@ -94,11 +95,12 @@ GLSNavierStokesVOFAssemblerCore<dim>::assemble_matrix(
...
             Parameters::SimulationControl::TimeSteppingMethod::steady ?
-          1. / std::sqrt(std::pow(2. * density_eq * u_mag / h, 2) +
-                         9 * std::pow(4 * dynamic_viscosity_eq / (h * h), 2)) :
-          1. / std::sqrt(std::pow(sdt, 2) +
-                         std::pow(2. * density_eq * u_mag / h, 2) +
-                         9 * std::pow(4 * dynamic_viscosity_eq / (h * h), 2));
+          calculate_navier_stokes_gls_tau_steady(u_mag,
+                                                 viscosity_eq,
+                                                 h,
+                                                 density_eq) :
```
jeannejoachim, line 101: good catch, I like that you are calling a common function! This is so much better than code copying

Also it makes really clearer the `steady?` condition (if true, we use `calculate_navier_stokes_gls_tau_steady`, if false, we use `calculate_navier_stokes_gls_tau_transient`)

source/solvers/navier_stokes_vof_assemblers.cc
```diff
@@ -445,14 +450,12 @@ GLSNavierStokesVOFAssemblerBDF<dim>::assemble_rhs(
         velocity[p + 1] = scratch_data.previous_velocity_values[p][q];
 
 
-      std::vector<double> densities(number_of_previous_solutions(method) + 1);
 
       densities[0] = scratch_data.density[q];
 
-      for (unsigned int p = 1; p < number_of_previous_solutions(method) + 1;
-           ++p)
+      for (unsigned int p = 0; p < number_of_previous_solutions(method); ++p)
```
jeannejoachim, line 456: yeah good catch, no idea why this loop was different from the others

jeannejoachim: can you change it also on l. 385?

jeannejoachim (commented):
Neat, just a loop to change and we are good to go. I like that it calls a common function, really less blurry than before! Good catch :)

source/solvers/navier_stokes_vof_assemblers.cc
blaisb: Yup will do

jeannejoachim (approved):
Perfect, good to merge :)

PR #606, opened 2022-10-07
Vof pre solve

jeannejoachim:
# Description of the problem

The free surface (through VOF) was solved after the fluid dynamics. This was not theoretically correct, and prevented the use from higher order time schemes.

The `presolve` mechanism existed but was not usable in the current percolate mechanism.

# Description of the solution

* Modify the `percolate_time_vectors` for multiphysics, so that it can be used before or after the fluid dynamics has been solved
* Merge the `pre_solve` and `post_solve` inside a general `solve` method
* Remove the auxiliary physics `finish_time_step` methods, which sole purpose was to call `percolate_time_vectors`

# How Has This Been Tested?

All tests using VOF has been updated (they display marginally better results)

# Documentation

None, the order of resolution was not specified in the documentation

# Future changes

I will see if bdf2 schemes work on VOF simulations now and maybe add a test on that!


blaisb:
@jeannejoachim I would first check to see if BDF2 work with this implementation now for the VOF equations. Also these changes seem to have broken the VANS and the CFD-DEM solver

source/solvers/heat_transfer.cc
```diff
@@ -906,7 +899,7 @@ HeatTransfer<dim>::set_initial_conditions()
                            newton_update);
   nonzero_constraints.distribute(newton_update);
   present_solution = newton_update;
-  finish_time_step();
+  percolate_time_vectors();
```
blaisb, line 902: Wait I don't understand something here. I though we were only percolating the time vectors manually from the Navier-Stokes solver here? But post-process percolates the time vector also? Does this not risk a double percolation?

include/solvers/multiphysics_interface.h
```diff
@@ -90,138 +90,51 @@ class MultiphysicsInterface
...
+   * @param fluid_dynamics_solved Boolean that states if the fluid dynamics has been
+   * already solved or not. See the map `solve_pre_fluid` to know which
+   * subphysics are solved before the fluid dynamics.
    *
    * @param time_stepping_method Time-Stepping method with which the assembly is called
    */
   void
-  post_solve(const Parameters::SimulationControl::TimeSteppingMethod
-               time_stepping_method)
+  solve(bool fluid_dynamics_solved,
```
blaisb, line 102:
```suggestion
  solve(const bool fluid_dynamics_solved,
```

include/solvers/multiphysics_interface.h
```diff
@@ -90,138 +90,51 @@ class MultiphysicsInterface
...
   {
     // Loop through all the elements in the physics map. Consequently, iphys is
     // an std::pair where iphys.first is the PhysicsID and iphys.second is the
     // AuxiliaryPhysics pointer. This is how the map can be traversed
     // sequentially.
     for (auto &iphys : physics)
       {
-        // If iphys.first should be solved after fluid dynamics OR if is not
+        // If iphys.first should be solved BEFORE fluid dynamics
+        if (fluid_dynamics_solved == false &&
```
blaisb, line 113:
```suggestion
        if (!fluid_dynamics_solved &&
```

include/solvers/multiphysics_interface.h
```diff
@@ -90,138 +90,51 @@ class MultiphysicsInterface
...
-            if (verbosity != Parameters::Verbosity::quiet)
-              {
-                // Embellish the console output
-                pcout << "-----------------------" << std::endl;
-                announce_physics(iphys.first);
-              }
-
-            solve_physics(iphys.first, time_stepping_method);
-          }
+        else if (fluid_dynamics_solved == true &&
```
blaisb, line 119:
```suggestion
        else if (fluid_dynamics_solved  &&
```

include/solvers/multiphysics_interface.h
blaisb: I would maybe use a more verbose name than fluid_dynamics_solved . maybe fluid_dynamics_has_been_solved or fluid_dynamics_already_solved

blaisb (changes requested):
Look pretty much ok to me, but when I tested the VOF with BDF2 on my machine it segfaulted automatically which is a bit weird. I think we need to investigate this more (this was not an issue before)

source/solvers/heat_transfer.cc
jeannejoachim: Yes, I wondered that myself! I am not sure why this is done so, but I like that having removed `finish_time_step` clarifies what is being done. Once I fix the tests I will try to remove that to see how it goes

blaisb (approved):
All good for me. Just add the tag when you are ready for merge :)!

oguevremont (approved):
Very clean!

PR #607, opened 2022-10-10
Rpt reconstruction parallelization

blaisb:
# Description of the problem

- The RPT code L2 projection was slow and did not use parallelization

# Description of the solution

- Make it parallel
- 
# How Has This Been Tested?

- All examples and tests work as before
- Code is significantly (a lot) faster now.

# Documentation

- @ghazalemir Can you double check if the examples need to be changed? The code is now split into two executable (rpt_l2_projection and rpt_fem_reconstruction) but this was this way before right?

# Future changes

- Improve documentation as we go. The model is very good now.
-


ghazalemir (approved):
Hey Bruno :)
To address your questions:
The updated branch works well.
The example doesn't need any change, since it is the original experimental data and code works well on it.
Also I guess it is a good idea to keep the .prm file separated for L2projection and reconstruction.

PR #608, opened 2022-10-11
fix gidaspow drag model

tonielgeitani:
# Description of the problem

- The Gidaspow drag model was not correct. I used the same approach implemented by fluent and shown in the paper by Marchelli et al. (2020).
- Quadrature centered method (QCM) for void fraction calculation is limited to a single reference sphere size.
-  I also noticed that we were creating a finish_time_step_fd function in gls_vans and not using it. We were using the one navier_stokes_base but the one in the vans percolated the void fraction. 

# Description of the solution

- Change the formulation of Gidaspow according to the previous reference.
- Introduced parameters to specify the reference sphere size for the quadrature centred method (QCM).
- I used the finish_time_step_fd that includes void fraction percolation created in the gls_vans for both the vans and the coupling solvers. This did not cause any test to fail.

# How Has This Been Tested?

-For both the Gidaspow model and the QCM, a pseudo-2D flat spouted bed has been simulated where the formation of the bubble after 0.2 s of central gas injection has been compared with experimental results as well as other simulations. The size and shape of the bubble as well as the height of the bed were compared.

- No specific test has been created.

# Documentation

- The documentation changed is the void fraction parameter section in the Unresolved CFD-DEM parameter section.

- doc/source/parameters/unresolved-cfd-dem/void-fraction.rst

# Future changes

-N/A

# Comments

- N/A


doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -26,11 +26,20 @@ In this subsection, all parameters required for the calculation of the void frac
...
-* The ``particle refinement factor`` is only required for the ``spm``. It allows to determine the number of pseudo-particles that we want to divide our particle into. By default, it is set to 0 refinements, and results in no refinement of the original meshed particle (division into 7 particles in 3D). Every additional refinement results in a :math:`2^{dim}` times more particles. The figure below shows how the number of pseudo-particles change with every refinement. Every cell in the particle mesh represents a pseudo-particle in the satellite point method
+* The ``sphere element equal volume`` determines whether or not we want to use a reference sphere with the same volume as the element in which it is located. If it is disabled, then each sphere will have a radius equal to the size of the element in which it is located.
+* The ``reference sphere diameter`` allows us to fix the diameter of all reference spheres in the simulation to a given value. If this option is used, it overrides the previous option of equal volumes.
```
voferreira, line 42: Same comment as before. Additionally, It is not clear that when the previous parameter is set to true, this one does not work

source/fem-dem/gls_vans.cc
```diff
@@ -644,120 +669,138 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
-            else if (dim == 3)
-              R_sphere = (pow(3 * cell->measure() / (4 * M_PI), 1. / 3.));
-
-            // Array of real locations for the quadrature points
-            std::vector<Point<dim>> quadrature_point_location;
+          double reference_sphere_volume;
 
-            quadrature_point_location =
-              fe_values_void_fraction.get_quadrature_points();
+          if (this->cfd_dem_simulation_parameters.void_fraction
```
voferreira, line 674: I think here I would do the opposite. If sphere_element_equal_volume it does not use reference_sphere_diameter. Then if it was false, you would use reference_sphere_diameter. In my opinion, it would be cleaner and would not depend on a double

voferreira:
Also, the reference for gidaspow in the documentation needs to be updated

source/fem-dem/gls_vans.cc
tonielgeitani: In this case, i don't believe so. Because if the dp = 0 then we do it automatically, if dp = value then it uses this values. If dp =0, we have two options: if boolean is true, we use the same value for the sphere as the cell, if it is false, then we use radius of sphere size of the cell.

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -26,11 +26,20 @@ In this subsection, all parameters required for the calculation of the void frac
...
-* The ``particle refinement factor`` is only required for the ``spm``. It allows to determine the number of pseudo-particles that we want to divide our particle into. By default, it is set to 0 refinements, and results in no refinement of the original meshed particle (division into 7 particles in 3D). Every additional refinement results in a :math:`2^{dim}` times more particles. The figure below shows how the number of pseudo-particles change with every refinement. Every cell in the particle mesh represents a pseudo-particle in the satellite point method
+* The ``sphere element equal volume`` determines whether or not we want to use a reference sphere with the same volume as the element in which it is located. If it is disabled, then each sphere will have a radius equal to the size of the element in which it is located.
```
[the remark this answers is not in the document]
tonielgeitani: done

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
tonielgeitani: done

source/fem-dem/parameters_cfd_dem.cc
```diff
@@ -45,6 +45,16 @@ namespace Parameters
...
+    prm.declare_entry(
+      "qcm sphere diameter",
+      "0",
+      Patterns::Double(),
+      "The diameter of the reference sphere for QCM void fraction scheme");
+    prm.declare_entry(
+      "qcm sphere equal cell volume",
+      "false",
+      Patterns::Bool(),
+      "Specify whether the virtual sphere has same volume as element");
```
voferreira, line 57:
```suggestion
      "Specify whether the virtual sphere has the same volume as mesh element");
```

source/fem-dem/gls_vans.cc
blaisb: I like Toni approach, but plaese do not use "if conditions" with double. use seomthing like < 1e-16 or somethign similar to prevent weird issues.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1260,40 +1257,35 @@ GLSVansAssemblerGidaspow<dim>::calculate_particle_fluid_interactions(
...
       else
         {
           // Assuming the sphericity of particles = 1
           momentum_transfer_coefficient =
-            150 * pow((1 - cell_void_fraction), 2) * viscosity /
-              (cell_void_fraction *
-               pow(particle_properties[DEM::PropertiesIndex::dp], 2)) +
-            1.75 * (1 - cell_void_fraction) * average_relative_velocity.norm() /
-              particle_properties[DEM::PropertiesIndex::dp];
+            (150 * (1 - cell_void_fraction) / pow(cell_void_fraction, 2) +
```
blaisb, line 1279: might be a good idea to use instead https://www.dealii.org/current/doxygen/deal.II/namespaceUtilities.html#a834c6b595ada6f8e73c78d9bbd8ec551
when you have fixed powers

source/fem-dem/parameters_cfd_dem.cc
blaisb, line 57: So the default behavior is to use the cell size as the radius is that is?

blaisb (changes requested):
First round of reviews. Will continue tomorrow or tonight

source/fem-dem/parameters_cfd_dem.cc
tonielgeitani: yes

voferreira (approved):
I think it is good to go. It is only missing rebasing.

source/fem-dem/gls_vans.cc
```diff
@@ -558,13 +560,34 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
               //***********************************************************************
               for (unsigned int n = 0; n < active_neighbors.size(); n++)
                 {
+                  double reference_sphere_volume;
```
blaisb, line 563: Why is the reference sphere volume calculated this deeply into the loops? Can't you calculate it at the beggining of the assembler instead? This does not depend on the neighbhour and on the particle (or maybe the particle, but not the neighbhour. I would pre-calculate everything I can to more the cost away.

tonielgeitani: The reference sphere volume, depending on the parameter we choose might be defined using the cell size, the diameter of the particle, or cell volume. Therefore, I have to calculate it after the loop for the cells and the loop over particles, since it changes with these. It is determined twice, the first time when looping over neighbouring cells and neighbouring particles to determine the particle contribution and the second time to calculate the void fraction. I tried defining it before, but there are unknowns that are only known once you get inside these loops

blaisb (approved):

PR #614, opened 2022-10-19
Fix containers and rename namespace

blaisb:
# Description of the problem

- Some containers defined in the data containers of the DEM where not used everywhere
- Some maps were remaining that needed to be refactored
- I don't think it is a good idea to have a seperate namespace for the container, moved them to the DEM namespace

# Description of the solution

- Fixed the container usage
- Renamed the namespace to DEM

# How Has This Been Tested?

- All tests pass so this is good to go.

### DIFF
--- a/include/dem/data_containers.h
+++ b/include/dem/data_containers.h
@@ -42,7 +42,7 @@
  * are here, but eventually all the data containers of the DEM solver will be
  * moved here
  */
-namespace dem_data_containers
+namespace DEM
 {
   /**
    * Operator overloading to enable using triangulation cells as map keys.
@@ -203,18 +203,20 @@ namespace dem_data_containers
       cells_total_neighbor_list;
 
     // <boundary id, (tensor, point)>
-    typedef std::map<types::boundary_id, std::pair<Tensor<1, 3>, Point<3>>>
+    typedef std::unordered_map<types::boundary_id,
+                               std::pair<Tensor<1, 3>, Point<3>>>
       boundary_points_and_normal_vectors;
 
     // <int, <boundary id, tensor>>
-    typedef std::map<unsigned int, std::map<types::boundary_id, Tensor<1, 3>>>
+    typedef std::unordered_map<unsigned int,
+                               std::map<types::boundary_id, Tensor<1, 3>>>
       vector_on_boundary;
 
     // <cell id, periodic cells info>
-    typedef std::map<types::global_cell_index,
-                     periodic_boundaries_cells_info_struct<dim>>
+    typedef std::unordered_map<types::global_cell_index,
+                               periodic_boundaries_cells_info_struct<dim>>
       periodic_boundaries_cells_info;
   };
 
-} // namespace dem_data_containers
+} // namespace DEM
 #endif
--- a/include/dem/data_containers.h
+++ b/include/dem/data_containers.h
@@ -45,6 +45,13 @@
 namespace DEM
 {
   /**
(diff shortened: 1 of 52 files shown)


include/dem/output_force_torque_calculation.h
```diff
@@ -43,14 +44,12 @@ write_forces_torques_output_locally(
...
-  std::map<unsigned int, std::map<unsigned int, dealii::Tensor<1, 3>>>
-    &forces_boundary_information,
-  std::map<unsigned int, std::map<unsigned int, dealii::Tensor<1, 3>>>
-    &torques_boundary_information)
+  const std::string                                filename,
+  const unsigned int                               output_frequency,
+  const std::vector<unsigned int>                  boundary_index,
+  const double                                     time_step,
+  DEM::dem_data_structures<3>::vector_on_boundary &forces_boundary_information,
+  DEM::dem_data_structures<3>::vector_on_boundary &torques_boundary_information)
```
acdaigneault, line 52: vector_on_boundary is `std::unordered_map<unsigned int, std::map<types::boundary_id, Tensor<1, 3>>>`
From the following line, the key of the 2nd map [i] is the # of processor and not a boundary id. Unsigned int is right, but not the typedef.

include/dem/data_containers.h
```diff
@@ -203,18 +203,20 @@ namespace dem_data_containers
...
     // <boundary id, (tensor, point)>
-    typedef std::map<types::boundary_id, std::pair<Tensor<1, 3>, Point<3>>>
+    typedef std::unordered_map<types::boundary_id,
+                               std::pair<Tensor<1, 3>, Point<3>>>
       boundary_points_and_normal_vectors;
 
     // <int, <boundary id, tensor>>
-    typedef std::map<unsigned int, std::map<types::boundary_id, Tensor<1, 3>>>
+    typedef std::unordered_map<unsigned int,
+                               std::map<types::boundary_id, Tensor<1, 3>>>
```
acdaigneault, line 212: Why not 2 unordered_map or unordered_map of vector?

source/dem/grid_motion.cc
```diff
@@ -77,7 +77,8 @@ GridMotion<dim, spacedim>::
       types::particle_index,
       std::map<types::boundary_id, particle_wall_contact_info_struct<spacedim>>>
       &particle_wall_pairs_in_contact,
-    const std::map<types::boundary_id, std::pair<Tensor<1, 3>, Point<3>>>
+    const typename DEM::dem_data_structures<
+      spacedim>::boundary_points_and_normal_vectors
```
acdaigneault, line 81: The key of that seems to be a global_face_id (see line 100) and not a boundary_id as define in the data_structures of boundary_points_and_normal_vectors. I know sometimes there's imaginary face id that are negative, is it not possible to have issues related to that?

include/dem/output_force_torque_calculation.h
blaisb: That's weird, it means this "bug" was kinda always there because in the past the vector_on_boundary was being cast into this. I'll look more into this and get back to you.

include/dem/data_containers.h
blaisb: Fixed. still, this will continue to require refactoring maybe in a follow up PR. :).

blaisb:
Everything should be fixed now back with map for the forces. I'll merge once CI passes @acdaigneault. It's really a partial cleanup, more will need to be done, but I'll get back into it.

PR #615, opened 2022-10-19
Add custom cylinder meshes

acdaigneault:
# Description of the problem

- Mesh geometries in Lethe were only deal.II functions or GMSH files (or periodic hills which is a specific geometry from a benchmark case).

# Description of the solution

- Creation of a mesh type called `cylinder` which allow to create 4 different cylinder grids where the axis and configuration fit the subdivided cylinder from deal.II. One of them, `classic`, is equivalent to use the deal.II function subdivided_cylinder().

# Documentation

- `mesh.rst` explains that the grid arguments are the same as subdivided_cylinder() and shows some results of grid types.

# Test

- Working with mapping

# Future

- Maybe make the usage of simplex works with those grids

# Comment

- Maybe move the code from grids.cc to another file regards the reviews and comments


source/core/grids.cc
```diff
@@ -109,6 +110,142 @@ attach_grid_to_triangulation(
...
+                                                 subdivisions,
+                                                 radius,
+                                                 half_height);
+            }
+          else
+            {
+              // Create a temporary 2d mesh
+              Triangulation<dim - 1, spacedim - 1> temporary_triangulation;
+
+              // Creat a spherical manifold for 2d mesh
```
voferreira, line 169:
```suggestion
              // Create a spherical manifold for 2d mesh
```

source/core/grids.cc
```diff
@@ -109,6 +110,142 @@ attach_grid_to_triangulation(
...
+              // Create a subdivided cylinder from deal.ii
+              GridGenerator::subdivided_cylinder(*triangulation,
+                                                 subdivisions,
+                                                 radius,
+                                                 half_height);
+            }
+          else
+            {
+              // Create a temporary 2d mesh
+              Triangulation<dim - 1, spacedim - 1> temporary_triangulation;
```
voferreira, line 167: I would put 2 here instead of dim-1, because it is explicitly a 2d mesh. Not a big deal, but I find defensive programming always important.

source/core/grids.cc
```diff
@@ -109,6 +110,142 @@ attach_grid_to_triangulation(
...
+
+              // Creat a spherical manifold for 2d mesh
+              Point<2>                                       center(0.0, 0.0);
+              const SphericalManifold<dim - 1, spacedim - 1> m0(center);
```
voferreira, line 171: Here, same comment as before

blaisb:
As my dear friend @peterrum  pointed out to me. The squared cylinder is going to be eventually ill-posed as you refine the mesh (three points will align along a line, making an ill-posed jacobian transformation). I am not sure I would keep it, if you do, at least put a warning in the documentation :).

source/core/grids.cc
acdaigneault: Yeah, I was ambiguous about it, I will change all `dim` and `dim - 1` to 3 and 2 in that code

source/core/grids.cc
```diff
@@ -109,6 +110,142 @@ attach_grid_to_triangulation(
...
+
+              if (mesh_parameters.grid_type == "regularized" ||
```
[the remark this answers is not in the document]
acdaigneault: I will still have the second if since it have to be after the set_manifold useless I put the 

`             temporary_triangulation.reset_all_manifolds();
              temporary_triangulation.set_all_manifold_ids_on_boundary(0);
              temporary_triangulation.set_manifold(0, m0);`

in all cases.

blaisb (changes requested):
Nothing to say for me. However, can you add a quick application_test that uses at least one of these meshes? For example you can copy the Poiseuille application test in 3D and redo it with this mesh . Just once will be enough, but I like to have tests for features such as this one.

source/core/grids.cc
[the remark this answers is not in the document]
voferreira: I would rather do this second option for clarity reasons, but it is not a big deal. Keeping it this ways is fine as well

doc/source/parameters/cfd/mesh.rst
```diff
@@ -5,25 +5,37 @@ This subsection provides information of the simulation geometry and its mesh. Th
...
+    * ``cylinder``: if this type is chosen, the ``grid type`` must be chosen according to the following figure. The ``classic`` type is equivalent to a subdivided cylinder from Deal.II and ``grid arguments`` must follow their `related documentation <https://www.dealii.org/current/doxygen/deal.II/namespaceGridGenerator.html#a95f6e6a7ae2fe3a862df035dd2cb4467>`_ and this applies to all cylinder grid types.
+
+    .. warning::
+        The squared cylinder may eventually leads to ill-posed jacobian transformation. As said in the `deal.II documentation <https://www.dealii.org/current/doxygen/deal.II/namespaceGridTools.html#a3f129213c63c92a6ed84c6f2a906048b>`_ : "The four cells that were originally the corners of a square will give you some troubles during computations, as the jacobian of the transformation from the reference cell to those cells will go to zero, affecting the error constants of the finite element estimates".
+        This type of mesh is available, but is not recommended.
```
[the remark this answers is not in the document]
blaisb: actually should be its usage and not it's

voferreira (approved):
Good to go for me. Nice one!

blaisb (approved):

PR #618, opened 2022-10-21
Fix void fraction calculation mode in examples

blaisb:
# Description of the problem

-  The CFD-DEM examples were not running anymore because of the void fraction calculation mode

# Description of the solution

- Fixed them. Improved the outputs at the same time

# How Has This Been Tested?

- Ran them on my machine, all is well


tonielgeitani (approved):
All is good for me

PR #623, opened 2022-10-25
Fix CI for test which have restarts and disable 9.4.0 CI

blaisb:
# Description of the problem

- All tests that had restarts have been broken by the latest change to deal.II master

# Description of the solution

- I had to disable the CI for dealii 9.4.0. I am not sure if this will need to be permanent. Restarts are not compatible between the versions...
- I had to remake all the restarts file to be compatible with the newer deal.II version

# How Has This Been Tested?

- If all tests pass... we are good.

# Comments

- This is annoying to say the least.


bangerth:
Ah, sorry for breaking stuff. We tend to think of restart files as something that only works with the very same version of deal.II (not release, but commit version) and it did not occur to me that people would keep those stored somewhere to use between versions.

PR #625, opened 2022-10-28
Add post-processing capabilities when using log output with DEM

blaisb:
# Description of the problem

- The Log when running DEM simulations only output the number of the iteration. This is problematic since it does not allow one to monitor how the simulation is going
- There was a mistake in how the kinetic energy was calculated for the particles

# Description of the solution

- Add a few statistics to monitor the DEM. Min/Max/Average/Total kinetic energy for translation and rotation as well as for translational and rotational velocity.

# How Has This Been Tested?

- This does not change the code, just allows for more monitoring.

### DIFF
--- a/source/dem/dem.cc
+++ b/source/dem/dem.cc
@@ -826,8 +826,6 @@ DEMSolver<dim>::report_statistics()
                          (simulation_control->get_step_number()) *
                          simulation_control->get_log_frequency();
 
-  // Update statistics on time
-
   // Calculate statistics on the particles
   statistics translational_kinetic_energy = calculate_granular_statistics<
     dim,
--- a/source/dem/dem.cc
+++ b/source/dem/dem.cc
@@ -14,9 +14,10 @@
  * ---------------------------------------------------------------------
  */
 
+#include <core/utilities.h>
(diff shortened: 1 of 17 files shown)


source/dem/dem.cc
```diff
@@ -816,6 +810,80 @@ DEMSolver<dim>::post_process_results()
...
+  contact_list.max =
+    std::max(number_of_list_built_since_last_log, contact_list.max);
+  contact_list.min =
+    std::min(number_of_list_built_since_last_log, contact_list.min);
+  contact_list.total += number_of_list_built_since_last_log;
+  contact_list.average = contact_list.total /
+                         (simulation_control->get_step_number()) *
+                         simulation_control->get_log_frequency();
+
+  // Update statistics on time
```
tonielgeitani, line 829: What does this comment mean here? There is no code that updates in time afterwards.

tonielgeitani (approved):
All is good for me :)

source/dem/dem.cc
blaisb: 100%  right. Fixed it :)!

PR #627, opened 2022-10-30
Optimization dem

blaisb:
# Description of the problem

- Some parts of the DEM were not correctly optimized. The force calculation notably was "refetching" information about particle_one of the contact when this could have been kept constant.

# Description of the solution

- Fixed this
- Improved Verlet integrator also
- Fixed some typos

Overall, the code is about 10% faster now on some of the benchmark I ran.

# How Has This Been Tested?

- All regular tests passed except two which were very slightly changed. These tests are always fragile so it's not alarming, one particle change very very slightly. That's because of the order of operation

# Comments

- I will keep on working to make this a bit more optimal.


source/dem/particle_particle_nonlinear_force.cc
```diff
@@ -190,21 +206,9 @@ ParticleParticleHertzMindlinLimitOverlap<dim>::
...
-                    particle_two->get_id();
-#else
-                  types::particle_index particle_one_id =
-                    particle_one->get_local_index();
                   types::particle_index particle_two_id =
                     particle_two->get_local_index();
-#endif
-                  Tensor<1, 3> &particle_one_torque = torque[particle_one_id];
                   Tensor<1, 3> &particle_two_torque = torque[particle_two_id];
-                  Tensor<1, 3> &particle_one_force  = force[particle_one_id];
```
acdaigneault, line 207: Same as for the linear forces, is the directives at line 312 still relevant if you removed those here?

source/dem/velocity_verlet_integrator.cc
```diff
@@ -15,20 +15,17 @@ VelocityVerletIntegrator<dim>::integrate_half_step_location(
   const std::vector<Tensor<1, 3>> &force,
   const std::vector<double> &      MOI)
 {
+  Tensor<1, dim> particle_acceleration;
```
acdaigneault, line 18: Since the `particle_acceleration` is removed for the loop, would it be a good practice to clear it in the loop before updating it at line 30?

source/dem/velocity_verlet_integrator.cc
```diff
@@ -15,20 +15,17 @@ VelocityVerletIntegrator<dim>::integrate_half_step_location(
...
-      auto particle_position   = particle->get_location();
-#if (DEAL_II_VERSION_MAJOR < 10 && DEAL_II_VERSION_MINOR < 4)
-      types::particle_index particle_id = particle->get_id();
-#else
-      types::particle_index particle_id = particle->get_local_index();
-#endif
-      Tensor<1, dim> particle_acceleration;
+      auto                  particle_properties = particle->get_properties();
+      auto                  particle_position   = particle->get_location();
+      types::particle_index particle_id         = particle->get_local_index();
```
acdaigneault, line 28:
```suggestion
      types::particle_index particle_id         = particle->get_local_index();
      particle_acceleration.clear();
```

source/dem/velocity_verlet_integrator.cc
```diff
@@ -63,45 +60,53 @@ VelocityVerletIntegrator<dim>::integrate(
...
-      double MOI_inverse  = 1 / MOI[particle_id];
+
+      double dt_mass_inverse = dt / particle_properties[PropertiesIndex::mass];
+      double dt_MOI_inverse  = dt / MOI[particle_id];
 
       if constexpr (dim == 3)
         particle_position = particle.get_location();
 
       if constexpr (dim == 2)
         particle_position = point_nd_to_3d(particle.get_location());
```
acdaigneault, line 83: No lambda cool stuff here? :(

source/dem/velocity_verlet_integrator.cc
```diff
@@ -63,45 +60,53 @@ VelocityVerletIntegrator<dim>::integrate(
   std::vector<Tensor<1, 3>> &      force,
   const std::vector<double> &      MOI)
 {
+  Point<3>     particle_position;
+  Tensor<1, 3> dt_g = g * dt;
```
acdaigneault, line 64:
```suggestion
  const Tensor<1, 3> dt_g = g * dt;
```

source/dem/velocity_verlet_integrator.cc
acdaigneault: and at the end of the code

acdaigneault (approved):
Just a few comments. I think you forgot to modify some code for the ghost loops in force related .cc files. You know my opinion about having the same loop for local and ghost, I think the loop should be moved in a function, I could do it in refactor PR.

source/dem/velocity_verlet_integrator.cc
blaisb: That would not be an issue because we are using equality afterwards

blaisb:
> Just a few comments. I think you forgot to modify some code for the ghost loops in force related .cc files. You know my opinion about having the same loop for local and ghost, I think the loop should be moved in a function, I could do it in refactor PR.

One thing I am a bit wary about making it into a function is the performance overhead. Glad to look into it in another PR though :).

include/dem/particle_particle_contact_force.h
```diff
@@ -129,16 +131,83 @@ class ParticleParticleContactForce
...
+    contact_info.tangential_relative_velocity =
+      contact_relative_velocity -
+      (normal_relative_velocity_value * normal_unit_vector);
+
+    // Calculation of new tangential_overlap, since this value is
+    // history-dependent it needs the value at previous time-step
+    // This variable is the main reason that we have iteration over
+    // two different vectors (pairs_in_contact and
+    // contact_pair_candidates): tangential_overlap of the particles
+    // which were already in contact (pairs_in_contact) needs to
```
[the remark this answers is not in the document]
blaisb: Fixed the comment not to mention this anymore. Referring to things which are really away does not help readability.

source/dem/particle_particle_linear_force.cc
```diff
@@ -77,17 +77,17 @@ ParticleParticleLinearForce<dim>::ParticleParticleLinearForce(
   if (dem_parameters.model_parameters.rolling_resistance_method ==
       Parameters::Lagrangian::ModelParameters::RollingResistanceMethod::
         no_resistance)
-    rolling_reistance_model =
```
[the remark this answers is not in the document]
blaisb: Agreed :)

source/dem/particle_particle_linear_force.cc
```diff
@@ -112,32 +112,47 @@ ParticleParticleLinearForce<dim>::calculate_particle_particle_contact_force(
...
+          auto particle_one       = first_contact_info->second.particle_one;
+          auto particle_one_properties = particle_one->get_properties();
+
+          types::particle_index particle_one_id =
+            particle_one->get_local_index();
+          Tensor<1, 3> &particle_one_torque = torque[particle_one_id];
+          Tensor<1, 3> &particle_one_force  = force[particle_one_id];
+
+          // Fix particle one location for 2d and 3d
+          Point<3> particle_one_location = [&] {
```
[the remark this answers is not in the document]
blaisb: So do I, so do I :)

source/dem/particle_particle_nonlinear_force.cc
```diff
@@ -128,32 +128,48 @@ ParticleParticleHertzMindlinLimitOverlap<dim>::
...
+              Point<3> particle_two_location = [&] {
+                if constexpr (dim == 3)
+                  {
+                    return particle_two->get_location();
+                  }
+                else
+                  {
+                    return (point_nd_to_3d(particle_two->get_location()));
+                  }
+              }();
```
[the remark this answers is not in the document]
blaisb: Fixed.

source/dem/velocity_verlet_integrator.cc
blaisb: Not sure if the cool lambda stuff is more readable, but i'll add it since you seem to fancy that :P!

blaisb: The one at the end of the code is not an assignment so I am not sure the lambda is the best approach here :).

blaisb:
All you comments have been addressed @acdaigneault . thanks for the very good review :)!

PR #628, opened 2022-11-01
Add parameter to impose initial conditions to DEM particles + fix Magnus force parameter

voferreira:
# Description of the problem

- DEM particles were always inserted with velocities and omegas equal to 0;
- Magnus force on/off parameter was not parsed.

# Description of the solution

- Added parameters for initial velocities and omegas in the insertion info section of the DEM;
- Added Magnus force to parse_parameters.

# How Has This Been Tested?
- [x] CFD-DEM Sedimentation of single particle

# Tests added
- [x] launching of a single particle with an initial velocity.

# Documentation

- [x] Add parameters to DEM documentation

# Comments

- The parameter is applied equally to all particles. It works currently, but future changes can be applied to make it so that types are inserted with different initial conditions.

### DIFF
--- a/source/core/parameters_lagrangian.cc
+++ b/source/core/parameters_lagrangian.cc
@@ -320,6 +320,31 @@ namespace Parameters
                           "0",
                           Patterns::List(Patterns::Double()),
                           "List of particles z positions");
+
+        prm.declare_entry("velocity x",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial velocity x");
+        prm.declare_entry("velocity y",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial velocity y");
+        prm.declare_entry("velocity z",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial velocity z");
+        prm.declare_entry("omega x",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial omega x");
+        prm.declare_entry("omega y",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial omega y");
+        prm.declare_entry("omega z",
+                          "0.0",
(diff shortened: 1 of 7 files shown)


source/core/parameters_lagrangian.cc
```diff
@@ -320,6 +320,31 @@ namespace Parameters
...
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial omega x");
+        prm.declare_entry("omega y",
+                          "0.0",
+                          Patterns::Double(),
+                          "Initial omega y");
+        prm.declare_entry("omega z",
+                          "0.0",
+                          Patterns::Double(),
```
blaisb, line 346: This means that all particles are inserted with the same values of the velocity and omega right?

blaisb (commented):
Seems good. Don't forget to update documentation though.
Do you think you could add a small dem test? Like a falling particle which has an initial velocity

source/core/parameters_lagrangian.cc
voferreira: They are. I don't know how I could separate them

voferreira:
> Seems good. Don't forget to update documentation though. Do you think you could add a small dem test? Like a falling particle which has an initial velocity

Sure! I will document it and add the test.

source/core/parameters_lagrangian.cc
voferreira: ``Insertion info`` could be divided according to ``particle type``, but I can't think of a situation in which this would be important

blaisb: For the time being this is ok as a mechanism I think. The current input format make it difficult to have very flexible insertion mechanism.

blaisb (approved):

PR #629, opened 2022-11-01
Change default rolling friction model

blaisb:
# Description of the problem

- The default rolling friction was no friction which maybe was not the wisest decision

# Description of the solution

- Put the default rolling friction to the constant model

# How Has This Been Tested?

- All test pass. Some of them were using default rolling friction and I manually put them to no_friction

# Documentation

- Fixed the default value in the documentation

# Comments

- We  can keep on improving the theory guide on these issues :). We could also compare the different models.
-


voferreira (approved):
All good for me.

tonielgeitani (approved):
Everything is clear for me

blaisb:
> Seems good, but you forgot to add the parameters in dem examples. I did remove the rolling friction coefficients in the application tests and examples in another branch (I have modified my hopper example in the same branch, I shouldn't had). But right now, in master, if you don't specify the no_resistance, some dem examples will use those coefficients.

That's ok. This is what I want. Most examples should indeed use the rolling resistance :).

PR #632, opened 2022-11-04
Fix dem hopper example

acdaigneault:
# Description of the problem

- The DEM Hopper example had bad meshing at corners
- The example did not used any rolling friction method and the `no_resistance` was the default value when example was added.
- Some DEM examples and application tests did not have rolling friction methods, but had explicit rolling friction coefficients for p-p and p-w which was a bit incoherent.

# Description of the solution

- A new GMSH mesh was generated.
- The rolling resistance torque method was set to `constant_resistance`
- Unused rolling friction coefficients in application tests and example where removed

# How Has This Been Tested?

- Application tests passed
- Removing the rolling friction coefficients should not have changed the examples since `no_resistance` was the previous default method

# Comments

- Some examples should actually use a rolling friction method, but since results are shown in the documentation of the examples, the use of a rolling friction must need a rerun to verify that the new results are mostly the same with the previous results.


blaisb (commented):
Seems very good but don't you also need to modify the image you have in the documentation of your flow loaded?

acdaigneault:
@blaisb Still waiting for the simulation results, the plot also needs to be changed and the only mesh to be deleted.

acdaigneault:
@blaisb documentation have been modified

blaisb (approved):

PR #634, opened 2022-11-04
Copy grids to DEM

voferreira:
# Description of the problem

- Cylinder and periodic hill grids were not implemented in DEM.

# Description of the solution

- Copied the implementation from core/grids to dem/read_mesh.

# How Has This Been Tested?

- Sedimentation of particles for further liquid-fluidized bed simulations

# Documentation

- The current parameter guide works for this implementation.

# Future changes

- It would be good if we wouldn't need to repeat code in the DEM, but just use what we have in lethe core instead.

### DIFF
--- a/include/dem/read_mesh.h
+++ b/include/dem/read_mesh.h
@@ -37,7 +37,7 @@ using namespace std;
  * Defines or reads the mesh based on the information provided by the user
  * Gmsh files can also be read in this function
  *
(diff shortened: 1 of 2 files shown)


blaisb:
This is a good point. Right now this grid duplication is a bad thing. It's in my todo list, will just need time to get it done. In the meantime, this is a good temporary solution.

blaisb (approved):

PR #637, opened 2022-11-07
Change steady-bdf to steady_bdf in documentation

CaroleAnneDaunais:
Small fix to the right name of the parameter in the documentation.


jeannejoachim (approved):
Good catch :)

PR #639, opened 2022-11-10
Unique mesh reader

### LINKED ISSUE #561: Move mesh reading in DEM to the core grid utilities
Function `read_mesh()` in dem/read_mesh.cc is very similar to `attach_grid_to_triangulation()` in core/grids.cc.
It would be great to move the grid handling for DEM into the same file or rearrange functions in core/grids.cc to make it works for CFD and DEM meshes.

blaisb:
# Description of the problem

- The DEM module contained a lot of code duplication from the CFD module. The entire part related to the attachment of the grid to the triangulation was fully duplicated. This was a total mess.

# Description of the solution

- Refactor the reader to at least use the same attachment mechanism
- I also took the opportunity to clean the interface to use references instead of pointers. This make the code more portable and easier to use everywhere.

# How Has This Been Tested?

- All tests pass without a change

# Comments

- Ongoing process to clean the DEM code
- Closes #561

### DIFF
--- a/applications/initial_conditions/initial_conditions.cc
+++ b/applications/initial_conditions/initial_conditions.cc
@@ -49,7 +49,7 @@ void
 InitialConditionsNavierStokes<dim>::run()
 {
   read_mesh_and_manifolds(
-    this->triangulation,
+    *this->triangulation,
     this->simulation_parameters.mesh,
     this->simulation_parameters.manifolds_parameters,
     this->simulation_parameters.restart_parameters.restart,
--- a/include/core/grids.h
+++ b/include/core/grids.h
@@ -41,9 +41,22 @@ using namespace dealii;
 template <int dim, int spacedim = dim>
 void
 attach_grid_to_triangulation(
+  parallel::DistributedTriangulationBase<dim, spacedim> &triangulation,
+  const Parameters::Mesh &                               mesh_parameters);
+
+
+/**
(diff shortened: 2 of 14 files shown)


acdaigneault (commented):
Yay, enfin :)! Will review tonight or tomorrow. 
I just want to say that you didn't modify the lines 172-188 in the cfd_dem_coupling.cc. I had some issues with that lately.

voferreira (approved):
I think it is good to go. It is way cleaner like this. The only thing I think could be changed, but it does not need to be within this PR is to use either "mesh" or "grid" for everything.

acdaigneault (approved):
All seems good to me! Useful changes :)
I was wondering why all related functions for grid were not moved in grids.cc (meaning removing dem/read_mesh.cc). I mean having like an overload function with the `Parameters::Lagrangian::BCDEM &bc_params` parameter.

blaisb:
> All seems good to me! Useful changes :) I was wondering why all related functions for grid were not moved in grids.cc (meaning removing dem/read_mesh.cc). I mean having like an overload function with the `Parameters::Lagrangian::BCDEM &bc_params` parameter.

It's in my TODO, I just want to do it step by step with smaller PRs. The target is to eventually remove completely the Parameters::Lagrangian::BCDEM all together. In a next PR :).

PR #640, opened 2022-11-11
Output single grid DEM

voferreira:
# Description of the problem

- In DEM, the grid is output in every single output step;

# Description of the solution

- Create a member function for the grid output separated from the output of the particles data
- Move the member function to Lagrangian post-processing

# How Has This Been Tested?

- Sedimentation of multiple particles case - All scenarios

# Documentation

- Created post-processing parameters guide


include/dem/dem.h
```diff
@@ -231,6 +231,13 @@ class DEMSolver
   void
   write_output_results();
 
+  /**
+   * @brief write_output_grid
+   * Post-processing as parallel VTU files
+   */
+  void
+  write_output_grid();
```
blaisb, line 239: What is an output grid? is it the boundary? the grid of the domain?

blaisb (changes requested):
Can you just comment a bit better in the .h? Otherwise this should be ok

blaisb:
From my understanding, this prevents us from also outputting the grid to monitor how to load balancing is proceeding no? I feel that's a bad thing, because it is very important for many applications to see the load balancing progress... Do you think there's a possible compromise somewhere?

voferreira:
> From my understanding, this prevents us from also outputting the grid to monitor how to load balancing is proceeding no? I feel that's a bad thing, because it is very important for many applications to see the load balancing progress... Do you think there's a possible compromise somewhere?

You are right. I didn't think about that when I did it. I thought of it and I think the best way to address this is to add a parameter to control this, and call ``write_output_grid()`` within ``write_output_results()`` if this parameter is set to ``true``. What do you think?

source/dem/dem.cc
```diff
@@ -675,6 +679,7 @@ DEMSolver<dim>::write_output_results()
     solid_object->write_output_results(simulation_control);
 }
 
+
```
voferreira, line 682:
```suggestion
```

source/dem/dem.cc
```diff
@@ -627,30 +627,34 @@ DEMSolver<dim>::write_output_results()
...
-
-  write_vtu_and_pvd<dim>(grid_pvdhandler,
-                         background_data_out,
-                         folder,
-                         grid_solution_name,
-                         time,
-                         iter,
-                         group_files,
-                         mpi_communicator);
+  if (parameters.post_processing.write_grid == true)
```
blaisb, line 630: Do you think there is a way that this grid could be merged with the lagrangian post-processing or you think this is too difficult? E.g. if we have lagrangian post-processing the info would just be added on top of this grid

blaisb (changes requested):
Just a question, otherwise seems good to me.

doc/source/parameters/dem/post-processing.rst
```diff
@@ -0,0 +1,63 @@
...
+* The ``end step`` is used to choose when to stop the calculating either the particle's average velocity, the granular temperature, and/or the grid outputting.
+
+.. warning::
+ Since the default value for the ``end step`` is 0, the ``end step`` must be set according to the duration of the simulation or at least is higher than the ``initial step``, otherwise the particles' average velocity, granular temperature, and/or grid will not be output.
+
+* The ``output frequency`` is used to choose the frequency of creation of output files with the particles' average velocity, granular temperature, and/or grid outputting.
+
+* The ``particles velocity output name`` is used to set a name for the output file with the particle's average velocity. By default, the file is named ``particles_velocity``.
+
+* The ``granular temperature output name`` is used to set a name for the output file with the particle's granular temperature. By default, the file is named ``granular_temperature``.
```
blaisb, line 62: So I was wondering, do you think there is a way this data could be added to the grid files directly instead of being appended to different variables? E.g. make it a single file with all the possible lagrangian post processing capabilities instead of a single one? Or is it already like this? :). I know i'm asking a lot but this would be very useful.

blaisb (approved):
Very cool PR. I am thinking that in the future all Lagrangian post-processing files could merged into the grid written. Maybe not for this PR this is already a lot :).

blaisb:
@voferreira can you get a second review from someone else? Once that's done I'll merge

doc/source/parameters/dem/post-processing.rst
voferreira: No worries, it is not asking a lot. I am trying to do it. I will push it later today if I manage to do so. I think it is the best approach in order to clean a bit the solver.

voferreira:
I just refactored the postprocessing in order to create a single file for the granular temperature, the particles' velocity, and the grid. I like the new pattern because the dem.cc and the cfd_dem_coupling.cc files get cleaner, additionally, we have a single file for all grid-related data.

blaisb:
> I just refactored the postprocessing in order to create a single file for the granular temperature, the particles' velocity, and the grid. I like the new pattern because the dem.cc and the cfd_dem_coupling.cc files get cleaner, additionally, we have a single file for all grid-related data.

You are getting real good at this :)! Thanks, life will be much cleaner now

blaisb (approved):
Good job! Very cool. This is so much cleaner than it used to be.

doc/source/parameters/dem/post-processing.rst
```diff
@@ -0,0 +1,53 @@
...
+  set output frequency = 100000
+
+ end
+
+.. note::
+ The parameters default values are the ones presented above.
+
+* The ``Lagrangian post processing`` enables the use of ``calculate particles average velocity``, ``calculate granular temperature``, and ``write grid``.
+
+* The ``write grid`` parameter enables one to output the grid as VTU files. The output files can be used for post-processing. By default, it is ``true``. The files with the grid will have ``-grid`` in their names.
```
acdaigneault, line 36: Can you explain a bit what disabling `write grid` does? For my understanding after the conversation in this PR, enabling write grid will output grid as before the changes, but I'm not sure what happen when it's false?

doc/source/parameters/dem/post-processing.rst
```diff
@@ -0,0 +1,53 @@
...
+
+.. warning::
+ ``calculate particles average velocity``, ``calculate granular temperature``, and ``write grid`` will not work without setting ``Lagrangian post processing`` to ``true``.
+
+* The ``initial step`` is used to choose when to start the calculating either the particle's average velocity, the granular temperature, and/or the grid outputting.
+
+* The ``end step`` is used to choose when to stop the calculating either the particle's average velocity, the granular temperature, and/or the grid outputting.
+
+.. warning::
+ Since the default value for the ``end step`` is 0, the ``end step`` must be set according to the duration of the simulation or at least is higher than the ``initial step``, otherwise the particles' average velocity, granular temperature, and/or grid will not be output.
```
acdaigneault, line 50: It is possible to set the default value the same as the time end in the Simulation control subsection?

acdaigneault: I just saw that this is the step and not the time. Is there's a reason why? I feel the time is easier to use.

doc/source/parameters/dem/post-processing.rst
```diff
@@ -0,0 +1,53 @@
...
+ end
+
+.. note::
+ The parameters default values are the ones presented above.
+
+* The ``Lagrangian post processing`` enables the use of ``calculate particles average velocity``, ``calculate granular temperature``, and ``write grid``.
+
+* The ``write grid`` parameter enables one to output the grid as VTU files. The output files can be used for post-processing. By default, it is ``true``. The files with the grid will have ``-grid`` in their names.
+
+* The ``calculate particles average velocity`` chooses whether the average velocity of the particles will be output or not. It generates a file with the average velocity of the particles.
```
acdaigneault, line 38: Can you explain a bit how the average velocity is calculated and how it is represented in the output? I don't understand how it works since it's exported on a grid.

doc/source/parameters/dem/post-processing.rst
acdaigneault: I tried the parameter and it generates `-postprocess_data` files and not independent `-grid` files.

acdaigneault (commented):
Hey nice PR! I have a comment that I did not address here which may lead to a discussion.
I'm not sure that I like the grid output with the post-process with the current frequency/start step/end step default values. It seems a lot of thinking to set proper parameters only to have a grid visualization. I think that's very interesting to have access to parameters to help on the grid output. But, in my opinion, I think it would be nice to have something similar as before if `write grid` is enable without having to set the other parameters. What do you think?

voferreira:
> Hey nice PR! I have a comment that I did not address here which may lead to a discussion. I'm not sure that I like the grid output with the post-process with the current frequency/start step/end step default values. It seems a lot of thinking to set proper parameters only to have a grid visualization. I think that's very interesting to have access to parameters to help on the grid output. But, in my opinion, I think it would be nice to have something similar as before if `write grid` is enable without having to set the other parameters. What do you think?

Thanks! You mean having the previous parameters as default? If that's it I agree with you. The current defaults need more attention from the user than necessary. I will try to address this.

doc/source/parameters/dem/post-processing.rst
voferreira: I think it is because it is an integer, not a double. Ifs with doubles are always a problem

acdaigneault:
@voferreira Yeah for instance, let's say I want a grid from the beginning to the end of the simulation. I don't know what is the end step, I just know the time. This is why I don't like the "end step" but I understand it's better for the ifs. Is the end step relevant for the post-processing? If not, maybe it should be stopped?

blaisb:
I think when write grid is true and the lagragian post-processing is true, then the grid should always be written.
The time start and time end are really used to decide WHEN you average the velocity and the temperature, and not when you output.
It's a very good comment @acdaigneault

voferreira:
As discussed, I removed the excessive parameters from the code and explained better what the Lagrangian post-processing does. The idea now is that the post-processing is done altogether when ``Lagrangian post-processing = true``, and the output of the post-processing follows the output of the particles. I think it is better to keep it simple like it is now. Please @acdaigneault and @blaisb, let me know what you think of it.

blaisb:
One thing @voferreira don't forget to update the prm file of the example also :)!

source/dem/lagrangian_post_processing.cc
```diff
@@ -173,20 +173,31 @@ LagrangianPostProcessing<dim>::calculate_cell_average_particles_velocity(
...
-  std::vector<std::string> average_solution_names(1, "average_velocity_x");
+  std::vector<std::string> average_solution_names;
+
+  // Write particles' average velocity
+  calculate_average_particles_velocity(triangulation, particle_handler);
+
+  average_solution_names.push_back("average_velocity_x");
   average_solution_names.push_back("average_velocity_y");
   if constexpr (dim == 3)
     average_solution_names.push_back("average_velocity_z");
```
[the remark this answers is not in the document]
acdaigneault: I was thinking about the same thing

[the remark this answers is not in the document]
acdaigneault: @voferreira I think storage is [avg_x_0, avg_y_0, avg_z_0, avg_x_1, ...] You can look at line 1570 of `navier_stokes_base.cc`, but not 100% sure since it's DoF data and not cell data, but it's should be similar. :)

doc/source/parameters/dem/post-processing.rst
```diff
@@ -1,53 +1,26 @@
...
+* The spatial average of the particle's velocity (in x, y, z and its norm) for each cell.
 
-  # Enable calculate granular temperature
-  set calculate granular temperature = false
+* The granular temperature for each cell.
 
-  # Choose time step to start calculating particles average velocity and/or granular temperature
-  set initial step = 0
+.. important::
+ The particles' average velocity and the granular temperature are cell based results calculated instantaneously at each outputting time-step. This means that they are only instantaneous spatial averages, i.e, DO NOT correspond to temporal averages measured between the outputs. For a higher resolution of the results in time, we recommend the use of a lower ``output frequency`` (:doc:`../dem/simulation_control`).
```
acdaigneault, line 12: Yay! I like it

acdaigneault (approved):
Thanks for the changes, I like it! Doc is really useful! I gave maybe for tips for the Bruno's comment. Also, I tried it and it works well. I have to say that watching the average velocities in cells reminds me of Minecraft haha.

PR #641, opened 2022-11-11
L2 projection nodes output

CaroleAnneDaunais:
# Description of the problem

Some of the post-processes variables (namely the Qcriterion, the Vorticity, ...) are calculated in a cell-by-cell fashion, meaning that their values at the nodes are undefined. This patch of code aims to rebuild the variables' profile using an L2-projection on the nodes. 

# Description of the solution

This is done by assembling the mass matrix, assembling the right hand side based on the solution values (fluid solution), and to solve the matrix system. A base class PostProcessorSmoothing is implemented, from which are derived classes that represent postprocessed variables. 

# How Has This Been Tested?

No tests for this. Should there be?

# Documentation

The Post-processing documentation is updated

# Future changes

- The VorticityPostProcessorSmoothing class yet has to be implemented

# Comments

- A small change is made `examples/resolved-cfd-dem/sphere-carreau-with-sharp-interface/sphere-carreau-with-sharp-interface.prm` that does not have to do with this feature.


include/solvers/post_processors_smoothing.h
```diff
@@ -0,0 +1,181 @@
...
+
+// Numerics
+#include <deal.II/distributed/solution_transfer.h>
+
+#include <deal.II/numerics/data_postprocessor.h>
+
+// Rheological models
+#include <core/rheological_model.h>
+
+#include <solvers/simulation_parameters.h>
```
blaisb, line 38: Do you really need all these includes? It seems like a lot but if you think you do it's ok.

include/solvers/post_processors_smoothing.h
```diff
@@ -0,0 +1,181 @@
...
+
+using namespace dealii;
+
+
+/**
+ * A base class that holds the dof_handler, fe_values and the simulation output
+ * to calculate post-processed parameters
+ *
+ * @tparam dim An integer that denotes the dimension of the space in which
+ * the flow is solved
```
blaisb, line 52:
```suggestion
 * the variable being smoothed is solved
```

include/solvers/post_processors_smoothing.h
```diff
@@ -0,0 +1,181 @@
...
+ * @tparam mpi_communicator Allowing communication between cores
+ */
+
+template <int dim, typename VectorType>
+class PostProcessorSmoothing
+{
+public:
+  // Member functions
+  PostProcessorSmoothing(
+    std::shared_ptr<parallel::DistributedTriangulationBase<dim>> triangulation,
```
blaisb, line 66: If you don't need to store it locally, I would pass isntead just a reference to the DistributedTriangulationBase instead of passing a shared ptr. We are trying to have cleaner interfaces this way :).

include/solvers/post_processors_smoothing.h
```diff
@@ -0,0 +1,181 @@
...
+    const SimulationParameters<dim> &simulation_parameters,
+    const unsigned int &             number_quadrature_points,
+    const MPI_Comm &                 mpi_communicator);
```
blaisb, line 69: Can't you get the communicator straight from the triangulation itself?

source/solvers/navier_stokes_base.cc
```diff
@@ -1668,8 +1669,36 @@ NavierStokesBase<dim, VectorType, DofsType>::write_output_results(
...
+    this->triangulation,
+    this->simulation_parameters,
+    number_quadrature_points,
+    this->mpi_communicator);
+
+  if (!this->simulation_parameters.post_processing.smoothing)
+    {
+      data_out.add_data_vector(solution, qcriterion);
+    }
+  else
```
blaisb, line 1683: I think it would be easier to read it the other way. If smoothing, else. But that's your final call. :)

source/solvers/post_processors_smoothing.cc
```diff
@@ -0,0 +1,323 @@
...
+
+  //**********************************************
+  // Trillinos Wrapper ILU Preconditioner
+  //*********************************************
+  const double ilu_fill =
+    this->simulation_parameters.linear_solver.ilu_precond_fill;
+  const double ilu_atol =
+    this->simulation_parameters.linear_solver.ilu_precond_atol;
+  const double ilu_rtol =
+    this->simulation_parameters.linear_solver.ilu_precond_rtol;
```
blaisb, line 124: I am not sure I would take these parameters from the simulation parameters.
The L2 projection is a trivial matrix, so I would just put default values here and be done with it.
Fill level of 0, atol of 1e-15, rtol of 1 for example would be just fine.

source/solvers/post_processors_smoothing.cc
```diff
@@ -0,0 +1,323 @@
...
+{
+  return dof_handler;
+}
+
+template class PostProcessorSmoothing<2, TrilinosWrappers::MPI::Vector>;
+template class PostProcessorSmoothing<3, TrilinosWrappers::MPI::Vector>;
+template class PostProcessorSmoothing<2, TrilinosWrappers::MPI::BlockVector>;
+template class PostProcessorSmoothing<3, TrilinosWrappers::MPI::BlockVector>;
+
+template <int dim, typename VectorType>
```
blaisb, line 169: I am curious, how will you manage the fact that vorticity has multiple components? (e.g. ti's a vector)

blaisb (changes requested):
Few comments there and there, but very cool work :)!
I am just a bit wondering what you will do with Vorticity. Since vorticity is a vector (n components) how will you manage that? Maybe the best approach would be to template the base class with the number of components since this is always known at compile time?

include/solvers/post_processors_smoothing.h
CaroleAnneDaunais: Good question, how does one do that?

include/solvers/post_processors_smoothing.h
CaroleAnneDaunais: You are right, some of them were removed.

include/solvers/post_processors_smoothing.h
CaroleAnneDaunais: Good catch, it is changed.

source/solvers/navier_stokes_base.cc
CaroleAnneDaunais: Absolutely. I changed it.

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -22,6 +22,7 @@ This subsection controls the post-processing other than the forces and torque on
   set verbosity                   = quiet
   set calculation frequency       = 1
   set output frequency            = 1
+  set smoothing                   = false
```
Luckabarbeau, line 25: I think an other name than smoothing would be better. Maybe continues derivatives or make results continuous?

Luckabarbeau (commented):
Good job on this PR. A few things to address, but otherwise, everything looks good! This will allows us to do very nice figures!

doc/source/parameters/cfd/post_processing.rst
CaroleAnneDaunais: I was trying to fit the class name, but I understand the comment. How about `smoothed_output_fields` ?

source/solvers/post_processors_smoothing.cc
```diff
@@ -0,0 +1,320 @@
...
+    true);
+
+  TrilinosWrappers::SolverCG solver(solver_control);
+
+  //**********************************************
+  // Trillinos Wrapper ILU Preconditioner
+  //*********************************************
+  const double ilu_fill = 1.;
+  const double ilu_atol = 1e-15;
+  const double ilu_rtol = 0;
```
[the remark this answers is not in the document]
CaroleAnneDaunais: As Bruno made the opposite comment, I will keep these values. Also, I made a mistake in the parameter values, it is changed.

source/solvers/post_processors_smoothing.cc
CaroleAnneDaunais: I think that I will either way need to calculate the smoothed field 1 dim at a time anyways, is there a way for me to convert a 'TrilinosWrappers::MPI::Vector' to a 'std::vector<double>' ?

doc/source/parameters/cfd/post_processing.rst
blaisb: smoothed_output_fields has a nice ring to it.

source/solvers/post_processors_smoothing.cc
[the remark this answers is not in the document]
blaisb: I suggest using the following:
ilu_fill=0
illu_atol=1e-15
illu_rtol=1

The matrix is trivial to solve anyway, but rtol should be by default 1 not 0 :)

blaisb:
@CaroleAnneDaunais I think if you address the hanging comments this will be good for a merge. The extension to vorticity with multiple components i can take care of it later on :). It will require a bit of templating but nothing too big

include/solvers/post_processors_smoothing.h
```diff
@@ -0,0 +1,172 @@
...
+ * A base class that holds the dof_handler, fe_values and the simulation output
+ * to calculate post-processed parameters
+ *
+ * @tparam dim An integer that denotes the dimension of the space in which
+ * the variable being smoothed is solved
+ *
+ * @tparam triangulation Flow triangulation
+ * @tparam simulation_parameters The simulation parameters
+ * @tparam number_quadrature_points The number of quadrature points
+ * @tparam mpi_communicator Allowing communication between cores
```
lpsaavedra, line 46:
```suggestion
 * @tparam dim An integer that denotes the dimension of the space in which
 * the variable being smoothed is solved
 * @tparam triangulation Flow triangulation
 * @tparam simulation_parameters The simulation parameters
 * @tparam number_quadrature_points The number of quadrature points
 * @tparam mpi_communicator Allowing communication between cores
```

include/solvers/post_processors_smoothing.h
CaroleAnneDaunais: Done :)

source/solvers/post_processors_smoothing.cc
CaroleAnneDaunais: Makes sense. Let's do this.

blaisb (approved):

PR #648, opened 2022-11-21
New relaxation scheme resolved

Luckabarbeau:
# Description of the problem

- The relaxation scheme could be improved and have a more established mathematical formulation.
- Some values were not properly defined at the restart, and this played a bigger role in the new relaxation formulation

# Description of the solution

- Implement the new mathematical formulation.
- Define a value for the last velocity iteration, which was causing a bug with the new mathematical formulation at the restart if it was not properly defined.

# How Has This Been Tested?

- Use in the simulation of multiple particles. All previous tests were the particle dynamics is coupled as been revised and adjusted.


source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1738,17 +1745,20 @@ GLSSharpNavierStokesSolver<dim>::integrate_particles()
             -bdf_coefs[0] - 0.5 * 2. / 5 * volume * fluid_density *
                               particles[p].radius * particles[p].radius *
                               inv_inertia.norm() / dt;
-
+          // Evaluate the relaxation parameter with a generalization of the
+          // secante method.
```
lpsaavedra, line 1749:
```suggestion
          // Evaluate the relaxation parameter using a generalization of the
          // secant method.
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1637,14 +1637,21 @@ GLSSharpNavierStokesSolver<dim>::integrate_particles()
           double d_residual_dv =
             -bdf_coefs[0] -
             0.5 * volume * fluid_density / particles[p].mass / dt + DBL_MIN;
```
lpsaavedra, line 1639: I think the name of this variable is not clear, as I couldn't understand exactly what this was. After talking to Lucka, I suggest to change the name to something related to alpha, as this is no longer the original definition of `d_residual_dv`. The same comment applies to all the following definitions of `d_residual_dv` and `d_residual_domega`

lpsaavedra (commented):
Good work. I have only two minor comments.

source/fem-dem/gls_sharp_navier_stokes.cc
Luckabarbeau: All the variables name use in the calculation of the update have been updated to more verbose name

blaisb (approved):

blaisb:
@Luckabarbeau  can you fix conflicts and then I can merge?

PR #649, opened 2022-11-22
Fix crash in debug for Lagrangian post-processing and documentation

blaisb:
# Description of the problem

- The Lagrangian post-processing was crashing in debug because of a conflicting DataOut attachment

# Description of the solution

- Fixed it
- Also fixed a typo in the documentation


acdaigneault (approved):
It works :) As CI has flagged, there's now an used parameter. Thx for the fix!

PR #652, opened 2022-11-29
Optimize rbf

oguevremont:
# Description of the problem

- The RBF shapes in the Sharp IB solver were very slow, which was in part due to looping over every RBF node for distance evaluations.
- This PR adds a precomputing step allowing the solver to map every node only to the cells on which they have a non-null contribution, which uses the fact that the basis functions have a compact support (null when distance>support radius). The following evaluations are then much faster and should scale better.
- This PR also adds the analytical gradients for basis functions, which cut the gradient evaluation time compared to numerical gradient.

# Description of the solution

- A precomputing step was added for Sharp RBF
- The analytical gradients were added instead of using numerical gradients for Sharp RBF

# How Has This Been Tested?

- The RBF test file was visually inspected
- The computing time was compared before and after the changes for a computing costly case

# Documentation

- No documentation change

# Future changes

- An example with a complex shape showing the workflow from STL file (surface mesh) to RBF to flow profile will be added
- A minor bug regarding Sharp IB will be fixed soon: orientation definition is inverted
- A composite shape parameter sub-subsection for Sharp IB solver will be defined


source/core/shape.cc
```diff
@@ -609,6 +761,89 @@ RBFShape<dim>::initialize_bounding_box()
...
+      {
+        const auto cell_parent     = cell->parent();
+        const auto parent_iterator = likely_nodes_map.find(cell_parent);
+        if (parent_iterator != likely_nodes_map.end())
+          {
+            parent_found = true;
+            temporary_iterable_nodes.swap(iterable_nodes);
+            iterable_nodes.swap(parent_iterator->second);
+          }
+      }
```
Luckabarbeau, line 790: If I understand correctly, you check if the cell has a parent, and if it has a parent, you use the parent cell to find the iterable nodes for this cell. Then you check which iterable node applied to the new cell and only keeps those has your iterable nodes for the cell. 

This will be fast only if you start with a coarse mesh and then refine it on a single core. But the problem is that when you refine, you are not sure that the new locally owned cell matches the previous cells set before the refinement. In most cases, it won't help you a lot since the new processor does not necessarily have the relevant information from the parent cell.

I don't know if starting from level 0 of the mesh to find the relevant iterable nodes would not be significantly faster. since you could use the full potential of the tree of the cell. I think it would be possible to use a slightly modified version of find_cells_around_point_with_tree from Lethegridtools.

Tell me if I miss something!

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3230,7 +3234,6 @@ GLSSharpNavierStokesSolver<dim>::write_checkpoint()
         ".ib_particles";
       std::ofstream output(filename.c_str());
       this->simulation_control->save(prefix);
-      ib_particles_pvdhandler.save(prefix + ".ib_particles");
```
Luckabarbeau, line 3233: this should not be erased

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3361,7 +3364,6 @@ GLSSharpNavierStokesSolver<dim>::read_checkpoint()
     this->simulation_parameters.simulation_control.output_folder + prefix +
     ".ib_particles";
 
-  ib_particles_pvdhandler.read(prefix + ".ib_particles");
```
Luckabarbeau, line 3364: same as above

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3535,16 +3537,7 @@ GLSSharpNavierStokesSolver<dim>::read_checkpoint()
...
-  for (unsigned int p_i = 0; p_i < particles.size(); ++p_i)
-    {
-      particles[p_i].velocity_iter        = particles[p_i].velocity;
-      particles[p_i].impulsion_iter       = particles[p_i].impulsion;
-      particles[p_i].omega_iter           = particles[p_i].omega;
-      particles[p_i].omega_impulsion_iter = particles[p_i].omega_impulsion;
-      particles[p_i].residual_velocity    = DBL_MAX;
-      particles[p_i].residual_omega       = DBL_MAX;
-    }
+  // Finish the time step of the particle.
```
Luckabarbeau, line 3540: again same as above

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3763,6 +3756,10 @@ GLSSharpNavierStokesSolver<dim>::solve()
   define_particles();
   this->setup_dofs();
   this->box_refine_mesh();
+  for (unsigned int p_i = 0; p_i < particles.size(); ++p_i)
+    {
+      particles[p_i].update_precalculations(this->dof_handler, this->mapping);
+    }
```
Luckabarbeau, line 3762: I would add a timer to this function call or around it

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3794,6 +3791,11 @@ GLSSharpNavierStokesSolver<dim>::solve()
           refine_ib();
           NavierStokesBase<dim, TrilinosWrappers::MPI::Vector, IndexSet>::
             refine_mesh();
+          for (unsigned int p_i = 0; p_i < particles.size(); ++p_i)
+            {
+              particles[p_i].update_precalculations(this->dof_handler,
+                                                    this->mapping);
+            }
```
Luckabarbeau, line 3798: same as above

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3856,6 +3858,11 @@ GLSSharpNavierStokesSolver<dim>::solve()
...
```
Luckabarbeau, line 3865: same as above

Luckabarbeau (commented):
Nice job once this is done, it will make RBF_shapr much faster! but some work to do and some checks on the gain made with the new scheme to identify the iterable nodes. At the moment, the complexity of the algorithm is still O(n_rbf_nodes&*n_cell) after a refinement. I think it is possible to optimize a bit more the identification of the iterable nodes by starting the iteration from level 0 on all proc. This could give you an approximate complexity of  O(n_rbf_nodes&*n_cells_level_0*log(n_cells))

source/core/shape.cc
```diff
@@ -609,6 +761,89 @@ RBFShape<dim>::initialize_bounding_box()
...
+
+  double     distance, max_distance;
+  double     cell_diameter = cell->diameter();
+  Point<dim> centered_support_point;
+
+  likely_nodes_map[cell].reserve(max_number_of_nodes);
+  for (auto &node_id : iterable_nodes)
+    {
+      // We only check for one support point, but we use a high security
+      // factor. This allows not to loop over all support points.
```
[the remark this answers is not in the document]
oguevremont: This was from another code. Now there is no security factor, just cell_diagonal+support_radius

source/core/shape.cc
oguevremont: No you understood correctly! It could indeed be faster
I'll try it

blaisb (changes requested):
I have a bunch of comments there and there. Nothing too major

include/core/shape.h
```diff
@@ -656,6 +707,7 @@ class RBFShape : public Shape<dim>
     C0C2,
     C1C2,
     C2C2,
+    COSINUS,
```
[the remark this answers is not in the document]
oguevremont: Because otherwise there is overlap with the basis functions' names. I can rename them though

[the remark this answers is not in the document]
blaisb: Nah it's ok :). But in english it should be COSINE or why not just COS?

source/core/shape.cc
oguevremont: It's much faster

source/core/ib_stencil.cc
```diff
@@ -149,35 +213,94 @@ IBStencil<dim>::point_for_cell_detection(IBParticle<dim> & p,
...
 
+template <int dim>
+double
+IBStencil<dim>::ib_velocity(
+  IBParticle<dim> &                                     p,
+  const Point<dim> &                                    dof_point,
+  unsigned int                                          component,
+  const typename DoFHandler<dim>::active_cell_iterator &cell_guess)
+{
+  double v_ib = 0;
```
blaisb, line 229: What is v_ib? can you use a more verbose variable name?

source/core/ib_stencil.cc
```diff
@@ -154,6 +218,72 @@ IBStencil<dim>::point_for_cell_detection(IBParticle<dim> & p,
...
+  // Return the value of the IB condition for that specific stencil.
+  double v_ib = 0;
+
+  Tensor<1, 3, double> radial_vector;
+  Point<dim>           closest_point;
+  p.closest_surface_point(dof_point, closest_point, cell_guess);
+  Point<dim> position_to_surface;
+  position_to_surface    = closest_point - p.position;
+  double radial_distance = position_to_surface.norm();
+  if (dim == 2)
```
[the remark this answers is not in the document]
blaisb: comment still not adressed

source/core/ib_stencil.cc
```diff
@@ -149,35 +213,94 @@ IBStencil<dim>::point_for_cell_detection(IBParticle<dim> & p,
...
-  double radial_distance = position_to_surface.norm();
+  position_to_surface = closest_point - p.position;
+  radial_vector[0]    = position_to_surface[0];
+  radial_vector[1]    = position_to_surface[1];
+  // We have to do that conversion as there is no proper conversion from
+  // tensor of dim 2 to 3.
+  if constexpr (dim == 3)
+    radial_vector[2] = position_to_surface[2];
+
   if (dim == 2)
```
blaisb, line 302:
```suggestion
  if constexpr(dim == 2)
```

source/core/ib_stencil.cc
```diff
@@ -192,12 +315,6 @@ IBStencil<dim>::ib_velocity(IBParticle<dim> & p,
     }
   if (dim == 3)
```
blaisb, line 316:
```suggestion
  if constexpr(dim == 3)
```

blaisb (changes requested):
Few more comments to address

source/core/ib_stencil.cc
oguevremont: I renamed it to immersed_boundary_velocity, I think that's better

blaisb (approved):

PR #659, opened 2022-12-08
Fix naming and use set where needed

blaisb:
# Description of the problem

- The find cell neighbor function was really slow when meshes were big. This was actually a bottle neck that took 30% of simulation time in one of the benchmark I was doing. This was caused by a huge vector (n_local_cells size) being searched over constantly. 

# Description of the solution

- Replace this vector with a set. One benchmark case is now 12% faster on a single core because of this...

# How Has This Been Tested?

- All previous tests pass as before.

# Comments

- Continuing to try and improve the DEM.


include/dem/data_containers.h
```diff
@@ -221,7 +221,11 @@ namespace DEM
 
     // [cell iterators]
     typedef std::vector<typename Triangulation<dim>::active_cell_iterator>
-      cell_container;
+      cell_vector;
+
+    // [cell iterators]
+    typedef std::set<typename Triangulation<dim>::active_cell_iterator>
+      cell_set;
```
acdaigneault, line 228: Why not `std::unordered_set`?

blaisb: We don't have a hash function for active_cell_iterator right now so we cannot use unordered set. I'll work on fixing this in the mid-term, but for now we can only use sets

acdaigneault (approved):

PR #660, opened 2022-12-08
Fix wetting

jeannejoachim:
# Description of the problem

- pressure gradient condition in peeling/wetting incorrect (at the level of dim, rather than quadrature_points)
- when one fluid is more than a hundred times denser than the other, the wetting mechanism can result in the denser fluid crawling on the wall in a non-physical way
- in 2D, mass and volume monitored are in fact mass/length and surface. This was not clear for the users.

# Description of the solution

- condition on pressure gradient changed to be at the level of the quadrature_points
- enable to turn off wetting (and so peeling, for good measure), which gives better results (tested on industrial case)
- clarify monitoring output

# How Has This Been Tested?

Tests have been updated, there is no change when we consider the two phases to be water and air for instance.

# Documentation

Updated and clarified in volume of fluid

# Future changes

When we come up with an approach that is less of a heuristic, this fix can be removed. However for now it gives very satisfactory results on industrial case, while not breaking more academic cases (breaking-dam, bouncing fluid etc.)

### DIFF
--- a/doc/source/parameters/cfd/volume_of_fluid.rst
+++ b/doc/source/parameters/cfd/volume_of_fluid.rst
@@ -136,7 +136,7 @@ The default values of the VOF parameters are given in the text box below.
 * ``subsection peeling wetting``: Peeling and wetting mechanisms are very important to consider when there are solid boundaries in the domain, like a wall. If the fluid is already on the wall and its velocity drives it away from it, the fluid should be able to detach from the wall, meaning to `peel` from it. If the fluid is not already on the wall and its velocity drives it toward it, the fluid should be able to attach to the wall, meaning to `wet` it. This subsection defines the parameters for peeling and wetting mechanisms at the VOF boundaries, as defined in :doc:`boundary_conditions_multiphysics`. 
 
   .. important::
-    This peeling/wetting mechanism implementation is a heuristic. It has been developed to meet the need of specific projects and gave satisfactory results as is, but it has not been broadly tested nor demonstrated, so its results should be considered with cautions. Do not hesitate to write to the team through the `Lethe github page <https://github.com/lethe-cfd/lethe>`_ would you need assistance.
+    This peeling/wetting mechanism implementation is a heuristic. It has been developed to meet the need of specific projects and gave satisfactory results as is, but it has not been broadly tested nor demonstrated, so its results should be considered with caution. Do not hesitate to write to the team through the `Lethe GitHub page <https://github.com/lethe-cfd/lethe>`_ would you need assistance.
 
   .. warning::
 
--- a/doc/source/parameters/cfd/volume_of_fluid.rst
+++ b/doc/source/parameters/cfd/volume_of_fluid.rst
@@ -151,7 +151,7 @@ The default values of the VOF parameters are given in the text box below.
     The cell is then filled with the lower density fluid by changing its phase value progressively.
 
     .. important::
-      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered as the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
+      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
 
   * ``enable wetting``: controls if wetting mechanism is enabled. Wetting occurs in a cell where those conditions are met: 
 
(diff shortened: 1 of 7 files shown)


blaisb (approved):
Nothing to add on my side. Very good PR

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -135,54 +136,60 @@ The default values of the VOF parameters are given in the text box below.
...
 
     * the cell is in the domain of the higher density fluid,
     * the cell pressure value is below the average pressure of the ``monitored fluid`` (``fluid 1`` by default, see ``subsection mass conservation``), and
     * the pressure gradient is negative for more than half of the quadrature points.
 
     The cell is then filled with the lower density fluid by changing its phase value progressively.
 
-  * Wetting occurs in a cell where those conditions are met: 
+    .. important::
+      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered as the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
```
voferreira, line 154:
```suggestion
      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
```

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -135,54 +136,60 @@ The default values of the VOF parameters are given in the text box below.
...
+
+  * ``enable wetting``: controls if wetting mechanism is enabled. Wetting occurs in a cell where those conditions are met:
```
voferreira, line 156:
```suggestion
  * ``enable wetting``: controls if the wetting mechanism is enabled. Wetting occurs in a cell where those conditions are met: 
```

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -135,54 +136,60 @@ The default values of the VOF parameters are given in the text box below.
...
     * the cell pressure value is above the average pressure of the ``monitored fluid`` (``fluid 1`` by default, see ``subsection mass conservation``), and
     * the pressure gradient is positive for more than half of the quadrature points.
 
-    The cell is then filled with the higher density fluid by changing its phase value.
+    The cell is then filled with the higher density fluid by changing its phase value progressively.
 
-    .. tip::
-      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered as the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
+    .. tip ::
+      Using ``set enable wetting = false`` and relying on the ``diffusivity`` to wet the boundaries (see :ref:`improve wetting`) can give better results when the densities of the two fluids are of very different order of magnitude.
```
voferreira, line 165:
```suggestion
      Using ``set enable wetting = false`` and relying on the ``diffusivity`` to wet the boundaries (see :ref:`improve wetting`) can give better results when the densities of the two fluids are of a very different order of magnitude. 
```

doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -135,54 +136,60 @@ The default values of the VOF parameters are given in the text box below.
...
-    The cell is then filled with the higher density fluid by changing its phase value.
+    The cell is then filled with the higher density fluid by changing its phase value progressively.
 
-    .. tip::
-      Even if ``monitoring`` is not enabled, the ``monitored fluid`` (``fluid 1`` by default) will be considered as the fluid of interest for the average pressure calculation in the peeling/wetting mechanism.
+    .. tip ::
+      Using ``set enable wetting = false`` and relying on the ``diffusivity`` to wet the boundaries (see :ref:`improve wetting`) can give better results when the densities of the two fluids are of very different order of magnitude. 
 
-.. warning::
+      Typically, when one fluid is more than a hundred times denser than the other, the wetting mechanism can result in the denser fluid crawling on the wall in a non-physical way. Again, this is still a heuristic, so do not hesitate to write to the team through the `Lethe github page <https://github.com/lethe-cfd/lethe>`_ would you need assistance.
```
voferreira, line 167:
```suggestion
      Typically, when one fluid is more than a hundred times denser than the other, the wetting mechanism can result in the denser fluid crawling on the wall in a non-physical way. Again, this is still a heuristic, so do not hesitate to write to the team through the `Lethe GitHub page <https://github.com/lethe-cfd/lethe>`_ would you need assistance.
```

voferreira (approved):
Seems clear to me. Maybe I would add more comments to this block of code, but it is good like this.

jeannejoachim:
Thank you @voferreira for your suggestions :) :eagle: 
@blaisb once the CI is finished, this is ready to be merged ^^

PR #661, opened 2022-12-08
Optimize properties for force and remove mpi calls

blaisb:
# Description of the problem

- Force calculations were made using a unordered map of unordered map structure which became very slow for nothing.
- There was an extra synchronisation in the particle-wall force that was not necessary
- 
# Description of the solution

- Use a serialized std::vector<double> as the property. Use inlined integer function to get the right vectorized index
- Remove extra synchronisation
# How Has This Been Tested?

- All tests still work

# Future changes
- Am going crazy over this, but we have improved performance by 20-30% across the board in the last 1-2 days. I think I can get another 20-30% there and there.


source/dem/particle_particle_contact_force.cc
```diff
@@ -1048,6 +1058,16 @@ ParticleParticleContactForce<dim, contact_model>::
   particle_two_properties[PropertiesIndex::type] = 0;
   particle_two_properties[PropertiesIndex::dp]   = 2.0 * particle_two_radius;
 
+  n_particle_types = 1;
```
acdaigneault, line 1061: Why that?

acdaigneault: There's only 1 type available for IB?

acdaigneault (approved):
Nice improvements :) 
`particle_wall_linear_force` and `particle_wall_nonlinear_force` should be modified in a future PR to follow the `particle_particle_contact_force` code and improvements.

PR #665, opened 2022-12-12
Emplace back fine search

blaisb:
# Description of the problem

- Some inefficient containers were still being use
- Some allocation was being done that could be spared

# Description of the solution

- Use emplace instead of insert in location where it saves computational cost
- Removed unordered map for the wall-particle contact models

# How Has This Been Tested?

- All tests pass

# Comments

- This might be my last one for a short while. I am very happy with the speed-up gained in the last few weeks.


include/dem/particle_particle_contact_info_struct.h
```diff
@@ -24,15 +24,26 @@
...
- * This struct handles the information related to the calculation of the
+ * This class handles the information related to the calculation of the
  * particle-particle contact force
  */
 
 using namespace dealii;
 
 template <int dim>
-struct particle_particle_contact_info_struct
+class particle_particle_contact_info
```
acdaigneault, line 34: Would it be a good idea to add information of the particles (overlap, properties, and on..) that is destroyed after their usage (but not the tangential torque)? Just a question of readability and reducing the number of arguments.

include/dem/particle_particle_contact_info_struct.h
```diff
@@ -24,15 +24,26 @@
 #  define particle_particle_contact_info_struct_h
```
acdaigneault, line 24: I would change the name of the file (removing the "_struct")

acdaigneault (commented):
Sad that this PR for speeding things up might be the last one :p 
I just reviewed the code, did not test it yet, will do tomorrow, but it looks good!

include/dem/particle_particle_contact_info_struct.h
blaisb: Good question, I am not sure about it but it is something to investigate.

blaisb:
All comments have been applied @acdaigneault . Thanks for pointing them out :)

include/dem/particle_particle_contact_info.h
```diff
@@ -23,16 +21,43 @@
...
+    , particle_two(particle_two)
+  {}
+
+  /** @brief Dummy constructor for cases where the two particle iterators are unknown.
+   *  This constructor is only used in the resolved CFD-DEM and should be
+   * deprecated eventually.
+   */
+
+  inline particle_particle_contact_info()
+  {}
```
acdaigneault, line 58: Same here

source/dem/particle_wall_linear_force.cc
```diff
@@ -61,24 +62,20 @@ ParticleWallLinearForce<dim>::ParticleWallLinearForce(
...
-           (particle_rolling_friction_coefficient +
-            wall_rolling_friction_coefficient + DBL_MIN)});
+      this->effective_coefficient_of_restitution[i] =
+        2 * particle_restitution_coefficient * wall_restitution_coefficient /
+        (particle_restitution_coefficient + wall_restitution_coefficient +
+         DBL_MIN);
+
+      this->effective_coefficient_of_friction[i] =
+        2 * particle_friction_coefficient * wall_friction_coefficient /
+        (particle_friction_coefficient + wall_friction_coefficient + DBL_MIN);
```
acdaigneault, line 72: Just curious, why is there DBL_MIN? Is it to avoid division by 0 so a NaN result?

source/dem/particle_wall_nonlinear_force.cc
```diff
@@ -32,6 +33,8 @@ ParticleWallNonLinearForce<dim>::ParticleWallNonLinearForce(
   const double wall_rolling_friction_coefficient =
     dem_parameters.lagrangian_physical_properties.rolling_friction_wall;
 
+
+
```
acdaigneault, line 37: Unnecessary spaces

acdaigneault (approved):
I did add a few comments/questions here and there and test it. :)

source/dem/particle_wall_linear_force.cc
blaisb: Yup

source/dem/particle_wall_nonlinear_force.cc
blaisb, line 37:
```suggestion
```

PR #669, opened 2022-12-16
Add new primitive shapes and composites

oguevremont:
# Description of the problem

- Lethe didn't have an interface for composite shapes
- More primitive shapes were needed

# Description of the solution

- The cylinder, tube, cylindrical helix and composite shapes were added

# How Has This Been Tested?

- Two composite shapes tests
- Cylindrical tube test 

# Documentation

- Parameters documentation for Resolved CFD DEM was adapted

# Future changes

- In this PR both RBF and composite shapes initialization is mostly done in Parameters, but the following PR will move this to the IBParticle class
- Examples with composite and RBF shapes will be added


applications_tests/gls_sharp_navier_stokes_2d/couette2d_gls_poisson.output
```diff
@@ -7,26 +7,26 @@ Running on 1 MPI rank(s)...
 Steady iteration :        1/3
 *****************************************************************
 div u : 0
-L2 error velocity : 0.00418651 L2 error pressure: 0
```
blaisb, line 10: What happened here? Why is the error so big all of a sudden?

applications_tests/gls_sharp_navier_stokes_2d/couette2d_gls_poisson.output
```diff
@@ -7,26 +7,26 @@ Running on 1 MPI rank(s)...
...
+L2 error velocity : 0.171083 L2 error pressure: 0
 
 *****************************************************************
 Steady iteration :        2/3
 *****************************************************************
    Number of active cells:       4096
    Number of degrees of freedom: 12675
    Volume of triangulation:      4
-div u : 5.12312e-10
-L2 error velocity : 0.00109947 L2 error pressure: 1.05594e-08
```
blaisb, line 19: What happened here again? Why is the error freaking huge all of a sudden for the velocity field?

applications_tests/gls_sharp_navier_stokes_3d/tube_shape_test.prm
```diff
@@ -0,0 +1,140 @@
...
+	set id = 4
+        set type              = noslip
+    end
+    subsection bc 5
+	set id = 5
+        set type              = noslip
+    end
+end
+
+subsection particles
```
blaisb, line 65: Can you indent the PRM file as well as the other prm files?

applications_tests/gls_sharp_navier_stokes_3d/tube_shape_test.prm
```diff
@@ -0,0 +1,140 @@
...
+  set ilu preconditioner absolute tolerance  = 1e-20
+  set ilu preconditioner relative tolerance  = 1.00
+  set amg aggregation threshold                 = 1e-20
+  set amg n cycles                              = 1
+  set amg preconditioner ilu absolute tolerance = 1e-20
+  set amg preconditioner ilu fill               = 0
+  set amg preconditioner ilu relative tolerance = 1.00
+  set amg smoother overlap                      = 1
+  set amg smoother sweeps                       = 2
+  set amg w cycles                              = false
```
blaisb, line 137: Remove the options you don't need

blaisb (changes requested):
First batch of comments, will continue tomorrow

include/core/shape.h
```diff
@@ -603,31 +606,70 @@ class DeathStar : public Shape<dim>
...
+   */
+  CompositeShape<dim>(
+    std::vector<std::shared_ptr<Shape<dim>>> components_vector,
+    const Point<dim> &                       position,
+    const Tensor<1, 3> &                     orientation)
+    : Shape<dim>(0., position, orientation)
+  {
+    size_t number_of_components = components_vector.size();
+
+    // run for loop from 0 to vecSize
```
acdaigneault, line 660: Please explain what is done here. It's a bit unclear.

include/core/shape.h
```diff
@@ -1243,4 +1288,172 @@ class RBFShape : public Shape<dim>
...
+   * @param position position of the barycenter of the cylinder
+   * @param orientation orientation of the cylinder
+   */
+  CylindricalTube<dim>(double              radius_inside,
+                       double              radius_outside,
+                       double              half_length,
+                       const Point<dim> &  position,
+                       const Tensor<1, 3> &orientation)
+    : Shape<dim>((radius_outside + radius_inside) / 2, position, orientation)
+    , radius((radius_outside + radius_inside) / 2)
```
acdaigneault, line 1361:
```suggestion
    : Shape<dim>((radius_outside + radius_inside) / 2., position, orientation)
    , radius((radius_outside + radius_inside) / 2.)
```

source/core/parameters.cc
```diff
@@ -2375,13 +2383,179 @@ namespace Parameters
...
                 }
+              particles[i].initialize_shape(shape_type, shape_arguments);
+            }
+          else if (shape_type == "composite")
+            {
+              if (shape_arguments_str == "1") // Default case
+                {
+                  // Default
+                  shape_type      = "sphere";
+                  shape_arguments = std::vector<double>(1.);
```
acdaigneault, line 2394: Maybe it would be nice to add a warning that the filename of the composite is not provided and the default value is a sphere. No one will use a composite shape type to have a default sphere and not having the filename would be probably forgotten by the user.

source/core/parameters.cc
```diff
@@ -2375,13 +2383,179 @@ namespace Parameters
...
+                    unsigned int,
+                    std::tuple<typename CompositeShape<dim>::BooleanOperation,
+                               unsigned int,
+                               unsigned int>>
+                    operations;
+                  // In the file, we first loop over all component shapes, then
+                  // we loop over operations
+                  std::ifstream myfile(shape_name);
+                  // open the file.
+                  if (myfile.is_open())
```
acdaigneault, line 2413:
```suggestion
                  // Read file line by line for section names or arguments 
                  if (myfile.is_open())
```

source/core/parameters.cc
acdaigneault, line 2394:
```suggestion
                  shape_arguments = std::vector<double>(1.);
                  
                  particles[i].initialize_shape(shape_type, shape_arguments);
```

source/core/parameters.cc
```diff
@@ -2375,13 +2383,179 @@ namespace Parameters
...
+                    }
+                  else
+                    std::cout << "Unable to open file";
+                }
             }
           else
             {
               shape_arguments =
                 Utilities::string_to_double(shape_arguments_str_list);
+              particles[i].initialize_shape(shape_type, shape_arguments);
```
acdaigneault, line 2557: The previous default value for composite (sphere) won't work here since it's not initialized. (I suggested to add the initialization in the default part).

source/core/shape.cc
```diff
@@ -494,9 +563,9 @@ double
 CompositeShape<dim>::displaced_volume(const double fluid_density)
 {
   double solid_volume = 0;
-  for (const std::shared_ptr<Shape<dim>> &elem : components)
+  for (auto const &[component_id, component] : components)
     {
-      solid_volume += elem->displaced_volume(fluid_density);
+      solid_volume += component->displaced_volume(fluid_density);
```
acdaigneault, line 568: Just curious, when there is an operation, how is he solid volume not the sum of the 2 shapes and only the volume of the remaining solid?

acdaigneault (commented):
Nice feature to have here! This is my first batch of comments.

include/core/shape.h
```diff
@@ -603,31 +606,70 @@ class DeathStar : public Shape<dim>
...
  */
 template <int dim>
 class CompositeShape : public Shape<dim>
 {
 public:
+  enum class BooleanOperation : int
+  {
+    Union,
+    Difference,
+    Intersection,
```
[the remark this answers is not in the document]
oguevremont: Good suggestion. I tried refactoring the names but union is reserved as a special type of class by Cpp. We can leave Union or find another keyword

applications_tests/gls_sharp_navier_stokes_2d/couette2d_gls_poisson.output
oguevremont: I changed the omega by mistake, I'll fix it

source/core/parameters.cc
oguevremont: Good catch!

source/core/shape.cc
oguevremont: It's because it's an approximate volume, but as it's a function that it never called for non spheres I'll remove it. That way it isn't misleading.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1594,12 +1593,10 @@ LaplaceAssembly<dim>::assemble_matrix(
 
               // Laplacian on the velocity terms
```
acdaigneault, line 1594: I know Lucka explain to me why viscosity term was removed, but it would be nice to explain why the units are not consistent now

acdaigneault (approved):
Code looks good! I tried to test it, but I don't know how to visualize the particles in Paraview. I will be at Poly tomorrow if you want to show me! It could be a nice thing to have a section to explain it in the documentation if it's not already there (not for this PR of course)

oguevremont:
> Code looks good! I tried to test it, but I don't know how to visualize the particles in Paraview. I will be at Poly tomorrow if you want to show me! It could be a nice thing to have a section to explain it in the documentation if it's not already there (not for this PR of course)

Thanks! I'll address the comments now. For visualizing the particles (good idea to add a section in the documentation), you can create a Contour in Paraview, select the field "levelset" and choose isosurface at 0 (so the distance to the particle surface=0). I can show you tomorrow if you want.

blaisb (changes requested):

include/core/shape.h
```diff
@@ -603,31 +621,72 @@ class DeathStar : public Shape<dim>
...
   /**
    * @brief Constructs an assembly of shapes into a composite shape
    * @param components The shapes from which this composite sphere will be composed
    */
-  CompositeShape<dim>(std::vector<std::shared_ptr<Shape<dim>>> components)
-    : Shape<dim>(0.,
-                 components[0]->get_position(),
-                 components[0]->get_orientation())
+  CompositeShape<dim>(
+    std::map<unsigned int, std::shared_ptr<Shape<dim>>> components,
```
blaisb, line 645: I don,t think component is a good name here because we use components for vectors components.
How about constituent? item?  part?

blaisb (changes requested):

include/core/shape.h
```diff
@@ -1243,4 +1302,172 @@ class RBFShape : public Shape<dim>
...
+
+template <int dim>
+class CylindricalHelix : public Shape<dim>
+{
+public:
+  /**
+   * @brief Constructs a cylindrical helix by extruding a disk through a helicoidal path
+   * aligned with the z axis
+   * @param radius_helix the helicoidal path radius
+   * @param radius_tube the extruded disk radius
```
blaisb, line 1420: This needs a bit more explanation...

source/core/parameters.cc
blaisb: Agreed, I would just throw an exception here. For example you could have a default filename that is garbage (dumb_dumb_file, whatev) and if the filename is dumb_dumb_file you could throw an exception.

applications_tests/gls_sharp_navier_stokes_3d/cavatappi_composite_test.output
```diff
@@ -0,0 +1,21 @@
...
+   Number of degrees of freedom: 9344
+   Volume of triangulation:      64
+Initial refinement around IB particles - Step : 2 of 2
+   Number of active cells:       5557
+   Number of degrees of freedom: 27688
+   Volume of triangulation:      64
+
+*****************************************************************
+Steady iteration :        1/1
+*****************************************************************
```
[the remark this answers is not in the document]
oguevremont: It's the name of the pasta on which the shape is based on (a cylindrical helix with a hole along the whole length)

include/core/shape.h
oguevremont: I chose component because of the relationship between component&composite, but I agree it could be confusing. How about component_shapes? What you suggested could also work

source/core/parameters.cc
```diff
@@ -2382,13 +2390,183 @@ namespace Parameters
...
+                               unsigned int,
+                               unsigned int>>
+                    operations;
+                  // In the file, we first loop over all component shapes, then
+                  // we loop over operations
+                  std::ifstream myfile(shape_name);
+                  // open the file.
+                  // Read file line by line for section names or arguments
+                  if (myfile.is_open())
+                    {
```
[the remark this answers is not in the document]
oguevremont: We plan to create a static function that initializes any shape directly in shape.h for a future PR, and that will make the design much better. This function would be overloaded and take (str:shape_type,vector<double>) or (str:shape_type,str:filename) for flexibility, to handle Composite, RBF and OpenCascade formats.
Regarding moving the parsing to a function and writing a unit test for it, that could be done now. Should we create a parsing.h file? (or something similar)

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1193,7 +1222,7 @@ GLSSharpNavierStokesSolver<dim>::output_field_hook(DataOut<dim> &data_out)
       all_shapes.push_back(particle.shape);
     }
   std::shared_ptr<Shape<dim>> combined_shapes =
-    std::make_shared<CompositeShape<dim>>(all_shapes);
+    std::make_shared<CompositeShape<dim>>(all_shapes, Point<dim>(), Point<3>());
```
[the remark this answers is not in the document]
oguevremont: I think it could be a useful option, but that it might become heavy for multiple particles. We could set it as a parameter in the future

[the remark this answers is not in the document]
Luckabarbeau: I agree with @oguevremont it could be added as an option in the future same as the cut cell map output I am working on at the moment. Maybe these options could be under the same parameter like: detailed particles output.

tests/core/shape_composites.output
```diff
@@ -0,0 +1,16 @@
...
+DEAL::OK
+DEAL::Testing copy
+DEAL:: Position of original cone[0] = 0.00000
+DEAL:: Position of original cone[1] = 0.00000
+DEAL:: Position of original cone[2] = 0.00000
+DEAL:: Translation of copy cone
+DEAL:: Position of copy cone[0] = 0.200000
+DEAL:: Position of copy cone[1] = 0.00000
+DEAL:: Position of copy cone[2] = 0.300000
+DEAL::OK
```
blaisb, line 16: This is not a cone no? It's a composite shape no?

tests/core/shape_primitives.output
```diff
@@ -8,6 +8,9 @@ DEAL:: Torus , SD = 0.896586
 DEAL:: Cone , SD = 1.27456
 DEAL:: Cut Hollow Sphere , SD = 0.315918
 DEAL:: Death Star , SD = 1.27344
+DEAL:: Cylinder , SD = 0.00000
+DEAL:: Cylindrical Tube , SD = -0.280625
+DEAL:: Cylindrical Helix , SD = 0.202481
 DEAL::OK
```
blaisb, line 14: Nice, but are you sure of these values? it could be nice to eventually have the unit test print out just a simple level set of the geometry. Maybe in a next PR?

tests/core/shape_composites.cc
```diff
@@ -0,0 +1,128 @@
...
+          << std::endl;
+  deallog << " Intersection , SD = " << composite_intersection->value(p)
+          << std::endl;
+  deallog << "OK" << std::endl;
+
+  deallog << "Testing copy" << std::endl;
+  Point<3>                  translation({0.2, 0., 0.3});
+  std::shared_ptr<Shape<3>> copy_composite_union =
+    composite_union->static_copy();
+  copy_composite_union->set_position(translation);
```
blaisb, line 78:
```suggestion
  deallog << "Testing copy and translation" << std::endl;
  Point<3>                  translation({0.2, 0., 0.3});
  std::shared_ptr<Shape<3>> copy_composite_union =
    composite_union->static_copy();
  copy_composite_union->set_position(translation);
```

blaisb (changes requested):
few last comments, minor stuff really. This can be merged tomorrow

tests/core/shape_primitives.output
oguevremont: They make sense given the shapes that are defined but it would be worth it to check all of them (in a next PR)

source/core/parameters.cc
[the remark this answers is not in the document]
oguevremont: I created shape_parsing as suggested. We can move to a better parser soon, maybe even with muparser so it's already supported

blaisb:
@oguevremont ping me when this is ready for final review :)!

source/core/shape_parsing.cc
```diff
@@ -0,0 +1,351 @@
...
+ShapeGenerator::initialize_shape_from_file(const std::string   type,
+                                           const std::string   file_name,
+                                           const Point<dim> &  position,
+                                           const Tensor<1, 3> &orientation)
+{
+  std::shared_ptr<Shape<dim>> shape;
+  std::vector<double>         shape_arguments;
+  if (type == "rbf")
+    {
+      if (file_name == "1") // Default case
```
blaisb, line 145: I would have used another default name than "1", something like "dummy" or similar...

blaisb, line 145:
```suggestion
      if (file_name == "1") // Default case which is used for testing purposes
```
If it's only used for testing, it's not the regular case, but the exception no? Seems a bit off to put it first

blaisb (changes requested):
One or two tiny last changes and this will be ready for a merge

source/core/parameters.cc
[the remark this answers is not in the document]
blaisb: MuParser is there to parse equation. The ideal would be to use a secondary xml or json file or something similar which has a structure to it.

source/core/shape_parsing.cc
oguevremont: I put 1 because it's the same argument as the ones for other shapes, and the default for them is 1. I could add a new parameter that is called _file name_ though

blaisb: No that's ok, but describe this in the comment, otherwise the 1 appears to be fully random.
When something has an explenation, commenting this explanation is the best approach.

blaisb (approved):

blaisb:
Will merge once CI is back to life (out of my control, that's a deal.II docker issue) :).

blaisb:
> > Will merge once CI is back to life (out of my control, that's a deal.II docker issue) :).
> 
> Perfect! Thanks for all the suggestions, it really improved the PR



> > Will merge once CI is back to life (out of my control, that's a deal.II docker issue) :).
> 
> Perfect! Thanks for all the suggestions, it really improved the PR

Good work on the PR yourself. It was a tough cookie to crack. Obviously the Sharp-Edge model can still be improved, but It's going in the right track I believe.

PR #673, opened 2022-12-20
Optimize particle-wall fine search and emplace

blaisb:
# Description of the problem

- Tuples generated when creating the particle_wall_fine search where created then inserted.  Same was being done for the contract info
- Particle wall fine search did not use search radius

# Description of the solution

- Tuples are now emplaced directly in the container for everything that was possible related to particle-wall contact. This actually has a much larger impact on performance than I expect (like... close to 5-6% on some cases)
- I tried using the serach radius for the particle-wall contact, but it turns out this breaks cases when the walls are moving (and this makes sense, static particle + moving wall won't be registered as a contact. Consequently, I don't think this change is worth it for now.

# How Has This Been Tested?

- All tests should pass. If so, everything is good.

# Documentation

- No changes

# Comments

- Can't seem to give up on keeping on optimizing this thing.


include/dem/particle_wall_contact_info_struct.h
```diff
@@ -32,17 +32,59 @@
 using namespace dealii;
 
 template <int dim>
-struct particle_wall_contact_info_struct
+class particle_wall_contact_info
```
acdaigneault, line 35: You forgot to change the name of the .h and .cc :)

include/dem/particle_wall_contact_info_struct.h
```diff
@@ -32,17 +32,59 @@
...
+  particle_wall_contact_info()
+    : normal_vector({0, 0, 0})
+    , point_on_boundary({0, 0, 0})
+    , boundary_id(0)
+    , global_face_id(0)
+    , normal_overlap(0)
+    , normal_relative_velocity(0)
+    , tangential_overlap({0, 0, 0})
+    , tangential_relative_velocity({0, 0, 0})
+  {}
```
acdaigneault, line 87: I would explain why we need those 3 different constructors

source/dem/dem.cc
```diff
@@ -780,9 +780,8 @@ DEMSolver<dim>::solve()
     }
 
   // Find the smallest contact search frequency criterion between (smallest
-  // cell size - largest particle radius) and (security factor * (blab
-  // diamater
-  // - 1) *  largest particle radius). This value is used in
+  // cell size - largest particle radius) and (security factor * (blob
```
acdaigneault, line 783: Blob definitely better than blab :p

source/dem/particle_wall_fine_search.cc
```diff
@@ -55,23 +51,14 @@ ParticleWallFineSearch<dim>::particle_wall_fine_search(
...
-              particle_wall_pairs_in_contact[particle_id].insert(
-                {face_id, contact_info});
+              particle_wall_pairs_in_contact[particle_id].emplace(
+                face_id,
+                particle_wall_contact_info<dim>(
+                  particle,
+                  normal_vector_3d,
+                  point_on_boundary_3d,
+                  std::get<3>(particle_pair_candidate_content),
+                  std::get<4>(particle_pair_candidate_content)));
```
acdaigneault, line 61: Can you add a comment saying what is the variable of the get<3> and get<4>?

source/dem/particle_wall_broad_search.cc
```diff
@@ -49,19 +49,18 @@ ParticleWallBroadSearch<dim>::find_particle_wall_contact_pairs(
...
               particle_wall_contact_candidates[particles_in_cell_iterator
                                                  ->get_id()]
-                .insert({boundary_cells_content.global_face_id, map_content});
+                .emplace(
+                  boundary_cells_content.global_face_id,
+                  std::make_tuple(particles_in_cell_iterator,
+                                  boundary_cells_content.normal_vector,
+                                  boundary_cells_content.point_on_face,
+                                  boundary_cells_content.boundary_id,
+                                  boundary_cells_content.global_face_id));
```
acdaigneault, line 63: Why is the global_face_id stored twice? (inside and outside of the tuple) Also, almost all of the members of the boundary_cells_content are stored (not the cell it) + particles_in_cell_it. Why not store particles_in_cell_it and boundary_cells_content?

source/dem/particle_wall_fine_search.cc
```diff
@@ -156,29 +143,13 @@ ParticleWallFineSearch<dim>::particle_floating_wall_fine_search(
...
-
-
-                  particle_floating_wall_pairs_in_contact[particle_id].insert(
-                    {floating_wall_id, contact_info});
+                  particle_floating_wall_pairs_in_contact[particle_id].emplace(
+                    floating_wall_id,
+                    particle_wall_contact_info<dim>(particle,
+                                                    normal_vector_3d,
+                                                    point_on_floating_wall_3d,
+                                                    100,
```
acdaigneault, line 151: Would keep the comment                   

```
// The boundary ID of floating walls is set to 100, it should
// be modified after adding motion of floating walls
```

acdaigneault (commented):
Here are some comments. There is still some ".insert" kinda related of object that are used here in find_boundary_cells_information.cc, but I know those won't really affect the performance, but to be consistent. Same for particle-line, but could be for another PR :)

include/dem/particle_wall_contact_info_struct.h
blaisb: Good catch, fixed :)

source/dem/particle_wall_broad_search.cc
blaisb: After some investigation, turns out there is absolutely no reason why the global_face_id should be stored in the class. I'll work to remove it everywhere within this PR since it's not much work.
I think a next PR will be to merge the broad and the fine wall search because I am not sure we are not just doing duplicate work for nothing

blaisb: This is now removed :)

source/fem-dem/ib_particles_dem.cc
```diff
@@ -462,7 +462,6 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
                 }
 
               // BB temporary fix for unused variables
```
acdaigneault, line 464: BB?

acdaigneault (approved):

source/fem-dem/ib_particles_dem.cc
blaisb: Leftover comment when I was debugging something

PR #676, opened 2022-12-28
Dem optimize particle-particle forces by templating rolling friction model

blaisb:
# Description of the problem

- Rolling friction was still passed as a parameter and the model was checked using an if condition for all partice ij pair

# Description of the solution

- The type of rolling friction model has been templated. The if is now a if constexpr so this prevents billion of ifs checks

# How Has This Been Tested?

- All tests pass

# Comments

- I tried a ton of optimizations and I have not found anything else where I could increase the speed of the simulations. I think we are reaching the limit...

### DIFF
--- a/include/dem/particle_particle_contact_force.h
+++ b/include/dem/particle_particle_contact_force.h
@@ -122,9 +122,16 @@ public:
 };
 
 /**
- * Class that carry out the calculation of
+ * @brief Class that carries out the calculation of
  * particle-particle contact force including non-linear and linear contact
- * models
+ * models. Instead of using a inheritance hiearchy to distinguish between
+ * the contact model, the class is templated with the type of force model
+ * and rolling friction model. Consequently, the code for each
+ * combination of force model is generated at compile time.
+ *
+ * @tparam dim The dimension of the problem
+ * @tparam force_model The particle-particle contact force model
+ * @tparam rolling_friction_model The rolling resistance model used
  */
 template <
   int                                                       dim,
@@ -352,7 +359,7 @@ protected:
    * @param particle_one_force Force acting on particle one
    */
   inline void
-  apply_force_and_torque_on_ghost_particles(
+  apply_force_and_torque_on_single_local_particle(
     const Tensor<1, 3> &normal_force,
     const Tensor<1, 3> &tangential_force,
     const Tensor<1, 3> &particle_one_tangential_torque,
--- a/include/dem/particle_particle_contact_force.h
+++ b/include/dem/particle_particle_contact_force.h
@@ -358,11 +358,8 @@ protected:
     Tensor<1, 3> &      particle_one_torque,
     Tensor<1, 3> &      particle_one_force)
   {
-    // Calculation of total force
-    Tensor<1, 3> total_force = normal_force + tangential_force;
-
     // Updating the force and torque acting on particles in the particle handler
-    particle_one_force -= total_force;
+    particle_one_force -= normal_force + tangential_force;
     particle_one_torque +=
       -particle_one_tangential_torque + rolling_resistance_torque;
   }
--- a/include/dem/particle_particle_contact_force.h
+++ b/include/dem/particle_particle_contact_force.h
@@ -126,13 +126,15 @@ public:
  * particle-particle contact force including non-linear and linear contact
  * models
  */
-template <int                                                       dim,
-          Parameters::Lagrangian::ParticleParticleContactForceModel force_model>
+template <
+  int                                                       dim,
+  Parameters::Lagrangian::ParticleParticleContactForceModel force_model,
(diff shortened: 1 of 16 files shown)


source/dem/particle_wall_nonlinear_force.cc
```diff
@@ -86,22 +86,19 @@ ParticleWallNonLinearForce<dim>::ParticleWallNonLinearForce(
     }
 
   if (dem_parameters.model_parameters.rolling_resistance_method ==
-      Parameters::Lagrangian::ModelParameters::RollingResistanceMethod::
-        no_resistance)
+      Parameters::Lagrangian::RollingResistanceMethod::no_resistance)
```
acdaigneault, line 89: Why not a template for particle-wall forces too? Not worth it since there not enough contacts?

source/dem/set_particle_particle_contact_force_model.cc
```diff
@@ -13,43 +13,183 @@ set_particle_particle_contact_force_model(
...
+                std::make_shared<ParticleParticleContactForce<
+                  dim,
+                  Parameters::Lagrangian::ParticleParticleContactForceModel::
+                    linear,
+                  Parameters::Lagrangian::RollingResistanceMethod::
+                    viscous_resistance>>(dem_parameters);
+              break;
+            }
+          default:
+            throw std::runtime_error("Invalid contact force model combination");
```
acdaigneault, line 54: Would add "contact force model **and rolling resistance method** combination"

source/dem/set_particle_particle_contact_force_model.cc
```diff
@@ -13,43 +13,183 @@ set_particle_particle_contact_force_model(
...
+                std::make_shared<ParticleParticleContactForce<
+                  dim,
+                  Parameters::Lagrangian::ParticleParticleContactForceModel::
+                    hertz_mindlin_limit_overlap,
+                  Parameters::Lagrangian::RollingResistanceMethod::
+                    viscous_resistance>>(dem_parameters);
+              break;
+            }
+          default:
+            throw std::runtime_error("Invalid contact force model combination");
```
acdaigneault, line 100: Same here

source/dem/set_particle_particle_contact_force_model.cc
```diff
@@ -13,43 +13,183 @@ set_particle_particle_contact_force_model(
...
+                std::make_shared<ParticleParticleContactForce<
+                  dim,
+                  Parameters::Lagrangian::ParticleParticleContactForceModel::
+                    hertz_mindlin_limit_force,
+                  Parameters::Lagrangian::RollingResistanceMethod::
+                    viscous_resistance>>(dem_parameters);
+              break;
+            }
+          default:
+            throw std::runtime_error("Invalid contact force model combination");
```
acdaigneault, line 146: Same here

source/dem/set_particle_particle_contact_force_model.cc
```diff
@@ -13,43 +13,183 @@ set_particle_particle_contact_force_model(
...
+                std::make_shared<ParticleParticleContactForce<
+                  dim,
+                  Parameters::Lagrangian::ParticleParticleContactForceModel::
+                    hertz,
+                  Parameters::Lagrangian::RollingResistanceMethod::
+                    viscous_resistance>>(dem_parameters);
+              break;
+            }
+          default:
+            throw std::runtime_error("Invalid contact force model combination");
```
acdaigneault, line 191: Same here :p

acdaigneault (commented):
Nice to still see some improvement on the DEM performance hehe. Nice job. Here is some comments, I will test it next monday or tuesday.

source/dem/particle_wall_nonlinear_force.cc
blaisb: It will be worth it eventually, just in a next PR :).

include/dem/particle_particle_contact_force.h
```diff
@@ -358,11 +360,8 @@ class ParticleParticleContactForce
     Tensor<1, 3> &      particle_one_torque,
     Tensor<1, 3> &      particle_one_force)
   {
-    // Calculation of total force
-    Tensor<1, 3> total_force = normal_force + tangential_force;
-
     // Updating the force and torque acting on particles in the particle handler
-    particle_one_force -= total_force;
+    particle_one_force -= normal_force + tangential_force;
```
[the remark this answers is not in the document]
blaisb: Very good point. I have renamed it to apply_force_and_torque_on_single_local_particle

blaisb:
@acdaigneault  All your wise comments have been applied. I improved the comments a bit also.
The Particle-wall templating will wait in another PR. There are other optimizations I have in mind before that (so many optimizations are possible).

acdaigneault (approved):
Tested, all looks good!

PR #687, opened 2023-01-28
Add RBF sphere MMS test

oguevremont:
# Description of the problem

- There was no test for Sharp IB with RBF that had an analytical solution

# Description of the solution

- A MMS verification case set around a sphere defined as an RBF shape is added

# How Has This Been Tested?

- The test passes and convergence reaches expected order, given enough refinement

# Documentation

- Not applicable

# Future changes

- An additionnal RBF MMS test for superquadrics may be added


applications_tests/gls_sharp_navier_stokes_3d/rbf_sphere_mms.output
```diff
@@ -0,0 +1,67 @@
...
+  -Tolerance of iterative solver is : 1.309e-05
+  -Iterative solver took : 5 steps
+		alpha =      1 res = 4.913e-06
+Newton iteration: 2  - Residual:  4.913e-06
+  -Tolerance of iterative solver is : 4.913e-09
+  -Iterative solver took : 5 steps
+		alpha =      1 res = 3.225e-09
+div u : 1.241e-08
+error at the weak BC : 0.3751
+L2 error velocity : 0.3046 L2 error pressure: 1.342
```
blaisb, line 23: We do not add verbosity in the output of test because they are generally not portable. Please remove this.

blaisb (changes requested):
Some changes to be made to the output so that it is portable. Otherwise, good to go.
How long is the runtime of the test?

applications_tests/gls_sharp_navier_stokes_3d/rbf_sphere_mms.output
oguevremont: That's right, I didn't notice. Initially I removed the linear and non-linear solver sections to keep the .prm light, but it seems that verbose is the default option.

oguevremont: It takes about 94 seconds in Debug with the CI

blaisb:
Will merge after CI

PR #689, opened 2023-02-02
Implement VOF phase fraction filter

AmishgaAlphonius:
# Description of the problem

- Sharpening tends to pull the fluid together (similar to the effect of artificial surface tension forces)

# Description of the solution

- Implement a phase fraction filter. At the moment, there is 2 filter options ("none" and "tanh").
- Both filtered and unfiltered solutions are outputted.

# How Has This Been Tested?

- Visual verification with the Rayleigh-Taylor example
- Application test (applications_tests/gls_navier_stokes_2d/gls_vof_tanh_filter)

# Documentation

- [x] doc/source/parameters/cfd/volume_of_fluid.rst

# Future changes

- Use the filtered phase fraction when calculating the physical properties.

PR #690, opened 2023-02-03
Vof improve surface tension

hepap:
# Description of the problem

- Previously, the input parameters of the filters used to compute the filtered phase fraction gradient and the curvature in the surface tension model (VOF) were dependent of the cell size, which was inconvenient and most likely not suitable for adaptive mesh refinement.
- There was also a small mistake in the surface tension force computation: the filtered phase fraction gradient was used instead of the phase fraction gradient (unfiltered).
- The element order of the VOF solver was hard coded to 1 instead of the input parameter.

# Description of the solution

- Now, the filter value n_filter is computed following :  n_filter = beta*h^2 where beta is the `phase fraction gradient filter factor` or `curvature filter factor` given in the parameter file and h is the cell size, computed in Lethe when the filtered phase fraction gradient and the curvature are resolved. The input parameters previously called `phase fraction gradient filter` or `curvature filter` have been replaced by the `phase fraction gradient filter factor` or `curvature filter factor` parameters.
- The mistake in the surface tension force was corrected.
- The input parameter for the element order of the VOF solver is now taken into account.

# How Has This Been Tested?

- The tests for the surface tension model already existed but they were updated (the prm files were also re-indented):

- [gls_static_droplet_surface_tension_force.prm and .output]
- [gls_droplet_marangoni_effect.prm and .output]

# Documentation

- The documentation for the VOF solver (`subsection surface tension force`) was updated and the rising bubble example as well.

- [volume_of_fluid.rst]
- [rising-bubble.rst] 

# Future changes

- A smooth initial condition for the VOF model will be added to further decrease the spurious currents in the velocity field. 

# Comments

- I will rebase when PR #689 is merged :)


doc/source/parameters/cfd/volume_of_fluid.rst
```diff
@@ -53,10 +53,10 @@ The default values of the VOF parameters are given in the text box below.
...
 			set enable 	= false
 			set verbosity 	= quiet
-			set output auxiliary fields 	   = false
-			set surface tension coefficient    = 0.0
-			set phase fraction gradient filter = 0.5
-			set curvature filter 		   = 0.5	
+			set output auxiliary fields 	          = false
+			set surface tension coefficient           = 0.0
+			set phase fraction gradient filter factor = 4
+			set curvature filter factor 		  = 1
```
oguevremont, line 59: Out of curiosity, are the new default values for the parameters equivalent to the old ones? Or no, because the point was to use parameters that do not depend on mesh size?

applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
```diff
@@ -8,31 +8,31 @@ Enstrophy  : 0
...
+Enstrophy  : 24.4541
+Kinetic energy : 0.000528142
 
 *****************************************************************
-Transient iteration : 3        Time : 0.003    Time step : 0.001    CFL : 0.123131
+Transient iteration : 3        Time : 0.003    Time step : 0.001    CFL : 0.0810037
 *****************************************************************
-Enstrophy  : 195.707
-Kinetic energy : 0.00347251
+Enstrophy  : 98.8678
```
oguevremont, line 25: Is such a difference to be expected?

doc/source/examples/multiphysics/rising-bubble/rising-bubble.rst
```diff
@@ -103,53 +103,69 @@ In Lethe, the surface tension force (:math:`{\bf{F_{\sigma}}}`) is calculated us
...
-        set output auxiliary fields        = true
-      end
+  #---------------------------------------------------
+  # VOF
+  #---------------------------------------------------
+  subsection VOF
+    subsection interface sharpening
+      set enable              = true
+      set threshold           = 0.5
+      set interface sharpness	= 1.4
```
oguevremont, line 158: To align if possible

oguevremont (commented):
Good work!

doc/source/parameters/cfd/volume_of_fluid.rst
hepap: There are not equivalent. We used to give the filter value n_filter and it was recommended to select it base on the mesh size, e.g.  n_filter=h/10. Now, n_filter is defined as n_filter=beta*h^2, and only beta is given as the input.

applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
hepap: Good question! I think yes for two reasons:

1. For this case, I lowered the filter factor values compared to the previous version. I think this way, there is less dissipation in the fluid, so less enstrophy.
2. I changed the computation of the surface tension force F_s = sigma*k*grad_alpha. The filtered value of grad_alpha was previously used, and now it's the unfiltered value. So I think this way (again), there is less dissipation in the fluid, so less enstrophy.

doc/source/examples/multiphysics/rising-bubble/rising-bubble.rst
```diff
@@ -103,53 +103,69 @@ In Lethe, the surface tension force (:math:`{\bf{F_{\sigma}}}`) is calculated us
...
 
 .. tip::
 
-  Phase fraction gradient filter value (:math:`\eta_n`) and curvature filter value (:math:`\eta_k`) must be small values larger than 0. We recommend the following procedure to choose a proper value for these parameters: 
+  The phase fraction gradient filter value (:math:`\eta_n = \alpha h^2`) and curvature filter value (:math:`\eta_k = \beta h^2`) must be small values larger than 0. The values of :math:`\alpha` and :math:`\beta` are controlled respectively by the parameters ``phase fraction gradient filter factor`` and ``curvature filter factor``  in the parameter file.
+  We recommend the following procedure to choose a proper value for these parameters:
 
   1. Enable ``output auxiliary fields`` to write filtered phase fraction gradient and filtered curvature fields.
-  2. Choose a small value for :math:`\eta = h/10`, where :math:`h` is the smallest mesh size. 
+  2. Choose a small value, for example, :math:`\alpha = 4` and :math:`\beta = 1`.
```
blaisb, line 143:
```suggestion
  2. Choose a value close to 1, for example, the default values  :math:`\alpha = 4` and :math:`\beta = 1`.
```

blaisb (approved):
Tiny changes

PR #694, opened 2023-02-09
 Implement VOF phase fraction filter - physical properties calculation

AmishgaAlphonius:
# Description of the problem

Sharpening tends to pull the fluid together (similar to the effect of artificial surface tension forces).

# Description of the solution

- Modifications were made to both the NavierStokesScratchData class and HeatTransferScratchData class so that filtered phase fractions are used when calculating physical properties.
- The "none" default _type_ of  filter allows to get the same results as before implementing the filter. 
- Both filtered and unfiltered solutions are still outputted.

# How Has This Been Tested?

- Visual verification with 2D Dam Break and Rayleigh-Taylor examples
- Application test output was updated (applications_tests/gls_navier_stokes_2d/gls_vof_tanh_filter.output)


blaisb (approved):
Good work. Nothing to say, this is very nice!

hepap (approved):
Nice PR! Nothing to say except that an example would be nice in a future PR :)

PR #695, opened 2023-02-10
Use Ankerl::dense_unordered::map instead of std::unordered_map for DEM particle-particle container

blaisb:
# Description of the problem

- The particle data structure in the DEM was an unordered_map of unordered_map. The underlying storage structure is a double linked-list. When the data structure was being rewritten significantly (after load balance or transfer accross cores) it became very inefficient.

# Description of the solution

- There exist an alternative to unordered map and these are call dense_unordered map structures. They use a vector as an underlying memory storage. This PR includes one of these and replaces the first unordered map (the one that is changing all the time and significantly) with an ankerl::unordered_dense::map. This makes the code significantly more efficient (see benchmark here)
-
Figure_1

# How Has This Been Tested?

- The results of one single unit test have been slightly changed. I understand why, and it's not a significant change.
- The rest passes
- I reran all the examples. The performance increase is sometime tremendous, sometimes marginal. The rule of thumb is that the bigger the case and the more core you are using and the more load balancing you are doing, the better the results are now compared to the past.

# Future changes

- I will try to identify if some other structures in the DEM could benefit from such map. I am thinking about particle-wall and particle-floating walls contacts.

# Comments

- This is a small change, but also major because it includes a new library as include-only. I tried to be as careful as possible when testing it. Hence why it took me so long to make that PR.
-


acdaigneault (approved):
Sooo nice. I understand why some results of some tests are different as discussed. Would you like me to check what/if other umap containers could be good candidate too?

blaisb:
> Sooo nice. I understand why some results of some tests are different as discussed. Would you like me to check what/if other umap containers could be good candidate too?

In another branch maybe :). It can wait, whenever you see fit we can look at this together.

PR #700, opened 2023-02-16
Vof improve initial condition

hepap:
# Description of the problem

- When initializing the VOF phase fraction field, especially  with curved geometries, the free surface was in staircase, which was hindering the reach of the 2nd order of accuracy.
- There was an error in the order of the assemblers of in gls navier stokes when VOF and surface tension are enable.

# Description of the solution

- Now, the user can use a projection step (with a diffusion term)  to smooth the VOF initial condition.
- The order of the assemblers has been corrected

# How Has This Been Tested?

- [x]  Since the assemblers order has changed, the some VOF application tests were updated.
- For the initial condition, the static bubble test case was used to verify if the order of convergence in space in 2 (and yes, it is now).

- [x] gls_droplet_marangoni_effect .output
- [x] gls_static_droplet_surface_tension_force.output
- [x] gls_vof_heat-transfer_hydrostat.output
- [x] gls_vof_ht_heated-walls.output
- [x] gls_vof_hydrostat.output
- [x] gls_vof_tanh_filter.output

# Documentation

- The parametes for the VOF initial condition were added in the documentation

- initial_conditions.rst

# Future changes

- Other changes in the VOF solver will be added in a PR to improve the rising droplet case.
- An example and an application test will be added for the smooth initial condition.


applications_tests/gls_navier_stokes_2d/gls_vof_tanh_filter.output
```diff
@@ -72,15 +72,15 @@ Filtered phase values:
...
-  -Tolerance of iterative solver is : 0.00402676
-  -Iterative solver took : 4 steps
+  -Iterative solver took : 0 steps 
+  -Tolerance of iterative solver is : 0.00384557
+  -Iterative solver took : 4 steps 
   -Tolerance of iterative solver is : 1e-05
-  -Iterative solver took : 4 steps
-Filtered phase values:
+  -Iterative solver took : 4 steps 
+Filtered phase values:
```
blaisb, line 83: Remove tolerance of linear solver.

blaisb (approved):
Just fix the test. Once CI passes i'll merge. Good job :)!

applications_tests/gls_navier_stokes_2d/gls_vof_tanh_filter.output
hepap: Done :)

PR #702, opened 2023-02-22
Add option to remove manifold after initialization

oguevremont:
# Description of the problem

- In some cases where the grid has a manifold, it would be better to remove that manifold before adaptive refinement is done. The flow through a cylindrical channel is such a case, because with adaptive refinement (before this PR) the section is not constant along the length.

# Description of the solution

- An option to remove the manifold was added in the mesh section.
- When the option is activated, a temporary mesh is generated, then flattened (which removes hierarchy and manifolds) into the regular triangulation variable.

# How Has This Been Tested?

- [gls_navier_stokes_3d/poiseuille_adaptive_without_manifold ] The option is tested in a Poiseuille flow, and we can see that the volume does not evolve between refinements

# Documentation

- [parameters/cfd/mesh.rst] The option, its description and limitations were added to the documentation

# Future changes

- If necessary, a similar option could be added for grids that are not dealii quad/hex.


peterrum:
> The flow through a cylindrical channel is such a case, because with adaptive refinement (before this PR) the section is not constant along the length.

Is this a huge problem? Wouldn't it make sense to use higher order mapping? In this case, the difference of the sections should be smaller, shouldn't it?

oguevremont:
> > The flow through a cylindrical channel is such a case, because with adaptive refinement (before this PR) the section is not constant along the length.
> 
> Is this a huge problem? Wouldn't it make sense to use higher order mapping? In this case, the difference of the sections should be smaller, shouldn't it?

That's a good suggestion, I had not thought about this. Let me try it. Bruno and your ideas are surely cleaner than removing the manifold or flattening the triangulation.

PR #705, opened 2023-02-27
Dem wall container map

blaisb:
# Description of the problem

- The Ankerl map was only being used for the particle-particle data structure. I was wondering if it would be promising for the particle-wall data structure

# Description of the solution

- Switching to the new map for the particle-wall interaction has a good impact. Some cases are 2-3% faster, while some ass significantly more (the square hopper, for example, goes from 86 minutes to 74 minutes on my desktop at home)

# How Has This Been Tested?

- All previous DEM tests pass without any changes

# Comments

- It will be worthwhile to try with some of the other containers if there are not more possible gains to be obtained.
-

### DIFF
--- a/include/dem/data_containers.h
+++ b/include/dem/data_containers.h
@@ -157,7 +157,7 @@ namespace DEM
       particle_wall_candidates;
 
     // <particle id, <global_face_id, particle-wall info>>
-    typedef std::unordered_map<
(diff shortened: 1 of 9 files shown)


acdaigneault (approved):
Doesn't crash with fem-dem, it seems good!

PR #708, opened 2023-02-27
Fix strong form of the residual for stabilization

blaisb:
# Description of the problem

- There was a missing term in the VOF strong residual

# Description of the solution

- add missing term

# Comments

- I am currently satisfied with the VOF formulation. Maybe the DCDD stabilization is not required anymore, to be investigated.


hepap (approved):
Only one question : Do the application tests changed?

blaisb:
@hepap  Nope nothing changed

PR #711, opened 2023-03-01
Add phase filter to the 2d dam-break VOF example

acdaigneault:
# Description of the problem

- 2d VOF dam break example was not updated since a while
- Some new features are now available

# Description of the solution

- Add the VOF phase filter to the example
- Update the documentation
- Regenerate new post-processing data

# How Has This Been Tested?

- Simulations with and without the phase filter were compared. Post-processed results without the filter are not the same than the previous plot shown in the documentation, this is probably because VOF solver was been improved since the documentation of the example was done. The results with and without the filter has been compared and they are very similar.

# Documentation

- Documentation is updated

# Comments

- The feature is not really explain is the example documentation here since the results are not quite better or significant.


blaisb (approved):
Nice work!

PierreLaurentinCS (approved):
Nothing to report, except for the typo in the first file

acdaigneault:
@blaisb Ready to merge :)

PR #717, opened 2023-03-03
Fixed tests precision

oguevremont:
# Description of the problem

- Some Sharp IB tests had a log precision that was too high, which made them less portable

# Description of the solution

- Log precision was reduced
- The superfluous parameters were removed

# How Has This Been Tested?

- They were run and the output files were updated


blaisb (approved):

blaisb:
Will merge after CI. Thanks :)!

PR #719, opened 2023-03-06
Add check to ensure there is no Nan gradient for spheres

oguevremont:
# Description of the problem

- A bug in Sharp happened when a sphere center was located at an evaluation point, which made the gradient function return a NaN.

# Description of the solution

- A check was added in the function to ensure that no NaN is output, and numerical gradient is used as an alternative.

# How Has This Been Tested?

- All tests were run.
- Ouput results should not change.

### DIFF
--- a/source/core/shape.cc
+++ b/source/core/shape.cc
@@ -311,6 +311,12 @@ Tensor<1, dim>
 Sphere<dim>::gradient(const Point<dim> &evaluation_point,
                       const unsigned int /*component*/) const
 {
+  // We make sure that the evaluation point and the sphere center are different,
+  // because if they are the same the analytical gradient is not defined: the
+  // function returns a NaN. We use the numerical gradient if the points are the
+  // same.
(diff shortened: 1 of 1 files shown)


source/core/shape.cc
```diff
@@ -311,6 +311,12 @@ Tensor<1, dim>
 Sphere<dim>::gradient(const Point<dim> &evaluation_point,
                       const unsigned int /*component*/) const
 {
+  // We make sure that the evaluation point and the sphere center are different,
+  // because if they are the same the analytical gradient is not defined: the
+  // function returns a NaN. We use the numerical gradient if the points are the
+  // same.
+  if ((evaluation_point - this->position).norm() < 1e-12)
```
blaisb, line 318: You might want to make the scaling of this < operator depend on the radius of the sphere no? To be coherent with dimensions.
Otherwise all good.

blaisb (approved):

PR #720, opened 2023-03-08
Add disabling contacts of particles with negligible motion

acdaigneault:
# Description of the problem

- Some DEM/CFD-DEM type of simulation has areas in their domain where particles are almost motionless, this results of irrelevant computation of contact list, contact forces and velocity integration where all particles are constantly vibrating in place
- The potential of the calculation of the granular temperature was not fully exploited

# Description of the solution

- A feature that can associate status to cells regards the mobility of their particles has been implemented. When disabling particles' mobility is enabled, this feature is called after the "sort_particles_into_subdomains_and_cells"
- The general idea behind the algorithm : 
    - It uses the granular temperature to determinate if particles in a cell are mobile enough that the contact forces are worth computation. A cell having a granular temperature under a value (default is 1e-4) will have an inactive status which makes them rejected in the broad search step. This results in no contact forces or velocity integration in future steps for those particules, which make less computation cost for the simulation.
    -  Cells may be flagged as that called mobility status : mobile (everything is calculated like if feature is not enabled), active (motionless particles next to mobile particles, see above) or inactive (particles are not in a neighbor’s candidate list of cells around)
    -  There are some edge cases that need some attention. In these cases, particles in cells can't be deactivated : 
        - The solid fraction of the cell is under a value (default = 40%) : particles in the cell may have other forces that can move them and it is supposed that with this fraction, there is not enough particles around to hold them in their position.
        - Cells having empty cell neighbors : it means that particles are next to a floating wall/mesh and they may have a change of contact forces from it if the wall disappears or is moving.
        - The cell has cell neighbors which is flagged as mobile from the 3 criteria mentioned above (granular temperature, solid fractions or floating walls) : this is because motion is badly propagated to particles to the cell around without this additional "layer" of mobile cells.
        - The cell has cell neighbors which is flagged as mobile from the previous criterion (additional layer of mobile cells) : this is again to allow the propagation of motion, but only the particles in contact with particles from the mobile cells are considered for the contact force calculation but their position is not computed at the integration step. Those cells are flagged as active cells.
      - It works with the assignment and verification of the mobility status at nodes to check the status of the neighboring cells. (See in the comments section.)

# How Has This Been Tested?

- Computational gain is significant for the hopper case (2x faster) where some of the particles are motionless during loading and when there's an accumulation of particles at the bottom.
- Mass flow rate results for the hopper have no significant difference (<1%)
- An application test was developed to evaluate each case of status for dem_3d (mobility_status.output)
- An application test was developed for the load balancing with weight according to the mobility status (load_balancing_mobility_status.output)

# Documentation

- No full documentation for now : this feature will be improved in the future and tested more rigorously (especially for CFD-DEM) before.

# Future changes

- Modification regards testing results interpretation in CFD-DEM
- Add a full example (maybe pneumatic conveying with CFD-DEM)

# Comments

-  A trivial version with checks to the neighbor cells state (and not the node status) was first made, but this results in issues using multiprocessing. Since there are at least 3 stages of information of cells (e.g., 1st cell from granular temperature value, 2nd cell from the additional layer of mobile cells and the 3rd from the active cells layer), having the data from the locally owned and ghost cells is not enough (only 2 stages of information is available in some case for a processor), we need to exchange information. To exchange this information, having information at nodes allows using the LinearAlgebra::distributed::Vector<int> where data is exchanged with the easy call of update_ghost_values(). But this is harder to understand.
- Functions in broad search classes (p-p, p-w, ...) are copy pasted and slightly modify to evaluate the cell mobility status. This doesn't look good, in my opinion, but this allows preventing lots of "if calls" for no reason when the feature is not enabled.


include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+   * with cell id order
+   * status_to_cell is a vector of 3 sets of cell iterator (inactive, active,
+   * mobile), it can't be used as is in the pvd post-processing or any data out,
+   * it needs to be converted to a vector of mobility status by active cell
+   * index
+   *
+   * @param status The initiated vector for the conversion
+   */
+  void
+  get_mobility_status_vector(std::vector<unsigned int> &status)
```
voferreira, line 160: If status_to_cell was a map it would be easier to make this vector, as well.

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+private:
+  void
+  check_granular_temperature(
+    const typename DEM::dem_data_structures<dim>::cells_neighbor_list
+      &                                    cells_neighbor_list,
+    const Particles::ParticleHandler<dim> &particle_handler);
+
+  std::vector<typename DEM::dem_data_structures<dim>::cell_set> status_to_cell;
```
voferreira, line 226: I would use cell_mobility_status_map if you decide to use a map here. Otherwise, cell_mobility_status.

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  unsigned int
+  get_n_mobility_status()
+  {
+    return mobility_status::n_mobility_status;
+  }
+
+  void
+  set_limit_value(const double granular_temperature,
```
voferreira, line 212: I would use threshold instead of limit. But no biggie

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,334 @@
...
+                    (2.0 * dim);
+                }
+
+              // Calculate average velocity in the cell
+              for (int d = 0; d < dim; ++d)
+                velocity_cell_average[d] =
+                  velocity_cell_sum[d] / n_particles_in_cell;
+
+              // Calculate void fraction of cell
+              solid_fraction = solid_volume / cell_volume;
```
voferreira, line 101: This means you are using PCM for the solid fraction. Will it affect your results?

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,334 @@
...
+
+              // Second loop over particle to calculate the average granular
+              // temperature
+              for (typename Particles::ParticleHandler<
```
voferreira, line 105: Can't you do this in a single loop?

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,334 @@
...
+                           locally_relevant_dofs,
+                           mpi_communicator);
+  mobility_at_nodes = 0;
+
+  // Empty status (3) to nodes when no particle in cell
+  for (auto cell = active_ghost_cells.begin();
+       cell != active_ghost_cells.end();)
+    {
+      // Check if the cell has any particles
+      if (particle_handler.n_particles_in_cell(*cell) == 0)
```
voferreira, line 176: Could you pass by reference instead of using a pointer here to follow the pattern?

voferreira (commented):
Round 1. I will continue later.

include/core/parameters_lagrangian.h
```diff
@@ -241,6 +241,13 @@ namespace Parameters
         gear3
       } integration_method;
 
+      // Disabling particle contacts to optimize performance
+      bool   disabling_particle_contacts;
```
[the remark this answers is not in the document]
voferreira: Minor suggestion...

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Carries out the calculation of the granular temperature and solid
+   * fraction approximation in each local cell. Those values are a criteria
+   * for cell mobility
+   *
+   * @param triangulation The triangulation
+   * @param particle_handler The particle handler that contains all the particles
+   */
+  void
+  calculate_average_granular_temperature(
```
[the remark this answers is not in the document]
blaisb: Agreed

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+   */
+  inline unsigned int
+  check_cell_mobility(
+    const typename Triangulation<dim>::active_cell_iterator &cell) const
+  {
+    unsigned int status;
+    // Check if the cell is in the mobile status set
+    status = mobility_status::mobile;
+    auto cell_iterator_from_mobile_container =
+      status_to_cell[status].find(cell);
```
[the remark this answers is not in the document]
blaisb: Agreed. a map of set here would be a good container. Just be careful that your enum has not > comparison, so you might need an unordered map

blaisb (changes requested):
First batch of comment, will continue later this week.

source/dem/disable_particle_contact.cc
acdaigneault: Yes it's equivalent of PCM, but I don't really need precision in my case

source/dem/disable_particle_contact.cc
acdaigneault: No since I need the velocity cell average before

source/dem/disable_particle_contact.cc
acdaigneault: I think it didn't work and that's why I used a pointer. Maybe I should do something at line 47 So iteration works.

include/core/parameters.h
```diff
@@ -996,6 +996,13 @@ namespace Parameters
   struct Testing
   {
     bool enabled;
+
+    enum class TestType
+    {
+      none,
+      mobility_status,
+    } test_type;
```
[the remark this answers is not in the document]
acdaigneault: I will change to "default", what do you think?

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Carries out the identification of the mobility status of each cell thought
+   * processing at nodes.
+   *
+   * 4 checks (search loops) are done is order:
+   * 1. Check if the cell is empty (n_particle = 0), if so, nodes and cells are
+   * flagged as empty mobility status
+   * 2. Check if the cell is mobile by criterion (n_particle > 0, average
+   * granular temperature > threshold, solid fraction < limit, as one empty node
+   * from previous check), if so, nodes and cells are flagged as mobile mobility
```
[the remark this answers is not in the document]
acdaigneault:
```suggestion
   * granular temperature > threshold, solid fraction < limit, has at least one empty node
   * from previous check), if so, nodes and cells are flagged as mobile mobility
```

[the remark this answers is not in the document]
acdaigneault: Oops

source/dem/disable_particle_contact.cc
voferreira: I think you just need to use reference in line 172 instead

include/dem/disable_particle_contact.h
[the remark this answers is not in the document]
acdaigneault: I think Victor is suggesting a map to get the status from the cell id instead of searching in a set. 
key = cell id and value = status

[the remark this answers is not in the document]
voferreira: Why not use an unordered map being cell the key and status the value? You would skip the checking andreturn status being status = status_to_cell[cell]

include/dem/disable_particle_contact.h
acdaigneault: I perfer threshold too, but I don't know, it sounds more like a +/- range value, no?

include/dem/disable_particle_contact.h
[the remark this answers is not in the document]
acdaigneault: Ooh, you're right

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,239 @@
...
+   * need to be in contact candidates lists empty:    cell is empty, only useful
+   * for mobility at nodes
+   */
+  enum mobility_status
+  {
+    inactive,
+    active,
+    mobile,
+    empty,
+    n_mobility_status
```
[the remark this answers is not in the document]
acdaigneault: I use mobility_status straight as an unsigned int everywhere, I don't know why should I use enum class, can you explain why it would be better Bruno?

source/dem/disable_particle_contact.cc
acdaigneault: I would look into that, but for another PR

source/dem/disable_particle_contact.cc
acdaigneault: Yeah, it doesn't work, I think I tried to do that.
Actually, it could work with `(auto &cell : active_ghost_cells)`, but this iterator doesn't work with .erase().
I will keep this comment open, so I could look at that later

source/dem/dem.cc
```diff
@@ -272,24 +284,112 @@ DEMSolver<dim>::cell_weight(
...
+
+  switch (status)
+    {
+      case parallel::distributed::Triangulation<dim>::CELL_PERSIST:
+      case parallel::distributed::Triangulation<dim>::CELL_REFINE:
+        {
+          const unsigned int n_particles_in_cell =
+            particle_handler.n_particles_in_cell(cell);
+          return alpha * n_particles_in_cell * particle_weight;
+          break;
```
voferreira, line 321: Why break here is inside case and outside for parallel::distributed::Triangulation<dim>::CELL_COARSEN?

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,246 @@
...
+            if (cell->is_locally_owned())
+              {
+                status[cell->active_cell_index()] = i_set;
+              }
+          }
+      }
+  };
+
+  std::vector<typename DEM::dem_data_structures<dim>::cell_set>
+  get_mobility_status()
```
voferreira, line 202: if you have two containers for this information, I would use get_mobility_status_map()

source/dem/dem.cc
```diff
@@ -272,24 +284,112 @@ DEMSolver<dim>::cell_weight(
...
   background_dh.distribute_dofs(background_fe);
 }
 
 template <int dim>
 void
 DEMSolver<dim>::load_balance()
 {
+  if (has_disabled_contacts)
+    {
+      std::vector<unsigned int> mobility_status(triangulation.n_active_cells());
```
voferreira, line 364: mobility status can be either a map or a vector depending on the section of code you use it. I would name mobility_status_vector and mobility_status_map for the two data types.

source/dem/disable_particle_contact.cc
voferreira: Ok! Did you try dereferencing .begin and .end?

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+      // - granular temperature > limit or
+      // - solid fraction of cell < limit or
+      // - is next to an empty cell
+      const unsigned int cell_id = (*cell)->active_cell_index();
+
+      std::vector<types::global_dof_index> local_dofs_indices(dofs_per_cell);
+      (*cell)->get_dof_indices(local_dofs_indices);
+
+      // Check if cell has an empty status node
+      bool has_empty_neighbor = false;
```
voferreira, line 215: has_empty_status_neighbor I guess? "Empty neighbor" does not feel like mobility status.

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+              break;
+            }
+        }
+
+      if (granular_temperature_average[cell_id] >=
+            granular_temperature_threshold ||
+          solid_fractions[cell_id] <= solid_fraction_threshold ||
+          has_empty_neighbor)
+        {
+          // Insert cell in mobile status set
```
voferreira, line 230: This part may be simplified when you switch to a map container.

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+                std::max((int)mobility_status::mobile,
+                         mobility_at_nodes[dof_index]);
+            }
+        }
+      else
+        {
+          ++cell;
+        }
+    }
+
```
voferreira, line 250: I feel like this member function is a bit too long. What would you think of breaking it down into 4 or even 5 separate members? It would be easier to maintain I think.

source/dem/explicit_euler_integrator.cc
```diff
@@ -87,5 +87,22 @@ ExplicitEulerIntegrator<dim>::integrate(
     }
 }
 
+// Explicite Euler not implemented for disabling contacts
```
voferreira, line 90:
```suggestion
// Explicit Euler not implemented for disabling contacts
```

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+      bool has_mobile_nodes = false;
+
+      // Loop over nodes of cell
+      for (auto dof_index : local_dofs_indices)
+        {
+          // Check if node is mobile and assign mobile status to
+          // the cell
+          if (mobility_at_nodes[dof_index] == mobility_status::mobile)
+            {
+              status_to_cell[mobility_status::mobile].insert(*cell);
```
voferreira, line 268: This one also will be cleaner with map

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+          cell = active_ghost_cells.erase(cell);
+
+          // Assign empty status to nodes
+          for (auto dof_index : local_dofs_indices)
+            {
+              mobility_at_nodes(dof_index) = mobility_status::empty;
+            }
+        }
+      else
+        {
```
voferreira, line 192: Are those else necessary?

source/dem/disable_particle_contact.cc
```diff
@@ -0,0 +1,335 @@
...
+
+      bool has_active_nodes = false;
+      bool has_mobile_nodes = false;
+
+      // Loop over nodes of cell and check if cell has active or mobile
+      // nodes
+      // Active nodes and no mobiles nodes means that the cell is a neighbor
+      // of the layer of cell over mobily cell by criterion Active nodes
+      // with mobiles nodes means that the cell is part of this layer
+      for (auto dof_index : local_dofs_indices)
```
voferreira, line 314: Is this last loop necessary? I think it could be part of the last one.

voferreira (commented):
Another round of comments. I will keep up with it tomorrow.

source/core/parameters.cc
```diff
@@ -1920,6 +1920,13 @@ namespace Parameters
         "Enable testing mode of a solver. Some solvers have a specific"
         "testing mode which enables the output of debug variables. This"
         "testing mode is generally used only for the automatic testing bench using ctest.");
+      prm.declare_entry(
+        "type",
```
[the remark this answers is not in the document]
acdaigneault: Because in the future we could use it for any data (void fraction, granular temperature etc), so I kept it general even if there's no other option at the moment

source/dem/dem.cc
```diff
@@ -829,6 +1044,9 @@ DEMSolver<dim>::solve()
                                           triangulation,
                                           triangulation_cell_diameter);
 
+  // Setup background dof
+  setup_background_dofs();
```
blaisb, line 1048: I am not sure why you need  dofs here?

source/dem/dem.cc
```diff
@@ -888,22 +1116,39 @@ DEMSolver<dim>::solve()
           checkpoint_step = false;
 
           // Execute broad search by filling containers of particle-particle
+          // contact pair candidates and containers of particle-wall
           // contact pair candidates
-          container_manager.execute_particle_particle_broad_search(
-            particle_handler, has_periodic_boundaries);
+          if (has_disabled_contacts && contact_build_number > 1)
```
blaisb, line 1121: Your case is the exception, it is not the norm, so the order of the if-else should be reversed.

source/dem/dem.cc
```diff
@@ -969,21 +1214,35 @@ DEMSolver<dim>::solve()
...
       else
         {
-          integrator_object->integrate(
-            particle_handler,
-            parameters.lagrangian_physical_properties.g,
-            simulation_control->get_time_step(),
-            torque,
-            force,
-            MOI);
+          if (has_disabled_contacts && contact_build_number > 1)
```
blaisb, line 1225: Same comment as above

blaisb: It's also a bit weird to always carry the contact_build_number>1. Why not try to agglomerate this check into a small inlined function that is more verbosed.
contacts_are_disabled()
and it would return  has_disabled_contacts && contact_build_number > 1

source/dem/particle_particle_contact_force.cc
```diff
@@ -22,6 +22,8 @@
 
 #include <dem/particle_particle_contact_force.h>
 
+#include <deal.II/particles/particle_accessor.h>
```
blaisb, line 25: Why is this new include necessary?

source/dem/velocity_verlet_integrator.cc
```diff
@@ -130,5 +130,113 @@ VelocityVerletIntegrator<dim>::integrate(
...
+
+  for (auto &particle : particle_handler)
+    {
+      // Get the total array view to the particle properties once to improve
+      // efficiency
+      types::particle_index particle_id         = particle.get_local_index();
+      auto                  particle_properties = particle.get_properties();
+
+      // Check if the particle is mobile, if not, integration is skipped
+      auto &cell_particle = particle.get_surrounding_cell();
```
blaisb, line 156: This is an expensive call (getting the surrounding cell)
It might be better to write this integrate with the cell set instead by looping over the cells and then looping over the particles within these cells.

source/dem/velocity_verlet_integrator.cc
```diff
@@ -130,5 +130,113 @@ VelocityVerletIntegrator<dim>::integrate(
...
+{
+  Point<3>           particle_position;
+  const Tensor<1, 3> dt_g = g * dt;
+
+  for (auto &particle : particle_handler)
+    {
+      // Get the total array view to the particle properties once to improve
+      // efficiency
+      types::particle_index particle_id         = particle.get_local_index();
+      auto                  particle_properties = particle.get_properties();
```
blaisb, line 153: This is an expensive call, you want to do this only if the cell has the correct mobility status

source/dem/velocity_verlet_integrator.cc
```diff
@@ -130,5 +130,113 @@ VelocityVerletIntegrator<dim>::integrate(
...
+  Particles::ParticleHandler<dim> &particle_handler,
+  const Tensor<1, 3> &             g,
+  const double                     dt,
+  std::vector<Tensor<1, 3>> &      torque,
+  std::vector<Tensor<1, 3>> &      force,
+  const std::vector<double> &      MOI,
+  const typename DEM::dem_data_structures<dim>::cell_set
+    &mobile_cells_container)
+{
+  Point<3>           particle_position;
```
blaisb, line 145: I think you have the wrong logic here.
Instead of doing
loop over particles ->. get cell around particle. Which is super expensive
You would be better off doing
loop over cells -> if cell has correct status  get the particles therein -> integrate them

blaisb (changes requested):
Other batch of comment. I think some important rework needs to be done on the Verlet Integrator

voferreira (commented):
Checkpoint.

include/dem/disable_particle_contact.h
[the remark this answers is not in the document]
acdaigneault: Ok, I understand. Right now, the hole thing is based on the fact that the mobility status are unsigned integers, so I have to refactor that. Here is an example : 

> mobility_at_nodes[dof_index] =
                    std::max((int)mobility_status::active,  mobility_at_nodes[dof_index]);

Also it is useful for output since the value is already the value. 
I will change the vector of map thing to a unordered map, and most of it will have to change and I think it will be easier to change the num to enum class. I don't really understand how/why it is such a bad thing, but I will find out how to do what you ask :)

source/core/parameters_lagrangian.cc
```diff
@@ -514,6 +514,38 @@ namespace Parameters
           Patterns::Selection("velocity_verlet|explicit_euler|gear3"),
           "Choosing integration method"
           "Choices are <velocity_verlet|explicit_euler|gear3>.");
+
+        prm.declare_entry(
+          "enable dynamic disabling contacts",
```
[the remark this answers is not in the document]
acdaigneault: I don't really understand what you mean, why enable dynamic contact search?

source/dem/disable_particle_contact.cc
acdaigneault: Yes, at line 183, calling .erase() will return the next iterator, but if the particle is not emprty, we have to "manually" iterate on the next cell

source/dem/disable_particle_contact.cc
acdaigneault: Empty status or empty cell is kinda the same thing

source/dem/disable_particle_contact.cc
acdaigneault: Since I use a set of cells that I update during the loops, I can't really do that, otherwise I would have to return and use the updated set. It's a member function in the code write not, but in my next commit it would be a copy of the member (I would update this set only during load balancing)

source/dem/disable_particle_contact.cc
acdaigneault: No I can't because I need to know the value of every nodes after adding the additional layer of mobile to know what cells are active. Otherwise I can assign active cells from mobile node from the 2nd loop.

source/dem/velocity_verlet_integrator.cc
acdaigneault: Yeah, I didn't want to changed the structure here, but I know we talked about it. I will refactor that!

include/dem/dem_container_manager.h
```diff
@@ -133,6 +133,26 @@ class DEMContainerManager
     dealii::Particles::ParticleHandler<dim> &particle_handler,
     const bool                               has_periodic_boundaries = false);
 
+  /**
+   * Particle-particle broad search which finds the particle-particle contact
+   * pairs candidates, local or ghost, which shows the collision pairs.
```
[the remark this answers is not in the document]
acdaigneault: Bad copy and paste, will modify that

source/dem/dem_container_manager.cc
```diff
@@ -250,6 +250,24 @@ DEMContainerManager<dim>::execute_particle_particle_broad_search(
...
+  const DisableParticleContact<dim> &      disable_particle_contact_object,
+  const bool                               has_periodic_boundaries)
+{
+  particle_particle_broad_search_object.find_particle_particle_contact_pairs(
+    particle_handler, *this, disable_particle_contact_object);
+
+  if (has_periodic_boundaries)
+    {
+      particle_particle_broad_search_object
+        .find_particle_particle_periodic_contact_pairs(
```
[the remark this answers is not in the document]
acdaigneault: I would need to pass the `has_periodic_boundaries` again, it seems a lot. Also, this class use this pattern in other functions also for the has_floating_mesh or walls. It could be refactor, but not in this PR in my opinion

source/dem/disable_particle_contact.cc
acdaigneault: Just tried it, but still won't work with erase(), cell needs to be the iterator

source/dem/dem_container_manager.cc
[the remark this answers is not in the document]
voferreira: Agreed. We could add it to the to-do list

source/dem/velocity_verlet_integrator.cc
acdaigneault: It's so much easier and better actually

source/core/parameters_lagrangian.cc
```diff
@@ -514,6 +514,38 @@ namespace Parameters
           Patterns::Selection("velocity_verlet|explicit_euler|gear3"),
           "Choosing integration method"
           "Choices are <velocity_verlet|explicit_euler|gear3>.");
+
+        prm.declare_entry(
```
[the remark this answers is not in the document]
acdaigneault: I agree. We should also have subsection for load balancing also, what do you think? (for a refactor PR, not this one)

source/dem/dem.cc
```diff
@@ -272,24 +284,112 @@ DEMSolver<dim>::cell_weight(
...
+      alpha = parameters.model_parameters.active_load_balancing_factor;
+    }
+  else if (mobility_status == disable_contact_object.inactive)
+    {
+      alpha = parameters.model_parameters.inactive_load_balancing_factor;
+    }
+
+  switch (status)
+    {
+      case parallel::distributed::Triangulation<dim>::CELL_PERSIST:
```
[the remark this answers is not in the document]
acdaigneault: As done in the "original" function, this is because CELL_PERSIST and CELL_REFINE will both execute the line 318 to 320. A case not having a break will execute until there's a break.
https://stackoverflow.com/questions/33222437/switch-case-statement-without-break

source/dem/dem.cc
acdaigneault: It's a copy and paste from the original function cell_weight()

source/dem/dem.cc
```diff
@@ -272,24 +284,112 @@ DEMSolver<dim>::cell_weight(
...
+
+          for (unsigned int child_index = 0;
+               child_index < GeometryInfo<dim>::max_children_per_cell;
+               ++child_index)
+            n_particles_in_cell += alpha * particle_handler.n_particles_in_cell(
+                                             cell->child(child_index));
+
+          return n_particles_in_cell * particle_weight;
+        }
+        break;
```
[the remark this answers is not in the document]
acdaigneault: It's a copy and paste from the original function cell_weight()
I can change it for both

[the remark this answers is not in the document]
acdaigneault: Actually, the original function is a bit weird since return is kind of a break... it exits the function

source/dem/dem.cc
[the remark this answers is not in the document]
acdaigneault: Hey, you did that part :p Thanks to git blame

source/dem/dem.cc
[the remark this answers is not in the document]
acdaigneault: Also, does a cell will have a status no matter what? why there's a assert thrown as a default, but also a return = 0? I refer to the original function cell_weight(), it's the same here, I didn't see that.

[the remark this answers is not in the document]
acdaigneault: https://github.com/lethe-cfd/lethe/pull/438/files#diff-e3ac0044bbb88c3571a82f24f7468444e752c0e8e1f7fbd5f6356084073fed91
@blaisb you are the one that did the function, maybe there's a reason? 
It seems the pattern comes from some deal.II code : (line 4224)
https://github.com/dealii/dealii/blob/master/source/distributed/tria.cc

source/dem/dem.cc
```diff
@@ -387,23 +487,123 @@ DEMSolver<dim>::check_load_balance_dynamic()
...
+          // related to the mobility status of the cells
+          vector<double> process_to_load_weight(n_mpi_processes, 0.0);
+
+          // Get the particle weight
+          const unsigned int particle_weight =
+            parameters.model_parameters.load_balance_particle_weight;
+
+          std::vector<unsigned int> mobility_status_vector(
+            triangulation.n_active_cells());
+          disable_contact_object.get_mobility_status_vector(
```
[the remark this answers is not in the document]
acdaigneault: Good idea

source/dem/dem.cc
acdaigneault: I use it for the identify mobility status and for the and output

include/dem/disable_particle_contact.h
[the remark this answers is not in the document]
acdaigneault: I refactored the class and looked more into that and I think using integer helps a lot for the nodes value assignation without overwriting. Otherwise i would need to have many ifs. Per instance, this part will be way more complicated.

>               for (auto node_id : local_dof_indices) {                
                  // Possible cases: (nodes are initialized to inactive (0))
                  // node = max(active (1), inactive (0)) = active (1)
                  // node = max(active (1), empty (3))    = empty (3)
                  // node = max(active (1), mobile (2))   = mobile (2)
                  mobility_at_nodes[node_id] =  std::max((int)mobility_status::active,
                                                                                                mobility_at_nodes[node_id]);
                }

What do you prefer?

acdaigneault:
@voferreira @blaisb 

I did most of the modifications suggested by you guys :
- Change the vector of set to a unordered map
- Add more comment in the code
- Change the logic for the verlet integrator
- Change the parameter subsection for disabling contacts

What I have to do : 
- Add "official" doc for parameters (it won't be very explicit)
- Add an application test for load balancing with disabling contacts (or I will remove this feature from this PR I think)

About the suggestion to change the mobility_status from enum to enum class, I really don't know. The core of the node-based algorithm is kinda designed on integer comparison at node values. For me, using integers with a name was kinda convenient since we manipulate integers, but we read names (mobile, active, inactive). If you think this is bad programming or I should not do that, I can change it, but in my opinion, it's not that bad and changing it seems less clean (but not impossible). Let me know. Also, I replied to all the comments, the ones still unresolved are the one that I would like feedback or are responses to suggestions, feel free too resolve if there's nothing to say or to do.

Thanks a lot for your precious feedbacks!

voferreira (commented):
Checkpoint.

source/dem/particle_particle_broad_search.cc
```diff
@@ -132,6 +132,176 @@ ParticleParticleBroadSearch<dim>::find_particle_particle_contact_pairs(
...
+  // This includes the cell itself as well as the neighbouring cells that
+  // were identified.
+  // cell_neighbor_list_iterator is [cell_it, neighbor_0_it, neighbor_1_it, ...]
+  for (auto cell_neighbor_list_iterator = cells_local_neighbor_list.begin();
+       cell_neighbor_list_iterator != cells_local_neighbor_list.end();
+       ++cell_neighbor_list_iterator)
+    {
+      // The main cell & its mobility status
+      auto cell_neighbor_iterator = cell_neighbor_list_iterator->begin();
+      unsigned int main_cell_mobility_status =
```
voferreira, line 164: Wherever you have check_cell_mobility, you can just use the key->value of the unordered map instead.

include/dem/disable_particle_contact.h
```diff
@@ -0,0 +1,280 @@
...
+    const unsigned int                     n_active_cells,
+    MPI_Comm                               mpi_communicator);
+
+  /**
+   * @brief Find the mobility status of a cell
+   *
+   * @param cell The iterator of the cell that needs mobility evaluation
+   */
+  inline unsigned int
+  check_cell_mobility(
```
voferreira, line 190: With the new unordered_map, you will no longer need this. Just use the key->value mechanism.

source/dem/particle_particle_broad_search.cc
```diff
@@ -132,6 +132,176 @@ ParticleParticleBroadSearch<dim>::find_particle_particle_contact_pairs(
...
+      if (main_cell_mobility_status == DisableParticleContact<dim>::inactive)
+        continue;
+
+      // Get particles in the main cell
+      typename Particles::ParticleHandler<dim>::particle_iterator_range
+        particles_in_main_cell =
+          particle_handler.particles_in_cell(*cell_neighbor_iterator);
+
+      // Check to see if the main cell has any particles
+      const bool particles_exist_in_main_cell = !particles_in_main_cell.empty();
```
voferreira, line 178: You can use the map here also

source/dem/particle_particle_broad_search.cc
```diff
@@ -132,6 +132,176 @@ ParticleParticleBroadSearch<dim>::find_particle_particle_contact_pairs(
...
+      // If main cell has status "inactive", skip to next cell
+      if (main_cell_mobility_status == DisableParticleContact<dim>::inactive)
+        continue;
+
+      // Particles in the main cell
+      typename Particles::ParticleHandler<dim>::particle_iterator_range
+        particles_in_main_cell =
+          particle_handler.particles_in_cell(*cell_neighbor_iterator);
+
+      // Check to see if the main cell has any particles
```
voferreira, line 261: Same about the map

voferreira (changes requested):
Apart from the current comments, I think I don't have anything to add. After you have implemented our map idea I will give it a last pass.

doc/source/parameters/dem/model_parameters.rst
```diff
@@ -86,4 +96,6 @@ it calls load-balancing. :math:`{L}` and :math:`{\beta}` denote computational lo
 
 * Three rolling resistance models are available in Lethe-DEM: ``no_resistance``, ``constant_resistance``, ``viscous_resistance``.
 
-
+* ``dynamic disabling contacts`` subsection deals with the enabling of the feature that disables the particle contacts computation when particles in a cell have low motion according to their granular temperature.
+  ``enable dynamic disabling contacts`` enabled the feature, ``granular temperature threshold`` is the threshold of the granular temperature below which the contacts are disabled, and ``solid fraction threshold`` is the minimum solid fraction of the cell in which the contacts may be disabled.
```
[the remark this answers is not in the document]
acdaigneault: Yeah, this feature needs way more explaination. This is the minimal documentation just to keep track of it in the official doc, but I will do an example one day and I will extend the documentation here eventually. Is it ok with you @voferreira & @blaisb ?

source/dem/velocity_verlet_integrator.cc
```diff
@@ -130,5 +130,123 @@ VelocityVerletIntegrator<dim>::integrate(
...
+  const Tensor<1, 3> dt_g = g * dt;
+
+  for (auto &cell : triangulation.active_cell_iterators())
+    {
+      if (cell->is_locally_owned())
+        {
+          // Check if the cell is mobile, if not, integration is skipped
+          unsigned int mobile_status  = 2; // Value of the mobile status
```
blaisb, line 154: why not use the enum directly to set the value of the mobile status? This seems dangerous

blaisb (changes requested):
Last batch of tiny comments. Amazing work @acdaigneault

blaisb:
The following test crashed on debug on my machine
dem_3d/load_balancing_mobility_status.mpirun=2.debug
99: ##331     #:1   <== dummy
99: ##331     #:1   ==> 1
99: @                                                     @@
99: ##331           <==
99: ##331     #>2   ==> 8
99: @ Line 331 in file "/home/blaisb/work/lethe/lethe/lethe/applications_tests/dem_3d/load_balancing_mobility_status.mpirun=2.output" is shorter than expected!
99: ----------------
99: ##332     #:1   <== 1
99: ##332     #:1   ==> 0
99: @ Absolute error = 1.0000000000e+0, Relative error = Inf
99: ##332     #:2   <== 8
99: ##332     #:2   ==> 0
99: @ Absolute error = 8.0000000000e+0, Relative error = Inf
99: ##332           <==
99: ##332     #>3   ==> 0 0 0 0 0 0 
99: @ Line 332 in file "/home/blaisb/work/lethe/lethe/lethe/applications_tests/dem_3d/load_balancing_mobility_status.mpirun=2.output" is shorter than expected!
99: ----------------
99: ##333     #>1   <== 0 0 0 0 0 0 0 0 
99: ##333           ==>
99:

voferreira (approved):
All is good on my side after addressing Prof Blais' comments. As he said, you did an excellent job here. The algorithm, the implementation, and the documentation comments you came up with are amazing. It was fun to review it 😊

applications_tests/dem_3d/load_balancing_mobility_status.mpirun=2.output
```diff
@@ -0,0 +1,610 @@
...
+***************************************************************** 
+
+DEM time-step is 16.4177% of Rayleigh time step
+Warning: It is recommended to decrease the time-step
+Reading triangulation 
+
+Finished reading triangulation 
+Warning: expansion of particle-wall contact list is disabled. 
+This feature is useful in geometries with concave boundaries. 
+Warning: the requested number of particles for insertion (132) is higher than maximum expected number of particles (-12)
```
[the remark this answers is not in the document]
acdaigneault: Yeah, I don't know why it says that, do you have any clue? 
I tried to fix it, but it keep saying that, but if I change sightly the number of particles, I wasn't able to produce a simulation with every cases for the mobility status, so I kinda gave up.

doc/source/parameters/dem/model_parameters.rst
```diff
@@ -86,4 +96,6 @@ it calls load-balancing. :math:`{L}` and :math:`{\beta}` denote computational lo
 
 * Three rolling resistance models are available in Lethe-DEM: ``no_resistance``, ``constant_resistance``, ``viscous_resistance``.
 
-
+* ``dynamic disabling contacts`` subsection deals with the enabling of the feature that disables the particle contacts computation when particles in a cell have low motion according to their granular temperature.
```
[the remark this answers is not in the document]
acdaigneault: Will reformulate

source/dem/dem.cc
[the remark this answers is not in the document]
acdaigneault: Oh it sounds kinda passive-aggresive, not my intention hahahaha Sorry 🫠

doc/source/parameters/dem/model_parameters.rst
voferreira, line 99:
```suggestion
* ``dynamic disabling contacts`` subsection controls the disabling contact mechanism for performance enhancement. This feature dynamically searches for cells with low particle motion (granular temperature), disabling the computation of contacts for particles within these cells.
```

applications_tests/dem_3d/load_balancing_mobility_status.mpirun=2.output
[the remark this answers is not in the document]
blaisb: I'll try to take a look at it, seems weird though like an int overflow or a round-off error?

acdaigneault:
Ok, everything should be fixed! :dancers:

blaisb (approved):

PR #723, opened 2023-03-14
Flux postprocessing options

oguevremont:
# Description of the problem

- Momentum drop was needed in some situations were the inlet and outlet velocity profiles were different
- Flow rate output was needed as a debugging tool to ensure mass conservation

# Description of the solution

- Two CFD post-processing tools were added, with their documentation

# How Has This Been Tested?

- A few simulations were run to make sure they work properly

# Documentation

- [/parameters/cfd/post_processing.rst]

### DIFF
--- a/applications_tests/gls_vans_3d/beetstra_packed_bed.output
+++ b/applications_tests/gls_vans_3d/beetstra_packed_bed.output
@@ -8,7 +8,7 @@ Total pressure drop: 0 Pa
 *****************************************************************
 Transient iteration : 1        Time : 0.002    Time step : 0.002    CFL : 0.06743077177
 *****************************************************************
-Pressure drop: 39.99872011 Pa
+Pressure drop: 40.47276122 Pa
 Total pressure drop: 40.47276186 Pa
 Mass Source: 8.527914948e-13 s^-1
 Max Local Mass Source: 8.141667298e-08 s^-1
@@ -17,7 +17,7 @@ Average Void Fraction in Bed: 0.7930753133
 *****************************************************************
 Transient iteration : 2        Time : 0.004    Time step : 0.002    CFL : 0.2194770617
 *****************************************************************
-Pressure drop: 26.0868311 Pa
+Pressure drop: 26.08585282 Pa
 Total pressure drop: 26.08633753 Pa
 Mass Source: -3.432880331e-07 s^-1
 Max Local Mass Source: 7.964573385e-08 s^-1
@@ -26,7 +26,7 @@ Average Void Fraction in Bed: 0.7931023626
 *****************************************************************
 Transient iteration : 3        Time : 0.006    Time step : 0.002    CFL : 0.2196560133
 *****************************************************************
-Pressure drop: 26.46032496 Pa
+Pressure drop: 26.46432352 Pa
 Total pressure drop: 26.46383733 Pa
 Mass Source: 3.432967504e-07 s^-1
 Max Local Mass Source: 8.066553916e-08 s^-1
@@ -35,7 +35,7 @@ Average Void Fraction in Bed: 0.7930753133
 *****************************************************************
 Transient iteration : 4        Time : 0.008    Time step : 0.002    CFL : 0.221619556
 *****************************************************************
-Pressure drop: 26.03449776 Pa
+Pressure drop: 26.03064416 Pa
 Total pressure drop: 26.03112891 Pa
 Mass Source: -3.433050155e-07 s^-1
 Max Local Mass Source: 7.974216631e-08 s^-1
@@ -44,7 +44,7 @@ Average Void Fraction in Bed: 0.7931023626
 *****************************************************************
 Transient iteration : 5        Time : 0.01     Time step : 0.002    CFL : 0.2195549911
 *****************************************************************
-Pressure drop: 26.46016419 Pa
+Pressure drop: 26.4641568 Pa
 Total pressure drop: 26.46367054 Pa
 Mass Source: 3.432996425e-07 s^-1
 Max Local Mass Source: 8.068671768e-08 s^-1
@@ -53,7 +53,7 @@ Average Void Fraction in Bed: 0.7930753133
 *****************************************************************
 Transient iteration : 6        Time : 0.012    Time step : 0.002    CFL : 0.2216697475
 *****************************************************************
-Pressure drop: 26.03449141 Pa
+Pressure drop: 26.03063816 Pa
 Total pressure drop: 26.03112278 Pa
 Mass Source: -3.433007706e-07 s^-1
 Max Local Mass Source: 7.974885069e-08 s^-1
@@ -62,7 +62,7 @@ Average Void Fraction in Bed: 0.7931023626
 *****************************************************************
 Transient iteration : 7        Time : 0.014    Time step : 0.002    CFL : 0.2195735936
 *****************************************************************
-Pressure drop: 26.46016849 Pa
+Pressure drop: 26.46416159 Pa
 Total pressure drop: 26.46367516 Pa
 Mass Source: 3.433010221e-07 s^-1
 Max Local Mass Source: 8.068816679e-08 s^-1
@@ -71,7 +71,7 @@ Average Void Fraction in Bed: 0.7930753133
 *****************************************************************
 Transient iteration : 8        Time : 0.016    Time step : 0.002    CFL : 0.2216842662
 *****************************************************************
-Pressure drop: 26.03449833 Pa
+Pressure drop: 26.0306456 Pa
 Total pressure drop: 26.03113002 Pa
 Mass Source: -3.433029575e-07 s^-1
 Max Local Mass Source: 7.974819213e-08 s^-1
@@ -80,7 +80,7 @@ Average Void Fraction in Bed: 0.7931023626
 *****************************************************************
 Transient iteration : 9        Time : 0.018    Time step : 0.002    CFL : 0.2195840306
 *****************************************************************
-Pressure drop: 26.46017765 Pa
+Pressure drop: 26.46417131 Pa
 Total pressure drop: 26.46368464 Pa
 Mass Source: 3.433019734e-07 s^-1
 Max Local Mass Source: 8.068683777e-08 s^-1
@@ -89,7 +89,7 @@ Average Void Fraction in Bed: 0.7930753133
 *****************************************************************
 Transient iteration : 10       Time : 0.02     Time step : 0.002    CFL : 0.2216923848
 *****************************************************************
-Pressure drop: 26.03450845 Pa
+Pressure drop: 26.0306562 Pa
 Total pressure drop: 26.03114037 Pa
 Mass Source: -3.433010813e-07 s^-1
 Max Local Mass Source: 7.974663741e-08 s^-1
--- a/applications_tests/gls_vans_3d/beetstra_packed_bed.output
+++ b/applications_tests/gls_vans_3d/beetstra_packed_bed.output
@@ -9,7 +9,7 @@ Total pressure drop: 0 Pa
 Transient iteration : 1        Time : 0.002    Time step : 0.002    CFL : 0.06743077177
 *****************************************************************
 Pressure drop: 40.47276122 Pa
-Total pressure drop: 40.47276186 Pa
+Total pressure drop: 40.47276154 Pa
 Mass Source: 8.527914948e-13 s^-1
 Max Local Mass Source: 8.141667298e-08 s^-1
 Average Void Fraction in Bed: 0.7930753133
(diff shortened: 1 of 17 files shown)


source/solvers/navier_stokes_base.cc
```diff
@@ -1288,10 +1288,12 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
           Parameters::Verbosity::verbose)
         {
           this->pcout << "Pressure drop: "
+                      << std::setprecision(
+                           simulation_control->get_log_precision())
                       << this->simulation_parameters.physical_properties_manager
                              .get_density_scale() *
                            pressure_drop
-                      << " Pa" << std::endl;
+                      << " m^2/s^2" << std::endl;
```
blaisb, line 1296: Be careful the units are not m^2/s^2 if we multiply by density scale...

source/solvers/postprocessing_cfd.cc
```diff
@@ -159,6 +159,151 @@ calculate_pressure_drop<3, TrilinosWrappers::MPI::BlockVector>(
...
+  momentum_inlet_boundary =
+    Utilities::MPI::sum(momentum_inlet_boundary, mpi_communicator);
+  inlet_surface = Utilities::MPI::sum(inlet_surface, mpi_communicator);
+  momentum_outlet_boundary =
+    Utilities::MPI::sum(momentum_outlet_boundary, mpi_communicator);
+  outlet_surface = Utilities::MPI::sum(outlet_surface, mpi_communicator);
+
+  momentum_outlet_boundary = momentum_outlet_boundary / outlet_surface;
+  momentum_inlet_boundary  = momentum_inlet_boundary / inlet_surface;
+  momentum_drop            = momentum_inlet_boundary - momentum_outlet_boundary;
```
blaisb, line 262: I disagree here. This is wrong. The momentum drop should not be divided by the area.... The momentum drop is momentum at outlet - momentum at inlet just straight up. This is not the same as when we calculate the pressure drop in which we want to know the difference between the average pressure at the outlet minus the average pressure at the inlet.

source/solvers/postprocessing_cfd.cc
```diff
@@ -159,6 +159,151 @@ calculate_pressure_drop<3, TrilinosWrappers::MPI::BlockVector>(
...
+  for (const auto &cell : dof_handler.active_cell_iterators())
+    {
+      if (cell->is_locally_owned() && cell->at_boundary())
+        {
+          for (unsigned int face = 0; face < GeometryInfo<dim>::faces_per_cell;
+               face++)
+            {
+              if (cell->face(face)->at_boundary())
+                {
+                  if (cell->face(face)->boundary_id() == inlet_boundary_id)
```
blaisb, line 208: I feel there's a lot of code repetition in what follows (all the get functions values, etc.). Why not make this in a single block and then do the addition in a if depending on inlet or outlet? Would make the code easier to read.

source/solvers/navier_stokes_base.cc
```diff
@@ -1312,6 +1314,101 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
...
+           0) &&
+          this->this_mpi_process == 0)
+        {
+          std::string filename =
+            simulation_parameters.simulation_control.output_folder +
+            simulation_parameters.post_processing.momentum_drop_output_name +
+            ".dat";
+          std::ofstream output(filename.c_str());
+          momentum_drop_table.set_precision("time", 12);
+          momentum_drop_table.set_precision("momentum-drop", 12);
```
blaisb, line 1355: Does this really require an independent file or it could be in the same file as the pressure drop? Otherwise I feel we are generating a ton of file for no reason. E.g. when you calculate pressure drop we could also calculate the total momentum drop and put i in the same file

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -20,12 +20,18 @@ This subsection controls the post-processing other than the forces and torque on
...
     set initial time                  = 0.0
 
-    # Pressure drop calculation
+    # Pressure and momentum drop calculations
     set calculate pressure drop     = false
     set pressure drop name          = pressure_drop
+    set calculate momentum drop     = false
+    set momentum drop name          = momentum_drop
     set inlet boundary id           = 0
     set outlet boundary id          = 1
```
blaisb, line 29: Why not merge the momentum drop into the pressure drop file directly? I feel it's not necessarily a good idea to create more and more files and more and more options.  You could just add the momentum drop directly when you calculate pressure drop to the same file and that would be it. Would make it less parameters, less logic and less files and anyway when. you calculate the pressure drop you will most likely want to calculate the momentum drop.

blaisb (changes requested):
Few comments, nothing major, mostly aesthetic and some refactoring to do.

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -70,7 +74,7 @@ This subsection controls the post-processing other than the forces and torque on
 * ``calculate average velocities``: controls if calculation of time-averaged velocities is enabled.
     * ``initial time``: initial time used for the average velocities calculations.
 
-* ``calculate pressure drop``: controls if calculation of the pressure drop from the inlet boundary to the outlet boundary is enabled. 
+* ``calculate pressure drop``: controls if calculation of the pressure drop from the inlet boundary to the outlet boundary is enabled.
```
blaisb, line 77: I would add the formula here for the pressure drop and the "total pressure drop"

blaisb: I would also add the formula for the flow rate.

source/solvers/postprocessing_cfd.cc
```diff
@@ -159,6 +159,137 @@ calculate_pressure_drop<3, TrilinosWrappers::MPI::BlockVector>(
   const unsigned int                        inlet_boundary_id,
   const unsigned int                        outlet_boundary_id);
 
+template <int dim, typename VectorType>
+double
+calculate_total_pressure_drop(
```
blaisb, line 164: Maybe this calculation could have been merged with the regular calculate pressure drop? since any both are combined

blaisb (changes requested):
I would consider merging the calculation of the total pressure drop with the regular pressure drop. Otherwise everything is all chill

applications_tests/gls_vans_3d/beetstra_packed_bed.output
```diff
@@ -3,83 +3,94 @@ Running on 1 MPI rank(s)...
    Number of degrees of freedom: 3300
    Volume of triangulation:      5.656854249e-05
 Pressure drop: 0 Pa
+Total pressure drop: 0 Pa
 
 *****************************************************************
 Transient iteration : 1        Time : 0.002    Time step : 0.002    CFL : 0.06743077177
 *****************************************************************
-Pressure drop: 39.99872011 Pa
+Pressure drop: 40.47276122 Pa
```
blaisb, line 11: Something is weird h ere, why did the pressure drop results change?

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -70,9 +74,23 @@ This subsection controls the post-processing other than the forces and torque on
...
     * ``inlet boundary id`` and ``outlet boundary id``: define the IDs for inlet and outlet boundaries, respectively. 
     * ``pressure drop name``: output filename for pressure drop calculations.
+    * The pressure drop :math:`\Delta P` and total pressure drop :math:`\Delta P_\text{total}` are calculated as such, with :math:`\Gamma` representing the boundary, :math:`\pmb{u}` the velocity  and :math:`P` the pressure:
+
+.. math::
+    \Delta P =  \frac{ \int_{\Gamma_\text{inlet}} P d \Gamma}{\int_{\Gamma_\text{inlet}} d \Gamma} - \frac{ \int_{\Gamma_\text{outlet}} P d \Gamma}{\int_{\Gamma_\text{outlet}} d \Gamma}
```
blaisb, line 83:
```suggestion
    \Delta P =  \frac{ \int_{\Gamma_\text{inlet}} P d \Gamma}{\int_{\Gamma_\text{inlet}} 1 d \Gamma} - \frac{ \int_{\Gamma_\text{outlet}} P d \Gamma}{\int_{\Gamma_\text{outlet}} 1 d \Gamma}
```

source/solvers/postprocessing_cfd.cc
```diff
@@ -54,72 +54,112 @@ calculate_pressure_drop(const DoFHandler<dim> &       dof_handler,
...
-          // Gather pressure (values)
-          fe_values[pressure].get_function_values(evaluation_point,
-                                                  present_pressure_values);
-
-          for (unsigned int q = 0; q < face_n_q_points; ++q)
+          for (unsigned int face = 0; face < GeometryInfo<dim>::faces_per_cell;
+               face++)
             {
-              for (const auto &face : cell->face_iterators())
+              if (cell->face(face)->at_boundary())
```
blaisb, line 87: here I would also check if it's the inlet or the outlet using the face_id. Otherwise you are doing calculations for no reasons no?

blaisb (changes requested):
small final comments. Can you explain why some pressure drops results have changed?

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -70,9 +74,23 @@ This subsection controls the post-processing other than the forces and torque on
...
+* ``calculate pressure drop``: controls if calculation of the pressure drop from the inlet boundary to the outlet boundary is enabled.
     * ``inlet boundary id`` and ``outlet boundary id``: define the IDs for inlet and outlet boundaries, respectively. 
     * ``pressure drop name``: output filename for pressure drop calculations.
+    * The pressure drop :math:`\Delta P` and total pressure drop :math:`\Delta P_\text{total}` are calculated as such, with :math:`\Gamma` representing the boundary, :math:`\pmb{u}` the velocity  and :math:`P` the pressure:
+
+.. math::
+    \Delta P =  \frac{ \int_{\Gamma_\text{inlet}} P d \Gamma}{\int_{\Gamma_\text{inlet}} 1 d \Gamma} - \frac{ \int_{\Gamma_\text{outlet}} P d \Gamma}{\int_{\Gamma_\text{outlet}} 1 d \Gamma}
+
+.. math::
+    \Delta P_\text{total} =  \frac{ \int_{\Gamma_\text{inlet}} (P + \pmb{u} \cdot \pmb{u}) d \Gamma}{\int_{\Gamma_\text{inlet}} d \Gamma} - \frac{ \int_{\Gamma_\text{outlet}} (P + \pmb{u} \cdot \pmb{u}) d \Gamma}{\int_{\Gamma_\text{outlet}} d \Gamma}
```
Luckabarbeau, line 86: If I am not mistaken, the pressure drop component coming from the velocity should be multiplied by 1/2. Also, should we take only the square of the velocity normal to the boundary (I am not sure about this though).

source/solvers/postprocessing_cfd.cc
```diff
@@ -54,72 +54,116 @@ calculate_pressure_drop(const DoFHandler<dim> &       dof_handler,
...
+                      temp_static_pressure_at_boundary +=
+                        fe_face_values.JxW(q) * pressure_values[q];
+                      temp_dynamic_pressure_at_boundary +=
+                        fe_face_values.JxW(q) * velocity_values[q] *
+                        velocity_values[q];
                     }
-
-                  if (face->at_boundary() &&
-                      (face->boundary_id() == inlet_boundary_id))
+                  if (boundary_id == inlet_boundary_id)
```
Luckabarbeau, line 114: It is very weird that all your pressure drop from the static pressure as changes. I would try to find an explanation since from what I see here, it should be identical.

Luckabarbeau (changes requested):
Very nice work ! This will help a lot with debugging new features. Although, I think there is an error in the calculation of the total pressure drop on the velocity component.

source/solvers/postprocessing_cfd.cc
oguevremont: I believe it is because the pressure values that were that were previously used came from the cell, and not the face (old line 76), even if the quadrature points iteration was done on the face. I reversed the concerned lines and tested on my machine and it seems to be the problem. 
From what I understand it would make more sense to take the values from the face like I did, even if it changes the values. @blaisb

doc/source/parameters/cfd/post_processing.rst
blaisb: yeah I also think it should be the normal velocity component but I am also not 100% sure aboutthat.

source/solvers/postprocessing_cfd.cc
blaisb: Yeah definitely it should be the values of the face and not of the cells

blaisb:
@oguevremont you need to rebase the branch but then it should be ok with a merge

doc/source/parameters/cfd/post_processing.rst
oguevremont: I looked into it a bit, and it seems that since dynamic pressure relates to the kinetic energy of the fluid we should use the magnitude of the velocity.

blaisb:
@oguevremont Can you rerun CI, it should work now

blaisb (approved):

blaisb:
@Luckabarbeau your review is missing for a merge

PR #725, opened 2023-03-15
Fix cut cell bug

oguevremont:
# Description of the problem

- The cells which had all of their vertices shared with cut cells, but that were not cut themselves, made it so that there was non-physical mass creation/destruction at the neighboring non-overconstrained cells.

# Description of the solution

- During the mapping of the cut cells, an additional step was added. This step maps the non-cut cells that have all of their vertices being shared with cut or boundary cells and considers them as overconstrained cells.
- This step converts these overconstrained fluid cells to a different state (with a different mapping), and the mapping is used to ensure that no non-physical phenomena related to mass creation/descruction happens.
 
# How Has This Been Tested?

- Most tests using sharp immersed boundaries and that had particle-particle or particle-wall proximity had their results changed.
- [gls_sharp_navier_stokes_3d/almost_blocked_channels] A test with almost closed channels was added. At the first level of refinement, there is no flow because the cells are overconstrained in the channel. At the second refinement, there is now flow.

# Documentation

No change necessary.


applications_tests/gls_sharp_navier_stokes_3d/rbf_sphere_mms.output
```diff
@@ -6,17 +6,17 @@ Running on 1 MPI rank(s)...
 *****************************************************************
 Steady iteration :        1/3
 *****************************************************************
-div u : 1.241e-08
-error at the weak BC : 0.3751
-L2 error velocity : 0.3046 L2 error pressure: 1.342
+div u : 0
+error at the weak BC : 0
+L2 error velocity : 0 L2 error pressure: 0
```
blaisb, line 11: Why is this all zero now? it seems like there is no solution at this iteration?

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -576,6 +576,25 @@ Return a bool that describes  if a cell contains a specific point
            std::tuple<bool, unsigned int, unsigned int>>
     cut_cells_map;
 
+
+  /*
+   * This map uses the cell as the key, and stores the following information:
+   * if that cell is overconstrained (bool), what particle overconstrains this
```
blaisb, line 582:
```suggestion
   * if the cell is overconstrained (bool), what particle overconstrains this
```

blaisb (approved):
Just two tiny changes, this is very well done

applications_tests/gls_sharp_navier_stokes_3d/rbf_sphere_mms.output
oguevremont: It's because all the cells are either overconstrained or cut. I could remove this refinement level or leave it, as it shows that the solver doesn't crash in that case.

blaisb: Nah it's good for me, as long as you understand why it's doing that it's ok

PR #729, opened 2023-03-20
Improve silo example

blaisb:
# Description of the problem

- Silo example had very old results for the timing
- Performance of the code has severely improved since then

# Description of the solution

- Reran the example, fixed some parameters and updated the timing results
- Lowered the simulation time (40s was 10 too many)

# Comments

- The simulation time are very different now... It takes 13ish hours on 8 cores instead of the old 48 hours on 32 cores... So it seems the code is like... 4-8 times faster than before on these kind of cases...

### DIFF
--- a/doc/source/examples/dem/silo/silo.rst
+++ b/doc/source/examples/dem/silo/silo.rst
@@ -100,9 +100,9 @@ Model parameters
 
     subsection model parameters
       set contact detection method                = dynamic
-      set dynamic contact search size coefficient = 0.8
+      set dynamic contact search size coefficient = 0.9
       set load balance method                     = frequent
-      set load balance frequency                  = 100000
+      set load balance frequency                  = 10000
       set neighborhood threshold                  = 1.3
       set particle particle contact force method  = hertz_mindlin_limit_overlap
       set particle wall contact force method      = nonlinear
--- a/doc/source/examples/dem/silo/silo.rst
+++ b/doc/source/examples/dem/silo/silo.rst
@@ -51,7 +51,7 @@ An insertion box is defined inside and on the top of the silo.
     subsection insertion info
       set insertion method                               = non_uniform
       set inserted number of particles at each time step = 20000
-      set insertion frequency                            = 20000
+      set insertion frequency                            = 10000
       set insertion box minimum x                        = -0.37
       set insertion box minimum y                        = -0.042
       set insertion box minimum z                        = 0.9
@@ -116,8 +116,8 @@ Simulation control
 .. code-block:: text
 
     subsection simulation control
-      set time step        = 1e-5
-      set time end         = 40
+      set time step        = 2e-5
+      set time end         = 30
       set log frequency    = 1000
       set output frequency = 1000
     end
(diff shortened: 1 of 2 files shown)


doc/source/examples/dem/silo/silo.rst
```diff
@@ -169,14 +169,14 @@ In this simulation, we need a stopper (floating wall) in the filling stage (0-4
 
 Running the simulation
 ----------------------
-This simulation can be launched by (in parallel mode on 32 processes):
+This simulation can be launched by (in parallel mode on 8 processes):
```
voferreira, line 172:
```suggestion
This simulation can be launched in parallel (8 processes) running:
```

doc/source/examples/dem/silo/silo.rst
```diff
@@ -169,14 +169,14 @@ In this simulation, we need a stopper (floating wall) in the filling stage (0-4
...
 
 .. code-block:: text
 
-  mpirun -np 32 dem_3d silo-Golshan.prm
+  mpirun -np 8 dem_3d silo-Golshan.prm
 
 .. warning::
-	This example takes approximately 48 hours on 32 cores. This high computational cost is due to the long simulation time (40 s of real-time) and the high number of particles. The number of particles can be reduced to have a shorter simulation time. For example, the entire simulation could be run in less than an hour on an 8 cores processor if only 13 230 particles are inserted.
+	This example takes approximately 14 hours on 8 cores. This high computational cost is due to the long simulation time (30 s of real-time).
```
voferreira, line 179: Maybe use "computational time" instead of "computational cost"?

voferreira (approved):
There are still two parameters to be fixed. Good to go after changing the following in the .rst file:
Line 103: set dynamic contact search size coefficient = 0.8 -> 0.9
Line 105: set load balance frequency                  = 100000 -> 10000
Additionally, the `Lagrangian post-processing` is not mentioned in the .rst file, but I don't know if you need to add it.

blaisb:
> There are still two parameters to be fixed. Good to go after changing the following in the .rst file: Line 103: set dynamic contact search size coefficient = 0.8 -> 0.9 Line 105: set load balance frequency = 100000 -> 10000 Additionally, the `Lagrangian post-processing` is not mentioned in the .rst file, but I don't know if you need to add it.

I don't think it's necessary
Fixed the other parameters though. Thanks for the keen eye!

examples/dem/3d-silo/silo-Golshan.prm
```diff
@@ -17,19 +18,26 @@ end
 
 subsection restart
   set checkpoint = true
-  set filename   = sliding_restart
```
acdaigneault, line 20: Why did you remove it? Why not just change the name?

acdaigneault (approved):
Everything looks good!

examples/dem/3d-silo/silo-Golshan.prm
blaisb: I think it's better to leave the default restart file name...

PR #731, opened 2023-03-31
Composite with cell guess

oguevremont:
# Description of the problem

- The boolean operations used in the composite shape class used numerical gradient instead of analytical gradient, which was very costly (especially when RBF or OpenCascade shapes were used)
- There was some repeated code in how the boolean operations were applied
- A levelset gradient postprocessor would be nice for debugging purposes

# Description of the solution

- Functions to analytically apply the boolean operations for gradient calculation were implemented
- A levelset gradient postprocessor was added

# How Has This Been Tested?

- Some gls_sharp_navier_stokes_3d tests were changed due to usage of analytical boolean operations. 
- The cavatappi test is especially sensitive for the force and torque calculations, but the results of the test still make sense visually

# Documentation

- No documentation change needed


applications_tests/gls_sharp_navier_stokes_3d/cavatappi_composite_test.output
```diff
@@ -17,5 +17,5 @@ Steady iteration :        1/1
 +------------------------------------------+
 |  Force  summary particles                |
 +------------------------------------------+
-particle_ID  time    T_x     T_y    T_z     f_x    f_y    f_z    f_xv    f_yv   f_zv   f_xp    f_yp   f_zp  
-          0 1.0000 -0.6968 -4.4547 2.5233 49.1749 1.0278 7.1728 26.3108 0.3030 0.1214 22.8641 0.7248 7.0514
\ No newline at end of file
+particle_ID  time   T_x     T_y     T_z      f_x    f_y    f_z    f_xv    f_yv    f_zv    f_xp   f_yp    f_zp
+          0 1.0000 0.8377 -1.0770 -12.4452 36.4788 9.4690 8.0446 27.5955 -2.0269 -0.8735 8.8833 11.4959 8.9182
```
blaisb, line 21: Wow the results for this one has indeed changed dramatically.... Are you sure this is a good test?

source/core/shape.cc
```diff
@@ -925,47 +925,17 @@ CompositeShape<dim>::value(const Point<dim> &evaluation_point,
   Point<dim> centered_point = this->align_and_center(evaluation_point);
 
   // The levelset value of all constituent shapes is computed
-  std::map<unsigned int, double> components_value;
+  std::map<unsigned int, double>         components_value;
+  std::map<unsigned int, Tensor<1, dim>> dummy_components_gradient;
```
blaisb, line 929: Can you explain what is dummy_components_gradients ?

source/core/shape.cc
```diff
@@ -980,43 +950,67 @@ CompositeShape<dim>::value_with_cell_guess(
...
+      dummy_components_gradient[component_id] = Tensor<1, dim>{};
     }
 
-  // The boolean operations between the shapes are applied in order
-  // The last computed levelset value is considered to be the right value
-  double levelset = components_value[0];
-  for (auto const &[operation_id, op_triplet] : operations)
+  double levelset;
+  std::tie(levelset, std::ignore) =
+    apply_boolean_operations(components_value, dummy_components_gradient);
```
blaisb, line 964: The name of the variables you use to apply boolean operations here is very different from the one in the .h, Can you document it a bit more? What is dummy_components_gradients? Is it the gradients for each shape? Then use a more verbose name that describes that it is.

applications_tests/gls_sharp_navier_stokes_3d/cavatappi_composite_test.output
oguevremont: I'm not sure I have to admit. It's nice to have I'd say but not robust enough. I added an lines to the unit test for composites shapes.

PR #733, opened 2023-04-04
Update simulation control verbosity and tests

blaisb:
# Description of the problem

- The delimiter for the simulation control output to the terminal was always a bit off sometimes. Less * than needed, more than needed, and so on and so forth. @voferreira did not like that :)

# Description of the solution

- Use the announce string function to print it instead so that the width is always correct.

# How Has This Been Tested?

- Does not change anything. However, it modifies the output of ALL tests. Consequently, I decided to do this on this small branch.

# Future changes

- I think uniformity is good. So trying to normalize our outputs is a good idea.
-


include/core/utilities.h
```diff
@@ -231,20 +231,25 @@ create_output_folder(const std::string &dirname);
...
+ * @param expression string that will be printed
+ * @param delimiter the character used to delimit the printing. Default value is "-"
  */
 inline void
-announce_string(const ConditionalOStream &pcout, const std::string expression)
+announce_string(const ConditionalOStream &pcout,
+                const std::string         expression,
+                const char                del = '-')
 {
   for (unsigned int c = 0; c < expression.size() + 1; ++c)
```
voferreira, line 244: There is no need to loop here. You can use the string constructor and do `pcout << std::string(expression.size(), del)`

voferreira: And I would use `output` and `delimiter`

applications_tests/cfd_dem_coupling_3d/dynamic_contact_search.output
```diff
@@ -59,9 +59,9 @@ Finished 100 DEM iterations
 Global continuity equation error: -1.507798244e-18 s^-1
 Max local continuity error: 1.562921799e-11 s^-1
 
-*****************************************************************
-Transient iteration : 3        Time : 0.015    Time step : 0.005    CFL : 1.44013875e-07
-*****************************************************************
+**************************************************************************************
+Transient iteration : 3        Time : 0.015    Time step : 0.005    CFL : 1.44014e-07
```
voferreira, line 63: Why did you change the precision?

voferreira (changes requested):
@voferreira likes it now haha. I suggest you use the string constructor for the delimiter.

applications_tests/cfd_dem_coupling_3d/dynamic_contact_search.output
blaisb: I did not change the precision, just changed the computer.

voferreira (approved):
All clear. I will work on the other outputs.

PR #735, opened 2023-04-06
Update verbosity and tests

voferreira:
# Description of the problem

- The ouputs of the particles insertion, dem initial information, and other parts of the code were not normalized.

# Description of the solution

- Applied `announce_string` on the parts I found it would make sense to normalize the output.

# How Has This Been Tested?

- No tests needed. On the other hand, this pr modifies all application_tests outputs.

# Comments

- This PR follows PR #733 .


source/dem/CMakeLists.txt
```diff
@@ -32,8 +32,7 @@ add_library(lethe-dem
   particle_wall_nonlinear_force.cc
   periodic_boundaries_manipulator.cc
   post_processing.cc
-  print_initial_information.cc
-  read_checkpoint.cc
+        read_checkpoint.cc
```
blaisb, line 35: Please indent this line correctly

include/core/utilities.h
```diff
@@ -241,9 +241,9 @@ announce_string(const ConditionalOStream &pcout,
                 const std::string         expression,
                 const char                delimiter = '-')
 {
-  pcout << std::string(expression.size() + 1, delimiter) << std::endl;
+  pcout << std::string(expression.size(), delimiter) << std::endl;
   pcout << expression << std::endl;
-  pcout << std::string(expression.size() + 1, delimiter) << std::endl;
+  pcout << std::string(expression.size(), delimiter) << std::endl;
```
blaisb, line 246: Personnaly I thought it was cleaner with the +1., but if you think it's better without i am ok with it.

blaisb (approved):
The tests are currently not working. Do you have the l atest version of deal.II master ? If you want, I can run them on my machine and update them from there instead.

include/core/utilities.h
voferreira: Another option would be `expression.size() + 1` + start the string with white space in front. Which one do you prefer?

blaisb: > Another option would be `expression.size() + 1` + start the string with white space in front. Which one do you prefer?

Nah I don't think starting with a white space is the right approach :)

voferreira:
My deal.ii version is up to date. The only test that did not pass was `dem_3d/insert_periodic_boundary_gmsh`, But the problem is with my gmsh version. Could you please fix this one? I have fixed the indentation and `expression.size() + 1`.

blaisb:
The issue should not stem from your GMSH version since you don't have to redo the GMSH mesh... I'll re-update the tests later today

blaisb:
@voferreira I updated the breaking test. Should be sorted out now. Once CI passes I will merge if you are ok with that

voferreira:
@blaisb I think it is all done. Thank you for updating the test!

PR #737, opened 2023-04-11
Bound cell void fraction at 1 for drag model calculation

acdaigneault:
# Description of the problem

- All interpolated void fraction values should be bounded to 1 before usage in any drag models
- Koch and Hill drag model didn't work because of the `log(1 - cell_void_fraction)` that gave a -nan value in the rhs

# Description of the solution

- The void fraction of cells used in drag models are bounded at 1
- Fix some equations when (1 - void fraction) = 0 may gives nan (add a small value)

# How Has This Been Tested?

- All tests are good

### DIFF
--- a/source/fem-dem/vans_assemblers.cc
+++ b/source/fem-dem/vans_assemblers.cc
@@ -1084,9 +1084,9 @@ GLSVansAssemblerKochHill<dim>::calculate_particle_fluid_interactions(
       // Koch and Hill Drag Model Calculation
       if ((1 - cell_void_fraction) < 0.4)
         {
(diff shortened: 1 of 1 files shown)


acdaigneault:
@blaisb I was not able to do a good application tests to reproduce the bug (the case I use has PBC add it's currently not fixed in the master branch, so using it is not a good test). I don't think it's worth spending more time on that to have an unit test.

Able to reproduce the bug with Koch Hill and I made sure the fix works. Wasn't able to have a Nan value with Beestra model since it's sqrt(0), but I added the DBL_MIN just in case it's sightly < 0.

blaisb:
Then I'll merge after CI @acdaigneault . No biggie

PR #739, opened 2023-04-17
Fix opencacade solid shape

Luckabarbeau:
# Description of the problem

-Some steps could have more than one shell for one solid. In these cases, we were throwing an error. Now we simply use the solid.

# Description of the solution

-Now, we simply use the solid instead of throwing.

### DIFF
--- a/include/core/shape.h
+++ b/include/core/shape.h
@@ -931,7 +931,20 @@ public:
     // distance tool with just the shell.
     if (shells.size() > 0)
       {
-        if (shells.size() > 1)
+        if (solids.size() == 1)
+          {
+            shape = solids[0];
+            OpenCASCADE::extract_compound_shapes(
+              shape, compounds, compsolids, solids, shells, wires);
+            std::cout << "nb shell of solid 0" << shells.size() << std::endl;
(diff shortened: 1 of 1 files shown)


include/core/shape.h
```diff
@@ -931,7 +931,15 @@ class OpenCascadeShape : public Shape<dim>
     // distance tool with just the shell.
     if (shells.size() > 0)
       {
-        if (shells.size() > 1)
+        if (solids.size() == 1)
```
bibeauv, line 934: The first `if` statement uses the object `solids` while the second `if` statement uses the object `shells`. I just want to make sure that this is normal :)

bibeauv (commented):
Bravo Lucka :p
Only thing, maybe just add a comment before the new `if` statement to explain what you do if the size is 1.
B

include/core/shape.h
bibeauv, line 934: It could be good to add a comment to explain what you do in the `if` statement!

blaisb: Yeah clearly I would at least document this part.

blaisb (approved):

PR #743, opened 2023-04-25
WIP: Add home made gcr solver *experimental*

Luckabarbeau:
# Description of the problem

- The sharp solver is unstable with certain parameters with an ilu precondition but not with an identity preconditioner. Since the preconditioner and the iterative process are very opaque when using Trillinos vector, we implemented a variation on a homemade basic GCR  iterative solver. 
- Note : there are some modifications from the classic GCR implementation to reuse more information from the previous correction vector. I it is still named it GCR since alpha is defined in the same way as in the GCR algorithm if you store 0 previous correction vector

# Description of the solution

- The GCR solver was implemented to enable future debugging and modification to the iterative solver process for sharp IB and understand why it is unstable with ILU preconditioning in some specific cases.

# How Has This Been Tested?

- The static stokes case is now solved using the GCR solver.
- The periodic hill case benchmarked the solver compares to GMRES. A similar number of iterations and time per iteration was observed. 

# Documentation

 Still missing. 
- Update parameter.
- Theory guide of the solver.

### DIFF
--- a/source/solvers/gls_navier_stokes.cc
+++ b/source/solvers/gls_navier_stokes.cc
@@ -1298,7 +1298,7 @@ GLSNavierStokesSolver<dim>::solve_system_gcr_iterative(
   TrilinosWrappers::MPI::Vector variation_in_direction(this->locally_owned_dofs,
                                                        this->mpi_communicator);
 
-  // tol of solution
+  // Tol of solution
   const AffineConstraints<double> &constraints_used =
     initial_step ? this->nonzero_constraints : this->zero_constraints;
   const double linear_solver_tolerance =
@@ -1311,7 +1311,7 @@ GLSNavierStokesSolver<dim>::solve_system_gcr_iterative(
                   << linear_solver_tolerance << std::endl;
     }
 
-  // Define initial solution guess assuming diagonal matrix.
+  // Define initial solution guess.
   solution = 0;
 
   // Initialize residual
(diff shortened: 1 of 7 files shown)


source/solvers/gls_navier_stokes.cc
```diff
@@ -1262,6 +1269,267 @@ GLSNavierStokesSolver<dim>::setup_AMG()
   amg_preconditioner->initialize(system_matrix, parameter_ml);
 }
 
+
+template <int dim>
+void
+GLSNavierStokesSolver<dim>::solve_system_gcr_iterative(
```
peterrum, line 1275: What about moving this into deal.II? I worked on GCR a long time ago (https://github.com/dealii/dealii/pull/12760) but never finished the PR...

Luckabarbeau: Hey @peterrum,

This is a good idea. We could definitely add this solver to dealii once it is stable. However, this is a relatively immature implementation. I did some modifications from the classic GCR implementation to reuse more information from the previous correction vector. I made these modifications to learn about how iterative linear solvers work. As such, I don't know the exact properties of the solver other than some benchmarks I made. I still named it GCR since alpha is defined in the same way as in the GCR algorithm if you store 0 previous correction vector. 

Also, I wrote this piece of code to have a test platform that I can modify with additional information I have on the matrix structure in the case of the sharp solver in the future.

I will add some information about the exact algorithm in the theory guide.

Luckabarbeau: I finish adding a brief overview of the algorithm to the theory guide. If you have any comments or suggestions, let me know. It is my first time playing with an iterative solver, I may have missed some key elements.

blaisb:
@Luckabarbeau can you write an implementation in a whole separate file with a unit test for a matrix system? I'm open to the idea of a (temporary) iterative linear solver, but hard coding it inside one of the solver itself is not a good idea. I also agree with @peterrum that these type of solvers are a better fit for deal.II than for Lethe itself. I know that if we make a temporary version, we will never have a permanent deal.II version...

Luckabarbeau:
After some discussion, we decided that we will keep this branch as an experimental branch to perform the potential modification to the precondition related to the constraint. If there some improvement to be made. I will add this solver in dealii.

PR #744, opened 2023-04-25
Vof improve marangoni effect

hepap:
# Description of the problem

- While verifying the implementation of the Marangoni effect with a analytical solution (thermocapillary migration of a droplet), an error was found  in the formulation of the surface tension force term.
- The default value of parameter beta in the tanh phase filtration was leading to significant errors in the computation of the barycenter position and velocity for large domains and fine meshes.

# Description of the solution

- The formulation of the surface tension force term was corrected to take into account the variation of the surface tension coefficient with the temperature.
- The default value of parameter beta was increased to reduce the error.

# How Has This Been Tested?

- [ ] gls_navier_stokes_2d/gls_droplet_marangoni_effect : the test already exists but was changed

# Documentation

The default value of parameter beta was changed in the documentation

- [ ] doc/source/parameters/cfd/volume_of_fluid.rst

# Future changes

- I think an example and/or a section in the future VOF theory guide about the Marangoni effect would be important to add :)


source/solvers/navier_stokes_vof_assemblers.cc
```diff
@@ -601,42 +601,46 @@ GLSNavierStokesVOFAssemblerMarangoni<dim>::assemble_rhs(
...
+      const double temperature = scratch_data.temperature_values[q];
 
       // Gather temperature gradient
       const Tensor<1, dim> temperature_gradient =
         scratch_data.temperature_gradients[q];
 
       const double JxW_value = JxW[q];
 
+
+      const Tensor<1, dim> tmp_STF =
```
blaisb, line 625: I suggest using a more verbose name here.

blaisb (changes requested):
I would just improve the tmp_STF naming, otherwise all good to merge :)

source/solvers/navier_stokes_vof_assemblers.cc
hepap: I changed it in GLSNavierStokesVOFAssemblerMarangoni and GLSNavierStokesVOFAssemblerSTF

blaisb (approved):

blaisb:
Will merge one CI finishes. :)

PR #745, opened 2023-04-26
Enable Lagrangian post-processing PVD restart

acdaigneault:
# Description of the problem

- There's currently no restart checkpoints for grid PVD that is used for Lagrangian post-processing results
- When running restart simulations, the PVD file for post-processing results outputs was not updated and it was not possible to visualize results in the previous simulations with the PVD

# Description of the solution

- Handle the checkpoints for grid PVD as it is for fluid or particles in DEM and CFD-DEM

# How Has This Been Tested?

- Restart has been tested for DEM and CFD-DEM and it's now possible to visualize all the simulation outputs with the PVD file

# Comments

- I don't know if I should handle the grid PVD reading as with other PVD. For instance, if we run a restart of a simulation that didn't have Lagrangian post-processing enabled in the previous restart. I don't know if this is really relevant.
- @blaisb do you think I need another reviewer for that since it's kinda copy and paste.

### DIFF
--- a/include/dem/read_checkpoint.h
+++ b/include/dem/read_checkpoint.h
@@ -46,6 +46,7 @@ using namespace std;
  * @param dem_parameters Input DEM parameters in the parameter handler file
  * @param simulation_control Simulation control
  * @param particles_pvdhandler PVD handler
+ * @param grid_pvdhandler PVD handler for post-processing
  * @param triangulation Triangulation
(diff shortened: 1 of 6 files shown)


blaisb (approved):

blaisb:
Awesome as is. I 'll merge

PR #746, opened 2023-04-27
[wip] Modify the dynamic flow control and add it to CFD-DEM

acdaigneault:
# Description of the problem

- The dynamic flow control wasn't working for CFD-DEM since it didn't take the void fraction into account
-  It wasn't possible to adjust the flow convergence rate or the stability resulting of large flow rate/average velocity oscillations due to large difference between beta force and the new beta force values
- There was some issue with the void fraction periodicity with the GLS VANS solver

# Description of the solution

- Implementation of the flow control in the coupling CFD-DEM solver
- Modify the type of the target value: flow rate has been changed to average velocity. It is a better concept for CFD-DEM since we can't evaluate the flow rate at a boundary, we need to evaluate the velocities and the void fraction for the whole domain. In other to keep the flow controller model the same for CFD and CFD-DEM (volumetric flow rate vs average velocity), I choose to also change the type value evaluation for CFD.
- Add a relaxation coefficient (alpha) that may help control convergence or stability
- Add a parameter allowing beta force for particles: it is calculated as beta_p = beta_f * rho_f / rho_p and the force is apply to all particles
- Resolve the issue for the void fraction periodicity with `make_periodicity_constraints`
- Some refactoring of the dynamic flow control has been done : 
      `calculate_average_velocity()` (previously `calculate_flow_rate()`) and `calculate_beta()` are now executed in the same place in the code, at the end of the iteration. Before this modification, the flow rate n was calculated at the end of a step n and the beta n+1 was calculated at the beginning of the step n+1. Now, both flow rate n and beta n+1 are calculated at the end of the time step n. You may see that the printed average velocity (flow rate) of step n+1 is the value of n.

# How Has This Been Tested?

- All tests for flow control has been reviewed. They were modified and tested with the new parameters. 
- For the application tests, the target value is now average velocity = volumetric flow rate / area of the old test version so the calculated beta values are the same.
- The result for the unit tests has been calculated in a spreadsheet and the printed results are the same.
- The parameters of the Periodic Hills example has been modified and tested

# Documentation

- The documentation is updated with the new parameters (average velocity target, alpha, enable beta particle)
- The term "volumetric flow rate" is now "average velocity"
- Add the equation of the main controller in doc

# Comments

- The flow control for CFD-DEM has not been fully tested (not with qcm void fraction calculation method). This branch is a part of another branch that has other fix (qcm is not currently working on master in periodic).
- I'm planning to do a application test with the flow control with qcm in periodic on the branch that fix the qcm.


blaisb:
Why the [WIP] label? Is it ready for review or not? If you want to put it as WIP, it's better to use the github label WIP instead of putting it in the title.

applications_tests/gls_navier_stokes_3d/poiseuille_restart_files/poiseuille_restart.flowcontrol
```diff
@@ -1,5 +1,5 @@
 Flow control
 Previous_beta 25.9666
-Previous_flow_rate -10.0126
+Previous_average_velocity -3.20768
```
blaisb, line 3: The reason for the change is the switch from flow to average velocity right?

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
 Dynamic flow control
 ~~~~~~~~~~~~~~~~~~~~
 
-The purpose of this subsection is to enable dynamic flow control. It is important when we want to target a volumetric flow on a specific boundary. To control the volumetric flow, the code calculates a :math:`\beta`  coefficient at each time step that is used to keep the volumetric flow value inside an acceptable threshold. If the chosen wall boundary on which a volumetric flow is imposed is an inlet, the volumetric flow rate targeted must be negative since the outward normal vector is used for the calculation. Thereby, if the chosen wall boundary is an outlet, volumetric flow rate must be positive.
+The purpose of this subsection is to enable dynamic flow control. It is important when we want to simulate an average
+velocity on a specific boundary (CFD) or the whole domain (for CFD-DEM). To control the average velocity of the flow, the code
+calculates a :math:`\beta` coefficient at each time step that is used to keep the average velocity in an acceptable
+threshold. If the chosen wall boundary on which a average velocity is imposed is an inlet, the average velocity
```
blaisb, line 7:
```suggestion
threshold. If the boundary on which an average velocity is imposed is an inlet, the average velocity
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
...
+targeted must be negative since the outward normal vector is used for the calculation. Thereby, if the chosen wall
+boundary is an outlet, average velocity must be positive. In other to keep the sign consistency, the targeted
+velocity should have an negative sign when used with CFD-DEM even if no normal vector is used since the
+whole domain is taken into account.
+
```
blaisb, line 12:
```suggestion
targeted must be negative since the outward normal vector is used for the calculation. If the 
boundary is an outlet, average velocity must be positive. In other to keep the sign consistency, the targeted
velocity should have an negative sign when used with CFD-DEM even if no normal vector is used since the
whole domain is taken into account.

```
These are boundaries, they are not wall (otherwise nothing would flow through them)

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
...
+The main controller of the average velocity is the following equation and is based on Wang [1]:
```
blaisb, line 13:
```suggestion
The main controller of the average velocity is the following equation and is based on approach of Wang [1]:
```

blaisb: et al? are there many authors to the Wang paper?

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
...
 
 .. warning::
 
-  This value should be negative if imposed on an the inlet boundary, and positive if imposed on an outlet boundary.
+  This value should be negative if imposed on an the inlet boundary (or CFD-DEM), and positive if imposed on an outlet boundary.
 
 .. note::
 
-  If you are simulating a problem with periodic boundary conditions at inlet and outlet walls, it is important to control the volumetric flow rate specifying the boundary id that corresponds to the inlet only (where the flow would have a negative value).
+  If you are simulating a problem with periodic boundary conditions at inlet and outlet walls, it is important to control the average velocity specifying the boundary id that corresponds to the inlet only (where the flow would have a negative value).
```
blaisb, line 45:
```suggestion
  If you are simulating a problem with periodic boundary conditions, it is important to control the average velocity by specifying the boundary id that corresponds to the inlet only (where the flow would have a negative value).
```

Would it work if. you specify the outlet only? Why the inlet? isnt it just one of the boundary only and that's it?

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
...
 
-* The ``boundary id`` parameter is the wall boundary where the flow rate intended is controlled (inlet or outlet).
+* The ``boundary id`` parameter is the wall boundary where the average velocity intended is controlled (for CFD solvers only).
```
blaisb, line 47:
```suggestion
* The ``boundary id`` parameter is the boundary where the average velocity intended is controlled (for CFD solvers only).
```
Ain't a wall though :)

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,61 @@
...
 
 * The ``flow direction`` parameter sets the absolute direction. It should be ``0`` if the flow is in the :math:`x` direction, ``1`` in the :math:`y` direction and ``2`` in the :math:`z` direction.
 
-* The ``initial beta`` parameter sets an initial :math:`\beta` coefficient that may speed up the convergence of the flow rate targeted. Some tests should be done to find a good initial :math:`\beta` coefficient.
+* The ``initial beta`` parameter sets an initial :math:`\beta` coefficient that may speed up the convergence of the average velocity target. Some tests should be done to find a good initial :math:`\beta` coefficient.
 
 .. tip:: 
 
   A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will greatly speed up the convergence.
```
blaisb, line 55:
```suggestion
  A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will reduce the time it takes to reach the target velocity.
```

examples/incompressible-flow/3d-periodic-hills/periodic-hills.prm
```diff
@@ -58,12 +58,12 @@ end
...
 subsection flow control
-  set enable               = true
-  set boundary id          = 0
-  set volumetric flow rate = -9.1575 # bulk velocity = -1
-  set flow direction       = 0
-  set initial beta         = 7.66
-  set verbosity            = verbose
+  set enable                  = true
+  set boundary id             = 0
+  set average velocity target = -1
```
blaisb, line 63: targetted average velocity might read better? I am not sure. The wording you chose is also nice.

include/core/parameters.h
```diff
@@ -1110,19 +1110,16 @@ namespace Parameters
   };
 
   /**
-   * @brief FlowControl - Set volumetric flow rate on a boundary id
-   * toward the normal direction of this wall.
+   * @brief FlowControl - Set volumetric flow rate or average velocity on a
+   * boundary id toward the normal direction of this wall.
```
blaisb, line 1114:
```suggestion
   * boundary id toward the normal direction of this boundary.
```

blaisb (changes requested):
Some initial feedback. I'll finish reviewing before boarding on Monday I think

acdaigneault:
> Why the [WIP] label? Is it ready for review or not? If you want to put it as WIP, it's better to use the github label WIP instead of putting it in the title.

I might want a small more thing. I'm thinking about adding a threshold factor parameter that if the new beta is between a certain range (let say 0.5%), the new beta = beta. It can be useful when reusing matrix. What do you think about it?

applications_tests/gls_navier_stokes_3d/poiseuille_restart_files/poiseuille_restart.flowcontrol
acdaigneault: Yes, it's -10.0126/3.12145 (flow rate/area)

doc/source/parameters/cfd/dynamic_flow_control.rst
acdaigneault: Fair enough :p

doc/source/parameters/cfd/dynamic_flow_control.rst
acdaigneault: This is his/her thesis

blaisb: great

blaisb:
> > Why the [WIP] label? Is it ready for review or not? If you want to put it as WIP, it's better to use the github label WIP instead of putting it in the title.
> 
> I might want a small more thing. I'm thinking about adding a threshold factor parameter that if the new beta is between a certain range (let say 0.5%), the new beta = beta. It can be useful when reusing matrix. What do you think about it?

Beta does not appear in the matrix, only in the RHS ( it is a source term), so it should not affect matrix reuse very much.

doc/source/parameters/cfd/dynamic_flow_control.rst
acdaigneault: I don't remember very well why there's this comment. I think it would be better to use inlet only, so remove any mention of outlet boundary. Also, maybe I can remove the - sign for the normal vector, it makes more sense with the average velocity.

examples/incompressible-flow/3d-periodic-hills/periodic-hills.prm
acdaigneault: it can be only `average velocity`, shorter seems better

blaisb:
@acdaigneault ping me when you want another review :).

acdaigneault:
> @acdaigneault ping me when you want another review :).

@blaisb, it's finished, just need to fix the application test in debug :)

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,56 @@
...
+* The ``initial beta`` parameter sets an initial :math:`\beta` coefficient that may speed up the convergence of the average velocity target. Some tests should be done to find a good initial :math:`\beta` coefficient.
 
 .. tip:: 
 
-  A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will greatly speed up the convergence.
+  A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will reduce the time it takes to reach the target velocity.
+
+* The ``alpha`` parameter is a relaxation coefficient that is used to control the convergence speed or stability. The higher the value, the faster the convergence. However, if the value is too high, the convergence may be unstable.
```
oguevremont, line 50:
```suggestion
* The ``alpha`` parameter is a relaxation coefficient that is used to control the convergence speed or stability. The higher the value, the faster the convergence. However, if the value is too high, there might be oscillations and the simulation could be unstable.
```

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,56 @@
...
+
+* The ``beta threshold`` parameter is the threshold of the previous calculated beta force where the new calculated beta doesn't need to be updated. This prevents the reassembly of the matrix because of the force term when reuse matrix is used for the non-linear solver
```
oguevremont, line 52: I don't really understand this explanation. The beta threshold is the old beta, or is it a tolerance in the sense that if abs(old_beta-new_beta)<tolerance, we reuse the same matrix?

include/core/parameters.h
```diff
@@ -1110,19 +1110,16 @@ namespace Parameters
   };
 
   /**
-   * @brief FlowControl - Set volumetric flow rate on a boundary id
-   * toward the normal direction of this wall.
+   * @brief FlowControl - Set volumetric flow rate or average velocity on a
```
oguevremont, line 1113: If we can't control volumetric flow rate directly I would change this

tests/solvers/flow_control_01.cc
```diff
@@ -34,42 +34,42 @@ void
...
-      flow_control.calculate_beta(flow_rate, dt, step_number);
+      flow_control.calculate_beta(average_velocity, dt, step_number);
       beta = flow_control.get_beta();
 
-      // Calculating a fake new flow_rate (fake pressure drop and force)
-      // Note : flow rate is negative because of the outward normal vector.
-      // Flow rate is decreased by 1 % (fake pressure drop)
+      // Calculating a fake new average velocity (fake pressure drop and force)
+      // Note : velocity is negative because of the outward normal vector.
+      // Average velocity is decreased by 1 % (fake pressure drop)
```
oguevremont, line 65:
```suggestion
      // Average velocity is decreased by 1 % (fake pressure drop)
```

include/solvers/flow_control.h
```diff
@@ -47,30 +47,86 @@ class FlowControl
...
+  inline double
+  main_flow_controller(const double &average_velocity_n, const double &dt)
+  {
+    double beta_n1 = beta_n + alpha *
+                                (average_velocity_0 - 2 * average_velocity_n +
+                                 average_velocity_1n) /
+                                dt;
+
+    // If desired average velocity is reached, new beta only maintains the force
+    // to keep the flow at the desired value. Is so, if calculated beta is
```
oguevremont, line 92:
```suggestion
    // to keep the flow at the desired value. If calculated beta is
```

include/solvers/flow_control.h
```diff
@@ -47,30 +47,86 @@ class FlowControl
...
+    // negative it is set to 0 to avoided +/- force.
+    if (average_velocity_0 * beta_n1 < 0 && no_force == false)
+      return 0.0;
+    else
+      return beta_n1;
+  }
+
+  /**
+   * @brief get_beta. This function gives the beta forceof the step time
```
oguevremont, line 101:
```suggestion
   * @brief get_beta. This function gives the beta force of the step time
```

include/solvers/flow_control.h
```diff
@@ -47,30 +47,86 @@ class FlowControl
...
 
+  /**
+   * @brief get_beta_particle. This function gives the beta force of the
+   * CFD time step scaled for particles.
+   */
+  inline Tensor<1, 3>
+  get_beta_particles(const double &fluid_density,
+                     const double &particle_density)
+  {
+    // The beta force for particle must have a tensor of dim = 3 since the body
```
oguevremont, line 117:
```suggestion
    // The beta force for particles must have a tensor of dim = 3 since the body
```

source/solvers/flow_control.cc
```diff
@@ -5,54 +5,54 @@
...
-  // If flow rate is over the desired value at time step 1 and beta applied at
-  // time step 2 decreased it under the value, "no_force" is disable.
-  // As the previous condition, if flow rate is below the value after the small
-  // beta applied at time step 2, it means the pressure drop is too big. Then,
-  // slowing down the flow rate with the pressure drop is ineffective.
-  // This early disabling prevents to set a potential beta to 0 which could
+  // Check if disabling no_force is needed (step 3):
+  // If average velocity is over the target value at time step 1 and beta
+  // force applied at time step 2 decreased it under the value, "no_force" is
+  // disable.
```
oguevremont, line 48:
```suggestion
  // disabled.
```

source/solvers/flow_control.cc
```diff
@@ -5,54 +5,54 @@
...
     }
 
-  // (Only at step time 3)
-  // If flow rate is over the desired value at time step 1 and beta applied at
-  // time step 2 decreased it under the value, "no_force" is disable.
-  // As the previous condition, if flow rate is below the value after the small
-  // beta applied at time step 2, it means the pressure drop is too big. Then,
-  // slowing down the flow rate with the pressure drop is ineffective.
-  // This early disabling prevents to set a potential beta to 0 which could
+  // Check if disabling no_force is needed (step 3):
```
oguevremont, line 45: I don't see the step 2, is it written explicitly?

oguevremont (commented):
Nice PR! I have some comments that could be clearer, but nothing major

doc/source/parameters/cfd/dynamic_flow_control.rst
acdaigneault: `beta_threshold` is a threshold on the `beta_n` value (for instance 0.01 is 1%). The way it works is that `beta_n+1` is calculated as usual but if `abs(beta_n - beta_n+1) < abs(beta_n * beta_threshold)`, it changes the value of `beta_n+1` to `beta_n`. That prevent to update the source term (the beta force is added to force in scratch data I think). It doesn't means the same matrix is going to be reuse with the option enable (this is independent of the non-linear solver parameters), but it could help to do so since we force the code to keep the same value. I will try to reformulate this explaination

acdaigneault, line 52:
```suggestion
* The ``beta threshold`` parameter is the threshold on beta calculated at the previous time step that prevents the new calculated beta force to be updated, it is kept as the previous beta value. This could avoid the reassembly of the matrix because of the updated force term when reuse matrix option is enable for the non-linear solver
```

include/core/parameters.h
acdaigneault, line 1113:
```suggestion
   * @brief FlowControl - Set average velocity on a
```

source/solvers/flow_control.cc
acdaigneault: I removed this part, it wasn't really relevant anymore since we have more parameters to help flow control.

doc/source/parameters/cfd/dynamic_flow_control.rst
acdaigneault: I reformulated the doc if you want to look at it :)

doc/source/parameters/cfd/dynamic_flow_control.rst
```diff
@@ -1,39 +1,55 @@
...
+  A good method to find a reasonable initial beta is to test two or three different initial beta parameters, write down the given flow rate at the first time step in the simulation and do a regression. The correlation is linear and giving a proper value will reduce the time it takes to reach the target velocity.
+
+* The ``alpha`` parameter is a relaxation coefficient that is used to control the convergence speed or stability. The higher the value, the faster the convergence. However, if the value is too high, there might be oscillations and the simulation could be unstable.
+
+* The ``beta threshold`` parameter is the threshold on beta calculated at the previous time step that prevents the new calculated beta force to be updated, it is kept as the previous beta value. This could avoid the reassembly of the matrix because of the updated force term when reuse matrix option is enable for the non-linear solver
```
oguevremont, line 51:
```suggestion
* The ``beta threshold`` parameter is the threshold on beta calculated at the previous time step that prevents the new calculated beta force to be updated. If ``abs(beta_n - beta_n+1) < abs(beta_n * beta_threshold)``, the previous beta value is kept. This could avoid the reassembly of the matrix because of the updated force term when reuse matrix option is enabled for the non-linear solver.
```

doc/source/parameters/cfd/dynamic_flow_control.rst
oguevremont: I understand better with your explanation, thanks. I think the inequality should be there as it really helps to understand (added as suggestion)

oguevremont (approved):

include/core/parameters.h
```diff
@@ -1110,19 +1110,16 @@ namespace Parameters
...
-    // Type of verbosity for the flow control
-    Verbosity verbosity;
-
     // Enable flow control
     bool enable_flow_control;
 
-    // Volumetric flow rate (L^3/t)
-    double flow_rate_0;
+    // Average velocity target (L/t)
+    double average_velocity_0;
```
blaisb, line 1122: Why not call the variable average_velocity_target instead?

include/solvers/postprocessing_cfd.h
```diff
@@ -274,6 +273,62 @@ calculate_flow_rate(const DoFHandler<dim> &    dof_handler,
...
+                           const Mapping<dim> &       mapping);
+
+/**
+ * @brief calculate_average_velocity. This function calculates the average velocity of
+ * the domain for the solver GLS VANS.
+ *
+ * Ū = ∫ε(v⋅γ)dΩ/∫εdΩ
+ * Where Ū is the average velocity, ε is the void fraction, v is the velocity
+ * of the cell, γ is the vector of the cell (flow direction) and dΩ is the
+ * volume of the cell.
```
blaisb, line 306: I am not sure most text editor will like these characters. I would suggest using latex notation instead it's quite common.

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1191,6 +1173,71 @@ CFDDEMSolver<dim>::postprocess_fd(bool first_iteration)
...
+        }
+
+      // Showing results
+      if (this->simulation_parameters.flow_control.verbosity ==
+            Parameters::Verbosity::verbose &&
+          this->simulation_control->get_step_number() > 0 &&
+          this->this_mpi_process == 0)
+        {
+          announce_string(this->pcout, "Flow control summary");
+          this->pcout << "Fluid space-average velocity: " << average_velocity
```
blaisb, line 1230:
```suggestion
          this->pcout << "Fluid average velocity: " << average_velocity
```

blaisb: or instantaneous average velocity

tests/solvers/flow_control_01.cc
```diff
@@ -34,42 +34,40 @@ void
...
 
   deallog << "*************************************************" << std::endl;
-  deallog << " Time step :         " << dt << std::endl;
-  deallog << " Initial beta :      " << flow_control_parameters.beta_0
-          << std::endl;
-  deallog << " Desired flow rate : " << flow_control_parameters.flow_rate_0
+  deallog << " Time step:         " << dt << std::endl;
+  deallog << " Initial beta:      " << flow_control_parameters.beta_0
           << std::endl;
+  deallog << " Desired average velocity: "
```
blaisb, line 53: Desired or target?

blaisb (changes requested):
Few last final changes

include/core/parameters.h
acdaigneault: It's (m_dot/A_c)^0 in the reference, but I changed the equation a lot, so I guess I could you average_velocity_target now

include/solvers/postprocessing_cfd.h
acdaigneault: Oh ok then!

applications_tests/gls_navier_stokes_3d/poiseuille3d-flow-control_gls_bdf1.output
```diff
@@ -6,99 +6,89 @@ Running on 1 MPI rank(s)...
...
-+------------------------------------------+
-|  Flow control summary                    |
-+------------------------------------------+
-Inlet area : 3.06147
-Flow rate : -10.0508
-Beta applied : 70
+---------------------
+Flow control summary
+---------------------
+Space-average velocity: -0
```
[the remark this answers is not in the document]
acdaigneault: It was very unclear if it's "space-average" or "space-averaged". I looked on Google and It seems space-average is often used, but that's true that space-averaged sounds better.

Also, I use space-averaged because there's also the average velocities concept which is the time-averaged velocity and I wanted to distinguish the two more explicitly.

tests/solvers/flow_control_01.cc
acdaigneault: It means the same to me

blaisb: It means the same, but I just want the vocabulary to be consistent throughout the code.

blaisb (approved):

PR #747, opened 2023-05-01
Fix singular cells bug and pressure scaling

oguevremont:
# Description of the problem

- Some cells were left without equations, so the system was singular when preconditioner absolute tolerance is 0
- When pressure and velocity scales were very different, the linear solver had trouble converging

# Description of the solution

- The cells without equation at the boundary were re-treated with the boundary condition, and treatment of overconstrained cells was fixed
- Assemblers for NS were changed to take into account a new parameter "pressure scaling factor", and after the linear solving the newton correction was rescaled (at pressure dofs)

# How Has This Been Tested?

- Concerned tests and their outputs were adapted
- A new test with pressure scaling != 1 is added

# Documentation

- The new parameter was added to the parameters documentation


source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1424,7 +1447,9 @@ GLSSharpNavierStokesSolver<dim>::output_field_hook(DataOut<dim> &data_out)
   data_out.add_data_vector(this->present_solution, *levelset_postprocessor);
   Vector<float> cell_cuts(this->triangulation->n_active_cells());
   Vector<float> cell_overconstrained(this->triangulation->n_active_cells());
-
+  Vector<float> cell_has_singular_dof(this->triangulation->n_active_cells());
```
Luckabarbeau, line 1450: I wouldn't keep this here since the code is not supposed to have a singular cell. This was only a tool to find the bug.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1433,28 +1458,48 @@ GLSSharpNavierStokesSolver<dim>::output_field_hook(DataOut<dim> &data_out)
...
+              for (unsigned int j = 0; j < local_dof_indices.size(); ++j)
+                {
+                  if (abs(this->system_matrix.el(local_dof_indices[j],
+                                                 local_dof_indices[j])) <=
+                        1e-16 * dr &&
+                      this->locally_owned_dofs.is_element(local_dof_indices[j]))
+                    {
+                      cell_has_singular_dof(i) = 1.0;
+                    }
+                }
```
Luckabarbeau, line 1479: see previous comments

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3077,26 +3248,38 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
...
-                              this->evaluation_point(global_index_overwrite) *
-                                sum_line;
+                              component_i == dim ?
+                                (sum_line * v_ib - this->evaluation_point(
+                                                     global_index_overwrite) *
+                                                     sum_line) /
+                                  this->pressure_scaling_factor :
+                                sum_line * v_ib - this->evaluation_point(
+                                                    global_index_overwrite) *
+                                                    sum_line;
```
Luckabarbeau, line 3280: add a comment to explain why the pressure dof need to be scaled in a dummy equation

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3179,6 +3367,17 @@ GLSSharpNavierStokesSolver<dim>::sharp_edge()
...
+  unsigned int n_dof = this->dof_handler.n_dofs();
+  for (unsigned int i = 0; i < n_dof; ++i)
+    {
+      if (abs(this->system_matrix.el(i, i)) <= 1e-16 * dr &&
+          this->locally_owned_dofs.is_element(i))
+        {
+          std::cout << "DOF index : " << i << " has a 0 on the diagonal"
+                    << std::endl;
+        }
+    }
```
Luckabarbeau, line 3380: remove this check this is not needed anymore

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -3303,25 +3502,21 @@ GLSSharpNavierStokesSolver<dim>::assemble_local_system_matrix(
...
-  // The id of the particle that cut the cell. Returns 0 if the cell is
-  // not cut.
-  unsigned int ib_particle_id;
-  std::tie(cell_is_cut, ib_particle_id, std::ignore) = cut_cells_map[cell];
+  bool cell_is_cut                                = false;
+  std::tie(cell_is_cut, std::ignore, std::ignore) = cut_cells_map[cell];
 
   bool cell_is_overconstrained;
-  std::tie(cell_is_overconstrained, std::ignore) =
+  std::tie(cell_is_overconstrained, std::ignore, std::ignore) =
```
Luckabarbeau, line 3514: update the comments with the check on cell is overconstrained.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -280,50 +286,67 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
         {
-          overconstrained_fluid_cell_map[cell] = {false, -1};
-
-          bool cell_is_cut;
-          bool cell_is_inside;
-          cell->get_dof_indices(local_dof_indices);
-          std::tie(cell_is_cut, std::ignore, std::ignore) = cut_cells_map[cell];
-          std::tie(cell_is_inside, std::ignore) = cells_inside_map[cell];
-          if (!cell_is_cut && !cell_is_inside)
+          overconstrained_fluid_cell_map[cell] = {false, -1, -1};
```
Luckabarbeau, line 298: Ideally, we would fill the map only if the cell is overconstrained to avoid huge maps of cells. The same modification could be applied on the cut cell map maybe we should keep this for an other PR only on this small correction.

source/solvers/navier_stokes_base.cc
```diff
@@ -2104,6 +2105,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+                {
+                  const unsigned int component_index =
+                    this->fe->system_to_component_index(j).first;
+                  if (is_locally_owned_dof(global_id) && component_index == dim)
+                    {
+                      this->newton_update(global_id) =
+                        this->newton_update(global_id) *
+                        pressure_scaling_factor;
+                    }
+                  rescaled_dofs_map[global_id] = true;
```
Luckabarbeau, line 2185: use unordered

Luckabarbeau (commented):
First batch of comments I will add a bit more tomorrow. Over all this will fix a lot of the issues we have and also add a useful pressure scaling factor. Although I am worried about the overhead cost of all the loops over the cells that have been added. have you noticed an increase in simulation time when you scale the pressure? If it is the case maybe we should simply keep the scaled value for the pressure and only rechange the value in the post-processor. Otherwise, this is ok if the cost is negligible.

acdaigneault (commented):
Nice PR, always cool to review something not directly related to my work :) This is only my code review, I will try to run it and test it tomorrow with the vans solver

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -580,13 +580,20 @@ Return a bool that describes  if a cell contains a specific point
   /*
    * This map uses the cell as the key, and stores the following information:
    * if the cell is overconstrained (bool), what particle overconstrains this
-   * cell (unsigned int). The id of the particle that overconstrains the cell
+   * cell (unsigned int), and the distance to the closest surface (double).
+   * The id of the particle that overconstrains the cell
    * is the id of the particle with the lowest particle index.
```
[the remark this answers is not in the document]
acdaigneault: I think it may not be relevent anymore according to a comment below

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -106,6 +106,12 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
   cells_inside_map.clear();
   overconstrained_fluid_cell_map.clear();
   vertices_cut.clear();
```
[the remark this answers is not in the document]
oguevremont: It wasn't used anymore, so I removed the vertices_cut variable. Good catch!

source/fem-dem/gls_sharp_navier_stokes.cc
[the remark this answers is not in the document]
oguevremont: I removed the initialization in the end, so we save a step. I looked it up and it turns out that when we call overconstrained_fluid_cell_map[cell_not_initialized] we get {false, 0, 0}, which is fine.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -280,50 +286,67 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
-          overconstrained_fluid_cell_map[cell] = {false, -1};
-
-          bool cell_is_cut;
-          bool cell_is_inside;
-          cell->get_dof_indices(local_dof_indices);
-          std::tie(cell_is_cut, std::ignore, std::ignore) = cut_cells_map[cell];
-          std::tie(cell_is_inside, std::ignore) = cells_inside_map[cell];
-          if (!cell_is_cut && !cell_is_inside)
+          overconstrained_fluid_cell_map[cell] = {false, -1, -1};
+          if (this->simulation_parameters.fem_parameters.velocity_order == 1)
```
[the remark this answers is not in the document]
oguevremont: I'll add comment for this. The explanation is that when the FE order > 1 overconstrained cells are impossible, since there is always at least a DOF inside that's isn't overconstrained (we only have to check when FE_order == 1)

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -280,50 +286,67 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
+                  if (number_of_vertices_in_cell == number_of_vertices_cut)
+                    {
+                      for (unsigned int p = 0; p < particles.size(); p++)
+                        {
+                          potential_particle_distance =
+                            particles[p].get_levelset(cell->barycenter(), cell);
+                          if (abs(potential_particle_distance) <
+                              abs(current_particle_distance))
+                            {
+                              current_particle_candidate = p;
```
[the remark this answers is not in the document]
oguevremont: I like your suggestion better

applications_tests/gls_sharp_navier_stokes_3d/almost_blocked_channels.prm
```diff
@@ -100,6 +100,6 @@ subsection non-linear solver
 end
 
 subsection linear solver
-  set verbosity = quiet
-end
-
+  set verbosity         = quiet
+  set relative residual = 1e-1
```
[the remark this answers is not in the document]
oguevremont: It was because the problem was too stiff, so instead I reduced the forcing term. With the current setup the solver decreases the residual by 6 orders of magnitude.

source/solvers/navier_stokes_base.cc
```diff
@@ -2104,6 +2105,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+{
+  return this->locally_owned_dofs[0].is_element(global_id);
+}
+template <>
+inline bool
+NavierStokesBase<3, TrilinosWrappers::MPI::BlockVector, std::vector<IndexSet>>::
+  is_locally_owned_dof(const unsigned int global_id)
+{
+  return this->locally_owned_dofs[0].is_element(global_id);
+}
```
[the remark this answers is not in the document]
oguevremont: Good point, I think you're right also. I changed it to this->dof_handler.locally_owned_dofs().is_element(global_id) instead.

oguevremont:
I tested the pressure scaling by setting the default value for the pressure scaling factor = 2, recompiling and running the tests. The results of the tests change a bit, but I examined them and nothing critical (only decimals).

oguevremont:
> First batch of comments I will add a bit more tomorrow. Over all this will fix a lot of the issues we have and also add a useful pressure scaling factor. Although I am worried about the overhead cost of all the loops over the cells that have been added. have you noticed an increase in simulation time when you scale the pressure? If it is the case maybe we should simply keep the scaled value for the pressure and only rechange the value in the post-processor. Otherwise, this is ok if the cost is negligible.

About the overhead cost, I tested a few times and the cost is low. In one simulation at 1M dofs, for example, the cost is 0.14% according to the timer.

applications_tests/gls_sharp_navier_stokes_3d/flow_around_cuthollowsphere.output
```diff
@@ -13,8 +13,8 @@ Steady iteration:        1/2
 +------------------------------------------+
 |  Force  summary particles                |
 +------------------------------------------+
-particle_ID  time    T_x      T_y       T_z       f_x        f_y        f_z      f_xv      f_yv      f_zv       f_xp       f_yp      f_zp    
-          0 1.0000 0.678984 3.112272 63.626985 198.021894 -39.624418 -0.610866 26.339639 -4.238040 -0.335928 171.682254 -35.386378 -0.274938 
+particle_ID  time    T_x      T_y       T_z        f_x        f_y      f_z      f_xv      f_yv     f_zv      f_xp      f_yp      f_zp
+          0 1.0000 2.871845 9.793226 117.105376 321.947926 11.919979 4.093938 29.607130 0.480646 0.013561 292.340796 11.439332 4.080377
```
blaisb, line 17: Why have the rsults changed so much here???

applications_tests/gls_sharp_navier_stokes_3d/flow_around_ellipsoid_rectangle.output
```diff
@@ -13,19 +13,19 @@ Steady iteration:        1/2
 +------------------------------------------+
 |  Force  summary particles                |
 +------------------------------------------+
-particle_ID  time     T_x       T_y       T_z       f_x       f_y       f_z      f_xv      f_yv      f_zv      f_xp      f_yp      f_zp    
-          0 1.0000 -0.333785 -2.750365  3.626807 21.225583 -4.583983 -2.762347  8.723989 -0.466253 -0.004813 12.501594 -4.117730 -2.757534 
-          1 1.0000  2.874939 -7.255216 10.661657 40.242897 -3.090175  0.381153 18.752948 -1.154742  1.809280 21.489949 -1.935433 -1.428127 
+particle_ID  time     T_x       T_y       T_z       f_x       f_y       f_z      f_xv      f_yv      f_zv     f_xp      f_yp      f_zp
+          0 1.0000 -0.545861 -3.054774  2.932368 18.471709 -4.347702 -2.587468  6.098268 -0.167732 0.034750 12.373441 -4.179970 -2.622218
+          1 1.0000  2.078090 -8.337907 14.347876 37.381657 -2.239937  0.802546 15.761853 -0.692786 1.759641 21.619804 -1.547151 -0.957095
```
blaisb, line 18: Why have the results changed so much here?

applications_tests/gls_sharp_navier_stokes_3d/sphere_mms_pressure_scaling.output
```diff
@@ -0,0 +1,35 @@
...
+   Number of active cells:       64
+   Number of degrees of freedom: 500
+   Volume of triangulation:      8
+
+*****************************
+Steady iteration:        1/3
+*****************************
+div u : 0
+error at the weak BC : 0
+L2 error velocity : 0 L2 error pressure: 0
```
blaisb, line 11: Why is the error here zero? I would remove this iteration or this mesh if it is inadequate.

doc/source/parameters/cfd/stabilization.rst
```diff
@@ -22,3 +25,6 @@ There are three choices of stabilization strategy:
 
 * ``stabilization=grad_div`` assembles a grad-div penalization term in the momentum equation to ensure mass conservation. This is not a stabilization method per-say and should not be used with elements that are not LBB stable. This stabilization should only be used with the grad-div Navier-Stokes solver (``gd_navier_stokes_2d`` and ``gd_navier_stokes_3d``)
 
+The ``pressure scaling factor`` parameter is used as a multiplier for the pressure that is used in the equations to be solved; the inverse of the factor is applied to the pressure after solving, so the output values should be properly scaled. It helps the convergence for cases where pressure and velocity have very different scales, which would normally result in a very stiff system of equations.
```
blaisb, line 28:
```suggestion
The ``pressure scaling factor`` parameter is used as a multiplier for the pressure in the momentum equation; the inverse of the factor is applied to the pressure after solving. It helps the convergence of the linear solver  by decreasing the condition number for cases where pressure and velocity have very different scales.
```

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -580,20 +580,19 @@ Return a bool that describes  if a cell contains a specific point
...
-           std::tuple<bool, int>>
+           std::tuple<bool, unsigned int, double>>
     overconstrained_fluid_cell_map;
 
   /*
-   * This map is used to keep information about the vertices that are
-   * constrained. This information is used to convert cells that are not
-   * considered cut at first, but that have all their vertices constrained, into
-   * overconstrained cells.
+   * These vectors are used to keep track of the DOFs that are overconstraint
```
blaisb, line 592:
```suggestion
   * These vectors are used to keep track of the DOFs that are overconstrained
```

source/core/parameters.cc
```diff
@@ -373,6 +373,15 @@ namespace Parameters
...
         "Choices are <pspg_supg|gls|grad_div>.");
+
+      prm.declare_entry(
+        "pressure scaling factor",
+        "1",
+        Patterns::Double(),
+        "This parameter can be used to change the scale of pressure in the "
+        "equations to solve. When the velocity and pressure scales are very "
+        "different, using this parameter allows to reduce the condition number"
+        " and reach a solution.");
```
blaisb, line 384:
```suggestion
        "This parameter can be used to change the scale of pressure in the "
        "Navier-Stokes equations. When the velocity and pressure scales are very "
        "different, using this parameter allows to reduce the condition number"
        " and reach a solution.");
```

source/core/shape.cc
```diff
@@ -1321,34 +1321,35 @@ RBFShape<dim>::gradient(const Point<dim> &evaluation_point,
...
+    {
+      // Calculation of the dr/dx
+      relative_position = (evaluation_point - rotated_nodes_positions[node_id]);
+      distance          = (relative_position).norm();
+      normalized_distance = distance / support_radii[node_id];
+      if (distance > 1e-16)
+        dr_dx_derivative = relative_position / distance;
+      else
+        for (int d = 0; d < dim; d++)
+          // Can be proved by taking the limit (definition of a derivative)
```
blaisb, line 1341:
```suggestion
          // Can be proven by taking the limit (definition of a derivative)
```
Thisdoes not explain what is being done here. That's what the comment should give as an information.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -249,8 +254,8 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
               size_t id;
               for (unsigned int j = 0; j < local_dof_indices.size(); ++j)
                 {
-                  id               = local_dof_indices[j];
-                  vertices_cut[id] = {true, particle_id_which_cuts_this_cell};
+                  id                            = local_dof_indices[j];
+                  local_dof_over_constraint(id) = 1;
```
blaisb, line 258: can you give a better variable name over_constraint you mean over_constrained no?

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1433,24 +1459,28 @@ GLSSharpNavierStokesSolver<dim>::output_field_hook(DataOut<dim> &data_out)
...
-            cell_cuts(i) = -1;
-
-          bool cell_is_overconstrained;
-          std::tie(cell_is_overconstrained, particle_id) =
-            overconstrained_fluid_cell_map[cell];
-          if (cell_is_overconstrained)
-            cell_overconstrained(i) = particle_id;
-          else
-            cell_overconstrained(i) = -1;
+          if (cell->is_locally_owned())
```
blaisb, line 1465: Can you explain a bit what is being done here? More comments

source/solvers/navier_stokes_base.cc
```diff
@@ -74,6 +74,7 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
   , velocity_fem_degree(p_nsparam.fem_parameters.velocity_order)
   , pressure_fem_degree(p_nsparam.fem_parameters.pressure_order)
   , number_quadrature_points(p_nsparam.fem_parameters.velocity_order + 1)
+  , pressure_scaling_factor(p_nsparam.stabilization.pressure_scaling_factor)
```
blaisb, line 77: Why do you need to store this a s a member of the class? I mean the parameter is always available no?

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2085,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
 }
 
+template <>
+inline bool
+NavierStokesBase<2, TrilinosWrappers::MPI::Vector, IndexSet>::
+  is_locally_owned_dof(const unsigned int global_id)
+{
+  return this->locally_owned_dofs.is_element(global_id);
+}
+template <>
```
blaisb, line 2095:
```suggestion

template <>
```

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2085,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+inline bool
+NavierStokesBase<3, TrilinosWrappers::MPI::Vector, IndexSet>::
+  is_locally_owned_dof(const unsigned int global_id)
+{
+  return this->locally_owned_dofs.is_element(global_id);
+}
```
blaisb, line 2101:
```suggestion
}

```

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2085,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+template <>
+inline bool
+NavierStokesBase<2, TrilinosWrappers::MPI::BlockVector, std::vector<IndexSet>>::
+  is_locally_owned_dof(const unsigned int global_id)
+{
+  return this->dof_handler.locally_owned_dofs().is_element(global_id);
+}
+template <>
```
blaisb, line 2109:
```suggestion

template <>
```

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2085,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+  temp.reinit(this->locally_owned_dofs, this->mpi_communicator);
+
+  temp += rescaled_newton_update;
+  temp -= newton_update;
+  if (temp.norm_sqr() < 1e-8)
+    {
+      newton_update = rescaled_newton_update;
+      return;
+    }
+  TimerOutput::Scope t(this->computing_timer, "rescale_pressure");
```
blaisb, line 2138: I am not sure this is the best way to do this. I mean, the Newton update is calculated when we solve the linear system. So instead why not use a flag? Like when you calculate the newton update the flag "new_newton_update" member of the base is set to true or something. This equality cheecking is prohibitive I feel.

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2085,91 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+
+  const unsigned int                   dofs_per_cell = this->fe->dofs_per_cell;
+  std::vector<types::global_dof_index> local_dof_indices(dofs_per_cell);
+  // Map used to keep track of which DOFs have been looped over
+  std::unordered_map<unsigned int, bool> rescaled_dofs_map;
```
blaisb, line 2143: Why use a map with a bool? You can use a std::set
Ideally a std::unordered_set

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -364,5 +363,13 @@ NavierStokesScratchData<dim>::calculate_physical_properties()
     }
 }
 
+template <int dim>
+void
+NavierStokesScratchData<dim>::set_pressure_scaling_factor(
+  const double pressure_scaling_factor)
+{
+  this->pressure_scaling_factor = pressure_scaling_factor;
+}
```
blaisb, line 372: Why not pass the scaling factor to the constructor of the thing instead?

blaisb (changes requested):
Some changes. Especially the way you manage the fact that the newton update has been recalculated or not.

source/fem-dem/gls_sharp_navier_stokes.cc
oguevremont: You're right. Is dof_overconstrained fine?

applications_tests/gls_sharp_navier_stokes_3d/sphere_mms_pressure_scaling.output
oguevremont: That's because every cell was either cut or overconstrained, so the error calculation took into account 0 cell. I'll remove this first level.

source/solvers/navier_stokes_scratch_data.cc
oguevremont: I agree it would be cleaner this way. You mean the reinit of the constructor itself?

source/solvers/navier_stokes_base.cc
oguevremont: It was useful when I wasn't sure of when to call the rescaling function, but I agree that it's not the best design. I removed this part and only call rescale_pressure_dofs_in_newton_update() when necessary now.

blaisb (approved):
Sounds good to me. Ready to merge

applications_tests/gls_sharp_navier_stokes_3d/flow_around_cuthollowsphere.output
Luckabarbeau: There is a lot of overconstrained cells.

source/solvers/navier_stokes_base.cc
```diff
@@ -2084,6 +2082,84 @@ NavierStokesBase<dim, VectorType, DofsType>::write_checkpoint()
...
+}
+
+template <>
+inline bool
+NavierStokesBase<2, TrilinosWrappers::MPI::BlockVector, std::vector<IndexSet>>::
+  is_locally_owned_dof(const unsigned int global_id)
+{
+  return this->dof_handler.locally_owned_dofs().is_element(global_id);
+}
+
```
Luckabarbeau, line 2108: This function alone should be sufficient.

Luckabarbeau (commented):
A few small details to check and one last correction once it is done; it is good to go, in my opinion.

oguevremont:
> A few small details to check and one last correction once it is done; it is good to go, in my opinion.

I check the required time to treat overconstrained cells in generate_cut_cells_map(), and it's no more than 10%. I also put ifs around the part treating the overconstrained cells so that it is limited to when finite elements are of order 1.

Luckabarbeau (approved):

PR #748, opened 2023-05-09
Matrix free nonlinear poisson prototype

lpsaavedra:
# Description of the problem

This adds a matrix-based and a matrix-free nonlinear Poisson prototype with Dirichlet boundary conditions to the prototypes folder.

# Documentation

It does not apply to this PR since it is only a prototype.

# Comments

This is the first step towards matrix-free Lethe.


prototypes/matrix_based_non_linear_poisson/matrix_based_non_linear_poisson.cc
```diff
@@ -0,0 +1,1486 @@
...
+
+                for (unsigned int j = 0; j < dofs_per_cell; ++j)
+                  {
+                    const double         phi_j = fe_values.shape_value(j, q);
+                    const Tensor<1, dim> grad_phi_j =
+                      fe_values.shape_grad(j, q);
+
+                    cell_matrix(i, j) +=
+                      (grad_phi_i * grad_phi_j - phi_i * nonlinearity * phi_j) *
+                      dx;
```
blaisb, line 763: No need to do this here, but there's a hack you could do to only multiply grad_phi_i and grad_phi_j when component_i==component_J. No need to do it, just FYI.

prototypes/matrix_free_non_linear_poisson/matrix_free_non_linear_poisson.cc
```diff
@@ -0,0 +1,1221 @@
...
+  system_matrix.evaluate_newton_step(solution);
+
+  mg_transfer.interpolate_to_mg(dof_handler, mg_solution, solution);
+
+  using SmootherType =
+    PreconditionChebyshev<LevelMatrixType,
+                          LinearAlgebra::distributed::Vector<float>>;
+  mg::SmootherRelaxation<SmootherType,
+                         LinearAlgebra::distributed::Vector<float>>
+                                                       mg_smoother;
```
blaisb, line 764: Why use a different data type here?

blaisb:
I did not see anything that ticked my eye, but I am not sooo familiar with the MF framework. I'll need to catch up more on that :).

prototypes/matrix_free_non_linear_poisson/matrix_free_non_linear_poisson.cc
lpsaavedra: This is just a way of exploiting single precision. It has been shown that for these kinds of problems one can have a mixed-precision approach, where one can use double precision for the CG and single-precision for the multigrid as preconditioner. This preserves the accuracy and iteration number of a full double precision and it allows you take advantage of vectorization. I used this idea only for the Poisson prototype, the advection-diffusion one uses the same type for both the solver and the preconditioner...

blaisb: Interesting. Clearly this gives an acceleration both from a memory and FLOPS perspective. Cool stuff.

PR #749, opened 2023-05-10
Full optimized GLS stabilization with LSIC term

blaisb:
# Description of the problem

- The current GLS stabilization was a bit innefficient
- It was also missing the LSIC term

# Description of the solution

- Used the same optimized mechanism as the PSPG-SUPG assembly
- Added the LSIC term

# How Has This Been Tested?

- Tested on the unit test (whose results have chanegd slightly). Using the examples (Taylor-Couette, etc.) I have also tested the implementation at very high-order. Everything works well.

# Documentation

- This will need to be added to the theory guide eventually (over the course of the Summer).

# Comments
- Will need to check if this actually does something different on canonical turbulence benchmark
-


lpsaavedra (approved):
Nice. I think this term should only change things if we have very high Reynolds numbers, so let's check it during the summer with a turbulent benchmark :) I can quickly add it to the documentation in this PR if you are okay with it.

PR #750, opened 2023-05-11
Compartmentalizing

ghazalemir:
# Description of the problem

Compartmental modeling (CM) offers a streamlined and computationally efficient alternative to computational fluid dynamics (CFD) by utilizing a network of interconnected zones known as compartments. These compartments are considered to be ideally mixed, simplifying the modeling process.

In compartmental models, key decisions must be made regarding the number of compartments, their respective volumes, the connections between them, and the flow rates traversing these connections. By carefully designing and configuring these parameters, compartmental models can effectively capture the dynamics of the system under investigation.

Compared to CFD, compartment modeling reduces the complexity and computational burden associated with simulating fluid flows. It offers a pragmatic solution that balances accuracy and efficiency, making it an attractive choice for various applications where a simplified representation is sufficient.

# Description of the solution

The code functions as follows:

- It organizes cells within a meshed geometry based on their electric field values, derived from a COMSOL simulation.
- Cells with electric field values falling within a defined threshold are grouped into compartments.
- If a compartment contains cells that are not physically connected and unable to exchange flux, the code subdivides the compartment into smaller compartments, ensuring all cells within each smaller compartment are connected. This process establishes the initial compartments based on the electric field.
- The code then iterates over the compartments. For each compartment, it sorts cells based on their velocity values and further aggregates cells falling within a specified threshold. It also verifies if the resulting compartments remain physically connected. If not, the code further breaks them into smaller compartments. This step leads to the creation of an overlaid pattern and the final sets of compartments.
- Overall, the code systematically categorizes cells based on their electric field and velocity values, progressively refining the compartments to ensure physical connectivity.

After establishing the compartments, our code proceeds to calculate several key parameters for each compartment. These include the volume of each compartment, the average value of the electric field within each compartment, the flow rate at the inlet and outlet, and the flux between compartments. These calculated values are essential components that we will utilize to solve the system of equations and determine the temperature at each compartment using our Python code.

# How Has This Been Tested?

There is an example (lethe/examples/rpt/compartmentalization) containing the electric field and velocity data at the cell centers for the given geometry.


source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+    cp_parameters.cp_param.cylinder_half_length);
+  triangulation.refine_global(cp_parameters.cp_param.initial_refinement);
+  std::ofstream output("triangulation.vtk");
+  grid_out.write_vtk(triangulation, output);
+  std::vector<double>     cell_index_output;
+  std::vector<Point<dim>> points;
+  for (const auto &cell : triangulation.active_cell_iterators())
+    {
+      // std::vector<double> center;
+      Point<3, double> center;
```
acdaigneault, line 50: Is it only working is 3d? Here you don't use dim. If you don't plan to use it in 2D, you don't need to use the dim template :)

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+  myfile.close();
+}
+
+template <int dim>
+auto
+Compartmentalization<dim>::electric_field_stored_in_vector()
+  -> std::vector<std::vector<double>>
+{
+  std::vector<double> electric_field;
+  std::ifstream       fin("input_values_emw");
```
acdaigneault, line 84: What do emw means?

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+  for (auto &point : points)
+    {
+      for (int j = 0; j < 3; j++)
+        {
+          myfile << point[j] << sep;
+        }
+      myfile << "\n";
+    }
+  myfile.close();
+}
```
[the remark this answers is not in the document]
acdaigneault: Actually, everything related to cell center is not really related to the cylinder generation, meaning it doesn't fit in this function "generate_cylindrical_grid()" Those steps should be in another function or the function should be renamed properly :)

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+      fill_matrix.push_back(electric_field[cell_index]);
+      primer_matrix_index_emw.push_back(fill_matrix);
+      fill_matrix.clear();
+    }
+  return primer_matrix_index_emw;
+}
+
+template <int dim>
+auto
+Compartmentalization<dim>::electric_field_stored_in_map()
```
acdaigneault, line 105: For me that more "read_electric_field"

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+    {
+      electric_field.push_back(element);
+    }
+
+  // Fill the container with the electromagnetic field
+  for (const auto &[cell_index, cell] : cells_and_indices)
+    {
+      primer_map_cell_emw.insert({cell, electric_field[cell_index]});
+    }
+  return primer_map_cell_emw;
```
acdaigneault, line 122: I don't really get why the cells_and_indices. You can have the index with the cell, so why not looping over the cell of the triangulation instead?

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+}
+
+template <int dim>
+auto
+Compartmentalization<dim>::read_velocity_vector()
```
acdaigneault, line 127: I see you need to read a few files. In order to make it cleaner, I suggestion that you implement something related to the output vector in rpt_utilities.h. There's currently std::vector<double>, std::vector<std::vector<double>> and std::vector<Point<dim>>. They have a specific names (read_counts, read detectors...), but they are kinda general, there names could be modified to something more general "read_values" etc. But it would be nice if you add other file reader that you need there :)

acdaigneault: Here I'm pretty sure read_points() will work with the velocity vector

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+          velocity_vector.push_back(velocity_components[0]);
+          velocity_components.erase(velocity_components.begin());
+        }
+      // Push {cell,[vector of velocity]}
+      // in cells_and_indices[i], "i" is the key which is cell index
+      // cells_and_indices[i] in the world of C++ map, means the element which
+      // its key is "i"
+      map_of_all_cells_and_velocity_vectors.insert(
+        {cells_and_indices[i], velocity_vector});
+      velocity_vector.clear();
```
acdaigneault, line 148: I think instead of using map with the cell_id, you can use vectors with the index = the cell_id

acdaigneault: Also, you can use Tensor<1, dim> for the velocity components so the map (or vector if you choose to change it) will be a map/vector of Tensor

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+    }
+  return map_of_all_cells_and_velocity_vectors;
+}
+
+template <int dim>
+auto
+Compartmentalization<dim>::read_velocity_magnitude()
```
acdaigneault, line 155: Same comments as I did for the electric_field_stored_in_map()

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+      electric_field.push_back(element);
+    }
+
+
+  for (const auto &[cell_index, cell] : cells_and_indices)
+    {
+      std::vector<double> fill_matrix;
+      fill_matrix.push_back(cell_index);
+      fill_matrix.push_back(electric_field[cell_index]);
+      primer_matrix_index_emw.push_back(fill_matrix);
```
acdaigneault, line 97: If you know the number of element, it's always better to initialize it with the value and index it after. Also, you can do primer_matrix_index_emw[i] = ({cell_index, electric_field[cell_index]}); it's shorter. 
If primer_matrix_index_emw is a vector of pair instead of a vector of vector, you can you make_pair :)

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,963 @@
...
+  std::vector<double> velocity_magnitude;
+  std::ifstream       fin("input_values_velocity_magnitude");
+  double              element;
+  while (fin >> element)
+    {
+      velocity_magnitude.push_back(element);
+    }
+
+  //  this for loop fill the container with the properties
+  //  (velocity or electromagnetic field)
```
acdaigneault, line 168: It's only velocity here, right?

lpsaavedra (commented):
Good work! This is my first batch of comments :)

include/rpt/compartmentalization.h
```diff
@@ -0,0 +1,222 @@
...
+  test();
+
+private:
+  /**
+   * @brief Read the electric field at each cell, derived from COMSOL,
+   * and write the pair of cell_index and electric field into a vector of a
+   * vector to use it for compartmentalization
+   */
+  auto
+  electric_field_stored_in_vector() -> std::vector<std::vector<double>>;
```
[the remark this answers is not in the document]
ghazalemir: Done

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+                            (neighbour_velocity[1] + main_velocity[1]) / 2);
+                          average_velocity_vector.push_back(
+                            (neighbour_velocity[2] + main_velocity[2]) / 2);
+
+                          // calculate dot product to see the direction of
+                          // velocity vector and its value
+                          This = norm_of_cell[0] * average_velocity_vector[0] +
+                                 norm_of_cell[1] * average_velocity_vector[1] +
+                                 norm_of_cell[2] * average_velocity_vector[2];
+                          average_velocity = abs(This);
```
acdaigneault, line 742: Why do the sign is not important?

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+              if (d->face(i)->at_boundary())
+                {
+                  if (d->face(i)->boundary_id() == 1)
+                    {
+                      for (unsigned int j = 0; j < 6; j++)
+                        {
+                          if (d->face(j)->boundary_id() == 0)
+                            {
+                              in_flow =
+                                in_flow + 0; // very close to the wall, noslip;
```
acdaigneault, line 823: Maybe you can skip this check and just explain each case in comment

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+          // loop over face
+          for (unsigned int i = 0; i < 6; i++)
+            {
+              if (d->face(i)->at_boundary())
+                {
+                  if (d->face(i)->boundary_id() == 1)
+                    {
+                      for (unsigned int j = 0; j < 6; j++)
+                        {
+                          if (d->face(j)->boundary_id() == 0)
```
acdaigneault, line 820: Maybe before the big loop you could declare inlet = 1, outlet = 2 and wall = 0

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+                        }
+                    }
+                  if (d->face(i)->boundary_id() == 2)
+                    {
+                      for (auto &s : map_of_all_cells_and_velocity_vectors)
+                        {
+                          if (s.first == d)
+                            {
+                              velocity_vector_outlet = s.second;
+                            }
```
acdaigneault, line 840: What is d, what is s x)

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+    }
+}
+
+template <int dim>
+void
+Compartmentalization<dim>::inlet_outlet_flux()
+{
+  std::vector<double> velocity_vector_outlet;
+  std::vector<double> velocity_vector_inlet;
+  std::vector<double> velocity_vector_at_boundary;
```
acdaigneault, line 792: It would be better to use dealii tensor, it also have the norm() function

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+    }
+  filling_matrix.push_back(0);
+  // push it to matrix_of_flux
+  matrix_of_flux.push_back(filling_matrix);
+  filling_matrix.clear();
+
+  std::string combination;
+  std::string outlet_combination;
+  // loop over flux
+  for (unsigned int i = 1; i < overlaid_set.size() + 1; i++)
```
acdaigneault, line 884:
```suggestion
  for (unsigned int i = 1; i <= overlaid_set.size(); i++)
```

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+                         "overlaid_map",
+                         time,
+                         step_number,
+                         1,
+                         mpi_communicator);
+}
+
+template <int dim>
+void
+Compartmentalization<dim>::test()
```
acdaigneault, line 953: Maybe a solve function would sound better :p

acdaigneault (commented):
Great work! My review is finished, I run the example and it works. I highly recommend to review the code with and modify names of some variables. Lots of variables have generic names and it's hard to know what it is, even if it's temporary variables or iteration variables since there's a lot of data containers and usage of pair.first etc. which makes it hard to know what it is. Congrats for this work again :)

source/rpt/compartmentalization.cc
```diff
@@ -0,0 +1,968 @@
...
+  for (const auto &pair : final_set)
+    {
+      for (typename Triangulation<dim>::active_cell_iterator d : pair.second)
+        {
+          compartments_electric_field[d->active_cell_index()] =
+            compartment_final_id;
+        }
+      compartment_final_id++;
+    }
+  write_file_compartments_first_field(compartments_electric_field, 1.0, 1);
```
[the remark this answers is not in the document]
acdaigneault: It applies for every writing of files

blaisb:
What is the status on this branch @ghazalemir ? Please don't leave PR dangling like that for months

PR #751, opened 2023-05-12
VOF - Implement filtered phase in surface tension force calculation

AmishgaAlphonius:
# Description of the problem
After implementing the filtered phase for the calculation of physical properties in the VOF solver, a consistency issue was noticed in the calculation of the surface tension force (STF) as the filtered phase was not considered. 

# Description of the solution

In the STF term of the continuum equation, the phase fraction gradient is used to calculate the normal (in the curvature expression) and to estimate the Dirac delta see (static bubble example). The phase fraction gradient is now calculated with the filtered phase. Modifications were made to both the NavierStokesScratchData class and HeatTransferScratchData class.

# How Has This Been Tested?

- [x] gls_static_droplet_surface_tension_force.output was updated
- [x] Rising bubble and static bubble examples (comparing results)

# Documentation

A few minor corrections (typos) were made in the rising bubble and the static bubble example files.

- [x] rising-bubble.rst
- [x] static-bubble.rst


doc/source/examples/multiphysics/rising-bubble/rising-bubble.rst
```diff
@@ -39,8 +39,8 @@ The following schematic describes the geometry and dimensions of the simulation
     :width: 400
 
 .. note:: 
-    All the four boundary conditions are ``noslip``, and an external 
-    gravity field of :math:`-0.98` is applied in the y direction.
+    On the upper and bottom walls ``slip`` boundary conditions are applied, and on side walls the boundary conditions are ``noslip``.
+    Also, an external gravity field of :math:`-0.98` is applied in the y direction.
```
blaisb, line 43:
```suggestion
   An external gravity field of :math:`-0.98` is applied in the y direction.
```

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -135,6 +138,9 @@ NavierStokesScratchData<dim>::enable_vof(
     std::vector<std::vector<double>>(maximum_number_of_previous_solutions(),
                                      std::vector<double>(this->n_q_points));
   phase_gradient_values = std::vector<Tensor<1, dim>>(this->n_q_points);
+  // For STF calculation
```
blaisb, line 141: I wonder, why is there a enable_filtered_vof and an enable_vof now?
Anyway, All of the fields used are the filtered fields no? So why not in the enable_vof just use the filtered field (for phase and for gradient) and if we want to have no filter, we put filter to none?
What do you think @AmishgaAlphonius and @hepap ?
I feel this would make the code simpler. I feel the Navier-Stokes solver should just get the filtered fields and if we disable filtering, we do it everywhere. This way we will prevent confusions like the one we had.

blaisb (commented):
This is good work really well done.
The only thing I am not sure is this notion of the Navier-Stokes getting both filtered and not filtered phase.
Right now it's a bit weird:
- NS gets non-filtered phase, but filters it itself manually to get the values at the gauss points
- NS gets non-filtered phase gradients, but does not use them
- NS gets filtered phase, but uses it only for the gradient.

It seems a bit inconsistent to me
Two questions
1. Do we want to have the phase and filter it only when it has been interpolated at the gauss point, or would we rather use the nodal filtered phase value itself and use that instead. Maybe the first one is more accuracy? If so, we need to be sure this is consistent
2. We use the filtered phase for the gradient. That's good and that's what we want. However, maybe we should always gather both the filtered and the non-filtered phase or just gather the filtered one and then forget about the additional switch. it would make things cleaner and easier to read.

source/solvers/navier_stokes_scratch_data.cc
AmishgaAlphonius: To avoid further confusion regarding the "filtered" phase gradient, the old "filtered phase gradient" has been renamed with the term "projected".

hepap (commented):
Good work :) I think it's a good feature to add! 

I have just one interrogation (in the comments below)! Maybe you can explain it to me tomorrow :D

source/solvers/vof.cc
```diff
@@ -1328,158 +1328,161 @@ VolumeOfFluid<dim>::assemble_filtered_phase_fraction_gradient_matrix_and_rhs(
...
                   // Matrix assembly
                   for (unsigned int j = 0; j < dofs_per_cell; ++j)
                     {
-                      local_matrix_filtered_phase_fraction_gradient(i, j) +=
-                        (phi_filtered_phase_fraction_gradient[j] *
-                           phi_filtered_phase_fraction_gradient[i] +
+                      local_matrix_projected_phase_fraction_gradient(i, j) +=
+                        (phi_projected_phase_fraction_gradient[j] *
+                           phi_projected_phase_fraction_gradient[i] +
                          h * h * phase_fraction_gradient_filter_factor *
```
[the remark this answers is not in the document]
AmishgaAlphonius: Yes, that's a great suggestion, I'll do it tomorrow morning!

blaisb:
@AmishgaAlphonius can you rebase? Then I'll merge after CI :)

examples/multiphysics/rising-bubble/rising-bubble.prm
```diff
@@ -37,14 +37,14 @@ subsection VOF
...
   end
 
   subsection surface tension force
     set enable                                = true
     set surface tension coefficient           = 24.5
-    set phase fraction gradient filter factor = 4
-    set curvature filter factor               = 1
+    set phase fraction gradient diffusion factor = 4
+    set curvature diffusion factor               = 1
     set output auxiliary fields               = true
```
blaisb, line 48: Can you apply the prm indent here? :)

blaisb (approved):
Good work. A tiny detail on one prm and this is good for a merge IMO.

hepap (approved):

PR #752, opened 2023-05-12
Matrix-free advection-diffusion prototype

lpsaavedra:
# Description of the problem

This adds a matrix-based and a matrix-free advection-diffusion prototype with stabilization for different test cases.

# Comments

This is the second step towards matrix-free Lethe. The boundary conditions are a bit messy since there are different cases, however, we decided to leave it like this as it is only a prototype.


prototypes/matrix_based_advection_diffusion/matrix_based_advection_diffusion.cc
```diff
@@ -0,0 +1,1910 @@
...
+                          {
+                            h = std::pow(6 * cell->measure() / M_PI, 1. / 3.) /
+                                parameters.element_order;
+                          }
+
+                        double tau = std::pow(
+                          std::pow(1 / (parameters.peclet_number * h * h), 2) +
+                            std::pow(2 * advection_term_values[q].norm() / h,
+                                     2),
+                          -0.5);
```
blaisb, line 1166: not important here, but it's better to use utilities::fixed_int_power than std::pow for fixed integer power.
https://www.dealii.org/current/doxygen/deal.II/namespaceUtilities.html#a834c6b595ada6f8e73c78d9bbd8ec551

blaisb (approved):
Nothing to add. Quite clear. Great work @lpsaavedra . Onwards to MF Lethe.

PR #753, opened 2023-05-16
Vof update 3D dam break with obstacle example

AmishgaAlphonius:
# Description of the problem

After implementing the phase fraction filter, the 3D Dam Break With an Obstacle example needed to be updated.

# Description of the solution

Instead of using the sharpening method to sharpen the interface, the phase fraction "tanh" filter is now used.

# Documentation

The 3D Dam Break With an Obstacle example was updated.

- [x] doc/source/examples/multiphysics/3d-dam-break/3d-dam-break.rst
- [x] examples/multiphysics/3d-dam-break/3d-dam-break.prm


doc/source/examples/multiphysics/3d-dam-break/3d-dam-break.rst
```diff
@@ -194,7 +194,7 @@ In the ``source term`` subsection, we define the gravitational acceleration.
 VOF
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In the ``VOF`` subsection, we enable ``interface sharpening`` to reconstruct the interface and keep it sharp during the simulation. Here we use the default ``constant`` method for interface sharpening.
+In the ``VOF`` subsection, we select the ``tanh`` filter to filter the phase fraction and get a more defined interface. We set the value of beta to 10.
```
blaisb, line 197: the default value for beta is 20, so maybe we should leave it at that for this example.

examples/multiphysics/3d-dam-break/3d-dam-break.prm
```diff
@@ -31,12 +31,9 @@ end
...
 subsection VOF
-  subsection interface sharpening
-    set enable                  = true
-    set threshold               = 0.5
-    set interface sharpness     = 1.5
-    set frequency               = 10
-    set type                    = constant
+  subsection phase filtration
+    set type = tanh
+    set beta = 10
```
blaisb, line 36:
```suggestion
    set beta = 20
```

blaisb (changes requested):
Just a small comment on the beta value :). It won't change the results much so you don't need to rerun it :).

doc/source/examples/multiphysics/3d-dam-break/3d-dam-break.rst
AmishgaAlphonius: I tried running for beta=20 on Narval, but it failed .-.

blaisb:
@AmishgaAlphonius Let's wait for the next PR which fixes stabilization

hepap (approved):

blaisb (approved):

blaisb:
@AmishgaAlphonius I will merge this one after CI. Although there are changes coming up, they don't make the results at higher reynolds necessarily more physically better. it's just that the solver does not blow up any more ;)

PR #754, opened 2023-05-16
Add clang-format instructions in doc

lpsaavedra:
# Description of the problem

There was no description of how the clang-format should be applied when contributing to Lethe.

# Description of the solution

The instructions were added to the contributing tab of the documentation page.

# Documentation

- [x] source/contributing.rst

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ cmake-build-release/*
 
 # Doc
 doc/build/*
+contrib/utilities/programs/*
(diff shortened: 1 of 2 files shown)


tonielgeitani (approved):
That's good. You should also mention that the comment download_clang_format would install a filled called programs in the contrib directory that should be removed before committing but after indenting.

lpsaavedra:
> That's good. You should also mention that the comment download_clang_format would install a filled called programs in the contrib directory that should be removed before committing but after indenting.

That is a good point. I think a better solution would be to add the folder to the `.gitignore` file, so that nobody has to worry about that. I just updated that as well.

blaisb (approved):

PR #755, opened 2023-05-17
Hollow shell for MF GLS Navier-Stokes solver

lpsaavedra:
# Description of the problem

There was no matrix-free solver in Lethe. This PR adds a hollow shell for the first mf_gls_navier_stokes solver. 

# Description of the solution

The source and header files for the new solver were added, along with the corresponding 2d and 3d applications. The solver is a derived class of the NavierStokesBase class as all other Lethe solvers. 

# How Has This Been Tested?

- Two mms tests were added for the 2d and the 3d application. They only allow us to test the basic functionality of Lethe: set up mesh, dofs, iterate (without solving anything) , print basic information and output the results. 

# Future changes

- Following PRs will focus on the implementation of the matrix-free operator and the GMG preconditioner.

# Comments

- Some post-processing functions do not work for the deal.ii vector type yet, namely, the calculation of average velocities and smoothed fields; they need to be further generalized. 
- An additional problem was found with the solution transfer in the functions that refine the mesh when running the application in parallel; this is because the interpolation step must be accomplished differently depending on the vector type (see https://www.dealii.org/current/doxygen/deal.II/classparallel_1_1distributed_1_1SolutionTransfer.html). This still needs to be fixed.


blaisb (approved):
Very good. Once you fix the warnings I will merge :)

PR #756, opened 2023-05-18
Add high-order output for the boundaries

### LINKED ISSUE #727: Boundary mesh generated is not high-order
The boundary mesh files are not high-order, so we can't visualize it as a smooth mesh in Paraview. 
This is also the case when we use the subdivision parameter value as the highest order for FEM.
It would also be nice to have a parameter to only output one boundary mesh for a simulation since most of them don't have moving mesh. Actually, maybe it should be the default setting.

blaisb:
# Description of the problem

- High-order boundary mesh was not correctly exported when "output boundaries = true". This was an annoyance in some simulations as noted by @acdaigneault  (damn I want to find an alternative name like Chad, but google tells me I should use Stacy. So Stacy Audrey)

# Description of the solution

- Fixed it and this closes #727 

# How Has This Been Tested?

- Tested it on a dummy Poiseuille case. The results are convincing. Before and after pictures are below.
Screenshot_20230517_230150

Screenshot_20230517_230156

### DIFF
--- a/source/solvers/navier_stokes_base.cc
+++ b/source/solvers/navier_stokes_base.cc
@@ -884,7 +884,7 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_kelly()
   Vector<float> estimated_error_per_cell(tria.n_active_cells());
   const FEValuesExtractors::Vector velocity(0);
   const FEValuesExtractors::Scalar pressure(dim);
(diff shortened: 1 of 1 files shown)


acdaigneault (approved):
"Stacy Audrey" omg what? hahaha
It works on my side, also with the `cylinder` grid type.
Thank you :)

PR #757, opened 2023-05-18
Allow more particles sharp and add cache to composite shapes

oguevremont:
# Description of the problem

- When solving the flow around multiple particles with the Sharp solver, there was a too low limit on the number of particles (10)
- The alternative of using composite shapes to define the packing was costly because of the multiple boolean operations

# Description of the solution

- The limit on the number of particles was increased (10 000)
- The cache functionality was added for composite shapes

# How Has This Been Tested?

No test was added, but all of them still pass.

# Documentation

No change was required.

### DIFF
--- a/include/core/shape.h
+++ b/include/core/shape.h
@@ -856,7 +856,6 @@ public:
     std::map<unsigned int, double>         constituent_shapes_values,
     std::map<unsigned int, Tensor<1, dim>> constituent_shapes_gradients) const;
 
-
   /**
    * @brief
    * Clear the cache of the shape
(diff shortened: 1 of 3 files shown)


blaisb:
@oguevremont I don't understand why you need this feature? There is already a feature that loads N particles from a file (see https://lethe-cfd.github.io/lethe/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.html). Definitely we should write an example that uses this feature.
We can't instantiate 10k particles in the parameter file. It makes debugging the code impossible because generating the parameter file readout becomes a nightmare.
So this can't work. You either have to use the file mechanism (and if it does not work for your application, extend it) or find another solution...

PR #758, opened 2023-05-20
Fix lack of smoothing within the QCM method

blaisb:
# Description of the problem

- The QCM method did not allow for smoothing  of the particles
- Some intersection options/calculations did not have briefs and were not inlined

# Description of the solution

- Add smoothing to QCM, it leads to good results with very small smoothing
- Moved function to .h and inlined them

# How Has This Been Tested?

- All regular test results remain unchanged

### DIFF
--- a/include/fem-dem/gls_vans.h
+++ b/include/fem-dem/gls_vans.h
@@ -50,6 +50,62 @@
 using namespace dealii;
 
 /**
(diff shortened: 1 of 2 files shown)


tonielgeitani (approved):

voferreira (approved):
Good catch. All clear to me.

PR #759, opened 2023-05-23
Using cache with composite shapes

oguevremont:
# Description of the problem

- Using complex composite shapes was costly because the boolean operations had to be repeated for each evaluation.

# Description of the solution

- The cache functionality for shapes was implemented for composite shapes: now, boolean operations only have to be performed once for each evaluation point.

# How Has This Been Tested?

- All tests still run.

# Documentation

- Not applicable.


Luckabarbeau (approved):
Nothing to add on my side!  Good job

oguevremont:
I added the changes for the gradient calculation of rotated composite shapes. It some tests where this was used in the expected manner.

blaisb (approved):

PR #761, opened 2023-05-24
Fix refine mesh issues for deal.ii distributed vector

lpsaavedra:
# Description of the problem

There were two issues when refining the mesh and using the `LinearAlgebra::distributed::Vector<double>`. 1) The solution vectors needed to update ghosts before calling the `prepare_for_coarsening_and_refinement` function. 2) The temporary vector used before the interpolation by the solution transfer object needs to be a ghosted vector.

# Description of the solution

The update_ghost_values call was added before all `prepare_for_coarsening_and_refinement` calls. A new inline function `init_temporary_vector` was added to the NavierStokesBase class. This function reinits the temporary vector according to vector type to solve the second issue.

# How Has This Been Tested?

- The parallel output for the two existing tests of the mf_gls_navier_stokes applications in 2D and 3D was added.

# Comments

- This fixes an issue mentioned in #755.


blaisb (approved):
Great work and good catch

PR #763, opened 2023-05-25
Fix QCM for particles with PBC

acdaigneault:
# Description of the problem

- The current implementation of the quadrature-centered method has not working with particles at periodic boundaries. The QCM reference sphere was not able to take into account the particles in the cells at the periodic boundary.

# Description of the solution

- Mapping of cells having coinciding vertices (periodic vertices) was added to handle quadrature points and particles in periodic cells.
- Some refactor was done for the radius of the reference sphere in other to minimize the calculation when the value is user defined – independent of the current cell size. 

# How Has This Been Tested?

- The void fraction of fluid around a particle crossing a PCB was compared to the one around a particle in the domain. The particles are separated by half the length of the domain for a better comparison (meaning both halves of the domain should gives same results). The animation shows the results before and after modifications. Before the modifications, the void fraction values at quadrature points at PBC was not correct. Also when the particles crossed the PBC, the particles wasn't even in a reference sphere, so gives no volumetric contribution.

vf

- A application test has been added for CFD-DEM coupling 3D based on the previous case (animation), but capturing only a few time steps where the particle crossed the PBC. This test uses dynamic flow control as promised here : #746 

# Documentation

- The doc is not affected by this modification
- The default value of the bound void fraction was fixed in doc according to the default value in the parameter parsing
- Reference of Toni's paper about QCM was added in the QCM theory

# Comments

- I'm curious if there a better way to handle the cell neighbors instead of using the `vertices_cell_mapping` and `find_cells_around_cell`. I feel we could map the neighbor cells directly as done in DEM solver, but maybe this usage was done of a specific reason.
- I tried to change vertices to cells map to an ordered map, but since it was also used in resolved CFD-DEM, I didn't want to go there...


include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
   void
   read_dem();
 
+
+  /**
+   * @brief This function gives the periodic offset of the domain with is needed
```
tonielgeitani, line 153:
```suggestion
   * @brief This function gives the periodic offset of the domain which is needed
```

include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
...
+                        offset = periodic_face_center - face_center;
+
+                        return offset;
+                      }
+                  }
+              }
+          }
+      }
+
+    // A zero tensor in returned in case no cells are found on the periodic
```
tonielgeitani, line 201:
```suggestion
    // A zero tensor is returned in case no cells are found on the periodic
```

include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
...
+    // boundaries on this processor. This processor won't handle particle in
+    // cells at periodic boundaries, so I won't affect any computation.
```
tonielgeitani, line 203:
```suggestion
    // cells at periodic boundaries, so it won't affect any computation.
```

source/fem-dem/gls_vans.cc
```diff
@@ -62,15 +78,30 @@ GLSVANSSolver<dim>::setup_dofs()
...
             void_fraction_dof_handler,
             boundary_conditions.id[i_bc],
             boundary_conditions.periodic_id[i_bc],
-            boundary_conditions.periodic_direction[i_bc],
+            periodic_direction,
             void_fraction_constraints);
+
+          // Get periodic offset if void fraction method is qcm
+          if (this->cfd_dem_simulation_parameters.void_fraction->mode ==
+              Parameters::VoidFractionMode::qcm)
```
tonielgeitani, line 91: Why is this only needed for QCM, and not also the other void fraction schemes? Is it only because of the volume averaging sphere?

source/fem-dem/gls_vans.cc
```diff
@@ -533,97 +580,202 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
                       else
-                        // Volume of sphere based on R_s = h_omega
-                        reference_sphere_volume =
-                          M_PI *
-                          pow((2 *
-                               pow(active_neighbors[n]->measure(), 1. / dim)),
-                              dim) /
-                          (2 * dim);
+                        {
+                          // The radius is get from the volume of sphere based
```
tonielgeitani, line 629:
```suggestion
                          // The radius is obtained from the volume of sphere based
```

source/fem-dem/gls_vans.cc
```diff
@@ -533,97 +580,202 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
+                          R_sphere =
+                            0.5 *
+                            pow(2 * dim *
+                                  active_periodic_neighbors[n]->measure() /
+                                  M_PI,
+                                1. / dim);
+                        }
+                      else
+                        {
+                          // The radius is get from the volume of sphere based
```
tonielgeitani, line 707:
```suggestion
                          // The radius is obtained from the volume of sphere based
```

source/fem-dem/gls_vans.cc
```diff
@@ -663,56 +815,52 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
-                  M_PI * pow((2 * pow(cell->measure(), 1. / dim)), dim) /
-                  (2 * dim);
-            }
-          else
-            {
-              // Volume of sphere based on R_s = user defined
-              reference_sphere_volume =
-                M_PI * pow((qcm_sphere_diameter), dim) / (2 * dim);
+                {
+                  // The radius is get from the volume of sphere based
```
tonielgeitani, line 830:
```suggestion
                  // The radius is obtained from the volume of sphere based
```

tonielgeitani (approved):

source/fem-dem/gls_vans.cc
acdaigneault: Yes, kind of since it calculates the distance between the quadrature points on one side of the PBC and the particles at the other side of the PBC. Since PCM doesn't "care" about the particles in cell neighbors, no needs of the offset to correct the particle position. But I just realize that SPM also needs  modifications since it also use particle position mhm.

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -894,6 +889,11 @@ CFDDEMSolver<dim>::dem_contact_build(unsigned int counter)
       this->pcout << "DEM contact search at dem step " << counter << std::endl;
       contact_build_number++;
 
+      // Particles displacement if passing through a periodic boundary
+      periodic_boundaries_object.execute_particles_displacement(
```
voferreira, line 893: As we discussed, I would add the if here, outside the function scope. I feel like it is cleaner to call a function only if it makes sense to call it instead of deciding if it is used within itself. Maybe doing "if bc is periodic" would make it more readable, but if you have "periodic_boundaries_cells_information.empty()" in this scope you can use it here to prevent looping through BCs. If empty() is the choice here, I would comment on it just to make sure you are testing if there are periodic BCs.

include/core/lethe_grid_tools.h
```diff
@@ -48,6 +48,25 @@ namespace LetheGridTools
...
+   *
+   * @param dof_handler DofHandler of the triangulation on which to create the map
+   *
+   * @param vertices_cell_map The map of vertex ids at periodic boundary to the
+   * set of periodic neighbor cells
+   */
+
+  template <int dim>
+  void
+  vertices_cell_mapping_with_periodic_boundaries(
```
voferreira, line 64: Do you need this "with"? :P

include/core/lethe_grid_tools.h
```diff
@@ -37,7 +37,7 @@ namespace LetheGridTools
    *
    * @param dof_handler DofHandler of the triangulation on which to create the map
    *
-   * @param vertices_cell_map Object to which the data will be written
+   * @param vertices_cell_map The map of vertex ids to the set of neighbor cells
```
voferreira, line 40: I would name it "map container with vertex ids per cells in a set of neighbors"

include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
   void
   read_dem();
 
+
+  /**
+   * @brief This function gives the periodic offset of the domain which is needed
```
voferreira, line 153:
```suggestion
   * @brief This function returns the periodic offset of the domain which is needed
```

include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
...
+   * for the periodic boundary conditions using the QCM for void fraction.
+   *
+   * @param periodic_boundary_id The id of the periodic boundary 0
```
voferreira, line 156: You mean 0 is the default pbc? If so, it would be clearer if it was said "... boundary. Boundary 0 is used by default", or smth like this.

include/fem-dem/gls_vans.h
```diff
@@ -148,9 +148,65 @@ class GLSVANSSolver : public GLSNavierStokesSolver<dim>
...
+   */
+  inline Tensor<1, dim>
+  get_periodic_offset(unsigned int periodic_boundary_id) const
+  {
+    Tensor<1, dim> offset;
+
+    // Iterating over the active cells in the triangulation
+    for (const auto &cell : (*this->triangulation).active_cell_iterators())
+      {
+        if (cell->is_locally_owned() || cell->is_ghost())
```
voferreira, line 168: Is this if necessary?

include/fem-dem/gls_vans.h
voferreira: As we discussed, "offset" can be a bunch of things and from my perspective, the purpose of it is not very clear. If you could bring here the explanation you gave me on slack, it would be more than perfect.

voferreira (commented):
The first part of the review. I will refresh and keep up. I added some comments here and there but so far, so good.

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -12,7 +12,7 @@ In this subsection, all parameters required for the calculation of the void frac
       set l2 smoothing factor = 0
       set l2 lower bound = 0
       set l2 upper bound = 1
-      set bound void fraction = false
+      set bound void fraction = true
```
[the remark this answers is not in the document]
acdaigneault: I didn't change the default value in Lethe, I just changed the documentation so it fits with the default value in the code (true)

doc/source/theory/unresolved_cfd-dem/unresolved_cfd-dem.rst
```diff
@@ -164,7 +164,7 @@ where :math:`n_{sp}` is the number of pseudo-particles j belonging to particle i
...
+The Quadrature Centered Method (QCM) `[7] <https://doi.org/10.1021/acs.iecr.3c00172>`_  is an analytical method that decouples the averaging volume from the mesh cells. It constructs an averaging sphere centered at each quadrature point in a given cell, and it calculates the void fraction directly in the averaging volume at the quadrature point. Since the sphere-sphere (particle-averaging sphere) intersection is analytically easier to calculate than sphere-polyhedron (particle-mesh cell), this method is less expensive than other analytical methods as the intersection does not involve the calculation of trigonometric functions at each CFD time step. The advantage of this method is that the void fraction varies within a cell. Additionally, particles in neighboring cells can affect the void fraction of the current cell. This allows the method to be continuous in both space and time. This is advantageous, especially in solid-liquid systems where the term :math:`\rho_f \frac{\partial \epsilon_f}{\partial t}` of the continuity equation is very stiff and unstable, when there exist even small discontinuities in the void fraction, and where it explodes when :math:`\Delta t_{CFD} \to 0`.
```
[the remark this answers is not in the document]
acdaigneault: @tonielgeitani Do you agree with this rephrasing?

include/core/lethe_grid_tools.h
acdaigneault: For me that's not the same since it's the global vertex ids to the set of cells that contains this vertex.

include/fem-dem/gls_vans.h
acdaigneault: It's not really a default value, this is just the one that is chosen between 0 and 1, it would work with 1 also.

acdaigneault: Actually, it's quite tricky because there's PBC in DEM and CFD. The GLS VANS has no idea about anything related to DEM, so it uses periodic for CFD which are "id" and "periodic_id" and not "periodic boundary 0" and "periodic boundary 1" as with DEM. I used the same name as CFD for DEM at first, but the "id" what used for outlet and other boundaries and it was confusing. I will just remove any mention of pbc 0 in this part.

include/fem-dem/gls_vans.h
acdaigneault: It will crash if not locally owned or ghost

source/fem-dem/gls_vans.cc
```diff
@@ -663,56 +817,52 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
-              reference_sphere_volume =
-                M_PI * pow((qcm_sphere_diameter), dim) / (2 * dim);
+                {
+                  // The radius is obtained from the volume of sphere based
+                  // on R_s = h_omega
+                  double reference_sphere_volume =
+                    M_PI * pow((2 * pow(cell->measure(), 1. / dim)), dim) /
+                    (2 * dim);
+                  R_sphere = 0.5 * pow(2 * dim * reference_sphere_volume / M_PI,
+                                       1. / dim);
```
blaisb, line 838: Shouldnt it be that the radius is the diameter of that sphere instead of its radius?
It seems a bit pessimistic to define it using the radius? Maybe we can revisit this in a next PR

blaisb (changes requested):
Just a few comments. Very good work!

source/fem-dem/cfd_dem_coupling.cc
acdaigneault: Since the function is also used in DEM, I prefer to add the check for periodic boundary, also it's more straight forward.

source/fem-dem/gls_vans.cc
```diff
@@ -533,97 +582,202 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
-                          (2 * dim);
+                        {
+                          // The radius is obtained from the volume of sphere
+                          // based on R_s = h_omega
+                          double reference_sphere_volume =
+                            M_PI *
+                            pow((2 *
+                                 pow(active_neighbors[n]->measure(), 1. / dim)),
+                                dim) /
+                            (2 * dim);
```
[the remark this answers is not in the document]
acdaigneault: Yes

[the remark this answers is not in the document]
acdaigneault: Actually, can I do it for 1./dim? Because it won't compile with 1./dim, but with 1/dim. But it won't give the good value (int divided by a int). Am I right?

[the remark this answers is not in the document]
acdaigneault: I changed to fixed_power<dim> everything that was ^dim

source/fem-dem/gls_vans.cc
acdaigneault: You mean that the idea of this radius definition may be wrong or the modification is wrong?

voferreira (commented):
Another round. I am about to finish it.

source/fem-dem/gls_vans.cc
```diff
@@ -733,8 +889,86 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
+                    active_periodic_neighbors[m]);
+                  for (auto &particle : pic)
+                    {
+                      double distance            = 0;
+                      auto   particle_properties = particle.get_properties();
+                      const double r_particle =
+                        particle_properties[DEM::PropertiesIndex::dp] * 0.5;
+                      double single_particle_volume =
+                        M_PI * Utilities::fixed_power<dim>(r_particle * 2) /
+                        (2 * dim);
```
voferreira, line 945:
```suggestion
                        (2.0 * dim);
```

source/core/lethe_grid_tools.cc
```diff
@@ -42,6 +42,99 @@ LetheGridTools::vertices_cell_mapping(
...
+  const DoFHandler<dim> &dof_handler,
+  std::map<unsigned int,
+           std::set<typename DoFHandler<dim>::active_cell_iterator>>
+    &vertices_cell_map)
+{
+  vertices_cell_map.clear();
+  const auto &cell_iterator = dof_handler.active_cell_iterators();
+  const auto &triangulation = dof_handler.get_triangulation();
+
+  // A map of coinciding vertices labeled by an arbitrary element from them
```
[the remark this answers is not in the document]
acdaigneault: I used the same vocabulary as Deal.II 😬
https://www.dealii.org/current/doxygen/deal.II/group__manifold.html#ga83d12f8567a8a310620f2326942e4ce8

source/fem-dem/gls_vans.cc
```diff
@@ -733,8 +889,86 @@ GLSVANSSolver<dim>::quadrature_centered_sphere_method(bool load_balance_step)
...
+                  const auto pic = particle_handler.particles_in_cell(
+                    active_periodic_neighbors[m]);
+                  for (auto &particle : pic)
+                    {
+                      double distance            = 0;
+                      auto   particle_properties = particle.get_properties();
+                      const double r_particle =
+                        particle_properties[DEM::PropertiesIndex::dp] * 0.5;
+                      double single_particle_volume =
+                        M_PI * Utilities::fixed_power<dim>(r_particle * 2) /
```
voferreira, line 944:
```suggestion
                        M_PI * Utilities::fixed_power<dim>(r_particle * 2.0) /
```

voferreira (approved):
I think it is all clear. Nothing else to add. Good job, friend!

source/core/lethe_grid_tools.cc
```diff
@@ -42,6 +42,99 @@ LetheGridTools::vertices_cell_mapping(
...
+              std::set<typename DoFHandler<dim>::active_cell_iterator>
+                periodic_neighbor_cells;
+
+              // Check if vertex is at periodic boundary, should be a key if so
+              if (vertex_to_coinciding_vertex_group.find(v_index) !=
+                  vertex_to_coinciding_vertex_group.end())
+                {
+                  // Get the coinciding vertex key to the current vertex
+                  unsigned int coinciding_vertex_key =
+                    vertex_to_coinciding_vertex_group.at(v_index);
```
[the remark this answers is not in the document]
acdaigneault: Good catch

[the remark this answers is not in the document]
acdaigneault: I now use the iterator value instead of .at (.at will throw exception if the key is not in map)

source/fem-dem/gls_vans.cc
```diff
@@ -62,15 +78,32 @@ GLSVANSSolver<dim>::setup_dofs()
...
             boundary_conditions.id[i_bc],
             boundary_conditions.periodic_id[i_bc],
-            boundary_conditions.periodic_direction[i_bc],
+            periodic_direction,
             void_fraction_constraints);
+
+          // Get periodic offset if void fraction method is qcm or spm
+          if (this->cfd_dem_simulation_parameters.void_fraction->mode ==
+                Parameters::VoidFractionMode::qcm ||
+              this->cfd_dem_simulation_parameters.void_fraction->mode ==
```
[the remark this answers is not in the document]
acdaigneault: I changed it in the last commit

blaisb (approved):

PR #765, opened 2023-05-25
Improve naca

PierreLaurentinCS:
# Description of the problem

The lift and drag coefficients in the example were wrong because of an unwanted coefficient (rho=1.3 instead of rho=1 in the formulas)

# Description of the solution

- I removed the aforementioned coefficient and updated the doc

# Documentation

- If applicable, list the documentation files changed and their changes

File modified in the doc :

- 2d-naca0012-low-reynolds.rst


blaisb:
@PierreLaurentinCS I think something broke down in your PR. You should not have 50 modified files!

PierreLaurentinCS:
Opened a new branch to modify the NACA0012 example, this PR was a mistake and shouldn't be considered

PR #772, opened 2023-06-05
Disabling contacts of particles with advection

acdaigneault:
# Description of the problem

- Periodic nodes were not properly handled with the last implementation: the periodic nodes had information on the mobility status on the coincident nodes. This was spotted when there has particles at one side of the periodic cells and the other periodic ones were empty. Those cells where not mobile, but this is one of the criteria for mobile.

- Disabling contacts of particles was functional in CFD-DEM: cells with low granular temperature deactivated the contacts. However, it is possible for particles within a cell to have a low variance in velocity and a significant average velocity. This is frequent for CFD-DEM simulations where particles may be carried by the fluid by bulk.

# Description of the solution

- Periodic nodes:
   - Mapping periodic nodes was achieved with the constraints since we can identify the periodic nodes from it (so called identity constraints).
   - Some attempted as been done to apply the periodic constraints to the `mobility_at_nodes`, without any success. The main issue is that constraints force values at PB1 = values at PB2. In our case, we need to check of the prevailing value, so sometime it is PB2 = PB1, with was not compatible with straight constraints operations. "[...] it adds the DoFs of face_1 to the list of constrained DoFs in constraints and adds entries to constrain them to the corresponding values of the DoFs on face_2. " see Deal.II doc [here](https://www.dealii.org/current/doxygen/deal.II/namespaceDoFTools.html#a929249499b1e5624728d212e90a8e037)
   
- Disabling contacts with advection:
   - New mobility status has been added to the disabling contacts: advected and advected_active
   - The new algorithm doesn't detect when particles are advected, this is a user-defined parameter. It could be useful for DEM when particles are moving in bulk by gravity, but this is more practical for CFD-DEM. The first idea was to implement  advected particles auto-detection, but this results in more checks, also the average velocities threshold is not easly configurable.
   - The way to "apply advection" is that the average velocity and acceleration of the cells are calculated or updated everytime the mobile cells are integrated. The average values are calculated in the same loop than the integration and it allows to always have the last calculated velocity values. For advected cells, all particles of a cell are integrated with an cell-averaged velocity and acceleration * time step. This velocity is also updated with the a*dt values.

# How Has This Been Tested?

- WIP

# Documentation

- WIP

# Future changes

- WIP

# Comments

- More information in the code
- Waiting for comments for improvement


source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1246,22 +1266,53 @@ void
...
+                 << ",  "
+                 << "v_y: " << particle_properties[DEM::PropertiesIndex::v_y]
+                 << ",  "
+                 << "vz: " << particle_properties[DEM::PropertiesIndex::v_z];
+
+              announce_string(this->pcout, ss.str(), '-');
+            }
+        }
+      usleep(100);
+      MPI_Barrier(this->mpi_communicator);
```
acdaigneault, line 1315: We could use the print_xyz in function, or print in the same manner (list and not 1 header per particle)

source/dem/velocity_verlet_integrator.cc
```diff
@@ -155,92 +292,161 @@ VelocityVerletIntegrator<dim>::integrate(
...
+                          {
+                            return (point_nd_to_3d(particle.get_location()));
+                          }
+                      }();
+
+                      // Calculate acceleration * dt and add velocity value for
+                      // average acceleration computation
+                      Tensor<1, 3> a_dt =
+                        dt_g + particle_force * dt_mass_inverse;
+                      average_a_dt += a_dt;
```
voferreira, line 343: I prefer to use "acc" as a variable name in these situations and then calculate the average, for safety. But it is up to you.

voferreira (approved):
I think it is all clear. Good job!

acdaigneault:
This draft PR will be reopened when ready for final review and merge

PR #776, opened 2023-06-15
Fix Rayleigh critical time step equation

acdaigneault:
# Description of the problem

- The critical time step was implemented as:
$$\Delta t_c =\frac{\pi d_i}{2(0.8766 + 0.1631\nu)} \sqrt{\frac{2\rho(2+\nu)(1-\nu)}{E}}$$

Where the shear modulus $G$ definition is wrong.

# Description of the solution

- New implementation fixes the issue and is expressed as followed:
$$\Delta t_c =\frac{\pi d_i}{2(0.8766 + 0.1631\nu)} \sqrt{\frac{\rho}{G}} = \frac{\pi d_i}{2(0.8766 + 0.1631\nu)} \sqrt{\frac{2\rho(1+\nu)}{E}}$$

# How Has This Been Tested?

- All the Rayleigh time step outputs in test are changed.

### DIFF
--- a/applications_tests/dem_2d/circle_restart.output
+++ b/applications_tests/dem_2d/circle_restart.output
@@ -2,7 +2,7 @@
 *********************
 Running on 1 rank(s)
 *********************
-DEM time-step is 2.32182% of Rayleigh time step
+DEM time-step is 2.58386% of Rayleigh time step
 Reading triangulation
 
 Finished reading triangulation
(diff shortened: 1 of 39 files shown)


OGaboriault (approved):
Good catch

blaisb (approved):
Good catch like always @acdaigneault

PR #780, opened 2023-06-19
Make workflow verbose on failure

blaisb:
# Description of the problem

- The current CI is verbose for all unit test. This makes it difficult to identify why it crashed

# Description of the solution

- Fix the ctest command to only output on failure.

### DIFF
--- a/.github/workflows/main_debug.yml
+++ b/.github/workflows/main_debug.yml
@@ -64,7 +64,7 @@ jobs:
           # Print the tests to be executed
           ctest -N --exclude-regex ${{ env.MULTI_CORE_TESTS_REGEX }}
           # Run in parallel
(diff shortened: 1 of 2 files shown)


acdaigneault (commented):
I tried the `--output-on-failure` command on my computer, it can be quite verbose, I'm not sure it helps that much. But I feel it could be relevant when it fails on CI but not on personal computer.
image

blaisb:
> I tried the `--output-on-failure` command on my computer, it can be quite verbose, I'm not sure it helps that much. But I feel it could be relevant when it fails on CI but not on personal computer. image

Well the current Verbose would output the same information (which is what is used on the CI, -V)
The only different is that now it would be only verbose when there is an error instead of being always verbose. (You can see if you look at the CI right now for example)

acdaigneault (approved):

PR #784, opened 2023-06-27
Superquadric shapes

oguevremont:
# Description of the problem

- Superquadric shapes were not implemented.
- Superquadric shapes have a surface function that is distorted (gradient isn't systematically of norm 1), which means it can't be used directly to find the gradient and closest surface points.

# Description of the solution

- An iterative way to find the closest point, and subsequently value and gradient of the distance function, is added.

# How Has This Been Tested?

- A few new lines were added in core/shape_primitives.cc.
- Visual inspections with different parameters were made.

# Documentation

- The new shape was added to the documentation.

# Future changes

- For now, the gradient and values found are a bit wrong: gradient is not perpendicular to the surface by construction, so the values can be a bit overestimated. For the purpose of sharp immersed boundary solver this will not be a problem, but a modified objective function to find the closest point could be used to find the true closest points (and so gradient and value). This will be the subject of another PR.

image

### DIFF
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -211,7 +211,7 @@ The following parameter and subsection are all inside the subsection ``particle 
 
     * Death Star: *sphere radius*, *hole radius*, *distance between centers*; the effective radius is the *sphere radius*;
 
-    * Superquadric: *x half length* (or :math:`a`), *y half length* (or :math:`b`), *z half length* (or :math:`c`), *x exponent* (or :math:`r`), *y exponent* (or :math:`s`), *z exponent* (or :math:`t`); the effective radius is the Euclidian norm of the half lengths. The exponents represent the blockiness in each direction. The surface is implicitly described by :math:`|\frac{x}{a}|^r + |\frac{y}{b}|^s + |\frac{z}{c}|^t - 1 = 0`;
+    * Superquadric: *x half length* (or :math:`a`), *y half length* (or :math:`b`), *z half length* (or :math:`c`), *x exponent* (or :math:`r`), *y exponent* (or :math:`s`), *z exponent* (or :math:`t`); the effective radius is the Euclidian norm of the half lengths. The exponents represent the blockiness in each direction. The surface is implicitly described by :math:`\left|\frac{x}{a}\right|^r + \left|\frac{y}{b}\right|^s + \left|\frac{z}{c}\right|^t - 1 = 0`;
 
     * Composite: *file name*.
    
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -164,7 +164,7 @@ The following properties are used if the particle impact one of the boundaries o
 .. warning::
     Currently this feature works only for spherical particles.
 
-* The ``particles file`` The file from which the particles are defined. Each line corresponds to a particle and all the relevant variables. The file must contain the following information for each particle (the header must be defined accordingly): type shape_argument_0 shape_argument_1 shape_argument_2 p_x p_y p_z v_x v_y v_z omega_x omega_y omega_z orientation_x orientation_y orientation_z density inertia pressure_x pressure_y pressure_z youngs_modulus restitution_coefficient friction_coefficient poisson_ratio rolling_friction_coefficient. The particle type is defined by the shape index. The shape indices are as follows: sphere=0, hyper rectangle=1, ellipsoid=2, torus=3, cone=4, cylinder=5, cylindrical tube=6, cylindrical helix=7, cut hollow sphere=8, death star=9. Currently, the composite, the RBF, and the OpenCascade shapes cannot be loaded from a file.
+* The ``particles file`` The file from which the particles are defined. Each line corresponds to a particle and all the relevant variables. The file must contain the following information for each particle (the header must be defined accordingly): type shape_argument_0 shape_argument_1 shape_argument_2 p_x p_y p_z v_x v_y v_z omega_x omega_y omega_z orientation_x orientation_y orientation_z density inertia pressure_x pressure_y pressure_z youngs_modulus restitution_coefficient friction_coefficient poisson_ratio rolling_friction_coefficient. The particle type is defined by the shape index. The shape indices are as follows: sphere=0, hyper rectangle=1, ellipsoid=2, torus=3, cone=4, cylinder=5, cylindrical tube=6, cylindrical helix=7, cut hollow sphere=8, death star=9, superquadric=10. Currently, the composite, the RBF, and the OpenCascade shapes cannot be loaded from a file.
 
 The following parameter and subsection are all inside the subsection ``particle info 0`` and have to be redefined for all particles separately.
 
@@ -188,7 +188,7 @@ The following parameter and subsection are all inside the subsection ``particle 
 
 * The ``pressure location`` parameter is used to define the X, Y, and Z coordinate offsets of the pressure reference point relative to the center of the particle. These parameters are used when the ``assemble Navier-Stokes inside particles`` parameter is set to ``true`` to define the pressure reference point.
 
-* The ``type`` parameter is used to define the geometry type of the particle. The alternatives in 2D are: ``sphere``, ``ellipsoid``, ``hyper rectangle``. In 3D, in addition to the previous shapes, alternatives include: ``cone``, ``death star``, ``cut hollow sphere``, ``torus``, ``cylinder``, ``cylindrical tube``, ``cylindrical helix``, ``composite``, ``rbf``, ``opencascade``. An ``rbf`` geometry is a flexible object described by a weighted sum of radial basis functions. The RBF data of an object can be generated from an STL file using a `bitpit <https://github.com/optimad/bitpit>`_-based script, namely example `RBF_example_00001 <https://github.com/optimad/bitpit/blob/master/examples/RBF_example_00001.cpp>`_.
+* The ``type`` parameter is used to define the geometry type of the particle. The alternatives in 2D are: ``sphere``, ``ellipsoid``, ``hyper rectangle``. In 3D, in addition to the previous shapes, alternatives include: ``cone``, ``death star``, ``superquadric``, ``cut hollow sphere``, ``torus``, ``cylinder``, ``cylindrical tube``, ``cylindrical helix``, ``composite``, ``rbf``, ``opencascade``. An ``rbf`` geometry is a flexible object described by a weighted sum of radial basis functions. The RBF data of an object can be generated from an STL file using a `bitpit <https://github.com/optimad/bitpit>`_-based script, namely example `RBF_example_00001 <https://github.com/optimad/bitpit/blob/master/examples/RBF_example_00001.cpp>`_.
 
 * The ``shape arguments`` parameter is used to define the parameters of the shape in the form of a list separated by ``;``. The required arguments and the effective radius, used for near-particle refinement, are:
     * Sphere: *radius*; the effective radius is the *radius*;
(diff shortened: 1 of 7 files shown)


doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
```diff
@@ -209,7 +209,9 @@ The following parameter and subsection are all inside the subsection ``particle
 
     * Cut Hollow Sphere: *radius*, *cut height*, *wall thickness*; the effective radius is the *radius*;
 
-    * Death Star: *sphere radius*, *hole radius*, *distance between centers*; the effective radius is the *sphere radius*.
+    * Death Star: *sphere radius*, *hole radius*, *distance between centers*; the effective radius is the *sphere radius*;
+
+    * Superquadric: *x half length* (or :math:`a`), *y half length* (or :math:`b`), *z half length* (or :math:`c`), *x exponent* (or :math:`r`), *y exponent* (or :math:`s`), *z exponent* (or :math:`t`); the effective radius is the Euclidian norm of the half lengths. The exponents represent the blockiness in each direction. The surface is implicitly described by :math:`|\frac{x}{a}|^r + |\frac{y}{b}|^s + |\frac{z}{c}|^t - 1 = 0`;
```
PierreLaurentinCS, line 214:
```suggestion
    * Superquadric: *x half length* (or :math:`a`), *y half length* (or :math:`b`), *z half length* (or :math:`c`), *x exponent* (or :math:`r`), *y exponent* (or :math:`s`), *z exponent* (or :math:`t`); the effective radius is the Euclidian norm of the half lengths. The exponents represent the blockiness in each direction. The surface is implicitly described by :math:`\left|\frac{x}{a}\right|^r + \left|\frac{y}{b}\right|^s + \left|\frac{z}{c}\right|^t - 1 = 0`;
```
Slightly more aesthetic

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
 #endif
 };
 
+/**
+ * @class This class defines superquadric shapes. Their signed distance function
+ * is:
+ * |\frac{x}{a}|^r  + |\frac{y}{b}|^s + |\frac{z}{c}|^t - 1 = 0
```
PierreLaurentinCS, line 384:
```suggestion
 * :math:`\left|\frac{x}{a}\right|^r + \left|\frac{y}{b}\right|^s + \left|\frac{z}{c}\right|^t - 1 = 0`
```

Slightly more aesthetic

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+{
+  auto point_in_string = this->point_to_string(p);
+  auto iterator        = this->closest_point_cache.find(point_in_string);
+  if (iterator == this->closest_point_cache.end())
+    {
+      // We align and center the evaluation point according to the shape
+      // referential.
+      Point<dim> centered_point = this->align_and_center(p);
+
+      Point<dim>   current_point = centered_point;
```
Luckabarbeau, line 401: add some comments about this being the initial guess

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
...
+   */
+  Superquadric<dim>(const Tensor<1, dim> half_lengths,
+                    const Tensor<1, dim> exponents,
+                    const Point<dim> &   position,
+                    const Tensor<1, 3> & orientation)
+    : Shape<dim>(half_lengths.norm(), position, orientation)
+  {
+    this->half_lengths = half_lengths;
+    this->exponents    = exponents;
+    this->epsilon      = 1e-12;
```
Luckabarbeau, line 406: Epsilon is hardcoded to a very small value at the moment. I think this should be a parameter

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
...
+   * @param half_lengths The half-lengths of each direction
+   * @param exponents The blockiness in each direction
+   * @param position The superquadric center
+   * @param orientation The superquadric orientation
+   */
+  Superquadric<dim>(const Tensor<1, dim> half_lengths,
+                    const Tensor<1, dim> exponents,
+                    const Point<dim> &   position,
+                    const Tensor<1, 3> & orientation)
+    : Shape<dim>(half_lengths.norm(), position, orientation)
```
PierreLaurentinCS, line 402:
```suggestion
    : Shape<dim>(half_lengths.norm(), position, orientation),half_lengths(half_lengths),exponents(exponents),epsilon(1e-12)
```
Is there a reason why half_lengths, exponents and epsilon are in the body of the constructor and not the initialization list ?

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
...
+  {
+    this->half_lengths = half_lengths;
+    this->exponents    = exponents;
```
PierreLaurentinCS, line 405:
```suggestion
```

include/core/shape.h
PierreLaurentinCS, line 406:
```suggestion
```

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+      unsigned int iteration_max = 1e4;
+
+      Point<dim> dx{}, distance_gradient{};
+      double     current_distance = superquadric(current_point);
+
+      // We scale the relaxation with the maximum blockiness, because higher
+      // blockiness makes convergence more difficult for the gradient descent.
+      const double shape_blockiness =
+        std::max(std::max(exponents[0], exponents[1]), exponents[2]);
+      double relaxation = 0.5 / shape_blockiness;
```
Luckabarbeau, line 412: Why the default is 0.5? Would it be better if it's 2 in which case for the ellipsoid it doesn't relax since it's a Euclidean distance?

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
...
+  void
+  closest_surface_point(
+    const Point<dim> &                                    p,
+    Point<dim> &                                          closest_point,
+    const typename DoFHandler<dim>::active_cell_iterator &cell_guess) override;
+  void
+  closest_surface_point(const Point<dim> &p,
+                        Point<dim> &      closest_point) const override;
+
+
```
PierreLaurentinCS, line 482:
```suggestion
```

include/core/shape.h
```diff
@@ -378,6 +378,178 @@ class Sphere : public Shape<dim>
...
+  /**
+   * @brief Computes the value of the superquadric from its equation
+   * @param centered_point point at which we make the evaluation, in the shape referential
+   */
+  inline double
+  superquadric(const Point<dim> &centered_point) const
+  {
+    return pow(abs(centered_point[0] / half_lengths[0]), exponents[0]) +
+           pow(abs(centered_point[1] / half_lengths[1]), exponents[1]) +
+           pow(abs(centered_point[2] / half_lengths[2]), exponents[2]) - 1;
```
PierreLaurentinCS, line 514:
```suggestion
           pow(abs(centered_point[2] / half_lengths[2]), exponents[2]) - 1.0;
```

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+    {
+      this->closest_surface_point(p, closest_point);
+
+      Point<dim> copy_closest_point{};
+      for (unsigned int d = 0; d < dim; d++)
+        copy_closest_point[d] = closest_point[d];
+      this->closest_point_cache[point_in_string] = copy_closest_point;
+    }
+  else
+    closest_point = iterator->second;
```
Luckabarbeau, line 385: When you have found the closest point, you can also directly store the value and gradient in the cache.

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+      Point<dim> closest_point{};
+      this->closest_surface_point(evaluation_point, closest_point);
+      Point<dim> copy_closest_point{};
+      for (unsigned int d = 0; d < dim; d++)
+        copy_closest_point[d] = closest_point[d];
+      this->closest_point_cache[point_in_string] = copy_closest_point;
+
+      double levelset                    = this->value(evaluation_point);
+      this->value_cache[point_in_string] = levelset;
+      return levelset;
```
Luckabarbeau, line 483: Same comments as above. You can store the gradient as well

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+    this->half_lengths, this->exponents, this->position, this->orientation);
+  return copy;
+}
+
+template <int dim>
+void
+Superquadric<dim>::clear_cache()
+{
+  Shape<dim>::clear_cache();
+  this->closest_point_cache.clear();
```
Luckabarbeau, line 503: This should just be added directly to the original clear cache. There is no reason to have a different clear cache for different shapes.

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+      Point<dim> closest_point{};
+      this->closest_surface_point(evaluation_point, closest_point);
+      Point<dim> copy_closest_point{};
+      for (unsigned int d = 0; d < dim; d++)
+        copy_closest_point[d] = closest_point[d];
+      this->closest_point_cache[point_in_string] = copy_closest_point;
+
+      Tensor<1, dim> gradient               = this->gradient(evaluation_point);
+      this->gradient_cache[point_in_string] = gradient;
+      return gradient;
```
Luckabarbeau, line 549: Same, you can store the value here.

source/core/shape.cc
```diff
@@ -363,6 +363,195 @@ Sphere<dim>::set_position(const Point<dim> &position)
 }
```
PierreLaurentinCS, line 365:
```suggestion
```

Luckabarbeau (commented):
Very good additions; some comments small comments nothing too big. I was wondering if the gradient descent function should be a general function of shape and just be used to find the closest point in cases where the shape is not defined by a distance field. In that case, defining epsilon as a general parameter for shape tolerance would make a lot of sense.

include/core/shape.h
oguevremont: I thought about this also, but I wasn't sure if it would be useful. I can add it as a 7th parameter.

source/core/shape.cc
oguevremont: I don't think I can because the function is const, but I'll check again to make sure

source/core/shape.cc
oguevremont: There is no closest point cache for other shapes. Should we add it you think?

include/core/shape.h
PierreLaurentinCS: What happens if one of the half_lenghts array element is equal to zero ? Maybe you could throw an error ?

PierreLaurentinCS (commented):
Very good work! I've left comments there and there, but it is nothing major

include/core/shape.h
```diff
@@ -378,6 +378,174 @@ class Sphere : public Shape<dim>
...
+          gradient[d] =
+            sign(centered_point[d]) * exponents[d] * (1 / half_lengths[d]) *
+            pow(abs(centered_point[d] / half_lengths[d]), exponents[d] - 1);
+      }
+    if (gradient.norm() > epsilon)
+      // Because it is possible that the centroid of the superquadric overlaps
+      // with the evaluation point, we consider this exception here and avoid
+      // normalizing if the norm is null
+      gradient = gradient / gradient.norm();
+    return gradient;
```
Luckabarbeau, line 536: The superquadric gradient shouldn't be normalized

source/core/shape.cc
oguevremont: Having some relaxation allows to find the "right" surface point. With relaxation = 1, it can happen that we find the surface point on the other side of the shape, especially for narrow concave shapes.
image

blaisb:
Is everybody good with the current state of this branch? For me it would be ready for merge. @PierreLaurentinCS @Luckabarbeau @oguevremont

PR #785, opened 2023-06-28
Add isothermal ideal gas density model

AmishgaAlphonius:
# Description of the problem

- When experimenting with the dam break with an obstacle example, it was noticed that the software performed poorly in mimicking the wave's natural shape after the impact. It was assumed that this is due to the lack of considering the air's natural compressibility.

# Description of the solution

- In an attempt to solve this issue, an isothermal compressible fluid dynamics model will be implemented. As a first step, an isothermal ideal gas density model is implemented.

# How Has This Been Tested?

- A unit test for the isothermal ideal gas density model was added.

- [x] tests/core/density_isothermal_ideal_gas.cc

# Documentation

- The physical properties page in the documentation was updated by adding the isothermal ideal gas density model.

- [x] doc/source/parameters/cfd/physical_properties.rst

# Future changes

- Currently, the isothermal ideal gas model can only be used in conjunction with the incompressible Navier-Stokes equations. However, it is meant to be used with the isothermal formulation of compressible Navier-Stokes equations to account for weakly compressible fluid flows. In a future PR, an isothermal compressible formulation of the equations will be implemented.

# Comments

- A DataPostprocessor object was implemented for density, and the density field is only outputted when it is not constant.
- Other minor corrections were made to unrelated files (mainly typos in the comments).

### DIFF
--- a/tests/core/carreau_rheology.cc
+++ b/tests/core/carreau_rheology.cc
@@ -16,8 +16,7 @@ test()
   Carreau rheology_model(5, 0, 1, 2, 0.5);
 
 
-  // field values can remain empty since the constant thermal conductivity does
-  // not depend  on any fields
+  // Field values must contain shear rate
   std::map<field, double> field_values;
 
   deallog << "Testing Carreau viscosity - nu" << std::endl;
--- a/tests/core/density_isothermal_ideal_gas.cc
+++ b/tests/core/density_isothermal_ideal_gas.cc
@@ -1,5 +1,5 @@
 /**
- * @brief Tests the isothermal ideal gas density model. This model should always return a constant.
+ * @brief Tests the isothermal ideal gas density model. This model should always return rho_ref + 1/(RT)*p
  */
 
 // Lethe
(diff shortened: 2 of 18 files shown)


tests/core/carreau_rheology.cc
```diff
@@ -16,8 +16,7 @@ test()
   Carreau rheology_model(5, 0, 1, 2, 0.5);
 
 
-  // field values can remain empty since the constant thermal conductivity does
-  // not depend  on any fields
+  // Field values must contain shear rate
```
blaisb, line 19: Yes, that is true :)!

tests/core/density_isothermal_ideal_gas.cc
```diff
@@ -0,0 +1,66 @@
+/**
+ * @brief Tests the isothermal ideal gas density model. This model should always return a constant.
```
blaisb, line 2:
```suggestion
 * @brief Tests the isothermal ideal gas density model. This model should always return rho_ref + 1/(RT)*p
```

:)

blaisb:
@AmishgaAlphonius Awesome. I'll merge after CI :).

PR #786, opened 2023-06-28
Fix issue when particles cross PBC

acdaigneault:
# Description of the problem

- When particles cross PBC, their location is updated but not the cell in which they are. No issues was encountered in the pass, bugs seem to happen when diameter is very small.

# Description of the solution

- Move the execution of the particle displacement at the same step as the sorting of particles in cells. This means the particles won't have their location updated as soon as the position > the location of the periodic wall. The particle displacement is already executed as is for CFD-DEM.

# How Has This Been Tested?

- Tested with hopper case and tests are good

### DIFF
--- a/source/dem/dem.cc
+++ b/source/dem/dem.cc
@@ -1133,6 +1133,11 @@ DEMSolver<dim>::solve()
       if (particles_insertion_step || load_balance_step ||
           contact_detection_step || checkpoint_step)
         {
+          // Particles displacement if passing through a periodic boundary
+          periodic_boundaries_object.execute_particles_displacement(
+            particle_handler,
+            container_manager.periodic_boundaries_cells_information);
+
           particle_handler.sort_particles_into_subdomains_and_cells();
 
           if (has_disabled_contacts && !simulation_control->is_at_start())
@@ -1309,11 +1314,6 @@ DEMSolver<dim>::solve()
                 disable_contacts_object.get_mobility_status());
             }
         }
-
-      // Particles displacement if passing through a periodic boundary
-      periodic_boundaries_object.execute_particles_displacement(
-        particle_handler,
-        container_manager.periodic_boundaries_cells_information);
 
       // Visualization
       if (simulation_control->is_output_iteration())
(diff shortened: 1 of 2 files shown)


blaisb:
Nice find. I suspectthe issue stems from the fact that once the particle crosses, then the periodic mapping of the particle becomes invalid because it has crossed, so the "-" factor to map it to the other side does not work anymore. It's a bit of an edge case, but it was actually bound to happen (I am surprised we never caught it to be honest). This is a good temporary fix, but it might decrease performance significantly no? Have you tested it in the periodic hopper case?

acdaigneault:
> Nice find. I suspect the issue stems from the fact that once the particle crosses, then the periodic mapping of the particle becomes invalid because it has crossed, so the "-" factor to map it to the other side does not work anymore. It's a bit of an edge case, but it was actually bound to happen (I am surprised we never caught it to be honest). This is a good temporary fix, but it might decrease performance significantly no? Have you tested it in the periodic hopper case?

This is @OGaboriault that found the bug with his simulation with very small particles (dp = 22.5e-6). He got popcorns when particles were crossing the periodic boundary at loading (high velocity because of the collision with wall). Never had this issue with the periodic hopper, I'm currently testing the performance.

The issue was that the location of the particle that crossed the PBC was updated, but not the cell parent. Meaning a failing of particle candidate in fine search (by radius).

Another solution could be to update the location of particle only when the contact list is updated (at the step it triggers right now) This solution would not affect the performance as much I think. Also this is how it works in CFD-DEM at the moment because of the qcm, I think I will do that. This is better than my quick fix, thanks :)

blaisb:
@acdaigneault I prefer your new solution actually than the previous one. So yeah, good idea :). Will review tonight before going to bed.

acdaigneault:
> @acdaigneault I prefer your new solution actually than the previous one. So yeah, good idea :). Will review tonight before going to bed.

Yes, the performance was crap with the old way :)

blaisb:
> > @acdaigneault I prefer your new solution actually than the previous one. So yeah, good idea :). Will review tonight before going to bed.
> 
> Yes, the performance was crap with the old way :)

Yeah but that's because the old way did not make sense. The difference between Lethe and Aspect is that the locate_particles_into_subdomain_and_cells is called every iteration in Aspect, whereas in Lethe we alternate between upate_ghosts and this. So actually we should only apply the flip of the particles (position correction) before we are going to call locate_particles. I feel dumb I did not think about this beforehand though.

blaisb (approved):
Awesome work. Will merge

PR #787, opened 2023-07-01
Update periodic hopper example

acdaigneault:
# Description of the problem

- The DEM solver was been improved since the creation of the hopper example, the results and computation time were not accurate anymore.
- A bug was fixed for periodic boundaries resulting of minor different behavior of particles' contacts in periodic cells.

# Description of the solution

- Update mass flow rate and computational time results were updated.

# Comments
 
- Figures were not updated since the differences are minor.


blaisb:
Will merge after CI

PR #797, opened 2023-07-12
Clear the tangential overlap after load balancing in DEM

acdaigneault:
# Description of the problem

- Tangential overlap and relative velocity stored values in contact info object were lost during load balancing only for contacts of particles that is moved on other processor.
- Disabling contacts algorithm had a small glitch after load balancing because the identification of the mobility status was not executed at the right place (before exchange of ghost).

# Description of the solution
- Force the resetting of the tangential overlap and relative velocity values during the update of the iterator since it loops on the map of the contact info object. It used the `load_balance_step` boolean to trigger this resetting.
- Move the identification of the mobility after the exchange of ghost particles


blaisb:
@acdaigneault some tests don't seem to work anymore?

include/dem/update_local_particle_containers.h
```diff
@@ -63,7 +63,8 @@ void
 update_contact_container_iterators(
   pairs_structure &pairs_in_contact,
   typename DEM::dem_data_structures<dim>::particle_index_iterator_map
-    &particle_container);
+    &        particle_container,
+  const bool load_balance_step = false);
```
blaisb, line 67: Is load balance_step the right variable name here or should it be "clear_contact_structures" In essence, the fact that we clear when we load balance is one thing, but the action being asked of the structure is to clear it.

source/dem/update_local_particle_containers.cc
```diff
@@ -83,6 +84,14 @@ update_contact_container_iterators(
                   unsigned int particle_two_id = adjacent_map_iterator->first;
                   adjacent_map_iterator->second.particle_two =
                     particle_container[particle_two_id];
+
+                  if (load_balance_step)
```
blaisb, line 88: I would definitely call that variable "clear"

blaisb (changes requested):
Tiny naming changes, nothing significant.

blaisb:
Awesome, will merge after CI

source/dem/update_local_particle_containers.cc
```diff
@@ -83,6 +84,14 @@ update_contact_container_iterators(
                   unsigned int particle_two_id = adjacent_map_iterator->first;
                   adjacent_map_iterator->second.particle_two =
                     particle_container[particle_two_id];
+
+                  if (clear_contact_structures)
+                    {
+                      // Update the particle id of the particle in contact
```
blaisb, line 90: Is this comment still valid? I am not sure it makes sense no? (or it should apply to the line above no?)

blaisb (approved):
One last thing to check and it will be ok to merge

source/dem/update_local_particle_containers.cc
acdaigneault, line 90:
```suggestion
```

PR #800, opened 2023-07-12
Interactive diagrams for the examples in the documentation

lpsaavedra:
# Description of the problem

There was no interactive way of going through the examples in the documentation. 

# Description of the solution

An interactive diagram/map for each of the categories was created. Now it is easier to find examples according to specific criteria. The title of some examples were modified (e.g., dimensions removed for incompressible flow examples, and only the first capital letter of the titles was kept with exception of names or capital letters needed for acronyms).

# Comments

The graphviz package must be installed for the diagrams to compile. The instructions were added in the contributing tab of the documentation and the github workflow for the documentation was also updated accordingly. Two images were added in the diagram for the incompressible flow as a possible way of making it look nicer.

### DIFF
--- a/.github/workflows/doc-github-pages.yml
+++ b/.github/workflows/doc-github-pages.yml
@@ -23,7 +23,6 @@ jobs:
           sudo apt-get install -y graphviz
           pip install sphinx==${{ env.SPHINX_VERSION }}
           pip install sphinx-copybutton
-          #pip install graphviz
           pip install -r doc/requirements.txt
 
       - name: Build HTML


.github/workflows/doc-github-pages.yml
```diff
@@ -20,8 +20,10 @@ jobs:
 
       - name: Setup Sphinx
         run: |
+          sudo apt-get install -y graphviz
           pip install sphinx==${{ env.SPHINX_VERSION }}
           pip install sphinx-copybutton
+          #pip install graphviz
```
blaisb, line 26:
```suggestion
```

doc/source/contributing.rst
```diff
@@ -70,6 +70,13 @@ To install additional packages:
 
   pip install sphinx-copybutton
 
+and:
+
+.. code-block:: shell
+  :class: copy-button
+
+  pip install graphviz
```
blaisb, line 78: Here I would suggest instead we install the system graphviz. Using pip is not sufficient :)!

blaisb (approved):
Two tiny changes then this is ready for merge IMO

PR #802, opened 2023-07-13
Refactor DEM model parameters' subsections

acdaigneault:
# Description of the problem

- The model parameters section has a lot of parameters for a few features 
- Not all DEM prm files were indented

# Description of the solution

- Add the `load balancing` and `contact detection` subsections
- Indent the prm files

# Comments

- All DEM and CFD-DEM unresolved parameter files from application tests and example have been updated


blaisb (approved):
Awesome. Very good work. Will merge

PR #805, opened 2023-07-13
Refactor the contact manager in DEM

acdaigneault:
# Description of the problem

- The container manager class was more a contact manager with extra random containers and it was a bit weird.

# Description of the solution

- Replace the DEMContainerManager class to DEMContactManager class
- Move some containers back to DEM or CFD-DEM coupling
- Fix a small error in the map of floating walls since it used particle_id type for floating walls

# How Has This Been Tested?

- Tests are good

### DIFF
--- a/include/dem/dem.h
+++ b/include/dem/dem.h
@@ -19,7 +19,7 @@
 #include <core/serial_solid.h>
 
 #include <dem/data_containers.h>
-#include <dem/dem_container_manager.h>
+#include <dem/dem_contact_manager.h>
 #include <dem/dem_solver_parameters.h>
 #include <dem/disable_contacts.h>
 #include <dem/find_boundary_cells_information.h>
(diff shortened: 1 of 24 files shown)


include/dem/dem.h
```diff
@@ -360,8 +359,23 @@ class DEMSolver
...
+  typename dem_data_structures<dim>::floating_mesh_information
+    floating_mesh_info;
+  typename dem_data_structures<dim>::boundary_points_and_normal_vectors
+    updated_boundary_points_and_normal_vectors;
+  typename dem_data_structures<dim>::vector_on_boundary
+    forces_boundary_information;
+  typename dem_data_structures<dim>::vector_on_boundary
+    torques_boundary_information;
+  typename DEM::dem_data_structures<dim>::periodic_boundaries_cells_info
+    periodic_boundaries_cells_information;
```
blaisb, line 373: Should these types be defined here or could they be defined in another .h file to be more neutral?

source/dem/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_library(lethe-dem
   # Sources
   data_containers.cc
   dem.cc
-  dem_container_manager.cc
+        dem_contact_manager.cc
```
blaisb, line 5: indent here is weird no?

blaisb (approved):
Some tiny comments, that's all. Great work!

include/dem/dem.h
acdaigneault: Mhm I don't know, they are not really related to each other, it's again a bit random to me. What do you think?

blaisb: Good for me then :)!

PR #807, opened 2023-07-14
Reuse velocity solution

oguevremont:
# Description of the problem

- It was difficult to couple problems where fluid dynamics reach pseudo-steady state much faster than auxiliary physics.

# Description of the solution

- Using average velocity field for heat transport advection was already possible, but it required keeping fluid dynamics active and it was only possible for heat transport.
- A modification was made that allows using average velocity field even with fluid dynamics disabled.
- It was made possible to do this with Tracer and VOF auxiliary physics.

# How Has This Been Tested?

- A short script running two simulations, one with and one without fluid dynamics, was used.
- This simple case will be made into an example.

# Documentation

- Documentation was adjusted in Multiphysics parameters.

# Future changes

- The simple case will be made into an example.


doc/source/parameters/cfd/multiphysics.rst
```diff
@@ -5,19 +5,20 @@ This subsection defines the multiphysics interface of Lethe and enables the solu
...
 
     # Tracer
-    set tracer 			= false
+    set tracer = false
 
     # Multiphase flow
-    set VOF 			= false
+    set VOF = false
+
+    set use time average velocity field = false
```
blaisb, line 21: why change the alignment here?

source/solvers/navier_stokes_base.cc
```diff
@@ -565,23 +565,34 @@ NavierStokesBase<dim, VectorType, DofsType>::iterate()
...
+          present_solution = this->local_evaluation_point;
+        }
+      else
+        {
+          // We get the solution via an initial condition
+          this->simulation_parameters.initial_condition->uvwp.set_time(
+            this->simulation_control->get_current_time());
+          set_initial_condition_fd(
+            this->simulation_parameters.initial_condition->type);
+        }
```
blaisb, line 592: Nice

blaisb (changes requested):
Cool stuff. it would be great to have a quick application_tests to ensure the feature remains stable.

blaisb:
@oguevremont Seems like this breaks a regular restart feature?

doc/source/parameters/cfd/multiphysics.rst
oguevremont: I ran this subsection in prmindent and copied the output. I can put it back the way it was though, the alignment was fine.

oguevremont:
> @oguevremont Seems like this breaks a regular restart feature?

I fixed it

PR #816, opened 2023-07-19
Optimize fine search

blaisb:
# Description of the problem

-One data structure still did not use the Ankerl map.

# Description of the solution

- User Ankerl map for this last data structure. This  increases performance by another 10%

# How Has This Been Tested?

- All unit tests pass and benchmark still give the same results.


acdaigneault (approved):
Nothing to say, nice optimization :)

PR #820, opened 2023-07-25
Matrix-free Navier-Stokes prototype

lpsaavedra:
# Description of the problem

This PR adds the first matrix free (MF)  steady-state Navier-Stokes prototype. It uses SUPG/PSPG stabilization and includes both global coarsening and local smoothing MG approaches as preconditioners for the linear iteration. 

# Comments

This is the fourth step towards Matrix-Free Lethe. It is the last MF prototype and now all the MF work will be done directly in the Lethe implementation.


prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+  AffineConstraints<number>              constraints;
+  mutable TrilinosWrappers::SparseMatrix system_matrix;
+  Settings                               parameters;
+  AlignedVector<VectorizedArray<number>> element_size;
+  Table<2, Tensor<1, dim + 1, VectorizedArray<number>>>
+    nonlinear_previous_values;
+  Table<2, Tensor<1, dim + 1, Tensor<1, dim, VectorizedArray<number>>>>
+    nonlinear_previous_gradient;
+  Table<2, Tensor<1, dim + 1, Tensor<1, dim, VectorizedArray<number>>>>
+    nonlinear_previous_hessian_diagonal;
```
peterrum, line 549: I would probably only store the global vector and recompute these quantities with `FEEvaluaion` (the Hessian is quite large).

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+                                               Functions::ZeroFunction<dim>(
+                                                 dim + 1),
+                                               constraint);
+      VectorTools::interpolate_boundary_values(mapping,
+                                               dof_handler,
+                                               3,
+                                               Functions::ZeroFunction<dim>(
+                                                 dim + 1),
+                                               constraint);
+      if constexpr (dim == 3)
```
peterrum, line 1272: Does the stokes example not support 3D?

peterrum (approved):

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+  return value<double>(p, component);
+}
+
+// Matrix-free helper function
+template <int dim, typename Number>
+VectorizedArray<Number>
+evaluate_function(const Function<dim> &                      function,
+                  const Point<dim, VectorizedArray<Number>> &p_vectorized)
+{
+  VectorizedArray<Number> result;
```
Luckabarbeau, line 419: I am not sure about the standard that we have for comments in prototypes, but I would add comments here to explain why it is unnecessary to initialize the VectorizedArray size.

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+}
+
+// Performs the transposed operator evaluation. Since we have
+// non-symmetric matrices this is different from the vmult call.
+// TODO: implement this correctly (let's start with something symmetric)
+template <int dim, typename number>
+void
+NavierStokesOperator<dim, number>::Tvmult(VectorType &      dst,
+                                          const VectorType &src) const
+{
```
Luckabarbeau, line 770: Not sure how this impact the resolution process. Can it have a significant impact to consider only a symmetric matrice?

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+             ls_mg_interface_out,
+             ls_mg_operators);
+}
+
+// Main class to solve the steady Navier-Stokes quations given by
+// using Newton's method and the matrix-free approach.
+template <int dim>
+class MatrixFreeStokes
+{
+public:
```
Luckabarbeau, line 1553: Maybe update the name to Navier stokes

Luckabarbeau (commented):
Multiple functions need more comments to explain the different steps executed inside the function in the main branch of Lethe. However, I am not sure what the standard upholds for a prototype. 

As for the contents, I don't think I am currently able to understand fully how the different parts of the codes interact with each other without taking a few days to play with the code seriously.

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
```diff
@@ -0,0 +1,2420 @@
...
+  unsigned int repetitions;
+  double       relaxation_parameter;
+  double       viscosity;
+  bool         output;
+  std::string  output_name;
+  std::string  output_path;
+};
+
+bool
+Settings::try_parse(const std::string &prm_filename)
```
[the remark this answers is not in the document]
lpsaavedra: This is the name chosen for all the previous prototypes, so I will leave it like this. It is only called try_parse due to the try catch inside of it, we won't be using this in the Lethe implementation

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
lpsaavedra: We are not using the transposed evaluation of the operator at the moment. So it does not have any impact right now. I will write a comment on this :)

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
lpsaavedra: That's a good point. I will leave the prototype like this, but I will take this into account for the actual implementation in Lethe :)

prototypes/matrix_free_navier_stokes/matrix_free_navier_stokes.cc
lpsaavedra: It does support 3D, but I probably didn't add the MMS. I will add it in a quick PR so that it is also complete.

PR #826, opened 2023-07-28
Implemented minimal refinement functionality

oguevremont:
# Description of the problem

- Until now we had to select a relative distance (relative to the effective radius of the shape) to define the crown refinement region
- When using a finer mesh, this meant having to decrease manually the parameters, or having a layer of refined cells thicker than necessary (we only need a layer of 2 uncut refined cells)

# Description of the solution

- This PR allows Lethe to define the distance automatically, so that the layer of uncut refined cells is always minimal: 2 cells thickness
- I added an option to define absolute distance instead of relative, and we call is_inside_crown with the absolute distance that depends on the smallest cell diameter

# How Has This Been Tested?

- A 2D test was added
- All tests still pass

# Documentation

- New feature was added to sharp parameters documentation

image

### DIFF
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -135,7 +135,7 @@ This subsection contains the parameters related to the sharp immersed boundary s
 	    The refined cells are all those for which at least one of the degrees of freedom (dof) location satisfies both the ``refine mesh inside radius factor`` and the ``refine mesh outside radius factor`` thresholds. Each cycle of refinement reduces the length of the elements by a factor two.
 
     .. note::
-        Using values ``refine mesh outside radius factor = 1`` and ``refine mesh inside radius factor = 1`` activates a minimal refinement mode. This enables the solver to select automatically the smallest region near the particle that is permitted by the method.
+        Using values ``refine mesh outside radius factor = 1`` and ``refine mesh inside radius factor = 1`` activates a minimal refinement mode. This enables the solver to select automatically the smallest region near the particle that guarantees stability of the solution.
 
     .. note::
 	    This near-particle zone will be systematically refined at each refinement step until reaching the ``max refinement level`` parameter (:doc:`../cfd/mesh_adaptation_control`).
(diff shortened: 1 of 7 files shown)


AmishgaAlphonius (approved):
Nice work! :)

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -864,16 +883,36 @@ GLSSharpNavierStokesSolver<dim>::refine_ib()
...
                       // and smaller than the outside radius of the refinement
                       // zone.
-                      if (particles[p].is_inside_crown(
+                      bool is_inside_crown;
+                      if (minimal_crown_refinement_enabled)
+                        {
+                          // The factor depends on dim, since cell->diameter()
+                          // returns the longest diagonal. The diagonal is
+                          // higher in 3D than in 2D. The chosen factor is equal
+                          // to the length of the smallest cell.
```
oguevremont, line 892: @AmishgaAlphonius I have an explanation for the factor. Does it make sense?

Luckabarbeau: I would say that you want the length of one of the edges of the cell. Since the cell diameter returns you the diagonal length of the cell, you assume a cubic cell and divide by the square root of the dimension to get the length of the edge.

doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
```diff
@@ -134,6 +134,9 @@ This subsection contains the parameters related to the sharp immersed boundary s
     .. note::
 	    The refined cells are all those for which at least one of the degrees of freedom (dof) location satisfies both the ``refine mesh inside radius factor`` and the ``refine mesh outside radius factor`` thresholds. Each cycle of refinement reduces the length of the elements by a factor two.
 
+    .. note::
```
Luckabarbeau, line 137: Nice usage of the parameters.

doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
```diff
@@ -134,6 +134,9 @@ This subsection contains the parameters related to the sharp immersed boundary s
...
+        Using values ``refine mesh outside radius factor = 1`` and ``refine mesh inside radius factor = 1`` activates a minimal refinement mode. This enables the solver to select automatically the smallest region near the particle that is permitted by the method.
```
Luckabarbeau, line 138: Instead of "permitted", I would say that it is the smallest region that guarantees stability of the solution.

Luckabarbeau (commented):
Very nice addition! Once the comments are addressed, I think it will be ready to merge.

PR #829, opened 2023-08-01
Reduce memory rbf

oguevremont:
# Description of the problem

- Memory requirements was too high for dense RBF shapes (high number of RBF nodes) in very fine grids.

# Description of the solution

- Nodes precalculations was changed: _before_, every cell had a vector with the IDs of the contributing RBF nodes, and _now_ every node is in only one vector, associated to a fine cell. The cells then have a list of shared pointers that reference these vectors of IDs.
-- The performance is barely affected in terms of computing speed, but very much improved for memory requirements even for the densest RBF tested and in fine grids.
-- A side effect of this (for the moment) is that RBF nodes that are in no cell are lost, so their contribution on the shape is as well. This happens only near the domain boundaries. An option to prevent mesh based precalculations was added for cases where accurate surface representation is important near the background grid boundary.

- Another change is that information of RBF nodes that don't affect locally owned cells are removed (a reload function is then used  when the shape moves).
- An additional change is that cache of constituents in CompositeShapes is not used anymore: this prevents duplication of information.

# How Has This Been Tested?

- Tests still pass.

# Documentation

- Changes are reflected in the sharp IB parameters section.

# Future changes

- A modification will be made to the algorithm to keep the effect of RBF nodes located outside the background grid.

### DIFF
--- a/applications_tests/gls_sharp_navier_stokes/flow_around_rbf.prm
+++ b/applications_tests/gls_sharp_navier_stokes/flow_around_rbf.prm
@@ -75,8 +75,9 @@ subsection particles
   end
 
   subsection particle info 0
-    set type            = rbf
-    set shape arguments = rbf_test_shape.input
+    set type                       = rbf
+    set shape arguments            = rbf_test_shape.input
+    set mesh based precalculations = false
     subsection position
       set Function expression = 0.01;0.01;0.01
     end
@@ -112,3 +113,4 @@ subsection linear solver
     set ilu preconditioner relative tolerance = 1.00
   end
 end
+
--- a/applications_tests/gls_sharp_navier_stokes/flow_around_rbf.prm
+++ b/applications_tests/gls_sharp_navier_stokes/flow_around_rbf.prm
@@ -77,7 +77,7 @@ subsection particles
   subsection particle info 0
     set type                       = rbf
     set shape arguments            = rbf_test_shape.input
-    set mesh based precalculations = false
+    set mesh-based precalculations = false
     subsection position
       set Function expression = 0.01;0.01;0.01
     end
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -8,6 +8,7 @@ This subsection contains the parameters related to the sharp immersed boundary s
 
     subsection particles
       set assemble Navier-Stokes inside particles = false
+      set levels not precalculated                = 0
       set number of particles                     = 1
       
       subsection extrapolation function
@@ -92,6 +93,8 @@ This subsection contains the parameters related to the sharp immersed boundary s
   end
 
 * The ``number of particles`` is the number of particles simulated by the sharp-edge IB.
+
+* The ``levels not precalculated`` parameter controls the number of layers of the hierarchical grid used by Lethe that are ignored by precalculations. It allows to reduce the memory footprint at the cost of an increased computing time. At the moment, this is used only for RBF shapes. The value should be increased when the RBF contains a lot of nodes and/or the grid is extremely fine.
 
 * The ``assemble Navier-Stokes inside particles`` parameter determines if the Navier-Stokes equations are solved inside the particles or not. If the Navier-Stokes equations are not solved (the parameter is false), the solver will solve a Poisson equation for each variable in the problem. This eliminates the need to define a reference value for the pressure.
 
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -59,10 +59,11 @@ This subsection contains the parameters related to the sharp immersed boundary s
       end
       
       subsection particle info 0
-        set type              = sphere 
-        set shape arguments   = 1
-        set integrate motion  = false
-        set pressure location = 0; 0; 0
+        set type                       = sphere
+        set shape arguments            = 1
+        set integrate motion           = false
+        set pressure location          = 0; 0; 0
+        set mesh based precalculations = true
         
         subsection position
           set Function expression = 0; 0; 0
@@ -251,6 +252,8 @@ The following parameter and subsection are all inside the subsection ``particle 
         As could be expected, using this type of shape requires that ``dealii`` be compiled with OpenCascade. This module can be installed with candi, by uncommenting the appropriate line in ``candi.cfg``.
 
 * The ``integrate motion`` parameter controls if the dynamics equations of the particles are calculated. If this parameter is set to false, the particles position, velocity, and angular velocity are defined directly by the functions. If ``integrate motion=true`` the position and the velocity will be defined by the integration of the particle dynamic.
+
+* The ``mesh based precalculations`` parameter controls if the mesh-based precalculations are applied. These precalculations are critical for good performance in medium to high detailed RBFs (and its composites), but can introduce deformations. These deformations appear when some RBF nodes are located outside of the background mesh.
 
 * The ``pressure location`` parameter is used to define the X, Y, and Z coordinate offsets of the pressure reference point relative to the center of the particle. These parameters are used when the ``assemble Navier-Stokes inside particles`` parameter is set to ``true`` to define the pressure reference point.
 
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -8,7 +8,6 @@ This subsection contains the parameters related to the sharp immersed boundary s
 
     subsection particles
       set assemble Navier-Stokes inside particles = false
-      set levels not precalculated                = 0
       set number of particles                     = 1
       
       subsection extrapolation function
@@ -93,8 +92,6 @@ This subsection contains the parameters related to the sharp immersed boundary s
   end
 
 * The ``number of particles`` is the number of particles simulated by the sharp-edge IB.
-
-* The ``levels not precalculated`` parameter controls the number of layers of the hierarchical grid used by Lethe that are ignored by precalculations. It allows to reduce the memory footprint at the cost of an increased computing time. At the moment, this is used only for RBF shapes. The value should be increased when the RBF contains a lot of nodes and/or the grid is extremely fine.
 
 * The ``assemble Navier-Stokes inside particles`` parameter determines if the Navier-Stokes equations are solved inside the particles or not. If the Navier-Stokes equations are not solved (the parameter is false), the solver will solve a Poisson equation for each variable in the problem. This eliminates the need to define a reference value for the pressure.
 
--- a/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
+++ b/doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
@@ -8,7 +8,6 @@ This subsection contains the parameters related to the sharp immersed boundary s
 
     subsection particles
       set assemble Navier-Stokes inside particles = false
-      set levels not precalculated                = 0
       set number of particles                     = 1
       
       subsection extrapolation function
@@ -92,8 +91,6 @@ This subsection contains the parameters related to the sharp immersed boundary s
   end
 
 * The ``number of particles`` is the number of particles simulated by the sharp-edge IB.
-
-* The ``levels not precalculated`` parameter controls the number of layers of the hierarchical grid used by Lethe that are ignored by precalculations. It allows to reduce the memory footprint at the cost of an increased computing time. At the moment, this is used only for RBF shapes. The value should be increased when the RBF contains a lot of nodes and/or the grid is extremely fine.
 
 * The ``assemble Navier-Stokes inside particles`` parameter determines if the Navier-Stokes equations are solved inside the particles or not. If the Navier-Stokes equations are not solved (the parameter is false), the solver will solve a Poisson equation for each variable in the problem. This eliminates the need to define a reference value for the pressure.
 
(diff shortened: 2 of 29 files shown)


source/core/shape.cc
```diff
@@ -1670,8 +1928,13 @@ RBFShape<dim>::determine_likely_nodes_for_one_cell(
...
 
-  bool                parent_found = false;
-  std::vector<size_t> temporary_iterable_nodes;
+  bool parent_found = false;
+  std::vector<
+    std::tuple<Point<dim>, double, std::shared_ptr<std::vector<size_t>>>>
+    temporary_iterable_nodes{1};
+  temporary_iterable_nodes[0] = {Point<dim>{},
+                                 0,
+                                 std::make_shared<std::vector<size_t>>()};
```
AmishgaAlphonius, line 1937:
```suggestion
    temporary_iterable_nodes{
      {Point<dim>{}, 0, std::make_shared<std::vector<size_t>>()}};
```

source/core/shape.cc
```diff
@@ -1719,81 +1988,166 @@ RBFShape<dim>::determine_likely_nodes_for_one_cell(
...
+  basis_functions.clear();
+  useful_rbf_nodes.clear();
+  rotated_nodes_positions.clear();
+  nodes_positions_x.clear();
+  nodes_positions_y.clear();
+  nodes_positions_z.clear();
+
+  // The following lines retrieve information regarding an RBF
+  // with a given file name. Then, it converts the information
+  // into one vector which is used to initialize the RBF shape.
```
AmishgaAlphonius, line 2004:
```suggestion
  // into a vector which is used to initialize the RBF shape.
```

source/core/shape.cc
```diff
@@ -1719,81 +1988,166 @@ RBFShape<dim>::determine_likely_nodes_for_one_cell(
...
+          temp_weights[counter] = weights[i];
+          counter++;
         }
+    }
+  weights.swap(temp_weights);
+  temp_weights.clear();
 
-      // We remove unnecessary cells
-      for (auto it = likely_nodes_map.cbegin(); it != likely_nodes_map.cend();)
+  // Nodes positions treatment
```
AmishgaAlphonius, line 2084:
```suggestion
  // Nodes' positions treatment
```

source/core/shape.cc
```diff
@@ -1637,17 +1680,21 @@ RBFShape<dim>::initialize_bounding_box()
...
   Tensor<1, dim> half_lengths = Tensor<1, dim>();
   for (int d = 0; d < dim; d++)
     {
       high_bounding_point[d] = std::numeric_limits<double>::lowest();
       low_bounding_point[d]  = std::numeric_limits<double>::max();
       for (size_t i = 0; i < number_of_nodes; i++)
         {
+          temp_point[0]       = nodes_positions_x[i];
+          temp_point[1]       = nodes_positions_y[i];
+          temp_point[dim - 1] = nodes_positions_z[i];
```
AmishgaAlphonius, line 1693: Is it always 3D, otherwise you are overwriting the y position using ``dim - 1``?
You could do a separate ``if constexpr (dim == 3)`` with a ``for`` loop inside (to minimize the number of ``if``).

source/core/shape.cc
```diff
@@ -1719,81 +1988,166 @@ RBFShape<dim>::determine_likely_nodes_for_one_cell(
...
-  for (unsigned int i = 0; i < nodes_positions.size(); ++i)
+  Point<dim> temp_point{};
+  rotated_nodes_positions.resize(weights.size());
+  for (unsigned int i = 0; i < weights.size(); ++i)
     {
-      rotated_nodes_positions[i] =
-        this->reverse_align_and_center(nodes_positions[i]);
+      temp_point[0]              = nodes_positions_x[i];
+      temp_point[1]              = nodes_positions_y[i];
+      temp_point[dim - 1]        = nodes_positions_z[i];
```
AmishgaAlphonius, line 2145: same comment as in ``displaced_volume()``

AmishgaAlphonius (changes requested):
Nice work! I have a few comments and suggestions :)

source/core/shape.cc
```diff
@@ -1660,6 +1707,217 @@ RBFShape<dim>::initialize_bounding_box()
...
+          // A fix for this will be implemented shortly
+        }
+      // We need to remove all cells that are not needed anymore
+      // from the map
+      for (auto it = temporary_nodes_portions_map.cbegin();
+           it != temporary_nodes_portions_map.cend();)
+        {
+          auto               cell       = it->first;
+          const unsigned int cell_level = cell->level();
+          bool cell_still_needed = cell->is_active() || cell_level == level;
```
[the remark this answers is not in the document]
oguevremont: I thought of it also, but for readability I'll leave it like this

source/core/shape.cc
```diff
@@ -1660,6 +1707,217 @@ RBFShape<dim>::initialize_bounding_box()
...
+          if (!cell_still_needed)
+            {
+              temporary_nodes_portions_map.erase(it++->first);
```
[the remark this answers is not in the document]
oguevremont: Good question. I searched a bit, but from my understanding we have to iterate before we remove the entry that the iterator refers to, otherwise the iterator becomes invalidated. I changed it a bit so that it's more readable, but I still have to place it++ inside the loop, at the latest when I call erase().
https://stackoverflow.com/questions/23353812/does-stdmaperaseit-maintain-a-valid-iterator-pointing-to-the-next-elemen

source/core/shape.cc
```diff
@@ -1660,6 +1707,217 @@ RBFShape<dim>::initialize_bounding_box()
...
+            {
+              if (cell->point_inside(temp_cell->barycenter()))
+                {
+                  temp_cell_tuple = std::make_tuple(temp_cell->barycenter(),
+                                                    temp_cell->diameter(),
+                                                    std::get<2>(it->second));
+                  likely_nodes_map[cell]->push_back(temp_cell_tuple);
+                }
+              else
+                for (const size_t &node_id : *temp_nodes)
```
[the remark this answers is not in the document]
oguevremont: It's a mistake on my part

include/core/shape.h
```diff
@@ -1002,12 +1016,26 @@ class CompositeShape : public Shape<dim>
...
+  update_precalculations(DoFHandler<dim> &updated_dof_handler,
+                         const bool       mesh_based_precalculations);
+
+  /**
+   * @brief Load data from file
    */
   void
-  update_precalculations(DoFHandler<dim> &  updated_dof_handler,
-                         const unsigned int levels_not_precalculated);
+  load_data_from_file();
```
Luckabarbeau, line 1029: Need more detail in the description and usage of this function

Luckabarbeau (approved):
Nothing to add other than what we discussed! Good Job!

AmishgaAlphonius (approved):

blaisb:
@oguevremont can you rebase and then I'll review?

applications_tests/gls_sharp_navier_stokes/flow_around_rbf.prm
```diff
@@ -75,8 +75,9 @@ subsection particles
   end
 
   subsection particle info 0
-    set type            = rbf
-    set shape arguments = rbf_test_shape.input
+    set type                       = rbf
+    set shape arguments            = rbf_test_shape.input
+    set mesh based precalculations = false
```
blaisb, line 80:
```suggestion
    set mesh based precalculations = false
```
I think this would require an hyphen to be gramatically correct but we can let this pass

include/core/ib_particle.h
```diff
@@ -224,12 +224,26 @@ class IBParticle
    * @brief Sets the proper dof handler, then computes/updates the map of cells
    * and their likely non-null nodes
    * @param updated_dof_handler the reference to the new dof_handler
-   * @param levels_not_precalculated the number of finer levels that won't be
-   * precalculated
+   * @param mesh_based_precalculations mesh-based precalculations that can lead to slight shape misrepresentation (if RBF typed)
```
blaisb, line 227:
```suggestion
   * @param mesh_based_precalculations mesh-based precalculations that can lead to slight shape misrepresentation (if type=RBF)
```

doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
```diff
@@ -255,6 +253,8 @@ The following parameter and subsection are all inside the subsection ``particle
 
 * The ``integrate motion`` parameter controls if the dynamics equations of the particles are calculated. If this parameter is set to false, the particles position, velocity, and angular velocity are defined directly by the functions. If ``integrate motion=true`` the position and the velocity will be defined by the integration of the particle dynamic.
 
+* The ``mesh based precalculations`` parameter controls if the mesh-based precalculations are applied. These precalculations are critical for good performance in medium to high detailed RBFs (and its composites), but can introduce deformations. These deformations appear when some RBF nodes are located outside of the background mesh.
```
blaisb, line 256: So should this be set to false for all RBF cases? In this case, I would suggest it there

include/core/shape.h
```diff
@@ -278,6 +281,15 @@ class Shape : public AutoDerivativeFunction<dim>
   std::string
   point_to_string(const Point<dim> &evaluation_point) const;
 
+  /**
+   * Sets whether the shape is part of a composite
+   */
```
blaisb, line 286: missing @param explanation. Missing also @brief. Definition could be clearer here.

include/core/shape.h
```diff
@@ -899,11 +912,12 @@ class CompositeShape : public Shape<dim>
     , constituents(constituents)
     , operations(operations)
   {
-    // Calculation of the effective radius
-    for (auto const &[component_id, component] : constituents)
+    // Calculation of the effective radius and setting of constituents' status
+    for (auto const &[constituent_id, constituent] : constituents)
```
blaisb, line 916: You could use a boost::value iterator adaptor to not have to declare the pair like this unused.

include/core/shape.h
```diff
@@ -1002,12 +1016,27 @@ class CompositeShape : public Shape<dim>
...
+
+  /**
+   * @brief Load data from file. To be called at initialization, after repartitioning or when shape has moved.
+   * This function is used only for RBF shapes and its composites at the moment.
+   */
+  void
+  load_data_from_file();
+
+  /**
+   * @brief Remove data that doesn't affect the cells owned by local process   *
```
blaisb, line 1033:
```suggestion
   * @brief Remove data that doesn't affect the cells owned by local process 
```

What are cells owned? Do you mean not locally_owned and not ghost? artificial? be a bit clearer

include/core/shape.h
```diff
@@ -1283,36 +1312,28 @@ class RBFShape : public Shape<dim>
...
    */
-  RBFShape<dim>(const std::vector<double> &shape_arguments,
-                const Point<dim> &         position,
-                const Tensor<1, 3> &       orientation);
+  void
+  load_data_from_file();
+
+  /**
+   * @brief Remove data that doesn't affect the cells owned by local process
+   */
```
blaisb, line 1333: Missing all @param

include/core/shape.h
```diff
@@ -1772,55 +1792,52 @@ class RBFShape : public Shape<dim>
...
-  std::vector<size_t>           nodes_id;
-  std::vector<double>           weights;
-  std::vector<Point<dim>>       nodes_positions;
-  std::vector<Point<dim>>       rotated_nodes_positions;
-  std::vector<double>           support_radii;
-  std::vector<RBFBasisFunction> basis_functions;
+  std::vector<double>     weights;
+  std::vector<double>     nodes_positions_x;
+  std::vector<double>     nodes_positions_y;
+  std::vector<double>     nodes_positions_z;
```
blaisb, line 1836: Just a question, but is it really better to have std::vectors of nodes_positions? Why not have a vector of Tensor<1,dim> or Tensor<1,3>?

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -370,6 +352,12 @@ void
 GLSSharpNavierStokesSolver<dim>::refinement_control(
   const bool initial_refinement)
 {
+  // We add a post-refinement check to update precalculations only if needed
+  bool triangulation_modified_here          = false;
+  bool triangulation_modified_at_least_once = false;
+  this->triangulation->signals.post_refinement.connect(
+    [&triangulation_modified_here]() { triangulation_modified_here = true; });
```
blaisb, line 359: Good usage of boost signals
But why is the triangulatio_modified_at_least_once not connected? I'm not sure I follow what is being done here. You need to comment it more please

source/core/shape.cc
```diff
@@ -1576,35 +1606,46 @@ RBFShape<dim>::gradient(const Point<dim> &evaluation_point,
...
-          // field is already imperfect and neighboring nodes will add their
-          // contribution at the evaluation point.
-          for (int d = 0; d < dim; d++)
-            dr_dx_derivative[d] = 0;
+          // Calculation of dr/dx
+          relative_position =
+            (evaluation_point - rotated_nodes_positions[node_id]);
+          distance            = (relative_position).norm();
+          normalized_distance = distance / support_radii[node_id];
+          if (distance > 1e-16)
```
blaisb, line 1621: Not the biggest fan of using tolerances which are not based on a dimension of the problem when it's a geometry thing

blaisb (changes requested):
First batch of changes. I will need to re-review the shape.cc file tomorrow

doc/source/parameters/sharp-immersed-boundary-solver/sharp-immersed-boundary-solver.rst
oguevremont: Not always, only for cases where some RBF nodes are outside of the background grid and where perfect shape representation is needed. For example, in MMS cases where we use symmetry and the RBF shape is cut by the grid boundary, we must deactivate mesh precalculations. However, for large simulations of complex shapes we must leave ``mesh-based precalculations=true``, otherwise all cells would have to loop on all RBF nodes, making it too costly. I plan to fix this issue soon and assign each outside node to its closest boundary cell so we can keep mesh-based precalculations active.

source/fem-dem/gls_sharp_navier_stokes.cc
oguevremont: It is to separate the update_precalculations and remove_superfluous_nodes steps. I added more comments to explain it. Is it clearer this way?

blaisb (approved):

PR #830, opened 2023-08-01
Add isothermal compressible NSVOF assemblers

AmishgaAlphonius:
# Description of the problem

- When experimenting with the dam break with an obstacle example, it was noticed that the software performed poorly in mimicking the wave's natural shape after the impact. It was assumed that this is due to the lack of considering the air's natural compressibility.

# Description of the solution

- In an attempt to solve this issue, an isothermal compressible fluid dynamics Volume of Fluid (VOF) model is implemented. In a previous PR, an isothermal ideal gas density model was implemented. As a second step, an isothermal compressible formulation of the Navier-Stokes (NS) equations was implemented. Only the continuity equation was adapted to take account of compressibility. Finally, in this PR, the compressible NSVOF assemblers has been implemented.
- As it was previously assessed, in the VOF model, the divergence of velocity is not necessarily equal to zero especially near the interface, the missing term (``phase * div_u``) has been added to the VOF equation as it appears in the theory guide.

# How Has This Been Tested?

- Two application tests were added with respectively one and two isothermal compressible fluids: 

  - [x] /applications_tests/gls_navier_stokes_2d/gls_vof_1_isothermal_compressible_fluid.prm (.output)
  - [x] /applications_tests/gls_navier_stokes_2d/gls_vof_2_isothermal_compressible_fluids.prm (.output)

- All VOF application test outputs were updated following the addition of the ``phase * div_u`` term (changes in value were minor)

  - [x] applications_tests/gls_navier_stokes_2d/gls_droplet_marangoni_effect.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_dirichlet_boundary_condition.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_heat-transfer_hydrostat.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_ht_adapt-dissipation.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat_initial_refine.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_hydrostat_mesh-adapt.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_static_droplet_surface_tension_force_with_tanh_filter.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_tanh_filter.output
  - [x] applications_tests/gls_navier_stokes_2d/gls_vof_wetting_mesh-adapt.output

# Documentation

- Documentation on the isothermal compressible density model was updated.

- [x] doc/source/parameters/cfd/physical_properties.rst

# Future changes

- The 2D and 3D dam break examples will have to be updated. For the 2D dam break example, it would be interesting to compare the results of a constant density versus a fluctuating density.
- For two-phase flow modeling, instead of calculating the dynamic viscosity from the kinematic viscosity, it may be better to add a dynamic viscosity parameter with its rheology model and therefore reduce the number of calculations.

# Comments

- The condition on continuity for peeling/wetting boundary conditions was not added as it is not used.
- The display of the vof_barycenter was fixed so that the output is correct.
- The density is outputted for every fluid that does not have a constant density.


source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -376,6 +383,31 @@ NavierStokesScratchData<dim>::calculate_physical_properties()
...
+            {
+              density_ref_0 = density_model_0->get_density_ref();
+              density_psi_0 = density_model_0->get_psi();
+              density_ref_1 = density_model_1->get_density_ref();
+              density_psi_1 = density_model_1->get_psi();
+
+              for (unsigned int q = 0; q < this->n_q_points; ++q)
+                {
+                  double filtered_phase_value = this->filtered_phase_values[q];
+                  // To be able to calculate of the equivalent dynamic viscosity
```
oguevremont, line 398:
```suggestion
                  // To be able to calculate of the equivalent dynamic density
```

oguevremont (commented):
Well done. Only a few comments/questions

lpsaavedra (approved):
Good job! Everything seems good to me.

source/solvers/CMakeLists.txt
```diff
@@ -49,6 +50,7 @@ add_library(lethe-solvers
   ../../include/solvers/heat_transfer_scratch_data.h
   ../../include/solvers/initial_conditions.h
   ../../include/solvers/isothermal_compressible_navier_stokes_assembler.h
+        ../../include/solvers/isothermal_compressible_navier_stokes_vof_assembler.h
```
PierreLaurentinCS, line 53:
```suggestion
  ../../include/solvers/isothermal_compressible_navier_stokes_vof_assembler.h
```

PierreLaurentinCS (commented):
Very good work, I've left some comments/suggestions that you can review

PR #831, opened 2023-08-02
Sloshing example update postprocess script

AmishgaAlphonius:
# Description of the problem

- The python script for postprocessing was using the pvd file name of the Re = 2 case. Therefore, when the Re = 2 case was not run the postprocessing script could not be used.

# Description of the solution

- Using the prm file name, which is an inputted argument when calling the postprocessing script, the pvd file of the correct case is generated.

# How Has This Been Tested?

- The different cases were runned and postprocessed.

# Documentation

- The example was updated (add figure, change reference format):

- [x] doc/source/examples/multiphysics/sloshing-in-rectangular-tank/sloshing-in-rectangular-tank.rst


OGaboriault (approved):
Everything seems to be fine. Good work.

Luckabarbeau (approved):
Same over here. Everything looks good to me!

PR #836, opened 2023-08-09
Add matrix free Navier Stokes operator

lpsaavedra:
# Description of the problem

This PR is a follow up on the matrix free Navier-Stokes solver shell created in PR #755. It adds a first working version of the matrix free Navier-Stokes solver in Lethe by implementing a matrix free operator. 

# Description of the solution
- The application was renamed and updated according to the new dimension parameter added in Lethe recently.
- For the matrix free operators, a base class was added and a SUPG-PSPG derived class was implemented.
- The MF Navier-Stokes solver was updated to use the system operator instead of the system matrix as in the traditional Navier Stokes solver. 

# How Has This Been Tested?

The application is tested in 2D and 3D by using a MMS. Two element orders are considered in 2D and one in 3D. The following tests were added:

- [x] applications_tests/mf_navier_stokes/mms2d_fe1.prm
- [x] applications_tests/mf_navier_stokes/mms2d_fe2.prm
- [x] applications_tests/mf_navier_stokes/mms3d_fe1.prm

# Documentation

- No documentation has been added as the solver is not fully functional yet from the user point of view. 

# Future changes

- At the moment, the solver supports only GMRES without preconditioning as linear solver. The next step is to implement the different MG preconditioners. This will lead to the refactoring of linear solvers throughout Lethe.


include/solvers/mf_navier_stokes.h
```diff
@@ -146,30 +150,22 @@ class MFGLSNavierStokesSolver
...
-  void
-  setup_GMG();
-
-  /**
-   * Members
-   */
 protected:
-  // TODO create operator: NSOperator<dim, double> system_matrix;
+  // Matrix-free operator
+  NavierStokesOperatorBase<dim, double> *system_operator;
```
peterrum, line 168: Raw pointer?

include/solvers/mf_navier_stokes_operators.h
```diff
@@ -0,0 +1,360 @@
...
+   * classes.
+   *
+   * @param integrator FEEvaluation object that allows to evaluate functions at
+   * quadrature points and perform cell integrations.
+   */
+  virtual void
+  do_cell_integral_local(FECellIntegrator &integrator) const;
+
+  /**
+   * @brief Loop over all cell batches withing certain range and perform a cell integral with access to global vectors, i.e., gathering and scattering values.
```
peterrum, line 242: Wrap line.

source/solvers/navier_stokes_base.cc
```diff
@@ -1456,6 +1456,9 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
           // Update the time of the exact solution to the actual time
           this->exact_solution->set_time(
             simulation_control->get_current_time());
+
+          present_solution.update_ghost_values();
```
peterrum, line 1460: Could you `zero_out_ghost_values()` at the end?

source/solvers/mf_navier_stokes.cc
```diff
@@ -408,46 +442,111 @@ MFGLSNavierStokesSolver<dim>::define_zero_constraints()
...
+  const double absolute_residual =
+    this->simulation_parameters.linear_solver.minimum_residual;
+  const double relative_residual =
+    this->simulation_parameters.linear_solver.relative_residual;
+
+  if (this->simulation_parameters.linear_solver.solver ==
+      Parameters::LinearSolver::SolverType::gmres)
+    solve_system_GMRES(initial_step, absolute_residual, relative_residual);
+  else
+    throw(std::runtime_error("This solver is not allowed"));
```
peterrum, line 457: AsserThrow?

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -0,0 +1,810 @@
...
+          const unsigned int count_global =
+            Utilities::MPI::sum(count, dof_handler.get_communicator());
+
+          const unsigned int count_cells_global =
+            Utilities::MPI::sum(matrix_free.n_cell_batches(),
+                                dof_handler.get_communicator());
+
+          if (Utilities::MPI::this_mpi_process(
+                dof_handler.get_communicator()) == 0)
+            std::cout << count_global << " " << count_cells_global << std::endl;
```
peterrum, line 194: Shouldn't this be optional?

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -0,0 +1,810 @@
...
+    }
+}
+
+template <int dim, typename number>
+void
+NavierStokesOperatorBase<dim, number>::compute_element_size()
+{
+  FECellIntegrator phi(matrix_free);
+
+  unsigned int n_cells =
```
peterrum, line 205:
```suggestion
  const unsigned int n_cells =
```

peterrum (approved):

include/solvers/mf_navier_stokes.h
blaisb: std::shared would be cleaner here

include/solvers/mf_navier_stokes_operators.h
```diff
@@ -0,0 +1,360 @@
...
+protected:
+  MatrixFree<dim, number>                matrix_free;
+  AffineConstraints<number>              constraints;
+  mutable TrilinosWrappers::SparseMatrix system_matrix;
+  AlignedVector<VectorizedArray<number>> element_size;
+  unsigned int                           fe_degree;
+  const Function<dim> *                  forcing_function;
+  double                                 viscosity;
+
+  // Variables needed from the last newton step vector
```
blaisb, line 295:
```suggestion
  // Variables needed from the last Newton step vector
```

include/solvers/mf_navier_stokes_operators.h
```diff
@@ -0,0 +1,360 @@
...
+ * @tparam dim An integer that denotes the number of spatial dimensions.
+ * @tparam number Abstract type for number across the class (i.e., double).
+ */
+template <int dim, typename number>
+class NavierStokesSUPGPSPGOperator
+  : public NavierStokesOperatorBase<dim, number>
+{
+public:
+  using FECellIntegrator = FEEvaluation<dim, -1, 0, dim + 1, number>;
+  using VectorType       = LinearAlgebra::distributed::Vector<number>;
```
blaisb, line 326: Does the using need to be remade for every inherited class?

source/solvers/mf_navier_stokes.cc
```diff
@@ -199,49 +217,65 @@ MFGLSNavierStokesSolver<dim>::update_boundary_conditions()
...
+        "Type of initial condition is not supported by MF Navier-Stokes");
+    }
 }
 
 template <int dim>
 void
-MFGLSNavierStokesSolver<dim>::assemble_system_matrix()
+MFNavierStokesSolver<dim>::assemble_system_matrix()
 {
   // TODO
```
blaisb, line 249: Is there really a TODO here? it should be a blank function forever no?

source/solvers/mf_navier_stokes.cc
blaisb: yup should be

source/solvers/mf_navier_stokes_operators.cc
blaisb: At least if the output is not optional, it should be a bit more verbose in what it means

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -0,0 +1,810 @@
...
+    matrix_free.n_cell_batches() + matrix_free.n_ghost_cell_batches();
+  element_size.resize(n_cells);
+
+  for (unsigned int cell = 0; cell < n_cells; ++cell)
+    {
+      for (auto lane = 0u;
+           lane < matrix_free.n_active_entries_per_cell_batch(cell);
+           lane++)
+        {
+          double h_k = matrix_free.get_cell_iterator(cell, lane)->measure();
```
blaisb, line 215:
```suggestion
          const double h_k = matrix_free.get_cell_iterator(cell, lane)->measure();
```

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -0,0 +1,810 @@
...
+      auto previous_hessian_diagonal =
+        this->nonlinear_previous_hessian_diagonal(cell, q);
+
+      // Calculate tau
+      VectorizedArray<number> u_mag = VectorizedArray<number>(0.0);
+      VectorizedArray<number> tau   = VectorizedArray<number>(0.0);
+      for (unsigned int k = 0; k < dim; ++k)
+        {
+          u_mag += previous_values[k];
+        }
```
blaisb, line 593:
```suggestion
      for (unsigned int k = 0; k < dim; ++k)
          u_mag += previous_values[k];
```
one line for loop are cleaning I think in this case

blaisb (changes requested):
Tiny change. Once addressed this will be easy to merge

source/solvers/mf_navier_stokes_operators.cc
lpsaavedra: Yes, it should be optional and I will add a comment explaining the purpose of it.

include/solvers/mf_navier_stokes_operators.h
lpsaavedra: I think they cannot be inherited as it is right now. The only way would be to declare both classes inside a namespace and declare the using directive right before it.

blaisb (approved):

peterrum (approved):

PR #839, opened 2023-08-09
Deprecate SDIRK time integrators

blaisb:
# Description of the problem

- The SDIRK time integrators were poorly maintained and gave a false illusion that they were actually useable in contexts in which they were not.
-  No one was actually using the time integrators and they have worse performance than the equivalent BDF one

# Description of the solution

- Remove the SDIRK integrator. If we actually need them in the future, we can do a better reimplementation.

# Documentation

- Updated documentation


include/core/sdirk.h
```diff
@@ -59,111 +59,5 @@ FullMatrix<double>
 sdirk_coefficients(const unsigned int order, const double time_step);
```
oguevremont, line 61: Are we keeping some lines, just in case?

source/core/parameters.cc
```diff
@@ -171,19 +171,13 @@ namespace Parameters
         method = TimeSteppingMethod::bdf2;
       else if (sv == "bdf3")
         method = TimeSteppingMethod::bdf3;
```
oguevremont, line 173: Sdirk options are still in the declare_parameters explanatory text, they should be removed

source/fem-dem/gls_sharp_navier_stokes.cc
oguevremont, line 1: There is still #include <core/sdirk.h>, maybe it needs to be removed. It is also in other files

include/solvers/navier_stokes_base.h
```diff
@@ -443,9 +443,6 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   // Previous solutions vectors
   std::vector<VectorType> previous_solutions;
 
-  // Intermediary solution stages for SDIRK methods
-  std::vector<VectorType> solution_stages;
```
lpsaavedra, line 447: I think all the multiphysics solvers have this parameter as well and needs to be deleted? Cahn Hilliard, VOF, heat transfer, tracer?

lpsaavedra, line 447: I think all the multiphysics (heat transfer, tracer, vof and cahn hilliard) and it should be removed as well?

source/core/parameters.cc
lpsaavedra, line 1: In this file, sdirk options need to be removed when declaring the parameters of simulation control (between lines 61 and 70)...

oguevremont (commented):
Good! Some questions/comments, but nothing major

source/solvers/cahn_hilliard_scratch_data.cc
lpsaavedra, line 1: Related to my comment above: `cahn_hilliard_scratch_data.h` still uses solution_stages in several places

source/solvers/heat_transfer_scratch_data.cc
lpsaavedra, line 1: Related to my comment above: `heat_transfer_scratch_data.h` still uses solution_stages in several places

source/solvers/tracer_scratch_data.cc
lpsaavedra, line 1: Related to my comment above: `tracer_scratch_data.h` still uses solution_stages in several places

source/solvers/vof_scratch_data.cc
lpsaavedra, line 1: Related to my comment above: vof_scratch_data.h still uses solution_stages in several places

include/core/sdirk.h
lpsaavedra, line 1: Do we still want to have `sdirk.h` and `sdirk.cc` in `source/core/CMakeLists.txt`? We could remove it if it is not being used. We should also avoid including the file if not used (remove the `#include <core/sdirk.h>` from all files) in source solvers.

lpsaavedra (changes requested):
Some questions/comments need to be addressed before merging. Also, there are two places in the documentation where sdirk is still mentioned and this should be updated:

- [x] `/doc/source/examples/incompressible-flow/2d-transient-flow-around-cylinder/2d-transient-flow-around-cylinder.rst` line 304.
- [ ] `/doc/source/examples/sharp-immersed-boundary-solver/sedimentation-1-particle/sedimentation-1-particle.rst` line 64.

include/core/sdirk.h
blaisb: Yeah I'd rather leave the file for the coefficients, but not use it anymore. I'll clean that up.

lpsaavedra (approved):
Seems good to me! Only two minor additional changes: the #include directive should be removed also from: `vans_assemblers.cc` and `gls_sharp_navier_stokes.cc`.

blaisb:
> Seems good to me! Only two minor additional changes: the #include directive should be removed also from: `vans_assemblers.cc` and `gls_sharp_navier_stokes.cc`.

Done! I'll merge after CI :)

oguevremont (approved):

PR #847, opened 2023-08-16
Find contact detection step interval

blaisb:
# Description of the problem

- Currently, we use MPI to communicate at every iteration if the contact detection structure should be rebuilt. This is not necessarily efficient when a large number of core is used.

# Description of the solution

- I adapted the frequency parameter to also be used in dynamic contact search. In this case, it indicates the frequency at which MPI communications should be done. I have found that increasing the frequency to 2 or 4 can decrease simulation cost by 5-10% in some cases when we use 8 or more cores.

# How Has This Been Tested?

- All tests still pass
- If applicable, describe future changes or features that will be added in subsequent pull requests


voferreira (approved):
Reviewed and tested! Just one small comment :)

doc/source/parameters/dem/model_parameters.rst
```diff
@@ -73,7 +73,8 @@ Lethe rebuilds the contact lists automatically. In this mode, Lethe stores the d
...
 
-* ``dynamic contact search size coefficient`` is a safety factor to ensure the late detection of particles will not happen in the simulations with ``dynamic`` contact search; and its value should be defined generally in the range of 0.5-1. 0.5 is a rather conservative value.
+* ``dynamic contact search size coefficient`` is a safety factor to ensure the late detection of particles will not happen in the simulations with ``dynamic`` contact search; and its value should be defined generally in the range of 0.5-1. 0.5 is a rather conservative value. The default value of 0.8 is adequate for most simulations.
+* ``frequency`` controls the frequency at which the dynamic contact search is carried out. For most cases, the default value of 1 should be maintained to ensure that the dynamic contact detection is refreshed accurately. Increasing this value between 2 and 5 can decrease the computational cost when a large (>16) number of cores is used since this diminishes the number of MPI communications.
```
OGaboriault, line 77: Knowing this, shouldn't we put line 19 of this file to 1 ? 

set frequency = 1 # instead of 10

blaisb: We should indeed.

PR #848, opened 2023-08-16
Announce Fluid Dynamics Solver

blaisb:
# Description of the problem

- Although all multiphysics are announced before being solved for, this is not the case for the fluid dynamics. This creates unncessary confusion when we have a multiphysics solver

# Description of the solution

- Announce the fluid dynamics. Regrettably this forced me to update a tremendous amount of test. Some numbers might have changed, but that's because the compiler I am using is not the same than the one that was used when creating the tests.

# How Has This Been Tested?

- All tests should pass

# Comments

- This is a first step before having a timer per Physics (for which I will also make a PR today)


lpsaavedra (commented):
Everything looks good! Only one remaining test needs to be updated: `lethe/applications_tests/gls_sharp_navier_stokes/opencascade/static_stokes_step.output`

source/solvers/navier_stokes_base.cc
```diff
@@ -558,6 +558,7 @@ NavierStokesBase<dim, VectorType, DofsType>::iterate()
                           simulation_parameters.simulation_control.method);
       multiphysics->percolate_time_vectors(false);
 
+      announce_string(this->pcout, "Fluid Dymamics");
```
AmishgaAlphonius, line 561: This is quite a sad typo, all tests will have to be rechanged :')
```suggestion
      announce_string(this->pcout, "Fluid Dynamics");
```

AmishgaAlphonius (commented):
Other than the last test, there's a little typo in "Dynamics" :')

source/solvers/navier_stokes_base.cc
blaisb: No worries, that takes 5 seconds really

AmishgaAlphonius (approved):

lpsaavedra (approved):

PR #853, opened 2023-08-21
Physic based linear solvers

lpsaavedra:
# Description of the problem
Until now, it was not possible in Lethe to specify different parameters for the linear solver for each of the physics involved in a simulation (e.g., fluid dynamics + VOF, fluid_dynamics + heat transfer). 

# Description of the solution
Instead of having one struct for the linear solver parameters, a map of structs was created in order to have a different subsection for all of the physics. Therefore, all of the linear solver parameters can be now specified for each physic and they all have the same default values. 

# How Has This Been Tested?

- All the parameter files in the application tests were updated in order to have a subsection in the linear solver parameters subsection that accounted for each of the physics involved. None of the outputs needed to be updated.
- The `tests/solvers/restart_01.cc` test was also updated as it uses the verbosity parameter of the linear solver parameters. 

# Documentation
- The main page describing the linear solver parameters was updated to mention the possibility of specifying parameters for each physic. 
- The parameter files of the examples were also updated accordingly. Therefore, the examples that explicitly mentioned the linear solver section in the documentation had to be updated as well. 

# Future changes
A following PR will refactor the linear solver across all applications, which will lead to the addition of a  "preconditioner" parameter to the linear solver parameters for each physic.

### DIFF
--- a/applications_tests/cfd_dem_coupling/dynamic_contact_search.prm
+++ b/applications_tests/cfd_dem_coupling/dynamic_contact_search.prm
@@ -217,12 +217,14 @@ end
 #---------------------------------------------------
 
 subsection linear solver
-  set method                                = gmres
-  set max iters                             = 5000
-  set relative residual                     = 1e-3
-  set minimum residual                      = 1e-10
-  set ilu preconditioner fill               = 0
-  set ilu preconditioner absolute tolerance = 1e-12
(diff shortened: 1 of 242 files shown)


blaisb (approved):
Very good job. I'll merge once CI passes so that this does not create conflicts all over the place.

PR #854, opened 2023-08-23
Refactor surface tension gradient parameter

AmishgaAlphonius:
# Description of the problem
- After creating a new material interaction class to set interface physical properties, a ``constant`` surface tension model was added (#818). However, to consider Marangoni effet, a linear surface tension model is needed. This model was implicitly implemented in the assemblers.
# Description of the solution
- With this PR, a temperature-dependent linear surface tension model is added to the surface tension models.
# How Has This Been Tested?
- The ``.prm`` file of the gls_droplet_marangoni_effect application test was updated and a new unit test for the linear surface tension model was added.
- [x] applications_tests/gls_navier_stokes/gls_droplet_marangoni_effect.prm
- [x] tests/core/surface_tension_linear.cc (.output)
# Documentation
- Documentation on surface tension was updated:
- [x] doc/source/parameters/cfd/physical_properties.rst
- [x] doc/source/parameters/cfd/volume_of_fluid.rst
# Future changes
- For non-time dependent temperature schemes, it would be useful to implement a position-dependent surface tension model that would avoid the need to solve the heat transfer physics to get the temperature.
- Review the formulation of the linear surface tension model, by adding a reference temperature for ``sigma_0``.
- Deprecate peeling and wetting. 
# Comments
- VOF parameters documentation was restructured with new subtitles to improve the reading experience.

PR #858, opened 2023-08-29
Add reference state temperature to linear surface tension model

AmishgaAlphonius:
# Description of the problem

- When working on PR #854, it was noticed that the linear surface tension model could only be used with a surface tension coefficient evaluated at T = 0 K.

# Description of the solution

- A ``reference state temperature`` parameter was add to SurfaceTensionParameters.

# How Has This Been Tested?

- All tests have passed.
- Unit test with linear surface tension model was updated: tests/core/surface_tension_linear.cc (.output)

# Documentation

- Physical properties documentation was updated: doc/source/parameters/cfd/physical_properties.rst

### DIFF
--- a/doc/source/parameters/cfd/physical_properties.rst
+++ b/doc/source/parameters/cfd/physical_properties.rst
@@ -61,6 +61,7 @@ Physical Properties
         # Surface tension
         set surface tension model                       = constant
         set surface tension coefficient                 = 0
+        set reference state temperature                 = 0
         set temperature-driven surface tension gradient = 0
       end
     end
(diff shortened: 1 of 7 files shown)


blaisb (approved):

doc/source/parameters/cfd/physical_properties.rst
```diff
@@ -122,6 +123,8 @@ Physical Properties
 
       where :math:`Re` is the Reynolds number, :math:`\mu_\text{ref}` and :math:`u_\text{ref}` are some reference viscosity and velocity characterizing the flow problem, and :math:`\sigma` is the surface tension coefficient.
 
+    * The ``reference state temperature`` parameter is the temperature of the reference state at which the surface ``surface tension coefficient`` is evaluated.
```
hepap, line 126:
```suggestion
    * The ``reference state temperature`` parameter is the temperature of the reference state at which the ``surface tension coefficient`` is evaluated. This parameter is used in the calculation of the surface tension using the ``linear`` surface tension model (see `Surface Tension Models`_).
```

hepap (approved):
Only a small comment, nice addition :D

PR #856, opened 2023-08-28
Preconditioner param linear solver

lpsaavedra:
# Description of the problem

The main linear solvers in Lethe were `gmres` for GMRES preconditioned with ILU and `amg` for GMRES preconditioned with AMG; this was confusing and did not represent the actual linear solver being used. Due to this the linear solvers were refactored.

# Description of the solution

A new parameter specifying the preconditioner was created. Now, it is possible to choose as method GMRES preconditioned by either AMG or ILU.  In the `gls_navier_stokes` solver and the `gd_navier_stokes` solver, the `solve_system_AMG` method was removed and now the preconditioner is chosen within the `solve_system_GMRES` call.

# How Has This Been Tested?

All the application tests that specified linear solver parameters were updated with the new parameter. If `gmres` was specified, the preconditioner was set to `ilu`, if `amg` was specified the method was set to `gmres` and the preconditioner to `amg`.

# Documentation

- [X] `doc/source/parameters/cfd/linear_solver_control.rst`: linear solver documentation in the parameters section
- [X] All the examples that contain a linear solver section in the documentation were updated as well (both their .prm file and the .rst file).

# Comments

This change not only improves the linear solvers in Lethe, but leaves a better infrastructure if other preconditioners want to be added (as it is the case for the matrix-free solvers).

PR #860, opened 2023-08-30
Fix if condition in preconditioner setup

lpsaavedra:
# Description of the problem

The preconditioner was being setup more than it should for the gls navier stokes solver.

# Description of the solution

Fix if condition and built the preconditioner only when really needed.

# Comments

The bug was introduced when the preconditioner parameter was introduced for the linear solver in PR #856.


blaisb (approved):

blaisb:
will merge once Release passes :)

PR #863, opened 2023-08-31
Physic based non linear solvers

hepap:
# Description of the problem
Until now, it was not possible in Lethe to specify different parameters for the non-linear solver for each of the physics involved in a simulation (e.g., fluid dynamics + VOF, fluid_dynamics + heat transfer).

# Description of the solution
Instead of having one struct for the non linear solver parameters, a map of structs was created in order to have a different subsection for all of the physics. Therefore, all of the linear solver parameters can be now specified for each physic and they all have the same default values.

# How Has This Been Tested?

- All the parameter files in the application tests were updated in order to have a subsection in the linear solver parameters subsection that accounted for each of the physics involved. None of the outputs needed to be updated.

# Documentation

- [x] The associated documentation was updated

# Comments

- verbosity  in the multiphysics_interface class was change also in order to announce the physics properly


lpsaavedra (approved):
Great work! No comments from my side...

blaisb (approved):
Ready for merge I think

PR #864, opened 2023-09-01
Fix matrix-free Navier-Stokes operator

lpsaavedra:
# Description of the problem

There was a bug in the operator that was only noticed when using Q3 elements for the current MMS problems in 2 and 3D. While debugging also a small error in the calculation of the u mag used in the stabilization parameter was noticed.

# Description of the solution

The problem was found in the SUPG terms of both the Jacobian and the residual expressions. The corresponding terms were fixed and now we get the same results (error and order of convergence) for matrix based and matrix free Navier-Stokes. These terms were fixed for both the MF Navier-Stokes prototype and the existing SUPG/PSPG operator in Lethe. In addition, the calculation of u mag for the stabilization parameter was also corrected.

# How Has This Been Tested?

The existing matrix free tests were updated with only minor changes in the results as they considered only Q1 and Q2 elements:
- [X] `application_tests/mf_navier_stokes/mms2d_fe1`
- [X] `application_tests/mf_navier_stokes/mms2d_fe2`
- [X] `application_tests/mf_navier_stokes/mms3d_fe1`

An additional test was added that considers an MMS in 2D with Q3 elements in order to avoid similar issues:
- [X] `application_tests/mf_navier_stokes/mms2d_fe3`

# Comments

Now that this bug was fixed, we can continue with the preconditioners implementation.


blaisb (approved):
Yay!

PR #868, opened 2023-09-06
Dem insertion list velocity, omega and diameters

OGaboriault:
# Description of the problem

- The list insertion method was only allowing the insertion of static particles. Now, every particle can be inserted with a specific transnational and rotational velocity. The diameter of every particles can also be specify. 

# Description of the solution

- 7 new parameters. 
- A new assign_particle_properties_for_list_insertion function has been created.

# How Has This Been Tested?

- Test will be added if requested.
- The bouncing particle was use to test the different parameters, with success.

# Documentation

- Insertion section.


blaisb:
@voferreira can you also review this? You know this part well

voferreira:
> @voferreira can you also review this? You know this part well
> 

@blaisb Sure thing!

source/core/parameters_lagrangian.cc
```diff
@@ -437,6 +478,53 @@ namespace Parameters
...
+
+        // Read the diameters list as a single string
+        std::string d_str = prm.get("list diameters");
+
+        // Convert the diameters list to a vector of strings
+        std::vector<std::string> d_str_list(
+          Utilities::split_string_list(d_str));
+
+        // Convert the diameters string vector to a double vector
+        list_d = Utilities::string_to_double(d_str_list);
```
voferreira, line 526: I think a good approach to list insertion would be to get the number of particles from the maximum size among vectors and then repeat the default value or a given single value for the remaining particles if values are not provided. Otherwise, you will always need the ``set list ... = 0.0, 0.0, 0.0......`` for each vector, and it is not necessary if you want to have the same value for all particles.

source/dem/list_insertion.cc
```diff
@@ -20,27 +45,64 @@ ListInsertion<dim>::ListInsertion(
...
+  const auto &list_vy = dem_parameters.insertion_info.list_vy;
+  const auto &list_vz = dem_parameters.insertion_info.list_vz;
+  const auto &list_wx = dem_parameters.insertion_info.list_wx;
+  const auto &list_wy = dem_parameters.insertion_info.list_wy;
+  const auto &list_wz = dem_parameters.insertion_info.list_wz;
+  const auto &list_d  = dem_parameters.insertion_info.list_d;
 
   Assert(list_x.size() == list_y.size(),
-         ListSizeCoherence(list_x.size(), list_y.size()));
+         PositionSizeCoherence(list_x.size(), list_y.size()));
```
voferreira, line 60: As before, I would use the maximum vector size here to check for coherence. Maybe you could create a map with list names as keys and sizes as values and then loop assert instead of doing it one by one. I think it would be cleaner this way and you could still print the name (key) of the incoherent list.

source/dem/list_insertion.cc
```diff
@@ -20,27 +45,64 @@ ListInsertion<dim>::ListInsertion(
...
+          insertion_velocities.emplace_back(
+            Tensor<1, 3>({list_vx[i], list_vy[i], 0.}));
+          insertion_angular_velocities.emplace_back(
+            Tensor<1, 3>({0., 0., list_wz[i]}));
+        }
       else
-        insertion_points.emplace_back(
-          Point<dim>({list_x[i], list_y[i], list_z[i]}));
+        {
+          insertion_points.emplace_back(
```
voferreira, line 98: The map would be useful here as well

include/dem/list_insertion.h
```diff
@@ -62,13 +62,26 @@ class ListInsertion : public Insertion<dim>
...
   unsigned int current_inserting_particle_type;
 
-  // Vector of the location used to insert the particles
-  std::vector<Point<dim>> insertion_points;
+  // Vector of the location, velocity and angular velocity used to insert the
+  // particles
+  std::vector<Point<dim>>   insertion_points;
+  std::vector<Tensor<1, 3>> insertion_velocities;
+  std::vector<Tensor<1, 3>> insertion_angular_velocities;
+  std::vector<double>       insertion_diameters;
```
blaisb, line 84:
```suggestion
  std::vector<Point<dim>>    positions;
  std::vector<Tensor<1, 3>> velocities;
  std::vector<Tensor<1, 3>> angular_velocities;
  std::vector<double>       diameters;
```

source/dem/list_insertion.cc
```diff
@@ -20,27 +45,64 @@ ListInsertion<dim>::ListInsertion(
...
+                                           list_wy.size(),
+                                           list_wz.size()));
     }
+  Assert(list_x.size() == list_d.size(),
+         DiameterSizeCoherence(list_x.size(), list_d.size()));
 
   // Generate vector of insertion position
   for (unsigned int i = 0; i < list_x.size(); ++i)
     {
       if (dim == 2)
```
blaisb, line 88:
```suggestion
      if constexpr (dim == 2)
```

source/dem/list_insertion.cc
```diff
@@ -20,27 +45,64 @@ ListInsertion<dim>::ListInsertion(
...
-        insertion_points.emplace_back(Point<dim>({list_x[i], list_y[i]}));
+        {
+          insertion_points.emplace_back(Point<dim>({list_x[i], list_y[i]}));
+          insertion_velocities.emplace_back(
+            Tensor<1, 3>({list_vx[i], list_vy[i], 0.}));
+          insertion_angular_velocities.emplace_back(
+            Tensor<1, 3>({0., 0., list_wz[i]}));
+        }
       else
```
blaisb, line 96: if constexpr (dim==3)

blaisb (changes requested):
few comments

voferreira (approved):
Good job! I pointed out some small typos here and there in the comments. Apart from that, the comparison with list_x size is something I really don't like, but it is not a big deal. But on my side, it is good to go.

include/dem/list_insertion.h
```diff
@@ -62,13 +62,26 @@ class ListInsertion : public Insertion<dim>
...
          const DEMSolverParameters<dim> &dem_parameters) override;
 
+
+
+  void
+  assign_particle_properties_for_list_insertion(
+    const DEMSolverParameters<dim> &  dem_parameters,
+    const unsigned int &              inserted_this_step_this_proc,
+    const unsigned int &              current_inserting_particle_type,
+    std::vector<std::vector<double>> &particle_properties);
```
blaisb, line 72: You need to add the @brief section and the description of all parameters ;)

source/core/parameters_lagrangian.cc
```diff
@@ -437,6 +461,69 @@ namespace Parameters
...
+          Utilities::split_string_list(vy_str));
+        std::vector<std::string> vz_str_list(
+          Utilities::split_string_list(vz_str));
+
+        // Convert vx,vy and vz string vector to double vector
+        list_vx = Utilities::string_to_double(vx_str_list);
+        list_vy = Utilities::string_to_double(vy_str_list);
+        list_vz = Utilities::string_to_double(vz_str_list);
+
+        // Fill the velocity vector with zeros to match the size of list_x
```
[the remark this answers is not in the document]
blaisb: bah it's the most likely thing people would vary first right
It makes absolutely zero sense to insert two particles at the same location, hence using list_x as a marker makes sense to me at least.
It's not a bad choice

source/dem/list_insertion.cc
```diff
@@ -20,27 +26,46 @@ ListInsertion<dim>::ListInsertion(
...
+  const auto &list_d  = dem_parameters.insertion_info.list_d;
 
   Assert(list_x.size() == list_y.size(),
-         ListSizeCoherence(list_x.size(), list_y.size()));
+         PositionSizeCoherence(list_x.size(), list_y.size()));
   if (dim == 3)
     {
-      Assert(list_y.size() == list_z.size(),
-             ListSizeCoherence(list_y.size(), list_z.size()));
+      Assert(list_x.size() == list_z.size(),
```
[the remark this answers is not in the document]
blaisb: We should kind-off force that list_x list_y and list_z be at least the same size. It makes no sense not to specify the positions

blaisb (changes requested):
Very cool.One tiny change, can you add an application test where the autocomplete feature for the properties not specified is actually used? This would amke this complete IMO

blaisb:
@OGaboriault from my perspective this is ready for a merge, just confirm it is and I will merge it

voferreira:
> @OGaboriault from my perspective this is ready for a merge, just confirm it is and I will merge it
> 

From mine too 😀

OGaboriault:
Can be merge after the CI checks. 

I have just remove the Assert that check the size of the position lists since they are always resize to the maximum size in the parsing function.

PR #869, opened 2023-09-08
Deprecate peeling wetting

blaisb:
# Description of the problem

- The peeling and wetting mechanism was very ad-hoc and not being used anymore. It was also not robust because of all the changes made to streamline the VOF solver. This is a right time to deprecate it and, if need be in the future or if we find people needing a similar mechanism, we can re-introduce it more robustly (using the compressible equation of state with it for example)

# Description of the solution

- Deprecate

# How Has This Been Tested?

- peeling-wetting tests have been removed

# Documentation

- Peeling-wetting documentation has been removed

# Comments

- It's cool to deprecate stuff from time to time. I think it's a healthy reflection on the state of the code
- I will deprecate the "conservative fluid" stuff next


hepap (commented):
There are comments mentioning the wetting/peeling
`navier_stokes_vof_assemblers.cc`, lines 13,244,677,908
`vof_assemblers.cc` lines 29,167
`vof_scratch_data.h` line 232

include/core/boundary_conditions.h
```diff
@@ -974,9 +970,9 @@ namespace BoundaryConditions
   {
     prm.declare_entry("type",
                       "none",
-                      Patterns::Selection("none|dirichlet|peeling/wetting"),
+                      Patterns::Selection("none|dirichlet"),
                       "Type of boundary condition for VOF"
-                      "Choices are <none|dirichlet|peeling/wetting>.");
+                      "Choices are <none|dirichlet.");
```
hepap, line 975:
```suggestion
                      "Choices are <none|dirichlet>.");
```

source/solvers/vof_scratch_data.cc
```diff
@@ -47,36 +47,6 @@ VOFScratchData<dim>::allocate()
...
-{
-  switch (properties_manager.get_number_of_fluids())
-    {
-        case 1: {
-          // Cannot proceed a two-phase flow simulations if only one fluid is
-          // defined. This error must have been thrown a long time ago by the
-          // code, it is repeated here for clarity.
-          throw std::runtime_error(
-            "Cannot run VOF simulations with only 1 fluid, review the Physical Properties section of the .prm");
-        }
```
hepap, line 68: Do we have this check somewhere else? I didn't find it :o

hepap (commented):
Only two things left I think :)

source/solvers/vof_scratch_data.cc
blaisb: Nope, that's a good point, but it also should not be there in the scratch data. I will add the assertion in the vof constructor, that's where it belongs

hepap (approved):

AmishgaAlphonius (approved):
Two last changes :)

include/solvers/vof.h
AmishgaAlphonius, line 1: line 243 mentions ``wetting/peeling``

PR #871, opened 2023-09-08
Remove unused AdditionalData object in the AMG preconditioner

lpsaavedra:
# Description of the problem

There was an Additional Data object being allocated for the AMG preconditioner that was not being used. 

# Description of the solution

The object was removed.

# How Has This Been Tested?

This change should not affect any tests.


blaisb (approved):
Good catch. Will merge after CI :)

PR #873, opened 2023-09-11
Deprecate conservative fluid parameter

blaisb:
# Description of the problem

- The conservative fluid parameter was used to neglect mass conservation in one of the fluid. It generally led to very poor results and added weird if statements everywhere in the VOF solver

# Description of the solution

- Deprecate it. Better results can be obtained with the equation of state module instead

# How Has This Been Tested?

- Tests results remain unchanged

# Documentation

- Documentation elements were removed.


AmishgaAlphonius (approved):
Only two small changes :)

hepap (approved):
I have nothing to add :)

PR #883, opened 2023-09-19
Replace 'adaptative' by 'adaptive' in interface sharpening methods

AmishgaAlphonius:
# Description of the problem

- In VOF, the adaptive sharpening method was named "adaptative" and both terms were used in the documentation leading to unnecessary confusion.

# Description of the solution

- The "adaptative" sharpening method was renamed "adaptive" sharpening method.

# How Has This Been Tested?

- All tests passed and ``applications_tests/lethe-fluid/interface_sharpening_blurred_circle_adaptive_mass_conservation.prm`` was updated.

# Documentation

- All relevant documentation was modified.

### DIFF
--- a/source/core/parameters_multiphysics.cc
+++ b/source/core/parameters_multiphysics.cc
@@ -15,7 +15,7 @@ DeclException1(
   double,
   << "Sharpening threshold max deviation : " << arg1
   << " is smaller than 0 or larger than 0.5." << std::endl
-  << "Adaptative interface sharpening requires a maximum deviation of the"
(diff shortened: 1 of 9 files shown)


source/core/parameters_multiphysics.cc
```diff
@@ -169,10 +169,10 @@ Parameters::VOF::parse_parameters(ParameterHandler &prm)
```
PierreLaurentinCS, line 170:
```suggestion
```

PierreLaurentinCS (approved):
Good work, nothing to say except for one small line

blaisb (approved):

PR #884, opened 2023-09-20
Add multigrid preconditioners to matrix-free application

lpsaavedra:
# Description of the problem

Until now the matrix-free application in Lethe did not have any preconditioner suitable for the matrix-free approach. 

# Description of the solution

This PR adds two geometric multigrid approaches: i) local smoothing (LSMG) and ii) global coarsening (GCMG). They were added to the matrix-free application directly. They are used as preconditioners for a GMRES solver; they use the relaxation smoother, and a GMRES coarse-grid solver that uses AMG as preconditioner. The parameters for the coarse-grid solver, the smoother and general MG can be specified via the parameter file. 

# How Has This Been Tested?

A simple MMS test in 3D and using FE degree 1 was added to test the LSMG preconditioner:
- [X] `applications_tests/lethe-fluid-matrix-free/mms3d_fe1_lsmg.prm`

The output of several tests unrelated to matrix-free had to be updated. The reason for this is that the triangulation used to have "mesh smoothing", but it was changed to "limit level difference at vertices" as required for the MG preconditioners. This changed slightly the number of cells and therefore the results in several tests.

# Documentation

The MG parameters were added to the linear solver section in the documentation:
- [X] `doc/source/parameters`

# Future changes
- The next step after merging this PR is to support the different possible BCs and to implement the transient part of the operator.

# Comments
- There seems to be a bug with the global coarsening algorithm. It runs, but it is extremely slow and the iteration counts suggest that something is not set correctly. A similar test than the one for LSMG can be added, however, at the moment it takes too long.


include/core/parameters.h
```diff
@@ -937,13 +938,16 @@ namespace Parameters
     // ILU or ILUT relative tolerance
     double ilu_precond_rtol;
 
-    // ILU or ILUT fill
+    // AMG parameters either as linear solver preconditioner or as
+    // preconditioner of a coarse-grid solver for LSMG or GCMG
+
+    // ILU or ILUT fill for smoother and coarsener
```
peterrum, line 944: what is a `coarsener`?

include/solvers/mf_navier_stokes.h
```diff
@@ -30,13 +37,15 @@ using namespace dealii;
  * @tparam dim An integer that denotes the dimension of the space in which
  * the flow is solved.
  */
-template <int dim>
+template <int dim, typename OperatorType>
```
peterrum, line 40: I would say that you don't need to make `OperatorType` a template parameter. Could you post the compilation error which forced to used this.

include/solvers/mf_navier_stokes_operators.h
```diff
@@ -67,7 +68,7 @@ class NavierStokesOperatorBase : public Subscriptor
   NavierStokesOperatorBase(const Mapping<dim>              &mapping,
                            const DoFHandler<dim>           &dof_handler,
                            const AffineConstraints<number> &constraints,
-                           const Quadrature<dim>           &quadrature,
+                           const Quadrature<1>             &quadrature,
```
peterrum, line 71: Let's keep a dim-quadrature. This is what you need for simplices.

source/core/parameters.cc
```diff
@@ -2083,6 +2083,53 @@ namespace Parameters
           "false",
           Patterns::Bool(),
           "A boolean that will force the linear solver to continue even if it fails");
+
+        prm.declare_entry("mg min level",
```
peterrum, line 2087: Oje. `declare_entry` is still used... I think one should prefer `add_parameter()`.

source/solvers/mf_navier_stokes.cc
```diff
@@ -264,23 +276,705 @@ MFNavierStokesSolver<dim>::assemble_system_rhs()
...
+  mg_operator->compute_inverse_diagonal(
+    chebyshev_additional_data.preconditioner->get_vector());
+  chebyshev_additional_data.constraints.copy_from(this->zero_constraints);
+  chebyshev_additional_data.degree               = 3;
+  chebyshev_additional_data.smoothing_range      = 20;
+  chebyshev_additional_data.eig_cg_n_iterations  = 40;
+  chebyshev_additional_data.eigenvalue_algorithm = ChebyshevPreconditionerType::
+    AdditionalData::EigenvalueAlgorithm::power_iteration;
+  chebyshev_additional_data.polynomial_type =
+    ChebyshevPreconditionerType::AdditionalData::PolynomialType::fourth_kind;
```
peterrum, line 311: Not all parameters need to be set here.

peterrum (commented):
Looks very good :)

source/solvers/mf_navier_stokes.cc
```diff
@@ -264,23 +276,705 @@ MFNavierStokesSolver<dim>::assemble_system_rhs()
...
+      .mg_coarse_grid_tolerance;
+  const double reduce =
+    this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
+      .mg_coarse_grid_reduce;
+  ReductionControl coarse_grid_solver_control(
+    max_iterations, tolerance, reduce, false, false);
+  SolverGMRES<VectorType>::AdditionalData solver_parameters;
+  solver_parameters.max_n_tmp_vectors =
+    this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
+      .mg_coarse_grid_max_krylov_vectors;
```
peterrum, line 532: not used

applications/lethe-fluid-matrix-free/mf_navier_stokes.cc
```diff
@@ -40,8 +40,17 @@ main(int argc, char *argv[])
           prm.parse_input(argv[1]);
           NSparam.parse(prm);
 
-          MFNavierStokesSolver<2> problem(NSparam);
-          problem.solve();
+          if (NSparam.stabilization.use_default_stabilization == true ||
+              NSparam.stabilization.stabilization ==
+                Parameters::Stabilization::NavierStokesStabilization::pspg_supg)
+            {
+              MFNavierStokesSolver<2, NavierStokesSUPGPSPGOperator<2, double>>
```
blaisb, line 47: Considering we will only have 2-3 alternative operator for a long time, this should not be that much of a big deal for the time being.

applications_tests/lethe-fluid-matrix-free/mms3d_fe1_lsmg.mpirun=2.output
```diff
@@ -0,0 +1,18 @@
...
+
+*****************************
+Steady iteration:        2/2
+*****************************
+   Number of active cells:       512
+   Number of degrees of freedom: 2916
+   Volume of triangulation:      8
+cells  error_velocity     error_pressure   
+   64 4.185618e-01    - 4.553725e-02     - 
+  512 1.099013e-01 1.93 1.771718e-01 -1.96
```
blaisb, line 18: After our discussion, I feel it's more coherent here to do the test with a sufficient number of initial cells to show that there is at least a degree of convergence for both variables

applications_tests/lethe-fluid-matrix-free/mms3d_fe1_lsmg.output
```diff
@@ -0,0 +1,18 @@
...
+
+*****************************
+Steady iteration:        2/2
+*****************************
+   Number of active cells:       512
+   Number of degrees of freedom: 2916
+   Volume of triangulation:      8
+cells  error_velocity     error_pressure   
+   64 4.185618e-01    - 4.553725e-02     - 
+  512 1.099013e-01 1.93 1.771718e-01 -1.96
```
blaisb, line 18: Same as previous

include/core/parameters.h
blaisb: Smoother when you are going down in the V cycle? I always saw it named coarsener, maybe that's just me though.

include/solvers/mf_navier_stokes_operators.h
blaisb: Good to know

source/core/parameters.cc
blaisb: Yeah we still use the old declare / parse approach and not the add parameter. For now let's keep it this way, this is an orthogonal refactoring we can do in the future (endless refactoring is my life)

blaisb (approved):
A few comments there and there, but nothing special on my side. Still confused about how global coarsening is not working correctly, but I don't have an idea right now what could be the cause.

source/solvers/mf_navier_stokes.cc
lpsaavedra: That's true. We haven't experimented much with this yet. I think I will leave it like this in this PR, but later we will check this again.

lpsaavedra:
All the comments have now been addressed with exception of the template of the `OperatorType`. The tests were updated and since GC was fixed, the GC tests were added as well. An additional bug was found in the `compute_element_size()` function of the operator, therefore, this function is not being used at the moment and the calculation of the element size is being done inside of the `do_cell_integral_local` and `local_evaluate_residual` functions. While not ideal, it will remain like this until another solution is found.

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -718,7 +744,7 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
       integrator.evaluate(EvaluationFlags::values | EvaluationFlags::gradients |
                           EvaluationFlags::hessians);
 
-      auto h = integrator.read_cell_data(this->get_element_size());
+      // auto h = integrator.read_cell_data(this->get_element_size());
```
peterrum, line 747: What is the issue? What about creating an issue and collecting there the TODOs?

peterrum (approved):

source/solvers/mf_navier_stokes_operators.cc
lpsaavedra: We have a TODO list in the MF Lethe Project in the projects section of github. I will give you access to it.

blaisb:
@lpsaavedra if the test are still timeouting after 1200 seconds or something you should reduce the number of mesh adapt by 1, even if they still run with a single mesh, they are still valid tests :). The CI is already very long so...

PR #885, opened 2023-09-20
Simulation parameters tidy

blaisb:
# Description of the problem

- There was an enable parameter in the Source term functionnalities that was poorly implemented. The enable true/false was only used in the Navier-Stokes solver,  nowhere else. This led to significant inconsistancies.

# Description of the solution

- Deprecate the enable parameter
- Deprecated the way the Source term works. Now the parameters are a class instead of a pointer to a class, but the functions themselves are pointer. This is significantly cleaner.

# How Has This Been Tested?

- All tests pass without change.

# Documentation

- Parameter was removed from documentation and examples
-  CHANGELOG was modified

# Comments

- Never ending refactoring of my stupid design choices from 2019. Damn you past Bruno
-


applications_tests/lethe-fluid/ggls-stab_gls.prm
```diff
@@ -78,14 +78,6 @@ subsection multiphysics
...
 
-#---------------------------------------------------
-# Analytical Solution
-#---------------------------------------------------
-
-subsection analytical solution
-  set enable = false
-end
-
 #---------------------------------------------------
```
lpsaavedra, line 81: The enable still exists for the analytical solution, no?

lpsaavedra (approved):
Good change!

PR #889, opened 2023-09-25
Dem particle wall tangential overlap

OGaboriault:
# Description of the problems

- The angular velocity of the particle was not taken into account correctly in the contact_relative_velocity calculation. (missing a minus sign)
- Particle-Particle and Particle-Wall contact were not using the same vector convention (i -> j)

# Description of the solution

- The code is now using the same convention for Particle-Particle and Particle-Wall. The normal_contact vector is going from the particle ( i ) to the wall ( j ). 
- The translational velocity induce by the rotation of the particle at the point of contact is now properly being calculated. 
- Comments in the code have been added explaining why minus sign and "-=" or "+=" operator are being use at certain places, which makes the calculation of the force way simpler to understand. 
- An application test have been added. This test `rolling_on_plane.prm` consisted of a single particle rolling perfectly on a plane without rolling resistance. The tangential_relative_velocity should remain at zero if the contact relation velocity calculation is done correctly.  

# How Has This Been Tested?

- New application test.
- Bouncing particle results are the same.
- Oblique wall impact collision results are the same. (Not exactly, which is normal) 
- The influence of the angular velocity of the wall has been tested using the small-rotating-drum example.  

# Comments

- none


source/dem/particle_wall_linear_force.cc
```diff
@@ -395,31 +399,32 @@ ParticleWallLinearForce<dim>::calculate_linear_contact_force_and_torque(
...
-    -2 * this->model_parameter_beta[particle_type] *
+    2 * this->model_parameter_beta[particle_type] *
     sqrt(particle_properties[DEM::PropertiesIndex::mass] *
          normal_spring_constant);
 
+  // REF :  R. Garg, J. Galvin-Carney, T. Li, and S. Pannala, “Documentation of
+  // open-source MFIX–DEM software for gas-solids flows,” Tingwen Li Dr., p. 10,
+  // Sep. 2012.
+  // There is a minus sign since the tangential force is apply in the opposite
+  // direction of the tangential_overlap
```
blaisb, line 413:
```suggestion
  // There is a minus sign since the tangential force is applied in the opposite
  // direction of the tangential_overlap
```

source/dem/particle_wall_linear_force.cc
```diff
@@ -437,14 +442,13 @@ ParticleWallLinearForce<dim>::calculate_linear_contact_force_and_torque(
         tangential_force / (tangential_spring_constant + DBL_MIN);
     }
 
-  // Calculation of torque
-  // First calculation of torque due to tangential force acting on
-  // particle
+  // Calculation torque caused by tangential force
+  // We add the minus sign here since the tangential_force apply on the particle
+  // is in the opposite direction
```
blaisb, line 447:
```suggestion
  // We add the minus sign here since the tangential_force is applied on the particle
  // is in the opposite direction
```

source/dem/particle_wall_nonlinear_force.cc
```diff
@@ -409,19 +413,26 @@ ParticleWallNonLinearForce<dim>::calculate_nonlinear_contact_force_and_torque(
...
     radius_times_overlap_sqrt;
+
+  // There is no minus sign here since model_parameter_beta is negative or
+  // equal to zero.
   double normal_damping_constant =
     1.8257 * this->model_parameter_beta[particle_type] *
     sqrt(model_parameter_sn * particle_properties[DEM::PropertiesIndex::mass]);
+
+  // There is a minus sign since the tangential force is apply in the opposite
+  // direction of the tangential_overlap
```
blaisb, line 424:
```suggestion
  // There is a minus sign since the tangential force is applied in the opposite
  // direction of the tangential_overlap
```

blaisb (approved):
Few tiny comments. Thanks for clearing that out. The next thing missing would be some explanation in the theory guide, but can leave that for later.

OGaboriault:
> Few tiny comments. Thanks for clearing that out. The next thing missing would be some explanation in the theory guide, but can leave that for later.

I'll add some explanations in my JKR branch since I'm already modifying the theory guide in it.

blaisb:
Seems like on test is still failing, I can help you check this up if it is still an issue tomorrow :)

Luckabarbeau (commented):
Very good work. As we discussed we can add to the test, I send you for IB particles and the bug fix required for the IB particles. Once it is added i will approve the PR

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1283,8 +1283,8 @@ CFDDEMSolver<dim>::print_particles_summary()
          << "v_x: " << particle_properties[DEM::PropertiesIndex::v_x] << ",  "
          << "v_y: " << particle_properties[DEM::PropertiesIndex::v_y] << ",  "
          << "vz: " << particle_properties[DEM::PropertiesIndex::v_z];
-
-      announce_string(this->pcout, ss.str(), '-');
+      this->pcout << ss.str() << std::endl;
+      // announce_string(this->pcout, ss.str(), '-');
```
Luckabarbeau, line 1287:
```suggestion

```

Luckabarbeau (approved):
Good job on the bug correction. There is nothing to add on my side!

blaisb (approved):

blaisb (approved):

PR #894, opened 2023-09-29
Add simple ALE module 

blaisb:
# Description of the problem

- For many melt-pool or other simulations, a small ALE module would be required to do simulations in a moving frame of reference

# Description of the solution

- Add said module

# How Has This Been Tested?

- A simple test for heat transfer is added. As the feature is more used, we will add more test.

# Documentation

- Documentation of the module was added


hepap (commented):
Just missing the time consideration if we want the velocity to change in time!

PR #896, opened 2023-10-01
Dem particle wall sliding velocity

blaisb:
# Description of the problem

- The calculation of the translational velocity on rotating walls was calculated inadequately [#896](https://github.com/lethe-cfd/lethe/pull/896): The velocity of rotating boundary conditions (e.g. boundaries of the mesh) was calculated inadequately in the case where the normal vector of the wall was aligned with the rotation axis. The whole calculation procedure was slightly messed up and only worked for cylinders. 

# Description of the solution

- This has been fixed and made general by using the vector from the contact point to the center of rotation. A new parameter has been added for this purpose.

# How Has This Been Tested?

- All test pass and give adequate results
- A new application test has been added to test this exact feature

# Documentation

- Documentation of the point on the axis has been updated

# Future changes

- I will continue to refactor the entire DEM particle-wall aspect to make it cleaner and more streamlined


doc/source/parameters/dem/boundary_conditions.rst
```diff
@@ -28,6 +28,9 @@ In this subsection, the boundary conditions of the DEM simulation are defined. F
       set rotational vector x = 1
       set rotational vector y = 0
       set rotational vector z = 0
```
OGaboriault, line 30: Shouldn't we also just change the way we define the rotation the `rotation vector` in the to be the same as the `point on rotation vector`. I find the new way more elegant.

source/dem/particle_wall_contact_force.cc
```diff
@@ -49,16 +50,34 @@ ParticleWallContactForce<dim>::update_contact_information(
   particle_angular_velocity[2] =
     particle_properties[DEM::PropertiesIndex::omega_z];
 
+  // Calculate approximation of the contact point using the normal vector
+  Point<3> contact_point =
+    particle_position +
+    0.5 * particle_properties[DEM::PropertiesIndex::dp] * normal_vector;
+
+  // Get vector pointing from the contact point to the axis
```
OGaboriault, line 58:
```suggestion
  // Get vector pointing from the contact point to the origin of the rotation axis
```

source/dem/particle_wall_contact_force.cc
```diff
@@ -49,16 +50,34 @@ ParticleWallContactForce<dim>::update_contact_information(
...
+  Tensor<1, 3> vector_to_rotating_axis =
+    contact_point - this->point_on_rotation_vector[boundary_id];
+
+  // Remove the rotating axis component of that vector
+  vector_to_rotating_axis =
+    vector_to_rotating_axis -
+    (vector_to_rotating_axis * this->boundary_rotational_vector[boundary_id]) *
+      this->boundary_rotational_vector[boundary_id] /
```
OGaboriault, line 66: this->boundary_rotational_vector is not already a unit vector?

OGaboriault: we are using the `double boundary_rotational_speed`, so technically `boundary_rotational_vector `should be a unit vector. We should add a step when creating that container to make sure that we are storing unit vectors in it. This would save the extra division (and addition at the denominator). 

Also, why are you using the norm_square?

tests/dem/particle_wall_contact_force_linear.cc
```diff
@@ -174,12 +174,7 @@ test()
     particle_wall_contact_list, particle_wall_contact_information);
 
   // Calling linear force
-  ParticleWallLinearForce<dim> force_object(
-    dem_parameters.boundary_conditions.boundary_translational_velocity,
-    dem_parameters.boundary_conditions.boundary_rotational_speed,
-    dem_parameters.boundary_conditions.boundary_rotational_vector,
-    grid_radius,
-    dem_parameters);
+  ParticleWallLinearForce<dim> force_object(dem_parameters);
```
OGaboriault, line 177: LOL. Ok I see what you mean hahaha. This is really funny.

OGaboriault (commented):
Good job on fixing this. Is was bothering me. 

My only big comment is on `vector_to_rotating_axis` calculation.

doc/source/parameters/dem/boundary_conditions.rst
blaisb: The new way is more elegant, but i want to introduce the change progressively. First fix the bug, then refactor the parameters in another PR. I would like to remove all of the x, y , z parameters as individual and always use lists, but I would rather do this in another PR

source/dem/particle_wall_contact_force.cc
blaisb: I am using the norm square because we do two scalar product. There is no garantee right now that the vector is a unit vector, (e.g. someone could write 1,1,1 and it should still work). We can refactor this, but this means changing what was the meaning of these parameters. I am all for this, but this should be done in another PR. This PR fixes the problem we had, it does not aim at refactoring the entire parameter structure.

OGaboriault (approved):

PR #900, opened 2023-10-03
Add matrix-free Navier-Stokes SUPG/PSPG transient operator

lpsaavedra:
# Description of the problem

Until now the matrix-free application in Lethe was only able to solve steady-state problems. 

# Description of the solution

This PR adds the first transient operator that uses SUPG/PSPG stabilization. The transient operator is now also compatible with both of the multigrid preconditioners introduced in PR #884. 

# How Has This Been Tested?

A simple test using the Taylor-Green-Vortex problem and the ILU preconditioner was added:

- [X]  `applications_tests/lethe-fluid-matrix-free/tgv_fe1_ilu.prm`

The results of this test differ from the ones of the lethe-fluid application, since MF does not support yet L2-projection as initial condition.

# Comments

The next MF PR will add the option to choose AMG or ILU for the coarse-grid solver preconditioner in the multigrid approaches.


source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
 
 template class NavierStokesSUPGPSPGOperator<2, double>;
 template class NavierStokesSUPGPSPGOperator<3, double>;
+
+/**
+ * @brief This function performs a cell integral, i.e., the jacobian of the discretization
+ * of the transient Navier-Stokes equations with SUPG PSPG stabilization in a
+ * cell batch. The equations are given as follows: (q,∇δu) + (v,∂t δu) +
+ * (v,(u·∇)δu) + (v,(δu·∇)u) - (∇·v,δp) + ν(∇v,∇δu) (Weak form jacobian)
```
peterrum, line 911:
```suggestion
 * cell batch. The equations are given as follows:
 * (q,∇δu) + (v,∂t δu) + (v,(u·∇)δu) + (v,(δu·∇)u) - (∇·v,δp) + ν(∇v,∇δu) (Weak form Jacobian)
```

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
...
+          gradient_result[i][i] += -value[dim];
+          // +(q,∇δu)
+          value_result[dim] += gradient[i][i];
+
+          for (unsigned int k = 0; k < dim; ++k)
+            {
+              // +(v,(u·∇)δu)
+              value_result[i] += gradient[i][k] * previous_values[k];
+              // +(v,(δu·∇)u)
+              value_result[i] += previous_gradient[i][k] * value[k];
```
peterrum, line 1027: I guess in the formula and implementation the terms are swapped.

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
...
 template class NavierStokesSUPGPSPGOperator<3, double>;
+
+/**
+ * @brief This function performs a cell integral, i.e., the jacobian of the discretization
+ * of the transient Navier-Stokes equations with SUPG PSPG stabilization in a
+ * cell batch. The equations are given as follows: (q,∇δu) + (v,∂t δu) +
+ * (v,(u·∇)δu) + (v,(δu·∇)u) - (∇·v,δp) + ν(∇v,∇δu) (Weak form jacobian)
+ * + (∂t δu +(u·∇)δu + (δu·∇)u + ∇δp - ν∆δu)τ·∇q (PSPG Jacobian)
+ * + (∂t δu +(u·∇)δu + (δu·∇)u + ∇δp - ν∆δu)τu·∇v (SUPG Part 1)
+ * + (∂t u +(u·∇)u + ∇p - ν∆u - f )τδu·∇v (SUPG Part 2)
```
peterrum, line 914: I guess you could compute the content of the bracket and after that multiply it with `u`/`δu`.

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
...
+ */
+template <int dim, typename number>
+NavierStokesTransientSUPGPSPGOperator<dim, number>::
+  NavierStokesTransientSUPGPSPGOperator()
+{}
```
peterrum, line 919:
```suggestion
  NavierStokesTransientSUPGPSPGOperator() = default;
```

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
...
+                {
+                  // +(v,(u·∇)u)
+                  value_result[i] += gradient[i][k] * value[k];
+                }
+              // + (v,∂t u)
+              value_result[i] += bdf_coefs[0] * value[i];
+              value_result[i] += previous_time_derivatives[i];
+            }
+
+          // PSPG term
```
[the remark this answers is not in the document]
lpsaavedra: All of this terms are tested by the gradient of q (`∇q` term in the comments), which corresponds to the `gradient_result[dim]` and `gradient_result[dim][i]` in the code.

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -877,3 +903,394 @@ NavierStokesSUPGPSPGOperator<dim, number>::local_evaluate_residual(
...
+            }
+          // +(∇p)τ∇·q
+          gradient_result[dim] += tau * gradient[dim];
+
+          // SUPG term
+          for (unsigned int i = 0; i < dim; ++i)
+            {
+              for (unsigned int k = 0; k < dim; ++k)
+                {
+                  // (-f)τ(u·∇)v
```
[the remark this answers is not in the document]
lpsaavedra: The f corresponds to the `source_value` which was extracted from the forcing function (f) a few lines before (line 1156).

source/solvers/mf_navier_stokes.cc
```diff
@@ -287,6 +304,40 @@ MFNavierStokesSolver<dim>::update_multiphysics_time_average_solution()
...
 
+template <int dim>
+void
+MFNavierStokesSolver<dim>::percolate_time_vectors_fd()
+{
+  for (unsigned int i = this->previous_solutions.size() - 1; i > 0; --i)
+    {
+      this->previous_solutions[i] = this->previous_solutions[i - 1];
+    }
+  this->previous_solutions[0] = this->present_solution;
```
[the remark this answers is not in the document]
lpsaavedra: I don't think I can use swap here, because I still need the present solution vector for other calculations...

source/solvers/mf_navier_stokes_operators.cc
lpsaavedra: Yes, they are swapped.

source/solvers/mf_navier_stokes.cc
```diff
@@ -287,6 +304,40 @@ MFNavierStokesSolver<dim>::update_multiphysics_time_average_solution()
...
+
+  calculate_time_derivative_previous_solutions();
```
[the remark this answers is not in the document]
blaisb: it should not be called in assemble_system_rhs if it is called here...

source/solvers/mf_navier_stokes.cc
```diff
@@ -272,6 +281,14 @@ template <int dim>
...
 MFNavierStokesSolver<dim>::assemble_system_rhs()
 {
+  if (is_bdf(this->simulation_control->get_assembly_method()))
+    {
+      calculate_time_derivative_previous_solutions();
+      this->time_derivative_previous_solutions.update_ghost_values();
+      this->system_operator->evaluate_time_derivative_previous_solutions(
+        this->time_derivative_previous_solutions);
+    }
+
```
blaisb, line 291: Should this be done here? Why not do it when you percolate time steps

blaisb (approved):
Just one comment, it's very good job

source/solvers/mf_navier_stokes.cc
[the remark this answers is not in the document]
lpsaavedra: Yes, it should not be called here. This should be done before iterating and after refining the mesh for it to be called only once.  I have updated the code and now it is done within the `solve() `function.

source/solvers/mf_navier_stokes_operators.cc
lpsaavedra: Yes, we could... we did this because there are so many terms and we wanted to keep it readable. Do you think it would be better to change it to improve performance?

peterrum (approved):

PR #903, opened 2023-10-10
Layer thickening

oguevremont:
# Description of the problem

- For some simulations, especially in porous media, in can be useful to artificially inflate or deflate a solid.
- This option is added in this PR, for all shapes.

# Description of the solution

- A member double is added to the Shape class, and not the IBParticle class: this is done so that the output, which calls Shape as a Function, can show the inflation/deflation.
- No change is made to the gradient (only value), because the inflation/deflation happens in the normal direction.

# How Has This Been Tested?

- [core/shape_thickening] Unit test that shows that value changes with thickening, but not the gradient.

# Documentation

- The parameter is added to the Sharp Immersed Boundary parameters page.


blaisb (approved):

AmishgaAlphonius (approved):
I only have a few suggestions that specify what a negative value would do. :)

PR #906, opened 2023-10-17
Add viscous and ramp initial conditions for MF

blaisb:
# Description of the problem

- The matrix free implementation did not allow for viscous or ramp initial conditions

# Description of the solution

- Added them.

# How Has This Been Tested?

- Tested it on a case and it works just fine. Not sure if it needs a dedicated application_test though. @lpsaavedra what's your take on this?

### DIFF
--- a/include/solvers/mf_navier_stokes.h
+++ b/include/solvers/mf_navier_stokes.h
@@ -30,6 +30,7 @@
 
 using namespace dealii;
 
+
 /**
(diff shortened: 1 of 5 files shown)


lpsaavedra (approved):
Everything seems good to me, good addition to the MF solver. I don't think we should add an application test for this, we do not have one in for the lethe-fluid solver either.

PR #907, opened 2023-10-18
Add ILU preconditioner for MG coarse grid solvers

lpsaavedra:
# Description of the problem

The coarse grid solver of the multigrid preconditioners used in the matrix free application only supported AMG as preconditioner. 

# Description of the solution

An ILU preconditioner was added as well. For this, a new parameter was created, where one can choose between AMG and ILU for the coarse-grid solver.

# How Has This Been Tested?

- No application test was added for this future. It was tested using the mms3d problem and it yield the same results.

# Documentation

- The parameter to choose the coarse grid solver preconditioner and the parameters specific to ILU were added in the MG section of the linear solver documentation:
- [X] `doc/source/parameters/cfd/linear_solver_control.rst`


blaisb (approved):
Seems good to me. There's always a slight bit of copy-paste when setting these preconditioner, but this is clean and tidy for me.
I am curious to compare the performance of both

PR #911, opened 2023-10-20
Greater number of solid objects + Case generator for LPBF + Post processing for LPBF

OGaboriault:
# Description of the modification

- The number of solid objects in a DEM simulation was to small for LPBF simulation. 10 -> 50 
- Case generator and Post-processing python codes are now on the repo.


blaisb:
@OGaboriault I think the cases and the case generator should migrate to the lethe-utils repository.
This is where we store this stuff usually
https://github.com/lethe-cfd/lethe-utils

blaisb (approved):

PR #915, opened 2023-10-25
Small improvements to matrix free operators

lpsaavedra:
# Description of the problem

In the matrix free operators several terms could be written in the same line, which should improve performance.

# Description of the solution

This PR writes several terms of the operator in one line, without affecting the readability of each of the parts of the Jacobian and right hand side. 

# How Has This Been Tested?

All the existing tests pass without issues.

# Future changes

Performance will be further improved in other PRs in the future, but for now, this is how the operators will be used to run several simulations.


blaisb (approved):
Very good. we might need to look at the various "order" of the loops to see if there could be some ordering/optimization to be done...

PR #918, opened 2023-10-26
Dem contact storage

blaisb:
# Description of the problem

- Tangential velocity was stored within the large containers for particle-particle contact history and this was not necessary. We only need to store the overlap.
- Some Tensors were kept as part of a member of the contact class for no reason because they were only being used locally. I made them local

# Description of the solution

- Removed tangential velocity from the large container
- Removed the class members and put them directly where they are used.

# How Has This Been Tested?

- All test pass as before
- The new architecture improved the performance by about 2-3% on the hopper case. So nothing at all really. I just find it much less confusing / cleaner.


include/dem/particle_particle_contact_force.h
```diff
@@ -865,16 +872,17 @@ class ParticleParticleContactForce
...
-    const Tensor<1, 3>                  &normal_unit_vector,
-    const double                         normal_overlap,
-    const ArrayView<const double>       &particle_one_properties,
-    const ArrayView<const double>       &particle_two_properties,
-    Tensor<1, 3>                        &normal_force,
-    Tensor<1, 3>                        &tangential_force,
-    Tensor<1, 3>                        &particle_one_tangential_torque,
-    Tensor<1, 3>                        &particle_two_tangential_torque,
-    Tensor<1, 3>                        &rolling_resistance_torque)
+    const Tensor<1, 3> & /*tangential_relative_velocity*/,
```
OGaboriault, line 875: We should add a TODO for the Hackathon. It is so weird to not have a tangential damping for the basic hertz model.

OGaboriault (approved):
Good job. I always enjoy seeing the DEM code getting faster... even though it is a small improvement. 

I have one tiny comment which is not related to this PR. :)

PR #920, opened 2023-10-31
Dem rotating boundary parsing

OGaboriault:
#Description of the problem

- This is a follow up to #896 
- The rotational vector for the rotational boundary condition in the lethe-particles solver is now define with one line in the parameters file.
- We make sure that the rotational vector is a unit vector. This way, we save the norm_square calculation, the addition and the division when computing the vector_to_rotating_axis variable. (Simply cleaner in my opinion)
-  We throw an error is the vector is defined as the null vector which wasn't the case previously. 

# How Has This Been Tested?

- All test pass.
- Throw error works

# Documentation

- The `set rotation vector` line in the documentation and examples were updated. 

# Future changes

- The translation speed should be define as such in the parameter file 

# Comments

- None


source/core/parameters_lagrangian.cc
```diff
@@ -1199,15 +1193,24 @@ namespace Parameters
...
-          double       rotational_speed = prm.get_double("rotational speed");
-          Tensor<1, 3> rotational_vector;
-
-          rotational_vector[0] = prm.get_double("rotational vector x");
-          rotational_vector[1] = prm.get_double("rotational vector y");
-          rotational_vector[2] = prm.get_double("rotational vector z");
+          BC_type                 = BoundaryType::rotational;
+          double rotational_speed = prm.get_double("rotational speed");
 
+          // Read the rotational vector from a list of doubles
```
blaisb, line 1199: Not necessary to do it now, but we should clearly do something like a string_to_tensor<dim> function to refactor this parsing.

blaisb (approved):

examples/dem/3d-rectangular-hopper/hopper.prm
```diff
@@ -10,8 +10,8 @@ set dimension = 3
 subsection simulation control
   set time step         = 1e-5
   set time end          = 7.5
-  set log frequency     = 1000
-  set output frequency  = 1000
+  set log frequency     = -1000
+  set output frequency  = -1000
```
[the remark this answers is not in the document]
OGaboriault: Woups, sorry...
I was testing the performance of the code with this example without output and I forgot to put to put it back.

source/core/parameters_lagrangian.cc
OGaboriault: I'll do the change in an other PR. Good idea

blaisb:
@OGaboriault You still have not reverted the change to the hopper example

blaisb (approved):

PR #921, opened 2023-11-02
Core: entry string to tensor function

OGaboriault:
# Description of the problem
- The entry_string to a variable type Tensor<1,3> or Point<3> was frequently being used in the parsing functions. Although this method is cleaner then the one used before (because it allows for the use of one line in the prm file), it was starting to become repetitive in the code. (String -> vector<string> -> 3X string_to_double -> Tensor ) 

# Description of the solution

- A new function called `entry_string_to_tensor3` has been created. This function does the same procedure has described earlier. 

# How Has This Been Tested?

- All test have pass. 
- The runtime_error has been tested.

# Documentation

- N/A

# Future changes

- We could do a simular function for the list_x, list_y, list_z, list_d, etc. for the List insertion vector. 

# Comments

- The name of the function can be change. 
- Not sure if the function is at the right place in the code. Maybe it could be in Utilities... although we need the `prm` parameter to my knowledge and the `deal.II/base/parameter_handler.h` is not included in the Utilities class.

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -1332,5 +1332,17 @@ namespace Parameters
     parse_parameters(ParameterHandler &prm);
   };
 
+  /**
+   * @brief Return the tensor of entry @p entry_string. If the entry was changed,
+   * then the changed value is returned, otherwise the default value is
+   * returned. If the entry is not a Tensor, an Throw_error will occur.
+   * This function can be use for Point variables.
+   *
+   * @param entry_string A declare string
+   */
+  Tensor<1, 3>
+  entry_string_to_tensor(ParameterHandler  &prm,
+                         const std::string &entry_string);
+
 } // namespace Parameters
 #endif
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -1335,8 +1335,8 @@ namespace Parameters
   /**
    * @brief Return the tensor of entry @p entry_string. If the entry is specified in the .prm file,
    * then the changed value is returned, otherwise the default value is
-   * returned. If the entry is not a Tensor, an AssertThrow will occur.
-   * This function can be use for Point variables.
+   * returned. If the entry is not equivalent to a Tensor<1,3>, an error will be
+   * thrown. This function can be use for Point<3> variables.
    *
    * @param prm A parameter handler which is currently used to parse the simulation information
    * @param entry_string A declare string in the parameter file.
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -1341,7 +1341,7 @@ namespace Parameters
    * @param entry_string A declare string
    */
   Tensor<1, 3>
-  entry_string_to_tensor(ParameterHandler  &prm,
+  entry_string_to_tensor3(ParameterHandler  &prm,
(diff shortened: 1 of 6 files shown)


include/core/parameters.h
```diff
@@ -1332,5 +1332,18 @@ namespace Parameters
...
+   * then the changed value is returned, otherwise the default value is
+   * returned. If the entry is not a Tensor, a Throw_error will occur.
+   * This function can be use for Point variables.
+   *
+   * @param prm A parameter handler which is used to parse the simulation information
+   * @param entry_string A declare string
+   */
+  Tensor<1, 3>
+  entry_string_to_tensor3(ParameterHandler  &prm,
+                          const std::string &entry_string);
```
blaisb, line 1346: That's an ok design, but why don't you just pass the string?
You could have string_to_tensor3 and pass the const string. You would call it by saying  string_to_tensor3(prm.get_string("parameter")). This would be a lot cleaner no? This way. you could move it to utilities.h and also inline the function.

blaisb (changes requested):
Just a small refactor suggestion, good work though.

include/core/parameters.h
```diff
@@ -1332,5 +1332,18 @@ namespace Parameters
     parse_parameters(ParameterHandler &prm);
   };
 
+  /**
+   * @brief Return the tensor of entry @p entry_string. If the entry was changed,
+   * then the changed value is returned, otherwise the default value is
```
hepap, line 1337:
```suggestion
   * @brief Return the tensor of entry @p entry_string. If the entry is specified in the .prm file,
   * then the changed value is returned, otherwise the default value is
```

hepap (commented):
Nice addition, which makes the parsing much cleaner. I agree with Bruno for the small refactor.

include/core/utilities.h
```diff
@@ -353,5 +353,15 @@ get_last_value_of_parameter(const std::string &file_name,
 unsigned int
 get_dimension(const std::string &file_name);
 
+/**
+ * @brief Return the entry_string as a Tensor<1,3>. If the entry_string is not
+ * equivalent to a Tensor<1,3>, a Throw_error will occur. This function is
```
blaisb, line 358:
```suggestion
 * equivalent to a Tensor<1,3>, an error will be thrown. This function is
```

source/core/utilities.cc
```diff
@@ -485,3 +485,24 @@ get_dimension(const std::string &file_name)
...
+    Utilities::split_string_list(entry_string));
+  std::vector<double> vector_of_double =
+    Utilities::string_to_double(vector_of_string);
+
+  if (vector_of_double.size() != 3)
+    {
+      throw(std::runtime_error(
+        "Invalid " + entry_string +
+        ". This should be a three dimensional vector or point."));
+    }
```
blaisb, line 502: I would prefer is you used the AssertThrow of deal.II instead of a raw throw like this...
It's more efficient in my perspective.

blaisb (changes requested):
Some slight changes

blaisb (approved):

PR #923, opened 2023-11-06
Streamline DEM contact model slightly

blaisb:
# Description of the problem

- There were unecessary calculations being done in the DEM model. Velocities were allocated and they were not required

# Description of the solution

- Streamline the calculations ever so slightly

# How Has This Been Tested?

- Some tests results were barely altered as a results of those changes becuase of floating point arithmetic.

# Comments

- I think this makes things clearer to read, but it does not have a real impact on performance to be honest.
-


OGaboriault (approved):
Good catch on those temporary variables. Honestly, I don't have comments on your modification. They are pretty straight forwards.

PR #928, opened 2023-11-15
Auto detect number of boundaries

blaisb:
# Description of the problem

- Current Lethe implementation did not allow for a undetermined number of boundary conditions. A max number of boundary conditions (14) were pre-declared and parsed. This was problematic because it prohibited checking the boundaries adequately, but also because it limited the number of boundary conditions

# Description of the solution

- Fix this by pre-parsing the number of boundary conditions before actually declaring them. This enables checking that the number of boundary conditions declared actually match the number given (this was a mistake present in many places in the tests it seems). I have tried to come up with a solution that is flexible my introducing a structure that, in the future, can be used to house variable size subsections before we pre-declare them. I think that the final solution is relatively elegant and is more robust than what we had.

# How Has This Been Tested?

- Some tests had to be modified because they were actually incorrect since there was an error in the boundary condition number.


applications/navier_stokes_parameter_template/navier_stokes_parameter_template.cc
```diff
@@ -23,7 +28,7 @@ main()
       {
         ParameterHandler        prm;
         SimulationParameters<3> nsparam;
-        nsparam.declare(prm);
+        nsparam.declare(prm, size_of_subsections);
```
lpsaavedra, line 31:
```suggestion

        nsparam.declare(prm, size_of_subsections);
```

source/core/utilities.cc
```diff
@@ -485,3 +530,18 @@ get_dimension(const std::string &file_name)
...
+int
+get_max_number_of_boundary_conditions(const std::string &file_name)
+{
+  int max_number_of_boundary_conditions =
+    get_max_value_of_parameter(file_name, "number");
+
+  AssertThrow(
+    max_number_of_boundary_conditions >= 0,
+    dealii::ExcMessage(
+      "Your parameter file does not contain any indication for the number of boundary conditions for any physics supported by Lethe. Since November 2023, Lethe requires that a \"boundary conditions\" subsection be present with at least \"number=0\" "));
```
lpsaavedra, line 544:
```suggestion
      "Your parameter file does not contain any indication for the number of boundary conditions for any physics supported by Lethe. Since November 2023, Lethe requires that a \"boundary conditions\" subsection is present with at least \"number=0\" "));
```

lpsaavedra (changes requested):
This solution is definitely better than declaring and parsing a fixed number of boundary conditions. We will have to be careful to not add any other parameter by the name "number", especially since the code is growing a lot lately.  Can you add an entry to the `CHANGELOG.md` file?

Luckabarbeau (approved):
Nothing to add on my side this is a very nice addition and we will be able to use a similar strategy for other parts of the code like the number of particle subsection in sharp.

blaisb:
@lpsaavedra I will definitely add a changelog entry. That's a good idea.

blaisb:
@lpsaavedra I also agree that number is a sucky parameter name. But the other choice was renaming all number to number of boundary conditions and I was hesitant to do this. We don't have any other "number" parameter right now.

lpsaavedra:
> @lpsaavedra I also agree that number is a sucky parameter name. But the other choice was renaming all number to number of boundary conditions and I was hesitant to do this. We don't have any other "number" parameter right now.

I understand. This is a good solution at the moment :)

source/core/utilities.cc
```diff
@@ -485,3 +530,18 @@ get_dimension(const std::string &file_name)
           "Lethe requires that the user explicitly specify the dimension of the problem within the parameter file. This can be achieved by adding set dimension = 2 or set dimension = 3 within the parameter file"));
     }
 }
+
+
+int
+get_max_number_of_boundary_conditions(const std::string &file_name)
+{
+  int max_number_of_boundary_conditions =
+    get_max_value_of_parameter(file_name, "number");
```
oguevremont, line 539: If I understand properly we get the maximal value of `number`, so if we have a different number of boundary conditions between physics we have to be careful, since the check if not as robust?

oguevremont (approved):
Very nice addition. I have nothing to add.

source/core/utilities.cc
blaisb: Yeah that's exactly it. If we have multiple physics, we will pre-declare the maximum number of boundary of all the physics. I could not figure out a way to robustly parse the number in another fashion... It's not the best, but clearly better.

lpsaavedra (approved):

PR #930, opened 2023-11-18
Dem insertion distribution object

OGaboriault:
# Description of the problem

- This PR is the first of two. The objective is to implement new types of diameter distributions for the DEM. Currently, the code has the uniform and normal distribution. These distribution are using the same snippet of code. The only different is that the uniform distribution uses a standard deviation equal to zero. Since both distribution are using the same snippet of code, the diameter generation was directly being made in the insert function, which doesn't allows for new distribution types to be implemented. 

# Description of the solution

- The "Insertion" base class has a new "Distribution_base" type attribute. This base class is purely virtual and has a function called "particle_size_sampling" which can be override by derived classes such as "NormalDistribution". 
- The "Insertion" class has a new constructor which is being used by its derived classes for the initialization of the distribution_object. 

# How Has This Been Tested?

- Since this is a refactoring and not a new feature, no new test has been implemented. 
- All current tests pass
- Some examples have been used to visualize the normal distribution. 

# Documentation

- N/A

# Future changes

- We could do a separate class for the uniform distribution, but I don't think this should be a priority. The code work the way it is currently. I my mind, this would just add new files for pretty much nothing.
- This refactoring will enable the implementation of new distribution (mainly log-normal and bar chart)


include/dem/insertion.h
```diff
@@ -120,9 +129,9 @@ class Insertion
     const DEMSolverParameters<dim> &dem_parameters,
     const ConditionalOStream       &pcout);
 
-  // Number of particles that is going to be inserted at each insetion step.This
-  // value can change in the last insertion step to reach the desired number of
-  // particles
+  // Number of particles that is going to be inserted at each insertion
+  // step.This value can change in the last insertion step to reach the desired
+  // number of particles
```
blaisb, line 134:
```suggestion
  // Number of particles that is going to be inserted at each insertion
  // step. This value can change in the last insertion step to reach the desired
  // number of particles
```

include/dem/normal_distribution.h
```diff
@@ -0,0 +1,56 @@
+/* ---------------------------------------------------------------------
+ *
```
blaisb, line 2: If you want you can also put all the distributions in the same file. No point in having one file per distribution IMO. This is what we did with the physical properties model

include/dem/normal_distribution.h
```diff
@@ -0,0 +1,56 @@
...
+   *
+   * @param average Average of the normal distribution.
+   * @param standard_deviation Standard_deviation of the normal distribution.
+   * @param particle_number Number of particles to insert at the current
+   * insertion time step.
+   */
+  NormalDistribution(
+    const std::unordered_map<unsigned int, double> normal_distribution_average,
+    const std::unordered_map<unsigned int, double>
+      normal_distribution_standard_deviation);
```
blaisb, line 40: Wait but there are only two params?

include/dem/normal_distribution.h
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Carries out the size sampling of every particles inserted at a insertion
+   * time step.
+   *
+   */
+  void
+  particle_size_sampling(const unsigned int particle_number,
+                         const unsigned int particle_type) override;
```
blaisb, line 49: Please add the @param for both parameters

blaisb (changes requested):
First set of comments will finish tomorrow

source/dem/insertion.cc
```diff
@@ -21,6 +21,28 @@
...
+  else if (dem_parameters.lagrangian_physical_properties
+             .size_distribution_type ==
+           Parameters::Lagrangian::LagrangianPhysicalProperties::
+             size_distribution_type::normal)
+    {
+      distribution_object = std::make_shared<NormalDistribution>(
+        dem_parameters.lagrangian_physical_properties.particle_average_diameter,
+        dem_parameters.lagrangian_physical_properties.particle_size_std);
+    }
+}
```
lpsaavedra, line 44: I don't understand this.... why are we doing the same if the distribution is uniform and normal? Or is it a WIP?

source/dem/volume_insertion.cc
```diff
@@ -40,7 +39,7 @@ VolumeInsertion<dim>::insert(
           ++current_inserting_particle_type);
     }
 
-  // Check to see if the remained uninserted particles is equal to zero or not
+  // Check to see if the remained un-inserted particles is equal to zero or not
```
lpsaavedra, line 42:
```suggestion
  // Check to see if the remaining un-inserted particles are equal to zero or not
```

tests/dem/insertion_non_uniform.cc
```diff
@@ -67,6 +67,9 @@ test()
   dem_parameters.insertion_info.inserted_this_step                   = 10;
   dem_parameters.insertion_info.distance_threshold                   = 2;
   dem_parameters.lagrangian_physical_properties.particle_type_number = 1;
+  dem_parameters.lagrangian_physical_properties.size_distribution_type =
+    Parameters::Lagrangian::LagrangianPhysicalProperties::
+      size_distribution_type::uniform;
```
lpsaavedra, line 72: I am not familiar with this tests but why would you have two tests with uniform distribution? Isn't one for non-uniform and one for uniform? At least this is what I get from the name of the tests...

include/dem/volume_insertion.h
```diff
@@ -12,13 +12,12 @@
...
  *
  * ---------------------------------------------------------------------
-
  *
- * Author: Shahab Golshan, Polytechnique Montreal, 2019
  */
 
 #include <dem/dem_solver_parameters.h>
 #include <dem/insertion.h>
+#include <dem/normal_distribution.h>
```
lpsaavedra, line 20: Is this needed? You are not using anything from the normal distribution in this class, are you?

lpsaavedra (changes requested):
I have only some comments/doubts that need to be addressed before approval

source/dem/insertion.cc
OGaboriault: Good question. An uniform distribution is just a normal distribution with a standard deviation equal to zero. This is how the code was working previously. (If you check in the parsing function,dem_parameters.lagrangian_physical_properties.particle_size_std  is set to zero when the distribution type is uniform)

We could add a new class called UniformDistribution with its own constructor, but I think that this would go in a other PR.

tests/dem/insertion_non_uniform.cc
OGaboriault: Are you talking about the name of the file? 

Previously (less than a month ago), we had three types of INSERTION methods called uniform, non-uniform and plane. The uniform insertion was not useful since you could obtained the same type of insertion location with the non-uniform one. The uniform insertion method got deprecated and the non-uniform insertion method got renamed volume insertion method. In brief, we could technically rename this test `insertion_volume.cc` and deprecated the `insertion_uniform.cc` 

Now, this PR is about size distribution, so the diameter value at which particles are being inserted. Currently we have two types of size distribution, uniform (every particle with the same diameter) and normal (particle diameters follow a normal distribution). 

Like I said in a previous comment, the difference between uniform and normal size distribution is the standard deviation (equal or not equal to zero). If you check the parsing function in `core/parameter_lagrangian.cc`, line 70, you will see that the `enum class size_distribution_type ` was previously never set to uniform or normal, but now it needs to be and this is why you see this modification in this test. 

I hope this helps.  :-)

source/dem/insertion.cc
lpsaavedra: Somehow I think it would be cleaner but I am not sure... maybe if we do not want to create a new class, then we can use the same one but I would add a comment within this if so it would be clear for anybody reading the code

tests/dem/insertion_non_uniform.cc
lpsaavedra: I see... in that case I do think that it would be better to rename the files? So that the name of the test actually reflects what is being tested...

source/dem/insertion.cc
blaisb: But then shouldnt at least the standard deviation be fixed to zero?
Maybe we should deprecate uniform completely and just ask the user to write normal with no standard deviation. Seems like a very weird option

tests/dem/insertion_non_uniform.cc
blaisb: I agree with @lpsaavedra

blaisb (changes requested):
Other comments

source/dem/insertion.cc
OGaboriault: When the distribution is set to unifom, the standard deviation is fixed to zero in the parsing function. The NormalDistribution constructor takes two std::unordered_map as inputs to support the uses of many types of particles, that is why it is set to zero there and not here. (Otherwise, I would have to create a std::unordered_map here, which would have been weird)

I could deprecated the uniform size distribution... not sure if it is a good thing, especially when you consider new types of size distribution that are going to be added in the following weeks.

tests/dem/insertion_non_uniform.cc
OGaboriault: Ok, will renamed the test. What should I do we the insertion_uniform.cc test? deprecate or renamed to insertion_volume_1.cc and insertion_volume_2.cc ?

blaisb: Yeah

source/dem/insertion.cc
blaisb: Well I think it's better to hardcode the parameter in the constructor than in the parser of the parameters
In all cases, uniform should always always have a std-dev of 0, no point in passing a st-dev assuming it has been set to zero previously

include/dem/insertion.h
```diff
@@ -120,9 +129,9 @@ class Insertion
     const DEMSolverParameters<dim> &dem_parameters,
     const ConditionalOStream       &pcout);
 
-  // Number of particles that is going to be inserted at each insetion step.This
-  // value can change in the last insertion step to reach the desired number of
-  // particles
+  // Number of particles that is going to be inserted at each insertion
```
voferreira, line 132:
```suggestion
  // Number of particles that are going to be inserted at each insertion
```

include/dem/normal_distribution.h
```diff
@@ -0,0 +1,56 @@
...
+
+class NormalDistribution : public Distribution
+{
+public:
+  /**
+   * The constructor store the parameters necessary to define the normal
+   * distribution
+   *
+   * @param average Average of the normal distribution.
+   * @param standard_deviation Standard_deviation of the normal distribution.
```
voferreira, line 33:
```suggestion
   * @param standard_deviation Standard deviation of the normal distribution.
```

include/dem/plane_insertion.h
```diff
@@ -32,7 +33,7 @@
  *
  * @note
  *
- * Paticle insertion using cells cut by a plane. Locally own cells that are
+ * Particle insertion using cells cut by a plane. Locally own cells that are
```
voferreira, line 36:
```suggestion
 * Particle insertion using cells cut by a plane. Locally owned cells that are
```

include/dem/plane_insertion.h
```diff
@@ -32,7 +33,7 @@
...
  * cut by the plane are flag. From those flag cells, we insert a particle at
```
voferreira, line 37:
```suggestion
  * cut by the plane are flagged. From those flagged cells, we insert a particle at
```

include/dem/plane_insertion.h
```diff
@@ -32,7 +33,7 @@
...
  * their center's if they are individually empty (contain no particle). This
```
voferreira, line 38:
```suggestion
 * their centroids if they are individually empty (contain no particle). This
```

source/core/parameters_lagrangian.cc
```diff
@@ -65,15 +65,17 @@ namespace Parameters
       const unsigned int &particle_type,
       ParameterHandler   &prm)
     {
-      const std::string size_distribution_type =
-        prm.get("size distribution type");
-      if (size_distribution_type == "uniform")
+      const std::string size_distribution = prm.get("size distribution type");
```
voferreira, line 68: I would use "distribution function" or "distribution function type" for the parameter instead of using size again here. And for the variable, distribution_function or distribution_type would make more sense imo

source/dem/insertion.cc
voferreira: I like more the idea of having a separate class for the uniform distribution without a standard deviation parameter. This way when other distributions would be implemented, it will be clear that another class need to be implemented and nothing would get mixed.

source/dem/insertion.cc
```diff
@@ -71,9 +89,10 @@ Insertion<dim>::assign_particle_properties(
     {
       double type     = current_inserting_particle_type;
       double diameter = 0.;
-      (particle_sizes[particle_counter] >= 0) ?
-        diameter = particle_sizes[particle_counter] :
-        -particle_sizes[particle_counter];
+      // We make sure that the diameter is positive
+      (this->distribution_object->particle_sizes[particle_counter] >= 0) ?
```
voferreira, line 93: I would use abs instead

voferreira (commented):
Good job! Just added some comments. I would really like to have distribution functions separated to have a cleaner and more sustainable structure. This would save time on future implementations of other size distribution functions

include/dem/normal_distribution.h
OGaboriault: Yep... I realized afterward that the third parameter wasn't useful and forgot to remove it.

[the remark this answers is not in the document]
OGaboriault: Got it

source/dem/insertion.cc
OGaboriault: I will do separate class, this will solve the problem.

tests/dem/insertion_non_uniform.cc
OGaboriault: This was not a yes or no question, but I will assume you said yes to the second option. 🫠

source/core/parameters_lagrangian.cc
OGaboriault: Are you talking about the string in the prm.get ?

voferreira: Both

OGaboriault: Ok, I'm not sure if I agree. I think it's fine the way it is right not. 

1. Changing the get string would break compatibility with all the prm file. It's not a problem for us, put for all the users out there it's better not to (except when we don't have the choice).  
2. For a random user, when you say "distribution function" it's not clear what the distribution is about. Like you could be talking about a distribution for the rolling resistance coefficient or something. I think keeping the word "size" or "diameter" is better. 

It could do this if you prefer: `const std::string size_distribution_type_str = prm.get("size_distribution_type");`

blaisb: Although we never use _ in parameter names
we use spaces

OGaboriault: My mistake, you are right. It's a typo

const std::string size_distribution_type_str = prm.get("size distribution type")


I have added a "_str", that is what I was talking about.

voferreira: I got it. Nope, using the type in the variable name is not necessary. You are right to keep it as it is

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+ */
+#include <cmath>
+#include <fstream>
+#include <iostream>
+#include <random>
+#include <string>
+#include <unordered_map>
+
+#ifndef distribution_h
+#  define distribution_h
```
lpsaavedra, line 25: This should be identical to the name of the file:

```suggestion
#ifndef distributions_h
#  define distributions_h
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+  NormalDistribution(
+    const std::unordered_map<unsigned int, double> d_averages,
+    const std::unordered_map<unsigned int, double> d_standard_deviations);
+
+  /**
+   * Carries out the size sampling of each particle inserted at a insertion
+   * time step.
+   *
+   * @param particle_number Number of particle inserted at a given insertion time step.
+   * @param particle_type The type of particles getting inserted.
```
lpsaavedra, line 63:
```suggestion
   * @param particle_type The type of particle being inserted.
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+
+class NormalDistribution : public Distribution
+{
+public:
+  /**
+   * The constructor stores the parameters necessary to define the normal
+   * distribution
+   *
+   * @param d_averages Average diameters for each types of particles.
+   * @param d_standard_deviations Standard deviation of the diameter for each types of particle.
```
lpsaavedra, line 52:
```suggestion
   * @param d_averages Average diameters for each type of particle.
   * @param d_standard_deviations Standard deviation of the diameter for each type of particle.
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param particle_number Number of particle inserted at a given insertion time step.
+   * @param particle_type The type of particles getting inserted.
+   */
+  void
+  particle_size_sampling(const unsigned int particle_number,
+                         const unsigned int particle_type) override;
+
+private:
+  // Average diameters for each particle types.
```
lpsaavedra, line 70:
```suggestion
  // Average diameters for each particle type.
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+  std::unordered_map<unsigned int, double> diameter_averages;
+  // Standard deviation of the diameter for each particle types.
```
lpsaavedra, line 72:
```suggestion
  // Standard deviation of the diameter for each particle type.
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+  std::unordered_map<unsigned int, double> standard_deviations;
+};
+
+class UniformDistribution : public Distribution
+{
+public:
+  /**
+   * The constructor store the parameters necessary to define the normal
+   * distribution
```
lpsaavedra, line 81:
```suggestion
   * The constructor store the parameters necessary to define the uniform
   * distribution
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+   * @param d_value Diameter values for each particle types.
+   */
+  UniformDistribution(const std::unordered_map<unsigned int, double> d_values);
+
+  /**
+   * Carries out the size sampling of every particles inserted at a insertion
+   * time step.
+   *
+   * @param particle_number Number of particle inserted at a given insertion time step.
+   * @param particle_type The type of particles getting inserted.
```
lpsaavedra, line 92:
```suggestion
   * @param particle_type The type of particle being inserted.
```

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+   */
+  void
+  particle_size_sampling(const unsigned int particle_number,
+                         const unsigned int particle_type) override;
+
+private:
+  // Diameter values for each particle types.
```
lpsaavedra, line 99:
```suggestion
  // Diameter values for each particle type.
```

include/dem/insertion.h
```diff
@@ -143,22 +150,12 @@ class Insertion
...
-  particle_size_sampling(std::vector<double> &particle_sizes,
-                         const double         average,
-                         const double         standard_deviation,
-                         const double         particle_number);
+  // A distribution object that carries out the attribution of diameter to every
+  // particle during an insertion time step
+  std::shared_ptr<Distribution> distribution_object;
 
+private:
+  // Stores particle diameter
```
lpsaavedra, line 158:
```suggestion
  // Stores particles diameters
```

lpsaavedra (changes requested):
Some last small comments from my side

include/dem/distributions.h
```diff
@@ -0,0 +1,103 @@
...
+  std::unordered_map<unsigned int, double> diameter_values;
+};
+
+#endif /* distribution_h */
```
lpsaavedra, line 103:
```suggestion
#endif /* distributions_h */
```

lpsaavedra (approved):

voferreira (approved):

blaisb (approved):

PR #932, opened 2023-11-22
Fix viscous initial condition

blaisb:
# Description of the problem

- Viscous initial condition was not working correctly in the Matrix Free solver

# Description of the solution

- Fixed the bug

### DIFF
--- a/include/core/parameters.h
+++ b/include/core/parameters.h
@@ -50,7 +50,6 @@ namespace Parameters
 
 
   /**
-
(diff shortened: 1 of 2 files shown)


include/core/parameters.h
```diff
@@ -50,6 +50,7 @@ namespace Parameters
 
 
   /**
+
```
blaisb, line 53:
```suggestion
```

lpsaavedra (approved):

PR #933, opened 2023-11-23
Dem fixing plane insertion distribution

OGaboriault:
# Description of the problem

- The plane insertion method was only supporting the uniform diameter distribution. 

# Description of the solution

- assign_particle_properties is now used in the PlaneInsertion::insert function. 
- A few attribute got removed because they weren't being used. 
- The assign_particle_properties function last input got removed. 

# How Has This Been Tested?

- All the tests passed. 
- The uniform type had to me specified in the dem/insertion_plane.cc test. 

# Documentation

- N/A

# Future changes

- N/A

# Comments

- The CHANGELOG file got modified because of an error I made two week ago. (The order was wrong)

### DIFF
--- a/source/dem/plane_insertion.cc
+++ b/source/dem/plane_insertion.cc
@@ -259,45 +259,13 @@ PlaneInsertion<dim>::insert(
           empty_cells_on_proc.erase(it); // Erase the first element
         }
 
-      // There's probably a better ways to define the properties of a particle,
-      // but for now it works...
-      double type     = this->current_inserting_particle_type;
-      double diameter = dem_parameters.lagrangian_physical_properties
-                          .particle_average_diameter.at(type);
-      double density =
-        dem_parameters.lagrangian_physical_properties.density_particle.at(type);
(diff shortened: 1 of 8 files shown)


source/dem/plane_insertion.cc
```diff
@@ -325,7 +293,7 @@ PlaneInsertion<dim>::insert(
                                            ref_point,
                                            starting_ID_on_proc++,
                                            cell,
-                                           properties_of_one_particle);
+                                           this->particle_properties.at(i++));
```
oguevremont, line 296: After this, will `this->particle_properties` be used again? If not, would it be useful (for memory) to reset the vector?

oguevremont (approved):
Nice

blaisb (approved):

source/dem/plane_insertion.cc
OGaboriault: I think I could remove this attribute from the class and create the vector directly in the insert() function.

oguevremont (approved):

PR #936, opened 2023-11-25
Add laser surface source flux

AmishgaAlphonius:
# Description of the problem

- The laser model previous implemented in Lethe was an exponentially decaying model (volumetric source). In laser powder bed fusion (LPBF) simulations, a surface contribution is rather preferred.

# Description of the solution

- A new assembler for a surface laser heat flux has been added to the code. To select the right laser model, a `type` parameter has to be specified in the `laser` subsection from now on.

# How Has This Been Tested?

- A new test was added (`applications_tests/lethe-fluid/heat_transfer_vof_interface_laser.prm`(.output))
- Example results haven't change (visual inspection).

# Documentation

- Parameter documentation and examples were updated.

- [x] `doc/source/parameters/cfd/laser_heat_source.rst`
- [x] `doc/source/examples/multiphysics/laser-heating/laser-heating.rst` (prm file also)
- [x] `doc/source/examples/multiphysics/laser-meltpool/laser-melt-pool.rst` (prm file also)

# Comments

- prm file indentations were fixed.
- A few parameter assertions were added, including one on the `penetration depth` parameter to avoid division by 0.

### DIFF
--- a/doc/source/parameters/cfd/laser_heat_source.rst
+++ b/doc/source/parameters/cfd/laser_heat_source.rst
@@ -90,7 +90,7 @@ Laser types
   where :math:`\phi'` is the filtered phase fraction.
 
   .. attention::
-    In this case, the heat affect the fluid initialized as ``fluid 1``.
+    In this case, the heat affects the fluid initialized as ``fluid 1``.
 
 * When ``type`` is set to ``heat_flux_vof_interface``, it **must be used in conjunction with the** :doc:`VOF auxiliary physic <./volume_of_fluid>`. This model is used to apply the heat flux, given by the expression below, only at the interface.
 


include/core/parameters.h
```diff
@@ -624,6 +624,14 @@ namespace Parameters
     // A boolean parameter that enables the calculations of laser heat source
     bool activate_laser;
 
+    // A boolean parameter that indicates if the penetration depth should be
+    // taken into account in heat source calculations.
```
AmishgaAlphonius, line 628:
```suggestion
    // Type of laser model
```

applications_tests/lethe-fluid/heat_transfer_laser.prm
```diff
@@ -32,6 +32,7 @@ end
 
 subsection laser parameters
   set enable               = true
+  set type                 = exponential decay
```
blaisb, line 35: generally parameters don't have space as far as I remember, but I am not sure? Do we have parameter values with spaces?

applications_tests/lethe-fluid/heat_transfer_vof_laser_material_interface.output
```diff
@@ -0,0 +1,64 @@
...
+Heat Transfer
+--------------
+Newton iteration: 0  - Residual:  1132
+  -Tolerance of iterative solver is : 1.132
+  -Iterative solver took : 11 steps 
+		alpha =      1 res = 0.4807
+Newton iteration: 1  - Residual:  0.4807
+  -Tolerance of iterative solver is : 0.0004807
+  -Iterative solver took : 13 steps 
+		alpha =      1 res = 0.0002065
```
blaisb, line 35: don't put the tolerances and the verbosity of the non-linear and linear solver within tests output,  those are rarely portable across machines

applications_tests/lethe-fluid/heat_transfer_vof_laser_material_interface.prm
```diff
@@ -0,0 +1,191 @@
...
+  end
+end
+
+#---------------------------------------------------
+# Linear Solver Control
+#---------------------------------------------------
+
+subsection linear solver
+  subsection fluid dynamics
+    set verbosity                             = verbose
```
blaisb, line 162: Put all verbosities to quiet please

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -8,10 +8,11 @@ If a laser heat source is present in a simulation, it can be added in this secti
...
     set enable               = false
+    set type                 = exponential decay
     set concentration factor = 2.0
     set power                = 100.0
     set absorptivity         = 0.5
-    set penetration depth    = 0.0
+    set penetration depth    = 1.0
     set beam radius          = 0.0
     set start time           = 0.0
     set end time             = 1.0
```
blaisb, line 18: so you changed the default value?

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -30,44 +31,74 @@ If a laser heat source is present in a simulation, it can be added in this secti
   end
 
 
-* The ``enable`` parameter is set to true if the problem has a laser heat source term and enables its calculation.
+* The ``enable`` parameter is set to ``true`` if the problem has a laser heat source term and enables its calculation.
 
-* Laser ``concentration factor`` parameter indicates the definition of the beam radius. In almost all the articles, it is assumed equal to 2.0.
+* The ``type`` parameter is set to ``exponential decay`` (default) if the media hit by the incident laser ray has a non-flat topography (e.g. powder bed). This enables the inclusion of the ``penetration depth`` parameter in the heat flux calculation. If the surface is flat, the ``type`` could be set at ``material interface vof`` and used in conjunction with the :doc:`VOF auxiliary physic <./volume_of_fluid>`. The different models are detailed :ref:`below <LaserTypes>`.
```
blaisb, line 36: I am not sure that's an adequate way of formualting it. Because if the surface is flat y ou could also use the surface model. The main thing here is how we want to model it, a volume contribution or surface contribution. I would not comment on the influence of the surface topology here :)

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -30,44 +31,74 @@ If a laser heat source is present in a simulation, it can be added in this secti
...
+* The ``power`` parameter sets the power of the laser :math:`[ML^2T^{-3}]`.
 
-* The ``penetration depth`` parameter determines the penetration depth of the laser in the simulation domain in the direction of emission.
+* The ``absorptivity`` parameter is defined as the fraction of the amount of incident radiation that is absorbed by the surface, and it is measured using diffuse reﬂectance spectroscopy (DRS). Generally, a constant value in the range of :math:`0.3`-:math:`0.8` (for welding processes with titanium) ise used in the literature. However, recent studies show that it varies with powder particle size distribution and the angle of incidence that changes due to the dynamic meltpool surface `[1] <https://doi.org/10.1016/j.optlastec.2018.08.012>`_.
+
+* The ``penetration depth`` parameter determines the penetration depth of the laser in the simulation domain in the direction of emission. The value should be grater than :math:`0`.
```
blaisb, line 44:
```suggestion
* The ``penetration depth`` parameter determines the penetration depth of the laser in the simulation domain in the direction of emission. The value should be greater than :math:`0`.
```

include/core/parameters.h
```diff
@@ -624,6 +624,13 @@ namespace Parameters
     // A boolean parameter that enables the calculations of laser heat source
     bool activate_laser;
 
+    // Type of laser model
+    enum class LaserType
+    {
+      exponential_decay,
+      material_interface_vof
```
blaisb, line 631: Can you add a quick comment next to the enum or in the description of the enum

include/solvers/heat_transfer_assemblers.h
```diff
@@ -317,8 +318,52 @@ class HeatTransferAssemblerLaser : public HeatTransferAssemblerBase<dim>
 };
 
 /**
- * @brief Class that assembles the laser heat source for the heat
- * transfer solver when VOF is enabled. Exponentially decaying model is
+ * @brief Class that assembles the laser heat source as a surface flux for the
+ * heat transfer solver when VOF is enabled. The laser heat source is only
+ * applied in the metal (when phase value is non-null) using the phase value
+ * alpha as a multiplying factor on the laser heat source.
```
blaisb, line 324: what does it mean when the phase value is non-null?

include/solvers/heat_transfer_assemblers.h
```diff
@@ -317,8 +318,52 @@ class HeatTransferAssemblerLaser : public HeatTransferAssemblerBase<dim>
...
+ *
+ * @tparam dim An integer that denotes the number of spatial dimensions
+ *
+ * @ingroup assemblers
```
blaisb, line 328: Can you add the @param to the constructor?

source/core/parameters.cc
```diff
@@ -49,6 +49,20 @@ DeclException3(MultipleAdaptationSizeError,
                << arg2 << ") does not correspond to the number of 'variables' ("
                << arg3 << ")");
 
+DeclException1(
+  PorousMediaDisabledError,
+  double,
+  << "You specified a penetration depth of " << arg1
+  << ". However, the 'type' parameter is not set to 'exponential decay'.");
```
blaisb, line 56: Is that necessary? I mean you can have a value of the parameter and not use the model, would that really pose au problem?

source/core/parameters.cc
```diff
@@ -1388,12 +1407,38 @@ namespace Parameters
...
+                          ParameterStrictlyGreaterThanError("penetration depth",
+                                                            penetration_depth,
+                                                            0.0));
+            }
+          else
+            {
+              AssertThrow(laser_type == LaserType::material_interface_vof &&
+                            penetration_depth == 0.0,
+                          PorousMediaDisabledError(penetration_depth));
+            }
```
blaisb, line 1438: No need to do this assertion right, or are you actually using the penetration depth in the vof model? That's a bit confusing from my end

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -786,24 +788,142 @@ HeatTransferAssemblerLaser<dim>::assemble_rhs(
...
+            laser_parameters->perpendicular_plane_coordinate_two);
+        }
+      laser_location[dim - 1] =
+        laser_scan_path.value(laser_location,
+                              laser_parameters->beam_orientation_coordinate);
+
+      // Get laser location on the operation surface
+      Point<dim - 1> laser_location_on_surface;
+      for (unsigned int d = 0; d < dim - 1; d++)
+        laser_location_on_surface[d] = laser_location[d];
```
blaisb, line 854: So in this model you still specify the laser focal point? @hepap do you think that'S the right thing? I wold have thought we would specify something like just the laser vector and a point on that vector and assume automatically that the focal point was the intersection between the surface and that laser vector. AKA no focal point. Otherwise it's tricky no?

blaisb (changes requested):
Some comments. I am not sure if using the focal point is the best idea. Like why not assume that the surfaces on the vector+point all see the vector and no need to specify a focal point per say. We could calculate the focal point dynamically. Otehrwise we will always wonder where the heck the surface is. Seems problematic to me. We can discuss this on Monday :)

doc/source/parameters/cfd/laser_heat_source.rst
AmishgaAlphonius: It was changed because it became incoherent in the sense that it does a division by 0 (by default) when the laser is activated. And the radiation application-test (``heat_transfert_vof_radiation``) was failing after I added my assertion on the penetration depth value. When talking with @hepap, she said we should be able to use the radiative cooling with or without the laser, and since the power is by default 0.0 changing the depth default value should't hinder much. However, I now realize that the default power value is not 0.0, that's my mistake. I could change it... but I'm now starting to wonder if the radiative cooling should belong in the ``laser`` subsection, since we should be able to use it without the laser also, even though that wouldn't be it's primary use.

doc/source/parameters/cfd/laser_heat_source.rst
AmishgaAlphonius: That was my understanding of why one would use a _volumetric flux_ instead of the _surface flux_, but it's true that both could be used. I'll change the formulation avoiding this explanation and simply explaining the difference as one being a volumetric source and the other a surface one!

include/core/parameters.h
AmishgaAlphonius: Like explaining further that ``exponential_decay`` is a volumetric contribution and ``material_interface_vof`` is a surface one?

source/core/parameters.cc
AmishgaAlphonius: It was more to avoid the user mistake of thinking that they were changing something by changing the ``penetration depth`` while using the ``material_interface_vof`` laser model... to avoid using mistakenly the wrong model. However, if you feel it is unnecessary, I could take it out.

source/core/parameters.cc
AmishgaAlphonius: This goes with my explanation on the comment of the exception.

include/solvers/heat_transfer_assemblers.h
AmishgaAlphonius: Oupss... this is not the appropriate @brief... I'll correct that!

AmishgaAlphonius (commented):
I'll apply tomorrow everything I said I would do (I'm having trouble connecting to my desktop from home). :)

doc/source/parameters/cfd/laser_heat_source.rst
blaisb: Yeah i guess teh radiative cooling would not really just belong to laser although right now it's the only context I can think of us using it to be honest...

blaisb:
> I'll apply tomorrow everything I said I would do (I'm having trouble connecting to my desktop from home). :)

Take your time. If my calendar is right, today is Sunday. No need to work on Sundays (although I do respect your desire to work on a Sunday if you wish to :) ).

source/core/parameters.cc
```diff
@@ -1339,6 +1353,12 @@ namespace Parameters
     prm.enter_subsection("laser parameters");
     {
       prm.declare_entry("enable", "false", Patterns::Bool(), "Activate laser");
+      prm.declare_entry("type",
+                        "exponential decay",
+                        Patterns::Selection(
+                          "exponential decay|material interface vof"),
+                        "Type of laser model used."
+                        "Choices are <exponential decay|surface vof>.");
```
hepap, line 1361:
```suggestion
                          "exponential decay|material interface vof"),
                        "Type of laser model used."
                        "Choices are <exponential decay|material interface vof>.");
```

source/solvers/heat_transfer.cc
```diff
@@ -274,12 +274,25 @@ HeatTransfer<dim>::setup_assemblers()
...
+          if (this->simulation_parameters.laser_parameters->laser_type ==
+              Parameters::Laser<dim>::LaserType::material_interface_vof)
+            {
+              this->assemblers.push_back(
+                std::make_shared<
+                  HeatTransferAssemblerLaserMaterialInterfaceVOF<dim>>(
+                  this->simulation_control,
+                  this->simulation_parameters.laser_parameters));
+            }
+          else // Laser is applied only to the metal phase (volumetric flux)
```
hepap, line 288:
```suggestion
          else // Laser is applied in the metal phase as a volumetric source
```

source/solvers/heat_transfer_assemblers.cc
hepap: I would assume that for the surface flux, we only need the distance  between the "laser vector" and the current quadrature point in think, while in the exponential decay the depth is required.

I think it is the same as saying that the 2D gaussian distribution is the same no matter the "depth" of the interface.

I will check that in more details tomorrow, but I think the implementation bellow is still good.

hepap (commented):
Nice work! I will check for the focal point thing Monday :)

include/core/parameters.h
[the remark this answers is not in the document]
AmishgaAlphonius: I'll go with `heat_flux_vof_interface`!

blaisb:
@hepap I'll let you give it another read then I will re-read also

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -786,24 +788,144 @@ HeatTransferAssemblerLaser<dim>::assemble_rhs(
...
+      // direction) between the laser focal point and the quadrature points.
+      // Here, the focal point is any given point on the laser's axis.
+      // Hence, we get the laser location (laser_location) as a Point<dim>, in
+      // which the first and second components show the position of the laser
+      // focal point in a plane perpendicular to the emission direction, and
+      // the (dim-1)th component denotes the position of the laser focal point
+      // in the direction of emission. Then we use dim-1 auxiliary variables
+      // (Point<dim-1> laser_location_on_surface) to store the position of the
+      // laser focal point in the perpendicular plane to the emission
+      // direction.
```
hepap, line 837: I think we don't need the part about the third component :)

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -786,24 +788,144 @@ HeatTransferAssemblerLaser<dim>::assemble_rhs(
...
+            laser_parameters->perpendicular_plane_coordinate_two);
+        }
+      laser_location[dim - 1] =
+        laser_scan_path.value(laser_location,
+                              laser_parameters->beam_orientation_coordinate);
+
+      // Get laser location on the operation surface
+      Point<dim - 1> laser_location_on_surface;
+      for (unsigned int d = 0; d < dim - 1; d++)
+        laser_location_on_surface[d] = laser_location[d];
```
hepap, line 856: So now I think we can just use laser_location[0] and laser_location[1] for the distance computation at l.893

hepap (commented):
Two small last comments :)

source/solvers/heat_transfer_assemblers.cc
AmishgaAlphonius: You're right! Nice catch! :)

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -30,44 +31,74 @@ If a laser heat source is present in a simulation, it can be added in this secti
...
+* ``subsection free surface radiation``: In additive manufacturing simulations, radiation at the interface between the air and the metal is a significant cooling mechanism. When this interface (i.e., free surface) is resolved by the :doc:`volume_of_fluid` solver, the ``free surface radiation`` subsection defines the parameters to impose this radiation cooling following the Stefan-Boltzmann law of radiation:
+
+  .. math::
+      q_\text{rad} = \epsilon \sigma (T^4 - T_\text{inf}^4)
 
-  * ``enable``: controls if the radiation cooling is enabled. The radiation sink is apply in the heat transfer solver in the cells where the phase fraction gradient norm :math:`|\nabla \alpha|` is non-null.
+  * ``enable``: controls if the radiation cooling is enabled. The radiation sink is applied in the heat transfer solver in the cells where the filtered phase fraction gradient norm :math:`|\nabla \phi'|` is non-null.
```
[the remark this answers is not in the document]
Luckabarbeau: I saw the equation and the code, so I understand better what you are saying here. But instead of saying that it is applied in the cells where the phase gradient is non-0, I would say that the norm of the phase gradient weights the sink term since you solve this on all cells anyway.

Luckabarbeau (commented):
Very good work! There are a few comments/questions here and there. But I think it is pretty much ready to merge after.

doc/source/parameters/cfd/laser_heat_source.rst
```diff
@@ -30,44 +31,74 @@ If a laser heat source is present in a simulation, it can be added in this secti
...
+
+  When the ``exponential_decay`` is used in conjunction with the :doc:`VOF auxiliary physic <./volume_of_fluid>` the equation takes the following form:
+
+  .. math::
+      q(x,y,z) = \frac{\phi' \eta \alpha P}{\pi r^2 \mu} \exp{\left(-\eta \frac{r^2}{R^2}\right)} \exp{\left(- \frac{|z|}{\mu}\right)}
+
+  where :math:`\phi'` is the filtered phase fraction.
+
+  .. attention::
+    In this case, the heat affect the fluid initialized as ``fluid 1``.
```
hepap, line 93:
```suggestion
    In this case, the heat affects the fluid initialized as ``fluid 1``.
```

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -786,24 +788,139 @@ HeatTransferAssemblerLaser<dim>::assemble_rhs(
...
+  const double absorptivity         = laser_parameters->laser_absorptivity;
+  const double laser_start_time     = laser_parameters->start_time;
+  const double laser_end_time       = laser_parameters->end_time;
+  const double beam_radius          = laser_parameters->beam_radius;
+  const double current_time = this->simulation_control->get_current_time();
+
+  if (current_time >= laser_start_time && current_time <= laser_end_time)
+    {
+      // Get laser path
+      Function<dim> &laser_scan_path = *(laser_parameters->laser_scan_path);
```
blaisb, line 823: This should be refactored later to be set outside of the assembler. It's not necessary to set the time at every cell...

blaisb (approved):

blaisb:
Seem CI is broken down because some stuff has changed in MatrixFree... Miam... I will merge but I will look at the rest next week. Anyway the broken part is unrelated to you

PR #939, opened 2023-11-27
Bugfix, dem fixing critical rayleigh time step

OGaboriault:
# Description of the problem

- The inspection of the DEM time step was wrong. It should be searching for the more restrictive rayleigh time, that is the minimum rayleigh time, not the maximum. 

# Description of the solution

- The rayleigh time step is initialize to infinity. 
- We use std::min instead of std::max. 

# How Has This Been Tested?

- One test has been updated. 
- All the other test are still succeeding 
- 
# Documentation

- N/A


OGaboriault:
Ehmm.... a multicore test is failling on dealii:master but not on dealii:v9.5.0

PR #940, opened 2023-11-27
Remove avg dia + fix documentation

blaisb:
# Description of the problem

- The standard deviation parameter was not clearly defined in the documentation
-  We had an average diameter and diameter parameter. This is confusing for no reason.

# Description of the solution

- Clear our the documentation and remove the average diameter parameter. It does not impact anything on the data structures anyway so yay!

# How Has This Been Tested?

-  All unit tests passed on my machine.

# Documentation

-  Changed documentation on the particle type themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## [Master] - 2023-11-27
 
+### Removed
+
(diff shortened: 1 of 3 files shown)


OGaboriault:
You should add a comment in the Changelog about the removal of "average diameter" parameter in the prm files.

blaisb:
> You should add a comment in the Changelog about the removal of "average diameter" parameter in the prm files.

Absolutely. You are absolutely right. This has been changed

PR #941, opened 2023-11-28
Dem refactor distribution

OGaboriault:
# Description of the problem

- Using many particle types with different size distribution wasn't possible. 
- The maximum_particle_size calculation was about to become false if we add new type of distribution. 

# Description of the solution

- New way to parse the size_distribution. 
- Distribution class have two new virtual function. 
- Insertion class stores a vector of Distribution. 
- find_maximum_particle_size  class has been removed. 

# How Has This Been Tested?

- All test are still pass.

# Future changes

- Add new type of distribution.

# Comments

- Probably a lot of stuff to change with this PR.


blaisb:
This HAKJSHLKAHSLK periodic test will need to be fixed. I'll try to fix it this week. Pisses me off to no avail :)

OGaboriault:
Ok, I went from 2k modified lines to 375 lines with this last push. I think this will be help you with your reviews.

source/core/parameters_lagrangian.cc
```diff
@@ -175,70 +172,71 @@ namespace Parameters
...
+      g[0] = prm.get_double("gx");
+      g[1] = prm.get_double("gy");
+      g[2] = prm.get_double("gz");
+
+      particle_type_number = prm.get_integer("number of particle types");
 
-        if (particle_type_number >= 1)
+      if (particle_type_number >= 1)
+        {
+          prm.enter_subsection("particle type 0");
```
blaisb, line 184: Instead of having a ton of ifs like that, why not just make a single for loop?

blaisb (changes requested):
First comment, will finish tonight

source/dem/dem.cc
```diff
@@ -203,26 +204,59 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
         parameters.model_parameters.solid_fraction_threshold);
     }
 
+  distribution_object_container.reserve(
+    parameters.lagrangian_physical_properties.particle_type_number);
+  maximum_particle_diameter = 0;
+  for (unsigned int counter = 0;
```
voferreira, line 210: "counter" doesn't feel like an iterator. I would use "type_number".

source/dem/dem.cc
```diff
@@ -203,26 +204,59 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
         parameters.model_parameters.solid_fraction_threshold);
     }
 
+  distribution_object_container.reserve(
```
voferreira, line 207: Given line 74, do you need to reserve?

source/dem/dem.cc
```diff
@@ -203,26 +204,59 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
...
+
+
   // Calling input_parameter_inspection to evaluate input parameters in the
   // parameter handler file, finding maximum particle diameter used in
   // polydisperse systems
-  maximum_particle_diameter =
-    find_maximum_particle_size(parameters.lagrangian_physical_properties,
-                               standard_deviation_multiplier);
+
+
```
voferreira, line 243: Extra line?

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1313,19 +1310,45 @@ CFDDEMSolver<dim>::dem_setup_contact_parameters()
...
+  distribution_object_container.reserve(
+    dem_parameters.lagrangian_physical_properties.particle_type_number);
+
+  maximum_particle_diameter = 0.;
+  for (unsigned int counter = 0;
+       counter <
+       dem_parameters.lagrangian_physical_properties.particle_type_number;
+       counter++)
+    {
+      if (dem_parameters.lagrangian_physical_properties.distribution_type.at(
```
voferreira, line 1322: Same comment about the variable

voferreira (changes requested):
Good Job. It is good to have distribution types separately like this. Just some small comments here and there but nothing fancy.

include/fem-dem/cfd_dem_coupling.h
```diff
@@ -272,6 +271,7 @@ class CFDDEMSolver : public GLSVANSSolver<dim>
   Visualization<dim>            visualization_object;
   BoundaryCellsInformation<dim> boundary_cell_object;
 
+
```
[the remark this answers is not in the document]
OGaboriault: I have add something and removed it afterward... sorry

include/dem/input_parameter_inspection.h
```diff
@@ -31,16 +32,15 @@ using namespace std;
  *
  * @param dem_parameters Input DEM parameters in the parameter handler file
  * @param pcout Printing in parallel
- * @param standard_deviation_multiplier Constant standard deviation multiplier.
- * It is defined in the dem constructor and it is equal to 2.5 to cover 99% of
- * the particle size distribution
+ * @param distribution_object_container Contain all the types of distribution being used for each type of particle.
```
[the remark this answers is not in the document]
OGaboriault: You are right, but clang-format doesn't do anything...

I will put this on two lines.

source/core/parameters_lagrangian.cc
OGaboriault: I will try to do something similar to the parsing of the solid object.

OGaboriault: Done

blaisb:
@OGaboriault can you rebase?

OGaboriault:
Done

voferreira (approved):

PR #945, opened 2023-12-01
Add missing include in MF operators

lpsaavedra:
# Description of the problem

Lethe was not compiling with the master version of deal.II due to the MGTools calls in the matrix-free operators.

# Description of the solution

The `deal.II/multigrid/mg_tools.h` needed to be included in `include/solvers/mf_navier_stokes_operators.h`  due to recent changes in deal.ii.


blaisb:
Will merge straight after CI @lpsaavedra  Thanks!

PR #953, opened 2023-12-06
Add heat flux postprocessor

mivaia:
# Description of the problem

- In the context of conjugated heat transfer simulations, there was no way to get the local heat fluxes of the domain.

# Description of the solution

- A new `DataPostprocessorVector` child class was made, namely, `HeatFluxPostprocessor`. This class postprocesses the local heat fluxes of the fluids and solids of the domain. The values for each fluid and solid is outputed individually.

# How Has This Been Tested?

- CI 

# Comments

An example using this HeatFluxPostprocessor values will be made soon.

### DIFF
--- a/include/solvers/postprocessors.h
+++ b/include/solvers/postprocessors.h
@@ -533,7 +533,7 @@ public:
     , thermal_conductivity_model(p_thermal_conductivity_model)
     , material_id(p_material_id)
   {}
-  
+
   virtual void
   evaluate_scalar_field(
     const DataPostprocessorInputs::Scalar<dim> &inputs,
--- a/include/solvers/postprocessors.h
+++ b/include/solvers/postprocessors.h
@@ -533,6 +533,7 @@ public:
     , thermal_conductivity_model(p_thermal_conductivity_model)
     , material_id(p_material_id)
   {}
+  
   virtual void
   evaluate_scalar_field(
     const DataPostprocessorInputs::Scalar<dim> &inputs,
--- a/include/solvers/postprocessors.h
+++ b/include/solvers/postprocessors.h
@@ -511,22 +511,26 @@ private:
 
 
 /**
- * @class Calculates heat flux of a material at every quadrature point.
- * @param p_thermal_conductivity_model Thermal conductivity model of the material
+ * @class Calculates heat flux (q = -k * grad T) of a material at every
+ * quadrature point.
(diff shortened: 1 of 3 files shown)


source/solvers/heat_transfer.cc
```diff
@@ -667,12 +667,43 @@ HeatTransfer<dim>::copy_local_rhs_to_global_rhs(
...
+
+  // Get thermal conductivity models
+  thermal_conductivity_models =
+    this->simulation_parameters.physical_properties_manager
+      .get_thermal_conductivity_vector();
+
+  const double n_fluids = this->simulation_parameters
+                            .physical_properties_manager.get_number_of_fluids();
+  const double n_solids = this->simulation_parameters
+                            .physical_properties_manager.get_number_of_solids();
```
oguevremont, line 684:
```suggestion
  const unsigned int n_fluids = this->simulation_parameters
                            .physical_properties_manager.get_number_of_fluids();
  const unsigned int n_solids = this->simulation_parameters
                            .physical_properties_manager.get_number_of_solids();
```

include/solvers/postprocessors.h
```diff
@@ -508,4 +508,66 @@ class DensityPostprocessor : public DataPostprocessorScalar<dim>
...
+    const std::string                         p_material_string = "f",
+    const unsigned int                        p_material_id     = 0,
+    const unsigned int                        p_id              = 0)
+    : DataPostprocessorVector<dim>("heat_flux_" + p_material_string +
+                                     Utilities::to_string(p_id, 2),
+                                   update_values | update_gradients)
+    , thermal_conductivity_model(p_thermal_conductivity_model)
+    , material_id(p_material_id)
+  {}
+  virtual void
```
oguevremont, line 536:
```suggestion
  {}
  
  virtual void
```

oguevremont (approved):
Very minor comments, good work! After my suggestions it is approved

blaisb (changes requested):
Minor minor changes, but please add a bit more comments

hepap (commented):
Nice addition! I agree with Bruno that a bit more comments would help :) Also, may be you can add a test to verify your implementation? I don't know if it is relevant at this point, but a pseudo 1D case with different layers of different materials could be interesting (there are analytical solutions for that in the BSL I think!)

AmishgaAlphonius (approved):

blaisb (approved):

PR #954, opened 2023-12-07
Dem histogram distribution

OGaboriault:
# Description of the new feature

- It`s now possible to insertion particles with a list of diameters and a list of probabilities. This will be really useful for particle not following a conventional size distribution. 

# Description of the solution

- A new distribution type has been added. 

# How Has This Been Tested?

- The small-rotating scale loading prm file has been modified to test the implementation. 
- All the previous test have been tested and have succeed on my machine. 

# Documentation

- The lagrangian_physical_properties.rst file have been updated following the same approach used with in insertion_info.rst file.
- Some default parameters in the documentation were not good. 

# Future changes

- I could add a test for this new distribution, but we are not using a random seed for the diameter generation, thus I'm not sure if this would work in a test.

### DIFF
--- a/doc/source/parameters/dem/lagrangian_physical_properties.rst
+++ b/doc/source/parameters/dem/lagrangian_physical_properties.rst
@@ -66,18 +66,19 @@ In this subsection, gravitational acceleration, and the physical properties of t
     If the particles in a simulation are monodispersed and have the same physical properties, the ``number of particle types`` should be equal to zero. For polydispersed systems, the ``number of particle types`` is selected equal to the number of particles types in the simulation. For each particle type, a separate subsection ``particle type n`` should be defined (n starts from zero to ``number of particle types`` - 1) which contains all the physical properties related to that particle type.
 
 * The ``size distribution type`` parameter specifies the size distribution for each particle type. For each particle type, three ``size distribution type`` can be defined: ``uniform``, ``normal`` and ``custom``.
+
   - For the ``uniform`` size distribution, the diameter of the particles is constant.
   - For the ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average diameter and a standard deviation.
   - For the ``custom`` size distribution, particle diameters are sampled from a list of diameters with a corresponding list of probabilities.
 
 .. note::
-    In the ``custom`` size distribution, the probability values are based on the volume taken by all the particles of the associated diameter, not to the total number of particles. For example, if a probability is equal to ``0.5`` , this means that haft of the total volume of inserted particles will be occupied by particle with the associated diameter value.
+    In the ``custom`` size distribution, the probability values are based on the volume fraction taken by all the particles of the associated diameter, not to the total number of particles. For example, if a probability is equal to ``0.5`` , this means that half of the total volume of inserted particles will be occupied by particle with the associated diameter value.
 
 * The ``diameter`` parameter defines the diameter of the particles in a ``uniform`` distribution. In the case of a ``normal`` distribution, this parameter indicates the average diameter.
 
 * For a ``normal`` distribution, the ``standard deviation`` parameter should be defined to indicate the standard deviation on the particle size distribution.
 
-* For a ``custom`` distribution, the ``custom diameters`` defines the different diameter values used when generating particles. The ``custom probabilities`` defines the probabilities corresponding to each diameter value previously declared based on volume, not on the number of particles. Both list must have the same length.
+* For a ``custom`` distribution, the ``custom diameters`` parameter defines the different diameter values used when generating particles. The ``custom probabilities`` parameter defines the probabilities corresponding to each diameter value previously declared based on volume fraction, not on the number of particles. Both list must have the same length.
 
 * The ``number of particles`` parameter defines the number of particles for each type.
 
--- a/doc/source/parameters/dem/lagrangian_physical_properties.rst
+++ b/doc/source/parameters/dem/lagrangian_physical_properties.rst
@@ -22,15 +22,15 @@ In this subsection, gravitational acceleration, and the physical properties of t
     # Entering particle type 0
     subsection particle type 0
 
-      # Choices are uniform, normal or histogram
+      # Choices are uniform, normal or custom
       set size distribution type            = uniform
 
       # If distribution type = uniform or normal
       set diameter                          = 0.001
 
-      # If distribution type = histogram
-      set histogram diameters list          = 0.001 , 0.0005
-      set histogram probabilities list      = 0.6   , 0.4
+      # If distribution type = custom
+      set custom diameters list          = 0.001 , 0.0005
+      set custom probabilities list      = 0.6   , 0.4
 
       # If distribution type = normal
       set standard deviation                = 0.0
@@ -65,16 +65,19 @@ In this subsection, gravitational acceleration, and the physical properties of t
 .. note::
     If the particles in a simulation are monodispersed and have the same physical properties, the ``number of particle types`` should be equal to zero. For polydispersed systems, the ``number of particle types`` is selected equal to the number of particles types in the simulation. For each particle type, a separate subsection ``particle type n`` should be defined (n starts from zero to ``number of particle types`` - 1) which contains all the physical properties related to that particle type.
 
-* The ``size distribution type`` parameter specifies the size distribution for each particle type. The acceptable choices are ``uniform`` and ``normal`` distributions.
+* The ``size distribution type`` parameter specifies the size distribution for each particle type. For each particle type, three ``size distribution type`` can be defined: ``uniform``, ``normal`` and ``custom``.
+  - For the ``uniform`` size distribution, the diameter of the particles is constant.
+  - For the ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average diameter and a standard deviation.
+  - For the ``custom`` size distribution, particle diameters are sampled from a list of diameters with a corresponding list of probabilities.
 
 .. note::
-    For each particle type, three ``size distribution type``s can be defined: ``uniform``, ``normal`` and ``histogram``. In ``uniform`` size distribution, the diameter of the particles is constant. In ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average diameter and a standard deviation of ``standard deviation``.  In the "histogram" size distribution, particle diameters are sampled from a list of diameters and a probability for each diameter value.
+    In the ``custom`` size distribution, the probability values are based on the volume taken by all the particles of the associated diameter, not to the total number of particles. For example, if a probability is equal to ``0.5`` , this means that haft of the total volume of inserted particles will be occupied by particle with the associated diameter value.
 
 * The ``diameter`` parameter defines the diameter of the particles in a ``uniform`` distribution. In the case of a ``normal`` distribution, this parameter indicates the average diameter.
 
 * For a ``normal`` distribution, the ``standard deviation`` parameter should be defined to indicate the standard deviation on the particle size distribution.
 
-* For a ``histogram`` distribution, the ``histogram diameters list`` defines each diameter to choose from when generating the diameter values. The ``histogram probabilities list`` defines the probability that a diameter value will be chosen for one particles. Both list must have the same length.
+* For a ``custom`` distribution, the ``custom diameters`` defines the different diameter values used when generating particles. The ``custom probabilities`` defines the probabilities corresponding to each diameter value previously declared based on volume, not on the number of particles. Both list must have the same length.
 
 * The ``number of particles`` parameter defines the number of particles for each type.
 
--- a/doc/source/parameters/dem/lagrangian_physical_properties.rst
+++ b/doc/source/parameters/dem/lagrangian_physical_properties.rst
@@ -29,8 +29,8 @@ In this subsection, gravitational acceleration, and the physical properties of t
       set diameter                          = 0.001
 
       # If distribution type = custom
-      set custom diameters list          = 0.001 , 0.0005
-      set custom probabilities list      = 0.6   , 0.4
+      set custom diameters                  = 0.001 , 0.0005
+      set custom probabilities              = 0.6   , 0.4
 
       # If distribution type = normal
       set standard deviation                = 0.0
--- a/doc/source/parameters/dem/lagrangian_physical_properties.rst
+++ b/doc/source/parameters/dem/lagrangian_physical_properties.rst
@@ -29,8 +29,8 @@ In this subsection, gravitational acceleration, and the physical properties of t
       set diameter                          = 0.001
 
       # If distribution type = histogram
-      set histogram diameters list           = 0.001
-      set histogram probabilities list      = 1.
+      set histogram diameters list          = 0.001 , 0.0005
+      set histogram probabilities list      = 0.6   , 0.4
 
       # If distribution type = normal
       set standard deviation                = 0.0
@@ -74,7 +74,7 @@ In this subsection, gravitational acceleration, and the physical properties of t
 
 * For a ``normal`` distribution, the ``standard deviation`` parameter should be defined to indicate the standard deviation on the particle size distribution.
 
-* For a ``histogram`` distribution, the ``histogram diameters list`` defines each diameter to choose from when generating the diameter values. The ``histogram probabilities list`` defines the probability that a diameter value will be chosen for one particles.
+* For a ``histogram`` distribution, the ``histogram diameters list`` defines each diameter to choose from when generating the diameter values. The ``histogram probabilities list`` defines the probability that a diameter value will be chosen for one particles. Both list must have the same length.
 
 * The ``number of particles`` parameter defines the number of particles for each type.
 
--- a/doc/source/parameters/dem/lagrangian_physical_properties.rst
+++ b/doc/source/parameters/dem/lagrangian_physical_properties.rst
@@ -21,56 +21,38 @@ In this subsection, gravitational acceleration, and the physical properties of t
 
     # Entering particle type 0
     subsection particle type 0
-      # Size distribution of particle type 0
+
+      # Choices are uniform, normal or histogram
       set size distribution type            = uniform
 
-      # Particle diameter
-      set diameter                          = 0.005
+      # If distribution type = uniform or normal
+      set diameter                          = 0.001
 
-      # Standard deviation
+      # If distribution type = histogram
+      set histogram diameters list           = 0.001
+      set histogram probabilities list      = 1.
+
+      # If distribution type = normal
       set standard deviation                = 0.0
 
-      # Number of particles in type 0
-      set number of particles               = 132300
-
-      # Particle density
-      set density particles                 = 2000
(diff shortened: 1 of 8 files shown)


source/core/parameters.cc
```diff
@@ -3260,6 +3260,20 @@ namespace Parameters
...
+  {
+    std::string              full_str = prm.get(entry_string);
+    std::vector<std::string> vector_of_string(
+      Utilities::split_string_list(full_str));
+
+    std::vector<double> vector_of_double =
+      Utilities::string_to_double(vector_of_string); // The error is thrown here
+
+
+    return vector_of_double;
```
AmishgaAlphonius, line 3274:
```suggestion
    std::vector<double> vector_of_double =
      Utilities::string_to_double(vector_of_string);

    return vector_of_double;
```

source/core/parameters_lagrangian.cc
```diff
@@ -11,13 +11,21 @@ namespace Parameters
...
                         Patterns::Double(),
                         "Particle diameter");
+      prm.declare_entry("histogram diameters list",
+                        "0.001 , 0.0005",
+                        Patterns::List(Patterns::Double()),
+                        "Diameter values for a histogram distribution");
+      prm.declare_entry("histogram probabilities list",
+                        "0.6 , 0.4",
+                        Patterns::List(Patterns::Double()),
+                        "Probabilities of each diameter");
```
AmishgaAlphonius, line 28:
```suggestion
      prm.declare_entry(
        "histogram probabilities list",
        "0.6 , 0.4",
        Patterns::List(Patterns::Double()),
        "Probabilities of each diameter of the histogram distribution");
```

source/core/parameters_lagrangian.cc
```diff
@@ -64,6 +72,23 @@ namespace Parameters
...
+      particle_histogram_diameter.at(particle_type) =
+        entry_string_to_vector(prm, "histogram diameters list");
+      particle_histogram_probability.at(particle_type) =
+        entry_string_to_vector(prm, "histogram probabilities list");
+
+      double probability_sum =
+        std::reduce(particle_histogram_probability.at(particle_type).begin(),
+                    particle_histogram_probability.at(particle_type).end());
+
+      // We make sure that the cummulative probability is equal to 1.
```
AmishgaAlphonius, line 84:
```suggestion
      // We make sure that the cumulative probability is equal to 1.
```

doc/source/parameters/dem/lagrangian_physical_properties.rst
```diff
@@ -86,12 +68,14 @@ In this subsection, gravitational acceleration, and the physical properties of t
 * The ``size distribution type`` parameter specifies the size distribution for each particle type. The acceptable choices are ``uniform`` and ``normal`` distributions.
 
 .. note::
-    For each particle type, two ``size distribution type``s can be defined: ``uniform`` and ``normal``. In ``uniform`` size distribution, the diameter of the particles is constant, while in ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average of ``average diameter`` and standard deviation of ``standard deviation``.
+    For each particle type, three ``size distribution type``s can be defined: ``uniform``, ``normal`` and ``histogram``. In ``uniform`` size distribution, the diameter of the particles is constant. In ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average diameter and a standard deviation of ``standard deviation``.  In the "histogram" size distribution, particle diameters are sampled from a list of diameters and a probability for each diameter value.
```
blaisb, line 71:
```suggestion
    For each particle type, three ``size distribution type``s can be defined: ``uniform``, ``normal`` and ``histogram``. In ``uniform`` size distribution, the diameter of the particles is constant. In ``normal`` size distribution, the particle diameters are sampled from a normal distribution with an average diameter and a standard deviation of ``standard deviation``.  In the "histogram" size distribution, particle diameters are sampled from a list of diameters and a probability for each diameter value.
```

Can you add an information regarding the fact that this is a probability is number based and not mass based. Ideally I would specify it's a number fraction.

source/core/parameters_lagrangian.cc
```diff
@@ -74,6 +99,10 @@ namespace Parameters
         {
           distribution_type.at(particle_type) = SizeDistributionType::normal;
         }
+      else if (size_distribution_type_str == "histogram")
+        {
+          distribution_type.at(particle_type) = SizeDistributionType::histogram;
+        }
       else
         {
           throw(std::runtime_error("Invalid size distribution type "));
```
AmishgaAlphonius, line 108:
```suggestion
          throw(std::runtime_error(
            "Invalid size distribution type. Choices are <uniform|normal|histogram>."));
```

doc/source/parameters/dem/lagrangian_physical_properties.rst
```diff
@@ -86,12 +68,14 @@ In this subsection, gravitational acceleration, and the physical properties of t
...
 
 * The ``diameter`` parameter defines the diameter of the particles in a ``uniform`` distribution. In the case of a ``normal`` distribution, this parameter indicates the average diameter.
 
 * For a ``normal`` distribution, the ``standard deviation`` parameter should be defined to indicate the standard deviation on the particle size distribution.
 
+* For a ``histogram`` distribution, the ``histogram diameters list`` defines each diameter to choose from when generating the diameter values. The ``histogram probabilities list`` defines the probability that a diameter value will be chosen for one particles. Both list must have the same length.
+
```
blaisb, line 78: Again, do you want mass fraction (volume fraction) or number fraction?
I would use a volume fraction as a specifier instead of a number fractoin

AmishgaAlphonius (commented):
Nice work! I have a few comments and suggestions :)

include/dem/distributions.h
```diff
@@ -14,6 +14,7 @@
  * ---------------------------------------------------------------------
  *
  */
+#include <algorithm>
```
blaisb, line 17:
```suggestion
```

source/core/parameters_lagrangian.cc
blaisb, line 28: To be changed to volume fraction

source/dem/distributions.cc
```diff
@@ -68,3 +68,59 @@ UniformDistribution::find_max_diameter()
 {
   return this->diameter_values;
 }
+
+HistogramDistribution::HistogramDistribution(
```
blaisb, line 72: Add algorithm in this .cc file and you would be good

blaisb (changes requested):
Tiny changes, nothing major

include/dem/distributions.h
```diff
@@ -128,4 +129,46 @@ class UniformDistribution : public Distribution
...
 
+class HistogramDistribution : public Distribution
+{
+public:
+  /**
+   * @brief The constructor stores the parameters necessary to define the histogram
+   * distribution.
+   * @param d_list
+   * @param d_probabilities
+   * @param number_of_particle_types
```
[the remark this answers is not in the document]
hepap: add also the **param**  for `number_of_particle_types`

include/dem/distributions.h
```diff
@@ -128,4 +129,46 @@ class UniformDistribution : public Distribution
...
+   * @brief Return the minimal diameter value from the @diameter_hist_value vector.
+   */
+  double
+  find_min_diameter() override;
+
+  /**
+   * @brief Return the maximal diameter value from the @diameter_hist_value vector.
+   */
+  double
+  find_max_diameter() override;
```
hepap, line 165: I think you can add the **return** description in the brief (just so that it is uniform, even if it is clear in the brief haha)

source/core/parameters_lagrangian.cc
```diff
@@ -64,6 +72,23 @@ namespace Parameters
...
+        entry_string_to_vector(prm, "histogram diameters list");
+      particle_histogram_probability.at(particle_type) =
+        entry_string_to_vector(prm, "histogram probabilities list");
+
+      double probability_sum =
+        std::reduce(particle_histogram_probability.at(particle_type).begin(),
+                    particle_histogram_probability.at(particle_type).end());
+
+      // We make sure that the cummulative probability is equal to 1.
+      if (probability_sum != 1.0)
```
hepap, line 85: I'm scared that it could be true even if the sum=1.0 because of the machine accuracy
```suggestion
      if (std::abs(probability_sum-1.0) > 1e-12)
```

source/core/parameters_lagrangian.cc
```diff
@@ -64,6 +72,23 @@ namespace Parameters
...
+        {
+          for (double &i : particle_histogram_probability.at(particle_type))
+            {
+              i = i / probability_sum;
+            }
+        }
```
hepap, line 91: I get that you are rescaling, but if it's a probability density function, should the sum be always 1? If it is not 1, I think it is an error in the .prm  :o

source/dem/distributions.cc
```diff
@@ -68,3 +68,59 @@ UniformDistribution::find_max_diameter()
...
+  const std::vector<double> &d_list,
+  const std::vector<double> &d_probabilities)
+{
+  diameter_hist_values = d_list;
+  std::vector<double> cumulative_probability_vector;
+  cumulative_probability_vector.reserve(d_probabilities.size());
```
hepap, line 78: Is there a reason for the use of reserve?

```suggestion
  const unsigned int n_diameters = d_probabilities.size();
  std::vector<double> cumulative_probability_vector(n_diameters);
```

source/dem/distributions.cc
```diff
@@ -68,3 +68,59 @@ UniformDistribution::find_max_diameter()
...
+
+  double cumulative_value;
+  for (double i_probability : d_probabilities)
+    {
+      cumulative_value += i_probability;
+      cumulative_probability_vector.push_back(cumulative_value);
+    }
```
hepap, line 85: I would use a typical `cummulative_probability_vector[i] = cummulative_value` instead of push_back since you know the size already.

doc/source/parameters/dem/lagrangian_physical_properties.rst
```diff
@@ -21,56 +21,38 @@ In this subsection, gravitational acceleration, and the physical properties of t
...
-      # Standard deviation
-      set standard deviation                = 0.0
+      # If distribution type = uniform or normal
+      set diameter                          = 0.001
 
-      # Number of particles in type 0
-      set number of particles               = 132300
+      # If distribution type = histogram
+      set histogram diameters list          = 0.001 , 0.0005
+      set histogram probabilities list      = 0.6   , 0.4
```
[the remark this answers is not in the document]
OGaboriault: Good point! 

I talked about it with Paul, and we think ``custom`` would be a better name.

source/core/parameters_lagrangian.cc
OGaboriault: But like you said previously, it may not be equal to 1.0 because of the machine accuracy.

OGaboriault: The other option would be with the random number generation. Currently I'm generating numbers between 0 and 1. 

I could generate number between z0. and the sum of all the probabilities.

source/dem/distributions.cc
OGaboriault: Allocating memory once and not every time I push_back the vector.

source/dem/distributions.cc
OGaboriault: Ok, I understand what you are saying. Both are equivalent in this case (Pretty sure... we could ask)

Memory is allocated once with my implementation. The only difference (with reserve) is that memory is allocated but the size of the vector is still zero, so the vector can grow without re-allocating memory (what we don't want) 

If you do : 
```
std::vector<double> cumulative_probability_vector(d_probabilities.size() );  
// Here it's a vector of size d_probabilities.size() filled with zeros.
  double cumulative_value;
  for (unsigned int i = 0; i < d_probabilities.size() ; i++ )
    {
      cumulative_value += d_probabilities[i];
      cumulative_probability_vector[i] =  cumulative_value ;
    }
```
Memory is allocated once, but you change the values in the vector

hepap: Ok I see! Thanks for the explaination :)

source/core/parameters_lagrangian.cc
hepap: In that case, does that mean you are allowing the user to provide a random PSD? For example :

dia  prob
20   10
35   50
57   7

source/core/parameters_lagrangian.cc
hepap: Even so, I think it is a good pratrice to avoid double comparison :)

source/core/parameters_lagrangian.cc
OGaboriault: Both would be equivalent statistically and yes the example you are giving me would be valid.

OGaboriault:
```
if (std::abs(probability_sum - 1.0) <1e-14)
        {
          for (double &i : particle_custom_probability.at(particle_type))
            {
              i = i / probability_sum;
            }
        }
```

OGaboriault: if (std::abs(probability_sum - 1.0) < 1.e-14)
        {
          throw(std::runtime_error(
            "Invalid custom probabilities. The sum of probabilities should be equal to 1.0 "));
        }

source/dem/distributions.cc
OGaboriault: I spoke with Bruno Bless and both are equivalent. :)

include/dem/distributions.h
```diff
@@ -55,77 +59,143 @@ class NormalDistribution : public Distribution
 public:
   /**
    * @brief The constructor stores the parameters necessary to define the normal
-   * distribution
+   * distribution.
    *
-   * @param d_averages Average diameters for each type of particle.
-   * @param d_standard_deviations Standard deviation of the diameter for each type of particle.
+   * @param d_average Average diameters for each type of particle.
+   * @param d_standard_deviation Standard deviation of the diameter for each type of particle.
```
blaisb, line 65: Why do you say for each type when what you pass is a double?

include/dem/distributions.h
```diff
@@ -55,77 +59,143 @@ class NormalDistribution : public Distribution
...
+
+private:
+  /**
+   * Vector containing all the diameters values.
+   */
+  const std::vector<double> diameter_custom_values;
+
+  /**
+   * Vector containing cummulative probabilities associated with de
+   * diameter_custom_values vector. The probabilities are based on the volume,
```
hepap, line 195:
```suggestion
   * diameter_custom_values vector. The probabilities are based on the volume fraction,
```

source/core/parameters_lagrangian.cc
```diff
@@ -64,6 +73,21 @@ namespace Parameters
...
+      particle_custom_diameter.at(particle_type) =
+        entry_string_to_vector(prm, "custom diameters");
+      particle_custom_probability.at(particle_type) =
+        entry_string_to_vector(prm, "custom probabilities");
+
+      double probability_sum =
+        std::reduce(particle_custom_probability.at(particle_type).begin(),
+                    particle_custom_probability.at(particle_type).end());
+
+      // We make sure that the cumulative probability is equal to 1.
```
hepap, line 85:
```suggestion
      // We make sure that the cummulative probability is equal to 1.
```

hepap (approved):
Only small things in the comments :)

doc/source/parameters/dem/lagrangian_physical_properties.rst
```diff
@@ -83,15 +65,20 @@ In this subsection, gravitational acceleration, and the physical properties of t
...
+    In the ``custom`` size distribution, the probability values are based on the volume taken by all the particles of the associated diameter, not to the total number of particles. For example, if a probability is equal to ``0.5`` , this means that haft of the total volume of inserted particles will be occupied by particle with the associated diameter value.
 
 * The ``diameter`` parameter defines the diameter of the particles in a ``uniform`` distribution. In the case of a ``normal`` distribution, this parameter indicates the average diameter.
 
 * For a ``normal`` distribution, the ``standard deviation`` parameter should be defined to indicate the standard deviation on the particle size distribution.
 
+* For a ``custom`` distribution, the ``custom diameters`` defines the different diameter values used when generating particles. The ``custom probabilities`` defines the probabilities corresponding to each diameter value previously declared based on volume, not on the number of particles. Both list must have the same length.
```
AmishgaAlphonius, line 80:
```suggestion
* For a ``custom`` distribution, the ``custom diameters`` parameter defines the different diameter values used when generating particles. The ``custom probabilities`` parameter defines the probabilities corresponding to each diameter value previously declared based on volume fraction, not on the number of particles. Both list must have the same length.
```

include/dem/distributions.h
```diff
@@ -31,20 +31,24 @@ class Distribution
   /**
    * @brief Carries out the size sampling of particles. This is the base class of
    * NormalDistribution, UniformDistribution, LogNormalDistribution and
```
AmishgaAlphonius, line 33: Is the "LogNormalDistribution" class actually implemented?

```suggestion
   * NormalDistribution, UniformDistribution and CustomDistribution classes.
```

include/dem/distributions.h
```diff
@@ -31,20 +31,24 @@ class Distribution
...
-   * HistogramDistribution classes.
+   * CustomDistribution classes.
```
AmishgaAlphonius, line 34:
```suggestion
```

include/dem/distributions.h
```diff
@@ -55,77 +59,143 @@ class NormalDistribution : public Distribution
...
+  find_max_diameter() override;
+
+private:
+  /**
+   * Vector containing all the diameters values.
+   */
+  const std::vector<double> diameter_custom_values;
+
+  /**
+   * Vector containing cummulative probabilities associated with de
```
AmishgaAlphonius, line 194:
```suggestion
   * Vector containing cumulative probabilities associated with de
```

AmishgaAlphonius (approved):
A few minor last comments :)

blaisb:
following @hepap comment i think it would be good to add a unit test and not an application test for this. It would help making this reproducible

include/dem/distributions.h
OGaboriault: Good catch. It was, but I have decided to implement custom first then log-normal, otherwise this pull request would have been to big

doc/source/parameters/dem/lagrangian_physical_properties.rst
OGaboriault: I'll use Amishga version.

include/dem/distributions.h
OGaboriault: You mean this? 

```suggestion
   NormalDistribution, UniformDistribution  and CustomDistribution classes.
```

OGaboriault: I should remove LogNormalDistribution, not CustomDistribution...

OGaboriault:
Rebase had been done. I will add the seed attribute to the Normal and the Custom distributions. Then, I'll create a test for each distribution.

PR #956, opened 2023-12-07
Fixed the parsing of manifolds, prevent unset ones

blaisb:
# Description of the problem

- There was an issue with the latest deal.II release when you have a triangulation which has a manifold id but no manifold is set on this id and an assertion was thrown

# Description of the solution

- This made me realize we were being kinda stupid with the way we were setting the manifolds. I fixed this. When we have a GMSH triangulation, we only set the manifolds ids on the places we want to put a manifold. In the end this is actually cleaner.

# How Has This Been Tested?

- All tests now pass correctly in debug and release on my machine

### DIFF
--- a/source/core/grids.cc
+++ b/source/core/grids.cc
@@ -58,16 +58,6 @@ attach_grid_to_triangulation(
           grid_in.attach_triangulation(triangulation);
           std::ifstream input_file(mesh_parameters.file_name);
           grid_in.read_msh(input_file);
-
-          // Give a manifold id to the faces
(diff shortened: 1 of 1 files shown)


lpsaavedra (approved):

oguevremont:
I was looking at how manifolds are defined in the parameter files, but could not find how from the documentation. Is it possible that this page doesn't exist yet?

blaisb:
Maybe.... actually maybe it does not.  If so, can you open an issue?

PR #964, opened 2023-12-11
Checkpointing the insertion object.

OGaboriault:
# Description of the problem

- The number of inserted particle wasn't being check-pointed. The "number of particle" parameter had to be adjusted in the prm file when restarting a simulation which is annoying especially if particle cross an output boundary condition before the restart and there's still particle left to insert in the simulation. 

# Description of the solution

- The attributes "particles_of_each_type_remaining" and "current_inserting_particle_type" are being check-pointed with a virtual void function. 

# How Has This Been Tested?

- Current tests have been updated in a way that the total number of particle at the end of the test says the same. In brief, the parameter "number of particle" describe the number of particle at the end of the test, which wasn't the case previously. 

# Documentation
- N/A

# Comments

- The circle_restart test output file had to be updated... it's hard to explain, but I think that the insertion method algorithm change after that the creation of the old restart file. This means that we are not able the generated the exact old initial condition of the test. I had to generated the ".insertion_object" restart file, thus running the simulation from t=0 to the checkpoint time step, but since the initial condition is not the same, the restart files and the output file are not the same.


OGaboriault:
I think I'll need to run the CFD-DEM examples since they are using the restart functionality...

CHANGELOG.md
```diff
@@ -3,10 +3,15 @@ All notable changes to the Lethe project will be documented in this file.
  
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+
 ## [Master] - 2023-12-11
 
 ### Fixed
 
+- MINOR  The number of remaining particle to insert of each tipe is being checkpointed adequately. This means that no modification are required to the "number of particle" parameter after restarting a simulation. [#964](https://github.com/lethe-cfd/lethe/pull/964)
```
blaisb, line 11:
```suggestion
- MINOR  The number of remaining particle to insert of each type is being checkpointed adequately. This means that no modification are required to the "number of particle" parameter after restarting a simulation. [#964](https://github.com/lethe-cfd/lethe/pull/964) 
```

include/dem/insertion.h
```diff
@@ -81,6 +84,12 @@ class Insertion
          const parallel::distributed::Triangulation<dim> &triangulation,
          const DEMSolverParameters<dim>                  &dem_parameters) = 0;
 
+  virtual void
+  serialize(boost::archive::text_oarchive &ar, const unsigned int) = 0;
+
+  virtual void
+  deserialize(boost::archive::text_iarchive &ar, const unsigned int) = 0;
+
```
blaisb, line 92: Documentation of the functions

include/dem/read_checkpoint.h
```diff
@@ -41,24 +42,24 @@ using namespace std;
...
 read_checkpoint(
   TimerOutput                                             &computing_timer,
-  const DEMSolverParameters<dim>                          &dem_parameters,
+  const DEMSolverParameters<dim>                          &parameters,
   std::shared_ptr<SimulationControl>                      &simulation_control,
   PVDHandler                                              &particles_pvdhandler,
   PVDHandler                                              &grid_pvdhandler,
   parallel::distributed::Triangulation<dim>               &triangulation,
   Particles::ParticleHandler<dim>                         &particle_handler,
+  std::shared_ptr<Insertion<dim>>                         &insertion_object,
```
blaisb, line 62: Documentation needs to be updated

blaisb (changes requested):
Some small comments and doc changes to address

source/dem/dem.cc
```diff
@@ -1115,6 +1115,17 @@ DEMSolver<dim>::solve()
...
+  // Store information about floating mesh/background mesh intersection
+  for (unsigned int i_solid = 0; i_solid < solids.size(); ++i_solid)
+    {
+      floating_mesh_info[i_solid] =
+        solids[i_solid]->map_solid_in_background_triangulation(triangulation);
+    }
+
+  // Set insertion object type before the restart. We need to do so, otherwise
+  // we don't know which kind of insertion in beging used.
+  insertion_object = set_insertion_type(parameters);
```
blaisb, line 1127:
```suggestion
  // Set insertion object type before the restart because the restart only rebuilds the member of the insertion object
  insertion_object = set_insertion_type(parameters);
```

blaisb (approved):
One last comment and this is ready for merge

PR #968, opened 2023-12-13
Add lpbf example

hepap:
# Description of the problem

- Few things were missing for the lpbf example. 

# Description of the solution

- Add a parameter for the universal gas constant to allow other units than the SI system.
- Add limiter on the expansion pressure term in the momentum flux to avoid division by 0.
- Add draft of the .prm of the upcoming example

# Future changes

- Documentation for evaporation
- LPBF example
- Thermocapillary droplet example


blaisb (approved):

blaisb:
Will merge after CI @hepap

PR #976, opened 2023-12-31
Documentation errors and small change to the custom distribution

OGaboriault:
Smalls changes to the custom distribution to prevent an edge case where the random number generated is equal to or bigger than the highest value in the cumulative probability vector. 

I also find errors in the documentation. In a previous Pull request, I forgot to change the old parameter names in the text below the code example.

Bonne année!


blaisb (approved):

blaisb:
Ci is crashing for something that's not your fault. I will merge this :)

PR #977, opened 2024-01-03
Darcy-like penalization for phase change

blaisb:
# Description of the problem

- Currently we only supported viscous penalization for phase change problem. This turns out to not be ideal to impose stasis and it is sometimes desirable to use a Darcy-like penality term
- The melting cavity example needed to be adapted to consider both approaches

# Description of the solution

- Add a darcy penalization which uses the phase change properties. This assembler currently only exists in single phase (it does not support VOF, this will be a next PR), but this is protected by assertions. I hesitated between this and making it a physical property, but tihs is really not a physical property per say and I would rather have a standalone assembler that takes the information from the physical properties themselves instead. I believe this is an adequate and safe compromise.
- Modify the melting cavity example to have an additional parameter file to launch a Darcy penalization simulation and compare the results in a new subsection
- Add an application tests which is a copy of the viscous phase change one, but that uses instead the Darcy penality

# How Has This Been Tested?

- A new application test has been added
- The melting cavity example can now also be run with this approach

# Documentation

- Everything is now fully documented in the VelocitySource documentation

# Future changes

- Extend to VOF, this will be an easy PR once this is merged.

# Comments

- This is a weird scenario where it is a real question we could ask ourselves whether this is or not a physical property. I would rather not consider it to be one, but that remains debatable I guess.


examples/multiphysics/melting-cavity/compare-melting-cavity.py
```diff
@@ -0,0 +1,81 @@
...
+temperature_limit = 29.75
+
+# List of name for the cases used in the comparison
+case_list=["Lethe - Viscous penality", "Lethe - Darcy penality", "Lethe - Hybrid penality"]
+
+
+
+
+
+
```
AmishgaAlphonius, line 51:
```suggestion
```

AmishgaAlphonius (commented):
Nice PR ! :)
I have a few comments and suggestions here and there and a more general one: 
I think it is more appropriate to use the term `penalty` rather than `penality`.

doc/source/parameters/cfd/physical_properties.rst
```diff
@@ -525,10 +525,10 @@ This model is parameterized using the following section:
     set solidus temperature  = 0
   
     # Thermal expansion of the liquid phase
-    set thermal expansion liquid = 0
+    set thermal expansion liquid = 1
   
     # Thermal expansion of the solid phase
-    set thermal expansion solid  = 1
+    set thermal expansion solid  = 0
```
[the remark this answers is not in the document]
blaisb: It should be set to 0. that's a mistake then

doc/source/parameters/cfd/velocity_source.rst
```diff
@@ -22,3 +26,17 @@ The default values are:
...
+
+A Darcy-like source term can be added to the simulation using the following parmeters:
+
+.. code-block:: text
+
+  subsection velocity source
+    set Darcy type          = none
+  end
+
+* The ``Darcy type`` parameter specifies the type of Darcy penalization term to be applied to the Navier-Stokes equations. The options are ``none`` or ``phase_change``. The ``phase_change`` model uses the values of the ``Darcy penality liquid``  and ``Darcy penality solid`` set-up within the ``phase change`` subsection of the :doc:`physical_properties`.
```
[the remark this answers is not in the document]
blaisb: The issue is that Phase Change is not its own section. Maybe for now we can keep it this way and we could eventually have a whole different section for phase change. I think we can leave it like that for now and when the feature is more mature we will know how to documentit the best way :)

include/core/parameters.h
```diff
@@ -218,6 +218,12 @@ namespace Parameters
     // kinematic viscosity solid - Units in in m^2/(s)
     double kinematic_viscosity_s;
 
+    // Darcy permeability liquid - Units in 1/(s)
+    double penality_l;
+
+    // Darcy permeabilitysolid - Units in 1/(s)
+    double penality_s;
```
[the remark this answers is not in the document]
blaisb: The comments will need to be update here.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+  // Loop over the quadrature points
+  for (unsigned int q = 0; q < n_q_points; ++q)
+    {
+      // Store JxW in local variable for faster access;
+      const double JxW = JxW_vec[q];
+      // Current temperature values
+      double current_temperature = temperatures[q];
+      // Depending on the value of the temperature, calculate the Darcy
+      // permeability This min(max) approach is used to calculate the liquid
+      // fraction without requiring if conditions
```
hepap, line 1758:
```suggestion
      // permeability. This min(max) approach is used to calculate the liquid
      // fraction without requiring if conditions.
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+                 std::max((current_temperature -
+                           phase_change_parameters.T_solidus) /
+                            (phase_change_parameters.T_liquidus -
+                             phase_change_parameters.T_solidus),
+                          0.));
+      double darcy_penality =
+        phase_change_parameters.penality_l * liquid_fraction +
+        (1. - liquid_fraction) * phase_change_parameters.penality_s;
+
+      strong_residual[q] += darcy_penality * velocities[q];
```
hepap, line 1770:
```suggestion
      double darcy_penalty =
        phase_change_parameters.penalty_l * liquid_fraction +
        (1. - liquid_fraction) * phase_change_parameters.penalty_s;

      strong_residual[q] += darcy_penalty * velocities[q];
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+
+      for (unsigned int j = 0; j < n_dofs; ++j)
+        {
+          strong_jacobian[q][j] += darcy_penality * scratch_data.phi_u[q][j];
```
hepap, line 1774:
```suggestion
          strong_jacobian[q][j] += darcy_penalty * scratch_data.phi_u[q][j];
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+  // Loop over the quadrature points
+  for (unsigned int q = 0; q < n_q_points; ++q)
+    {
+      // Store JxW in local variable for faster access;
+      const double JxW = JxW_vec[q];
+      // Current temperature values
+      double current_temperature = temperatures[q];
+      // Depending on the value of the temperature, calculate the Darcy
+      // permeability This min(max) approach is used to calculate the liquid
+      // fraction without requiring if conditions
```
hepap, line 1815:
```suggestion
      // Depending on the value of the temperature, calculate the Darcy
      // permeability. This min(max) approach is used to calculate the liquid
      // fraction without requiring if conditions.
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+                 std::max((current_temperature -
+                           phase_change_parameters.T_solidus) /
+                            (phase_change_parameters.T_liquidus -
+                             phase_change_parameters.T_solidus),
+                          0.));
+      double darcy_penality =
+        phase_change_parameters.penality_l * liquid_fraction +
+        (1. - liquid_fraction) * phase_change_parameters.penality_s;
+
+      strong_residual[q] += darcy_penality * velocities[q];
```
hepap, line 1827:
```suggestion
      double darcy_penalty =
        phase_change_parameters.penalty_l * liquid_fraction +
        (1. - liquid_fraction) * phase_change_parameters.penalty_s;

      strong_residual[q] += darcy_penalty * velocities[q];
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -1726,6 +1728,119 @@ BuoyancyAssembly<dim>::assemble_rhs(
...
+
+      // Assembly of the right-hand side
+      for (unsigned int i = 0; i < n_dofs; ++i)
+        {
+          const auto phi_u_i = scratch_data.phi_u[q][i];
+
+          // Laplacian on the velocity terms
+          local_rhs(i) -= darcy_penality * velocities[q] * phi_u_i * JxW;
```
hepap, line 1835:
```suggestion
          local_rhs(i) -= darcy_penalty * velocities[q] * phi_u_i * JxW;
```

hepap (commented):
Nice addition :)

As you said in the Description of the problem, I also don't think it should be a physical properties. I don't think that the penalty value for each phase should be either, because it's more a model parameter for me (it doesn't have a real physical value I think).

doc/source/examples/multiphysics/melting-cavity/melting-cavity.rst
```diff
@@ -249,6 +252,86 @@ contains the simulation results. In post-processing, the position of the solid-l
     :width: 500
 
 
+
+--------------------------------------------
+Darcy Penalization: An Alternative Approach
+--------------------------------------------
+
+Lethe supports an alternative strategy to impose statis (no motion) within the solidified material using a Darcy-like penalization. This penalization adds a forcing term to the momentum equation to prohibit the motion of the solid instead of increasing its viscosity. This has the advantage of leading to a better matrix conditioning, at the expense of potentially increased motion within the solid phase. To enable this forcing term, a velocity source term must be specified:
```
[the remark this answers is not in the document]
blaisb: Yes, but penalizing the value is less stringent than penalizing the gradient of the value. One is in L2 and the other one is in H1.
If you look at the number of iterative solver iterations, you can really see the difference.

doc/source/examples/multiphysics/melting-cavity/melting-cavity.rst
```diff
@@ -249,6 +252,86 @@ contains the simulation results. In post-processing, the position of the solid-l
...
+          set viscosity solid = 0.0007366698558439125
+    
+          # Thermal expansion of the liquid phase
+          set thermal expansion liquid       = 1
+    
+          # Thermal expansion of the solid phase
+          set thermal expansion solid        = 0
+    
+          # Permeability of the liquid phase
+          set Darcy penality liquid         = 0
```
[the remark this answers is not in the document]
blaisb: Well right now the penality model is phase_change. Nothing prevents us from implementing another Darcy penality.
However, in this current implementation, at least I can make it general to have "two penalties" kinda. 
I mean, I can remove the penalty liquid if you want, but the Darcy model will still require the Phase Change parameter to have the liquidus and the solidus temperature, so I kinda like the idea of putting it there, even if by default we never use the penalty in the liquid phase, the implementation will remain generic.

blaisb:
I will address the penalty parameter tonight and integrate that. Other than that @AmishgaAlphonius and @hepap are you ok with the current structure? I'd like to merge that quickly so we can work on the VOF implementation. Will most likely finish it tonight.

hepap:
Good for me @blaisb !

PR #979, opened 2024-01-04
Add 3D motion of non spherical objects

Luckabarbeau:
# Description of the problem

- This PR aims to add support for the coupled angular motion of any 3D object. This includes the coupling with the fluid and the DEM interaction between these objects.  The following elements are added: integration of the Euler angles in the DEM and in the CFD coupling, relevant parameters for the inertia and volume of the objects, A function to evaluate the distance between the two objects using the intersection of their level-set and their bounding box, a set of three boolean parameters that aimed at limiting the number of call to the DEM calculation.
- While implementing these features, two bugs were found: A bug that affected the evaluation of the force when two particles were in contact. This bug was changing the force evaluation on a particle depending on the number of processors that were used. This bug comes from overconstrained cells (cells with all their DOF imposed by boundary conditions were poorly treated in the force evaluation). This modification changed the results of a few tests where contact played a role. However, the solution's behavior does not change significantly, and new results still agree well with previous results for cases with particle contact. The modification to the function generate_cut_cell_map and force_evaluation only addresses this bug.
Another bug for the evaluation of the gradient of superquadric was found. The gradient had a sign error if evaluated inside of the particle. The evaluation of the supequadric was also optimized by avoiding overshoot.
- The new inertia required a modification of the load particle from the file option. To add the possibility of defining the inertia matrix in the load particle from the file option, we modified the syntax to be semicolon-separated instead of space-separated and added the volume variable.
# Description of the solution

- Add support for integrating the Euler angle in the coupling formulation and in the DEM. Move all the DEM calculations for contact force in the ib_particle_dem to take the contact point into account correctly and simplify the code.
- Add the new function in the shape classes called distance_to_shape. This function looks for the minimum of the level set defined by the intersection of two shapes. Using this minimum, we can evaluate the overlap of the two shapes. This function combines a Cartesian search with a gradient direction search and a numerical gradient descent.

# How Has This Been Tested?

- A new test was added that modeled two ellipsoids in contact with each other and a wall.
- All previous tests were verified, and an investigation was performed when the results were modified (e.g., pp_contact_test).
- Previous examples were tested and run successfully. 
- A modified example was performed to test multiple superquadric shapes (64).

# Documentation

- The documentation was updated to reflect the new options and add some tips on using the new feature.

# Future changes

- This feature version remains experimental, and more tests and validation should be performed. An example with a validation case should be added.


blaisb:
You'll need to rebase, No hurry though

include/core/parameters.h
```diff
@@ -1318,6 +1318,9 @@ namespace Parameters
     bool                         load_particles_from_file;
     std::string                  particles_file;
     bool                         enable_extra_sharp_interface_vtu_output_field;
+    bool                         explicit_contact_impulsion_calculation;
+    bool                         explicit_position_integration_calculation;
+    bool                         use_approximate_radius_for_contact;
```
blaisb, line 1323: Can you add comments as to what these parameters mean?

include/core/physics_solver.h
```diff
@@ -116,6 +116,11 @@ class PhysicsSolver
     auto &system_rhs = get_system_rhs();
     return system_rhs.l2_norm();
   }
+  virtual unsigned int
+  get_current_newton_iteration()
+  {
+    return non_linear_solver->get_current_newton_iteration();
+  }
```
blaisb, line 123: Does this need to be virtual?
Also needs a brief
Could also be inlined :)

blaisb (changes requested):
First batch of comments, will continue most likely early next week :)

include/core/parameters.h
```diff
@@ -1291,41 +1291,87 @@ namespace Parameters
...
     void
     parse_parameters(ParameterHandler &prm);
-
-    unsigned int                 nb;
-    unsigned int                 order;
-    unsigned int                 initial_refinement;
-    double                       inside_radius;
-    double                       outside_radius;
-    bool                         time_extrapolation_of_refinement_zone;
+    // Number of ib particles declared
```
oguevremont, line 1294: It could improve readability to make parameter ''paragraphs'' and add line jumps between categories.

include/core/shape.h
```diff
@@ -133,6 +134,952 @@ class Shape : public AutoDerivativeFunction<dim>
...
+  /**
+   * @brief Return the distance, the center point, and the normal between the current shape and the shape given in the argument. The center point is the point where both shapes are at the same distance from each other. The normal is defined using the closest surface point on the two shapes. By default, the function does not calculate the distance between the two shapes if their bounding boxes are not in contact. This behavior can be modified using the appropriate parameter.
+   * @param shape The shape with which the distance is evaluated
+   * @param cell The cell that is likely to contain the evaluation point
+   * @param candidate_points This is the initial guess points used in the calculation.
+   * @param precision This is the precision of the distance between the two shapes.
+   * @param exact_distance_outside_of_contact This is a boolean to force the exact distance evaluation if the shapes are not in contact.
+   */
+  virtual std::tuple<double, Tensor<1, dim>, Point<dim>>
+  distance_to_shape_with_cell_guess(
```
oguevremont, line 159: I could not understand the purpose of this function from only the name. Is there a clearer name that could be used? I understand that it does a lot, so finding a good name isn't easy.

source/core/shape.cc
```diff
@@ -366,6 +374,151 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+Plane<dim>::gradient(const Point<dim> &evaluation_point,
+                     const unsigned int /*component*/) const
+{
+  (void)evaluation_point;
+  //  We take the vector in z of the plane and rotate it in the world frame.
+  if constexpr (dim == 2)
+    {
+      return tensor_nd_to_2d(this->rotation_matrix * tensor_nd_to_3d(normal));
+    }
+
```
oguevremont, line 507:
```suggestion
```

source/core/shape.cc
```diff
@@ -366,6 +374,151 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+  Point<dim> projected_point =
+    current_point - dot_product / normal.norm_square() * normal;
+  auto rotate_in_globalpoint = this->reverse_align_and_center(projected_point);
+  if (dot_product > 0)
+    return (rotate_in_globalpoint - evaluation_point).norm();
+  else
+    return -(rotate_in_globalpoint - evaluation_point).norm();
+}
+
+
```
oguevremont, line 486:
```suggestion
```

source/core/shape.cc
```diff
@@ -366,6 +374,151 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+double
+Sphere<dim>::local_curvature_radius_with_cell_guess(
+  const Point<dim>                                     p,
+  const typename DoFHandler<dim>::active_cell_iterator cell)
+{
+  (void)cell;
+  return (this->position - p).norm();
+}
+
+
```
oguevremont, line 468:
```suggestion
```

source/core/shape.cc
```diff
@@ -124,6 +132,7 @@ Shape<dim>::align_and_center(const Point<dim> &evaluation_point) const
   return translated_point;
 }
 
+
```
oguevremont, line 135:
```suggestion
```

source/core/shape.cc
```diff
@@ -31,21 +31,28 @@
...
+  // function can be expensive if done frequently.
+  if (value_cache.size() > 0)
+    {
+      value_cache.clear();
+      gradient_cache.clear();
+      closest_point_cache.clear();
+    }
 }
 
+
```
oguevremont, line 55:
```suggestion
```

source/core/shape.cc
```diff
@@ -31,21 +31,28 @@
...
 double
-Shape<dim>::displaced_volume(const double /*fluid_density*/)
+Shape<dim>::displaced_volume()
 {
   StandardExceptions::ExcNotImplemented();
-  return 1.0;
+  return 0;
 }
 
+
```
oguevremont, line 40:
```suggestion
```

source/core/shape.cc
```diff
@@ -413,23 +566,27 @@ Superquadric<dim>::closest_surface_point(const Point<dim> &p,
...
       while (iteration < iteration_max && abs(current_distance) > epsilon)
         {
           distance_gradient = superquadric_gradient(current_point);
+          // limit the step size
+
           if (distance_gradient.norm() < epsilon)
             // Gradient can be null if the evaluation point is exactly on the
             // centroid of the shape. In this case it is also the closest point.
             break;
+          // This is a modify newton method that limit the step size when the
```
oguevremont, line 579:
```suggestion
          // This is a modified Newton method that limits the step size when the
```

source/core/shape.cc
```diff
@@ -2574,3 +2741,7 @@ template class RBFShape<2>;
 template class RBFShape<3>;
 template class OpenCascadeShape<2>;
 template class OpenCascadeShape<3>;
+template class Plane<2>;
+template class Plane<3>;
+// template class Shape<2>;
+// template class Shape<3>;
```
oguevremont, line 2747:
```suggestion
```

include/core/shape.h
```diff
@@ -327,18 +1415,23 @@ class Shape : public AutoDerivativeFunction<dim>
...
 
+
+  // The rotation matrix that describes the solid orientation.
+  Tensor<2, 3> rotation_matrix;
   // Layer thickening: used to artificially inflate/deflate the shape
   double layer_thickening;
 };
 
 
+
```
oguevremont, line 1434:
```suggestion
```

oguevremont (commented):
Nice PR! The comments I have so far. I still have some files to review

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -202,7 +212,11 @@ This subsection contains the parameters related to the sharp immersed boundary s
...
+        .. code-block:: text
+        
+            type; shape_argument; p_x; p_y; p_z; v_x; v_y; v_z; omega_x; omega_y; omega_z; orientation_x; orientation_y; orientation_z; volume ;density; inertia; pressure_x; pressure_y; pressure_z; youngs_modulus; restitution_coefficient; friction_coefficient; poisson_ratio; rolling_friction_coefficient; integrate_motion;
+            superquadric; 1: 1: 1: 3: 3: 3; 0.25; 0.25; 20.25; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.001953125; 0.0015; 7.6698974609375e-08; 0.0; 0.0; 0.0; 1000000.0; 0.9; 0.0; 0.3; 0.0; true
```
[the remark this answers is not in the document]
Luckabarbeau: Yeah, and it follows a more common pattern.

include/core/shape.h
```diff
@@ -133,6 +134,952 @@ class Shape : public AutoDerivativeFunction<dim>
...
+  double
+  smooth_max(double a, double b)
+  {
+    double smooth_factor = 10;
+    return (a * std::exp(a * smooth_factor) + b * std::exp(b * smooth_factor)) /
+           ((std::exp(a * smooth_factor) + std::exp(b * smooth_factor)));
+  }
+
+  /**
+   * @brief Return the distance, the center point, and the normal between the current shape and the shape given in the argument. The center point is the point where both shapes are at the same distance from each other. The normal is defined using the closest surface point on the two shapes. By default, the function does not calculate the distance between the two shapes if their bounding boxes are not in contact. This behavior can be modified using the appropriate parameter.
```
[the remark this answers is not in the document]
Luckabarbeau: I prefer to give a long brief than not since this function is relatively complex.

include/core/shape.h
Luckabarbeau: Well, it is named like this as its primary function is to return the distance of the current shape with the shape given in the argument.

include/core/shape.h
```diff
@@ -1199,6 +2622,17 @@ class OpenCascadeShape : public Shape<dim>
...
+    this->bounding_box_half_length[0] = 0;
+    this->bounding_box_half_length[1] = 0;
+    this->bounding_box_center[0]      = 0;
+    this->bounding_box_center[1]      = 0;
+    if constexpr (dim == 3)
+      {
+        this->bounding_box_half_length[2] = 0;
+        this->bounding_box_center[2]      = 0;
+      }
+
```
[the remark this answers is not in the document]
Luckabarbeau: True 0 was used for the plane as it simplifies some calculations, but fir this case, DBL_max are better.

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -164,8 +168,14 @@ This subsection contains the parameters related to the sharp immersed boundary s
 
     .. warning::
 	    If ``contact search radius factor`` :math:`\leq 1`, an error is thrown.
+	    
+	* The ``enable lubrication force`` parameter enables or disables the use of lubrication forces. This parameter must be set to ``false`` when using non-newtonian fluid.
     
-    * The ``enable lubrication force`` parameter enables or disables the use of lubrication forces. This parameter must be set to ``false`` when using non-newtonian fluid.
+    * The ``use explicit contact impulsion`` parameter enables or disables the use of explicit contact impulsion evaluation in the resolution of the coupling of the particle. When it is set to true, this parameter results in the code only performing the DEM calculation once per CFD time step.
```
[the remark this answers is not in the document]
Luckabarbeau: Yes exactly

include/core/parameters.h
Luckabarbeau: I split them into group that follow the same structure as in the prm file

include/fem-dem/ib_particles_dem.h
```diff
@@ -145,6 +160,46 @@ class IBParticlesDEM
...
+                        Point<dim>   &particle_two_position,
+                        Tensor<1, 3> &particle_two_velocity,
+                        Tensor<1, 3> &particle_two_omega,
+                        const double  particle_two_mass,
+                        const double  particle_two_radius,
+                        const double  particle_two_youngs_modulus,
+                        const double  particle_two_poisson_ratio,
+                        const double  particle_two_restitution_coefficient,
+                        const double  particle_two_friction_coefficient,
+                        const double  particle_two_rolling_friction_coefficient,
```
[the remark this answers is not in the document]
Luckabarbeau: I simplify the comment and put a general comment on the function's input parameter.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +801,328 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
-              best_indices[boundary_cell_iter.boundary_index].value =
-                boundary_index;
-              best_cells[boundary_cell_iter.boundary_index] =
-                boundary_cell_iter;
+              // Find the best candidate (the closest point) for each different
+              // wall.
+
+              // Find a way to use cell guess
+              // correctly!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Do not merge without
+              // final check on this
```
oguevremont, line 823: Don't forget

source/fem-dem/ib_particles_dem.cc
```diff
@@ -890,6 +1252,38 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+            }
+
+          /*if (Utilities::MPI::this_mpi_process(this->mpi_communicator) == 0)
+            {
+              std::cout<<"time "<< t <<std::endl;
+              std::cout<<"particle velocity "<<
+            dem_particles[0].velocity<<std::endl; std::cout<<"particle force"<<
+            contact_wall_force[0]<<std::endl; std::cout<<"particle torque"<<
+            contact_wall_torque[0]<<std::endl;
+            }*/
```
oguevremont, line 1286:
```suggestion
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -967,11 +1338,79 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+                       angular_velocity_in_particle_frame[0] *
+                       angular_velocity_in_particle_frame[1]) /
+                    dem_particles[p_i].inertia[2][2];
+
+                  // Rotate angular acceleration in world frame
+                  k_omega[p_i][step] =
+                    invert(dem_particles[p_i].rotation_matrix) *
+                    angular_acceleration_in_particle_frame;
+
+                  // Store est rk4 step of the angular contact impulsion
```
oguevremont, line 1381: What is `est` here?

source/fem-dem/ib_particles_dem.cc
```diff
@@ -1050,73 +1489,253 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+              unsigned int particle_id = it->first;
+              for (auto it2 = it->second.begin(); it2 != it->second.end();
+                   it2++)
+                {
+                  unsigned int wall_id      = it2->first;
+                  auto         contact_info = it2->second;
+                  // Check if all 4 of the RK steps are in the contact info. If
+                  // it is not the case, this means this is a new contact, or
+                  // the contact ended. In this case, we simply take the last RK
+                  // step register as the derivative for the integration.
```
oguevremont, line 1506:
```suggestion
                  // step registered as the derivative for the integration.
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -1050,73 +1489,253 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+              unsigned int particle_one_id = it->first;
+              for (auto it2 = it->second.begin(); it2 != it->second.end();
+                   it2++)
+                {
+                  unsigned int particle_two_id = it2->first;
+                  auto         contact_info    = it2->second;
+                  // Check if all 4 of the RK steps are in the contact info. If
+                  // it is not the case, this means this is a new contact, or
+                  // the contact ended. In this case, we simply take the last RK
+                  // step register as the derivative for the integration.
```
oguevremont, line 1535:
```suggestion
                  // step registered as the derivative for the integration.
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -1050,73 +1489,253 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+        {
+          unsigned int particle_id = it->first;
+          for (auto it2 = it->second.begin(); it2 != it->second.end(); it2++)
+            {
+              unsigned int wall_id      = it2->first;
+              auto         contact_info = it2->second;
+              // Check if all 4 of the RK steps are in the contact info. If it
+              // is not the case, this means this is a new contact, or the
+              // contact ended. In this case, we simply take the last RK step
+              // register as the derivative for the integration.
```
oguevremont, line 1650:
```suggestion
              // registered as the derivative for the integration.
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -1050,73 +1489,253 @@ IBParticlesDEM<dim>::integrate_particles_motion(const double dt,
...
+        {
+          unsigned int particle_one_id = it->first;
+          for (auto it2 = it->second.begin(); it2 != it->second.end(); it2++)
+            {
+              unsigned int particle_two_id = it2->first;
+              auto         contact_info    = it2->second;
+              // Check if all 4 of the RK steps are in the contact info. If it
+              // is not the case, this means this is a new contact, or the
+              // contact ended. In this case, we simply take the last RK step
+              // register as the derivative for the integration.
```
oguevremont, line 1697:
```suggestion
              // registered as the derivative for the integration.
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -90,6 +90,13 @@ template <int dim>
 void
 GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
 {
+  // Potential amelioration of this function. This function could try to use the
+  // information of the previous cut cell mapping as an initial guess of the new
+  // cell mapping by looking at the parent level of the cells to look for new
+  // cells cut.
```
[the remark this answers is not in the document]
Luckabarbeau: Yeah, but it is not easy to implement and make it robust.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -785,9 +816,11 @@ GLSSharpNavierStokesSolver<dim>::define_particles()
...
           if (particles[i].integrate_motion == true)
             {
-              if (typeid(*particles[i].shape) != typeid(Sphere<dim>))
-                throw std::runtime_error(
-                  "Shapes other than sphere cannot have their motion integrated through fluid-structure interaction");
+              /* if (typeid(*particles[i].shape) != typeid(Sphere<dim>))
+                 throw std::runtime_error(
+                   "Shapes other than sphere cannot have their motion integrated
+                 through fluid-structure interaction");
+                   */
```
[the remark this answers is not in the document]
Luckabarbeau: no, it is removed

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -769,9 +799,10 @@ GLSSharpNavierStokesSolver<dim>::define_particles()
             this->simulation_parameters.particlesParameters->particles[i];
           if (particles[i].integrate_motion == true)
             {
-              if (typeid(*particles[i].shape) != typeid(Sphere<dim>))
+              /*if (typeid(*particles[i].shape) != typeid(Sphere<dim>))
                 throw std::runtime_error(
-                  "Shapes other than sphere cannot have their motion integrated through fluid-structure interaction");
+                  "Shapes other than sphere cannot have their motion integrated
+                through fluid-structure interaction");*/
```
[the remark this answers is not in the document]
Luckabarbeau: No this is removed

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -4401,120 +4749,117 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
...
+            particles_data["friction_coefficient"][p_i]);
+          particles[p_i].poisson_ratio =
+            Utilities::string_to_double(particles_data["poisson_ratio"][p_i]);
           particles[p_i].rolling_friction_coefficient =
-            restart_data["rolling_friction_coefficient"][p_i];
+            Utilities::string_to_double(
+              particles_data["rolling_friction_coefficient"][p_i]);
           particles[p_i].initialize_previous_solution();
-          if (restart_data["integrate_motion"][p_i] == 0.0)
+          if (particles_data["integrate_motion"][p_i] == "false")
```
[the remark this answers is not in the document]
Luckabarbeau: Yeah now we can load composite as well

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -267,6 +281,9 @@ Here is the content of a file that defines a cylinder topped with a sphere:
...
 
+    .. warning::
+        Even though non-spherical particles can now have their dynamic coupled with the fluid, this feature is not yet fully validated and remains experimental at this point. We note the following limitation: Particles can only have one point of contact between them (this means nonconvex shape contact may be wrong at this point since these shapes can have more than 1 point of contact), Fluid entrapment between particles can happen more frequently for non-spherical shapes in 3D (Fluid entrapment occurs when a portion of the fluid domain become completely isolated from the rest of the fluid domain due to the imposition of the immersed boundary by multiple particles. A simple example of a case that causes fluid entrapment would be three circles in contact in 2D. Fluid entrapment leads to a zone without reference pressure, which is not a well-posed problem). In this case, the linear solver may fail to converge for a given Newton iteration.
```
OGaboriault, line 285:
```suggestion
        Even though non-spherical particles can now have their dynamic coupled with the fluid, this feature is not yet fully validated and remains experimental at this point. We note the following limitation: Particles can only have one point of contact between each other and with each wall. (this means contact detection for concave shapes may be wrong since these shapes can have more than 1 point of contact), fluid entrapment between particles can happen more frequently for non-spherical shapes in 3D (fluid entrapment occurs when a portion of the fluid domain become completely isolated from the rest of the fluid domain due to the imposition of the immersed boundary by multiple particles. A simple example of a case that causes fluid entrapment would be three circles in contact in 2D. Fluid entrapment leads to a zone without reference pressure, which is not a well-posed problem). In this case, the linear solver may fail to converge for a given Newton iteration.
```

OGaboriault, line 285: I think this would be cleaner if you would list each limitation, but I'm not sure if this is possible in a warning...

include/core/ib_particle.h
```diff
@@ -32,14 +32,13 @@
 
 using namespace dealii;
 
+
```
OGaboriault, line 35: `/**
  * @brief This class defines values related to a particle used in the sharp interface IB. These values are used in the
  * solver
  * 
  * @tparam dim  An integer that denotes the number of spatial dimensions
  */

include/core/ib_particle.h
```diff
@@ -32,14 +32,13 @@
...
 template <int dim>
 class IBParticle
 {
 public:
   /**
-   * @brief
-   * initialised the particle
-   *
+   * @brief This class defines values related to a particle used in the sharp interface IB. Each particle defined will have these value used in the solver.
```
OGaboriault, line 41:
```
```

include/core/ib_particle.h
OGaboriault, line 35:
```suggestion
/**
  * @brief This class defines values related to a particle used in the sharp interface IB. These values are used in the
  * solver
  * 
  * @tparam dim  An integer that denotes the number of spatial dimensions
  */
```

OGaboriault (changes requested):
Still not finish, on last file to review

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    // impulsion in the CFD-DEM coupling.
+    bool explicit_contact_impulsion_calculation;
+    // Boolean that turn on or off the explicit evaluation of the position of
+    // the particle in the CFD-DEM coupling.
+    bool explicit_position_integration_calculation;
+    // Boolean that turns on or off the evaluation of contact radius. If it is
+    // true, the exact calculations are turned off, and the approximate radius
+    // based on the particle effective radius is used.
+    bool use_approximate_radius_for_contact;
+    // The function that is used to define the gravitational acceleration in the
```
voferreira, line 1366:
```suggestion
    // Function defining the gravitational acceleration vector used by the CFD-DEM
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    // CFD-DEM calculation.
+    std::shared_ptr<Functions::ParsedFunction<dim>> f_gravity;
+    // The wall young's modulus.
```
voferreira, line 1369:
```suggestion
    // Young's modulus of the wall
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    double wall_youngs_modulus;
+    // The wall poisson ratio.
```
voferreira, line 1371:
```suggestion
    // Poisson ratio of the wall
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    double wall_poisson_ratio;
+    // The wall rolling friction coefficient.
```
voferreira, line 1373:
```suggestion
    // Rolling friction coefficient of the wall
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    double wall_rolling_friction_coefficient;
+    // The wall friction coefficient.
```
voferreira, line 1375:
```suggestion
    // Sliding friction coefficient of the wall
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    double wall_friction_coefficient;
+    // The wall restitution coefficient
```
voferreira, line 1377:
```suggestion
    // Coefficient of restitution of the wall
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    double wall_restitution_coefficient;
+
+    // Boolean that dictates if the particle has been defined in a file instead
```
voferreira, line 1380:
```suggestion
    // Boolean for loading particles from an independent file. If true, the definition of the particles in the Particle subsection in the .prm file is ignored.
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    // of the particle subsection.
```
voferreira, line 1381:
```suggestion
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    bool load_particles_from_file;
+    // The file name used if the particles are loaded from a file.
```
voferreira, line 1383:
```suggestion
    // Name of the independent file containing particles' information at insertion. Only used if load_particles_from_file is true
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,97 @@ namespace Parameters
...
+    std::string particles_file;
```
voferreira, line 1384:
```suggestion
```

voferreira (commented):
First batch of suggestions.

CMakeLists.txt
```diff
@@ -147,4 +147,5 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
   endif()
 endif()
 
-ADD_SUBDIRECTORY(doc/doxygen)
\ No newline at end of file
+ADD_SUBDIRECTORY(doc/doxygen)
+
```
blaisb, line 151: To address by rebasing. Should not be modified here.

include/core/non_linear_solver.h
```diff
@@ -46,7 +43,9 @@ class NonLinearSolver
                   const Parameters::NonLinearSolver &params);
 
   virtual ~NonLinearSolver()
-  {}
+  { // Initialized the outer_iteration_counter by default to 0.
+    outer_iteration = 0;
+  }
```
blaisb, line 48: This should be put in the  pre-constructor and not the constructor
After the : statement and not within {}

include/core/non_linear_solver.h
```diff
@@ -57,9 +56,21 @@ class NonLinearSolver
   virtual void
   solve(const bool is_initial_step) = 0;
 
+
+  /**
+   * @brief Return the current newton iteration.
+   *
+   */
+  inline unsigned int
+  get_current_newton_iteration()
```
blaisb, line 65:
```suggestion
  get_current_newton_iteration() const
```

This is a const function. Let's mark it as such.

include/core/physics_solver.h
```diff
@@ -117,6 +117,15 @@ class PhysicsSolver
     return system_rhs.l2_norm();
   }
 
+  /**
+   * @brief Return the current newton iteration of this physics solver.
+   */
+  inline unsigned int
+  get_current_newton_iteration()
```
blaisb, line 124:
```suggestion
  get_current_newton_iteration() const
```

include/fem-dem/ib_particles_dem.h
```diff
@@ -145,6 +160,40 @@ class IBParticlesDEM
...
+                        const double  dt);
```
blaisb, line 194: Ok this is a mess to have a function with so many argument
Why not make a struct that contains all this information and just pass the struct? Otherwise it becomes a bit unreadable to know what is what. I would have maybe a struct per particle one. Like struct contact_info and have one for particle 1 and one for particle 2.

include/fem-dem/ib_particles_dem.h
```diff
@@ -264,25 +314,20 @@ class IBParticlesDEM
...
-    pp_contact_map;
-  std::map<unsigned int,
-           std::map<unsigned int, particle_wall_contact_info<dim>>>
-    pw_contact_map;
+  std::map<unsigned int, std::map<unsigned int, ContactInfo>> pp_contact_map;
+  std::map<unsigned int, std::map<unsigned int, ContactInfo>> pw_contact_map;
 
   // A vector of vectors of candidate cells for each of the particle.
-  std::vector<std::vector<BoundaryCellsInfo>> boundary_cells;
+  std::vector<std::map<unsigned int, BoundaryCellsInfo>> boundary_cells;
```
blaisb, line 330: unordered map are generally faster than map if that's an issue here

source/core/ib_particle.cc
```diff
@@ -299,5 +317,80 @@ IBParticle<dim>::load_data_from_file()
...
 }
 
+
+
+template <int dim>
+void
+IBParticle<dim>::compute_local_inertia()
+{
+  // WIP This is implemented but not used since its application makes the
+  // evaluation of the levelset more complicated.
```
blaisb, line 327: I'm not a huge fan of leaving commented blocks long term in the code....

source/core/parameters.cc
```diff
@@ -2935,6 +2951,26 @@ namespace Parameters
           Patterns::Double(),
           "Smallest gap considered for the lubrification force calculation. This value is multiplied by the smallest cell size");
 
+        prm.declare_entry(
+          "use explicit contact impulsion",
```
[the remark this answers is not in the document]
blaisb: Agreed

source/core/parameters.cc
```diff
@@ -3147,9 +3194,44 @@ namespace Parameters
           particles[i].radius = particles[i].shape->effective_radius;
           prm.enter_subsection("physical properties");
           {
-            particles[i].inertia[0][0] = prm.get_double("inertia");
-            particles[i].inertia[1][1] = prm.get_double("inertia");
-            particles[i].inertia[2][2] = prm.get_double("inertia");
+            std::string              inertia_str = prm.get("inertia");
```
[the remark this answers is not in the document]
blaisb: Agreed

source/core/shape.cc
```diff
@@ -366,6 +371,148 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+Sphere<dim>::distance_to_shape_with_cell_guess(
+  Shape<dim>                                           &shape,
+  const typename DoFHandler<dim>::active_cell_iterator &cell,
+  std::vector<Point<dim>>                              &candidate_points,
+  double                                                precision,
+  bool exact_distance_outside_of_contact)
+{
+  (void)candidate_points;
+  (void)precision;
+  (void)exact_distance_outside_of_contact;
```
blaisb, line 385: You are doing this to avoid warnings right?
If so, add a comment  above to say this.

blaisb (changes requested):
Other bunch of comments. I only have 3 files left. Most likely tomorrow morning

include/core/ib_particle.h
```diff
@@ -372,20 +378,16 @@ class IBParticle
...
 
   // Location of the pressure reference point relative to the center of the
   // particle. This point is used to define a constant on the pressure.
   Point<dim> pressure_location;
+
+  // WIP implemented but not use: Location of the center of mass relative to the
+  // frame of reference of the particule. Point<3> center_of_mass_location
+
+  // Rotation matrix of the particle in the global space
+  Tensor<2, 3> rotation_matrix;
```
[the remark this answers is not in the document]
OGaboriault: It gives the orientation of an object in 3D space. It's really usefull when you want to pass from the particle frame of reference to a global frame of reference. I can give you more explanation about this in person if you want. 

Regardless, this is the appropriated name for the variable.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -167,117 +164,130 @@ IBParticlesDEM<dim>::calculate_pp_contact_force(
                 }
 
               // Calculation of normal overlap
-              double normal_overlap;
+              double                  normal_overlap;
```
OGaboriault, line 167: I would add some comments in this section like it's done for the normal DEM. 

// i is the particle, j is the wall.
// etc. 

Every time we go back in this code, it's always hard to remember the convention that is being used.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -286,10 +296,17 @@ IBParticlesDEM<dim>::calculate_pp_contact_force(
...
+                                      point_nd_to_3d(particle_one.position)),
+                                     -(normal_force + tangential_force));
                   contact_torque[particle_two.particle_id] =
                     contact_torque[particle_two.particle_id] -
-                    particle_two_tangential_torque - rolling_resistance_torque;
+                    particle_two_tangential_torque - rolling_resistance_torque +
+                    cross_product_3d((point_nd_to_3d(contact_point) -
+                                      point_nd_to_3d(particle_two.position)),
+                                     (normal_force + tangential_force));
+                  ;
```
OGaboriault, line 309:
```suggestion
                    particle_two_tangential_torque - rolling_resistance_torque +
                    cross_product_3d((contact_point_3d -
                                      point_nd_to_3d(particle_two.position)),
                                     (normal_force + tangential_force));
                  ;
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -476,163 +522,263 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
             {
-              BoundaryCellsInfo floating_wall_cell_info;
-              floating_wall_cell_info.point_on_boundary =
-                floating_walls_parameters->points_on_walls[i];
-              floating_wall_cell_info.normal_vector =
-                floating_walls_parameters->floating_walls_normal_vectors[i];
-              best_cells[lowest_floating_wall_indices + i + 1] =
-                floating_wall_cell_info;
+              if (floating_walls_parameters->time_start[i] < cfd_time &&
+                  floating_walls_parameters->time_end[i] > cfd_time)
```
OGaboriault, line 571:
```suggestion
              if (floating_walls_parameters->time_start[i] != cfd_time)
```

OGaboriault (changes requested):
First batch finish. Good work

include/core/utilities.h
```diff
@@ -265,6 +265,20 @@ fill_vectors_from_file(std::map<std::string, std::vector<double>> &map,
                        const std::string                           file_name,
                        const std::string delimiter = " ");
 
+/**
+ * @brief Function that read a file that was build from a dealii table and create a map with the key being the column name and the variable the vectors of data.
```
voferreira, line 269:
```suggestion
 * @brief Function that reads a dealii table and creates a map with keys being the column of names and variables being the column of vectors.
```

include/core/utilities.h
```diff
@@ -265,6 +265,20 @@ fill_vectors_from_file(std::map<std::string, std::vector<double>> &map,
...
+ * * @param map A map used to contain the data based on the columns name.
```
voferreira, line 270:
```suggestion
 * * @param map Container to be filled with the information in the dealii table.
```

include/core/utilities.h
```diff
@@ -265,6 +265,20 @@ fill_vectors_from_file(std::map<std::string, std::vector<double>> &map,
...
+ *
+ *   @param file The path the file that will be use to fill up the table.
```
voferreira, line 272:
```suggestion
 *   @param file Path to the dealii table file.
```

include/core/utilities.h
```diff
@@ -265,6 +265,20 @@ fill_vectors_from_file(std::map<std::string, std::vector<double>> &map,
...
+ *
+ *   @param delimiter The delimiter used to read the table.
```
voferreira, line 274:
```suggestion
 *   @param delimiter Character that separates table cells.
```

applications_tests/lethe-fluid-sharp/particles
```diff
@@ -1,6 +1,6 @@
...
-0.0 0.75 0.75 0.75 5 12.75 5 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.00112 1.0 0.0 0.0 0.0 10000000000.0 0.9 0.1 0.3 1.0 1.0
-0.0 0.75 0.75 0.75 7 12.75 5 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.00112 1.0 0.0 0.0 0.0 10000000000.0 0.9 0.1 0.3 1.0 1.0
-0.0 0.75 0.75 0.75 3 12.75 5 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.00112 1.0 0.0 0.0 0.0 10000000000.0 0.9 0.1 0.3 1.0 1.0
-0.0 0.75 0.75 0.75 5 10.75 5 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.00112 1.0 0.0 0.0 0.0 10000000000.0 0.9 0.1 0.3 1.0 1.0 
-0.0 0.75 0.75 0.75 5 8.75 5 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.00112 1.0 0.0 0.0 0.0 10000000000.0 0.9 0.1 0.3 1.0 1.0
+type; shape_argument; p_x; p_y; p_z; v_x; v_y; v_z; omega_x; omega_y; omega_z; orientation_x; orientation_y; orientation_z; volume; density; inertia; pressure_x; pressure_y; pressure_z; youngs_modulus; restitution_coefficient; friction_coefficient; poisson_ratio; rolling_friction_coefficient; integrate_motion
+sphere; 0.75; 5; 12.75; 5; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0.0; 0; 0.00112; 1.0; 0.0; 0.0; 0.0; 10000000000.0; 0.9; 0.1; 0.3; 1.0; true
```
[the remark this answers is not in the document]
voferreira: As such, shouldn't the default delimiter be ";" instead of " "?

voferreira (commented):
Second small batch

applications_tests/lethe-fluid-sharp/particles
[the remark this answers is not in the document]
voferreira: @Luckabarbeau ☝️

source/core/parameters.cc
```diff
@@ -2935,6 +2951,26 @@ namespace Parameters
           Patterns::Double(),
           "Smallest gap considered for the lubrification force calculation. This value is multiplied by the smallest cell size");
 
+        prm.declare_entry(
+          "explicit contact impulsion",
+          "false",
+          Patterns::Bool(),
+          "Bool to enable or disable the explicit evaluation of the contact impulsion. This means that if that parameter is set to true, the DEM is only run once, and the contact impulsion obtained is used for all Newton's iterations.");
```
voferreira, line 2958: As discussed, I think this could be reworded. Bringing the .rst explanation here would be a good approach. Same comment can be applied to the parameters that follow.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -216,19 +226,29 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
                             }
-                          number_of_particles_cutting_this_cell += 1;
+                          particles_cutting_this_cell.push_back(p);
+                          if (particles_cutting_this_cell.size() > 1)
+                            {
+                              for (unsigned int j = 0;
+                                   j < local_dof_indices.size();
+                                   ++j)
+                                {
+                                  dof_with_more_then_one_particle
```
voferreira, line 243:
```suggestion
                                  dof_with_more_than_one_particle
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -113,6 +120,7 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
   cut_cells_map.clear();
   cells_inside_map.clear();
   overconstrained_fluid_cell_map.clear();
+  dof_with_more_then_one_particle.clear();
```
voferreira, line 123:
```suggestion
  dof_with_more_than_one_particle.clear();
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -339,6 +359,14 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
                                 particle_candidate_distance;
                             }
                         }
+
+                      for (unsigned int j = 0; j < local_dof_indices.size();
+                           ++j)
+                        {
+                          dof_with_more_then_one_particle
```
voferreira, line 366:
```suggestion
                          dof_with_more_than_one_particle
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1101,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
-                              if (component_i == 0)
+                              const unsigned int component_i =
+                                this->fe->face_system_to_component_index(i)
+                                  .first;
+                              // Check if that dof already have been used to
+                              // extrapolate the fluid stress tensor on the IB
+                              // surface.
+                              if (force_eval.find(local_face_dof_indices[i]) ==
+                                    force_eval.end() or
+                                  dof_with_more_then_one_particle
```
voferreira, line 1226:
```suggestion
                                  dof_with_more_than_one_particle
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1101,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                                      auto cell_2 =
+                                        ib_done[local_face_dof_indices[i]]
+                                          .second;
+                                      // Check if we already have the cell used
+                                      // to defined the IB constraint of that
+                                      // dof. We always have that information
+                                      // except if the dof is not owned.
+                                      if (ib_done[local_face_dof_indices[i]]
+                                              .first == false or
+                                          dof_with_more_then_one_particle
```
voferreira, line 1255:
```suggestion
                                          dof_with_more_than_one_particle
```

source/fem-dem/gls_sharp_navier_stokes.cc
voferreira: Careful applying this comment, as I may not have cached all typos.

source/fem-dem/gls_sharp_navier_stokes.cc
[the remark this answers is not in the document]
voferreira: Not removed.

source/fem-dem/gls_sharp_navier_stokes.cc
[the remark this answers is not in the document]
voferreira: same

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1101,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                          local_face_projection_triangulation
+                            .set_all_manifold_ids(0);
+                          local_face_projection_triangulation.set_manifold(
+                            0, *particles[p].shape->get_shape_manifold());
+
+                          local_face_dof_handler.reinit(
+                            local_face_projection_triangulation);
+                          local_face_dof_handler.distribute_dofs(local_face_fe);
+
+                          // Defined the solution on  IB surface cell by using
```
voferreira, line 1206:
```suggestion
                          // Define the solution on IB surface cell by using
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -4401,120 +4749,117 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
...
+            Utilities::string_to_double(particles_data["orientation_y"][p_i]);
+          particles[p_i].orientation[2] =
+            Utilities::string_to_double(particles_data["orientation_z"][p_i]);
+
+
+          std::string shape_type = particles_data["type"][p_i];
+          std::string shape_arguments_str =
+            particles_data["shape_argument"][p_i];
+          // Here shape is supposed to be separated by : but the initializer
+          // take ; as separator so we replace : by ;.
```
voferreira, line 4777:
```suggestion
          // takes ; as separator so we replace : by ;.
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1101,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
-                            this->fe->face_system_to_component_index(i).first;
-                          // Check if that dof already have been used to
-                          // extrapolate the fluid stress tensor on the IB
-                          // surface.
-                          if (force_eval.find(local_face_dof_indices[i]) ==
-                              force_eval.end())
+                          // Create a vector to approximate the normal of the
+                          // cell to orient the projection of the face.
+                          Tensor<1, dim> approximate_surface_cell_normal;
+                          // Projects the face on the surface of the IB. This
```
voferreira, line 1157:
```suggestion
                          // Project the face on the surface of the IB. This
```

voferreira (commented):
One more batch of comments.

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -35,14 +35,17 @@ This subsection contains the parameters related to the sharp immersed boundary s
...
+        set alpha                              = 1
+        set contact search frequency           = 1
+        set contact search radius factor       = 3
+        set enable lubrication force           = true
+        set lubrication range max              = 2
+        set lubrication range min              = 0.1
+        set particle nonlinear tolerance       = 1e-6
+        set use explicit contact impulsion     = false
+        set use explicit position integration  = false
+        set use approximate radius for contact = false
```
OGaboriault, line 48:
```suggestion
        set explicit contact impulsion     = false
        set explicit position integration  = false
        set approximate radius for contact = false
```

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -164,8 +168,14 @@ This subsection contains the parameters related to the sharp immersed boundary s
 
     .. warning::
 	    If ``contact search radius factor`` :math:`\leq 1`, an error is thrown.
+	    
+	* The ``enable lubrication force`` parameter enables or disables the use of lubrication forces. This parameter must be set to ``false`` when using a non-newtonian fluid rheology.
     
-    * The ``enable lubrication force`` parameter enables or disables the use of lubrication forces. This parameter must be set to ``false`` when using non-newtonian fluid.
+    * The ``use explicit contact impulsion`` parameter enables or disables the use of explicit contact impulsion evaluation in the resolution of the coupling of the particle. When it is set to true, this parameter results in the code only performing the DEM calculation once per CFD time step and using the resulting contact impulsion to evaluate all the other Newton's iterations. This reduces the number of times the DEM calculation is made. However, since the position is still implicitly evaluated in the absence of contact, the cut cell mapping must be performed at each Newton iteration.
```
OGaboriault, line 174:
```suggestion
    * The ``explicit contact impulsion`` parameter enables or disables the use of explicit contact impulsion evaluation in the resolution of the coupling of the particle. When it is set to true, this parameter results in the code only performing the DEM calculation once per CFD time step and using the resulting contact impulsion to evaluate all the other Newton's iterations. This reduces the number of times the DEM calculation is made. However, since the position is still implicitly evaluated in the absence of contact, the cut cell mapping must be performed at each Newton iteration.
```

doc/source/parameters/sharp-immersed-boundary/sharp-immersed-boundary.rst
```diff
@@ -164,8 +168,14 @@ This subsection contains the parameters related to the sharp immersed boundary s
...
+
+    * The ``use explicit position integration`` parameter enables or disables the use of explicit position integration in the resolution of the coupling of the particles. When it is set to true, this parameter results in the code only performing the DEM calculation once and using the resulting position and orientation to evaluate all the other Newton's iterations. This reduces the number of times the cut cell mapping must be performed and the number of call to the DEM calculations. However, this can affect the stability of the scheme.
+
+    * The ``use approximate radius for contact`` parameter enables or disables the use of the approximate contact radius for contact calculation. When this parameter is true, the contact radius used in the contact force calculation is obtained through the effective contact radius. Otherwise, the curvature radius of the shape is evaluated at the contact point. In the case of a flat surface contact point, the contact radius is limited to 100 times the effective radius of the particle.
```
OGaboriault, line 178:
```suggestion
    * The ``explicit position integration`` parameter enables or disables the use of explicit position integration in the resolution of the coupling of the particles. When it is set to true, this parameter results in the code only performing the DEM calculation once and using the resulting position and orientation to evaluate all the other Newton's iterations. This reduces the number of times the cut cell mapping must be performed and the number of call to the DEM calculations. However, this can affect the stability of the scheme.

    * The ``approximate radius for contact`` parameter enables or disables the use of the approximate contact radius for contact calculation. When this parameter is true, the contact radius used in the contact force calculation is obtained through the effective contact radius. Otherwise, the curvature radius of the shape is evaluated at the contact point. In the case of a flat surface contact point, the contact radius is limited to 100 times the effective radius of the particle.
```

include/core/ib_particle.h
```diff
@@ -284,9 +281,11 @@ class IBParticle
   double rolling_friction_coefficient;
   // The particle mass.
   double mass;
+  // The particle volume.
+  double volume;
   // The surface energy of the particle
   double surface_energy;
-  // The particle rotational inertia. It is uniform in all directions.
+  // The particle moment of inertia in its reference frame
```
OGaboriault, line 288: So, this is the the inertia matrix of the particle in it`s frame of reference. If I understand, this a diagonal matrix right? 

```suggestion
  // The inertia matrix in the particle frame of reference
```

include/core/parameters.h
```diff
@@ -1291,42 +1291,99 @@ namespace Parameters
...
-
-    std::shared_ptr<Functions::ParsedFunction<dim>> f_gravity;
 
-    double      particle_nonlinear_tolerance;
-    double      length_ratio;
-    bool        enable_extrapolation;
-    double      alpha;
-    bool        print_dem;
+    // Number of declared IB particles
+    unsigned int nb;
```
OGaboriault, line 1298: Not sure if the variable name is really descriptive...

include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -164,19 +164,27 @@ class GLSSharpNavierStokesSolver : public GLSNavierStokesSolver<dim>
...
+          }
         integrate_particles();
         update_precalculations_for_ib();
+      }
+    // Do the cut cell mapping only if it is the first Newton iteration or
+    // if the explicit position integration is false.
+    if (this->get_current_newton_iteration() == 0 ||
+        this->simulation_parameters.particlesParameters
+            ->explicit_position_integration_calculation == false)
+      {
```
OGaboriault, line 179:
```suggestion
    if (this->get_current_newton_iteration() == 0 ||
        !(this->simulation_parameters.particlesParameters
            ->explicit_position_integration_calculation) )
      {
```

source/core/shape.cc
```diff
@@ -413,23 +559,27 @@ Superquadric<dim>::closest_surface_point(const Point<dim> &p,
       Point<dim> dx{}, distance_gradient{};
       double     current_distance = superquadric(current_point);
 
-      const double relaxation = 0.5;
+      double relaxation = 1;
```
OGaboriault, line 562:
```suggestion
      double relaxation = 1.;
```

source/core/shape.cc
```diff
@@ -522,9 +672,16 @@ Superquadric<dim>::gradient(const Point<dim> &evaluation_point,
...
+      Tensor<1, dim> gradient;
+      Point<dim>     centered_point = this->align_and_center(evaluation_point);
+      if (superquadric(centered_point) > 0)
+        gradient = (evaluation_point - closest_point) /
+                   ((evaluation_point - closest_point).norm() +
+                    1e-16 * this->effective_radius);
+      else
+        gradient = -(evaluation_point - closest_point) /
+                   ((evaluation_point - closest_point).norm() +
+                    1e-16 * this->effective_radius);
```
OGaboriault, line 684:
```suggestion
      Tensor<1, dim> gradient;
      Point<dim>     centered_point = this->align_and_center(evaluation_point);
      tensor<1,dim> point_distance = evaluation_point - closest_point;
      if (superquadric(centered_point) > 0)
        gradient =  point_distance / (point_distance.norm() +
                    1e-16 * this->effective_radius);
      else
        gradient = - point_distance / (point_distance.norm() +
                    1e-16 * this->effective_radius);
```

OGaboriault: Is there a reason why you add 1e-16 * this->effective_radius ? If it is for avoiding division by zero, why multiplying by this->effective_radius ? If this is the case, you should replace by DML_MIN

Maybe there's something I don't know...

source/fem-dem/ib_particles_dem.cc
OGaboriault: didn't see the were different...

OGaboriault (approved):

include/fem-dem/ib_particles_dem.h
```diff
@@ -145,6 +173,28 @@ class IBParticlesDEM
                              std::vector<Tensor<1, 3>> &contact_torque);
 
 
+  /**
+   * @brief Calculates non-linear (Hertzian) force between two objects. Take into input the physical properties and contact state of the two objects.
+   */
+  void
+  calculate_force_model(const double           normal_overlap,
```
blaisb, line 180: This needs to be refactored to use the new structs you defined no?

include/core/shape.h
```diff
@@ -133,6 +134,950 @@ class Shape : public AutoDerivativeFunction<dim>
...
 
+  /**
+   * @brief Return the smoothed maximum of two variables used for shape contact calculation.
+   * @param a first variable
+   * @param b second variable
+   */
+  double
+  smooth_max(double a, double b)
+  {
+    double smooth_factor = 10;
```
blaisb, line 145: seems like a magic number,. Maybe you could have the smooth max take an optional thirdargument with a default value of 10 ? This way you would get the best of both world :).

include/core/shape.h
```diff
@@ -133,6 +134,950 @@ class Shape : public AutoDerivativeFunction<dim>
...
+          }
+        else
+          {
+            search_direction.push_back(Tensor<1, dim>({1.0, 0.0, 0.0}));
+            search_direction.push_back(Tensor<1, dim>({-1.0, 0.0, 0.0}));
+            search_direction.push_back(Tensor<1, dim>({0.0, 1.0, 0.0}));
+            search_direction.push_back(Tensor<1, dim>({0.0, -1.0, 0.0}));
+            search_direction.push_back(Tensor<1, dim>({0.0, 0.0, 1.0}));
+            search_direction.push_back(Tensor<1, dim>({0.0, 0.0, -1.0}));
+          }
```
blaisb, line 570: What would be simpler here would be to push back the first four vector and push back the last two only if constexpr dim==3. This would make it more readable since the if would contain only the difference.

include/core/shape.h
```diff
@@ -133,6 +134,950 @@ class Shape : public AutoDerivativeFunction<dim>
...
+        // Loop over all the initial candidate points.
+        for (unsigned int i = 0; i < candidate_points.size(); ++i)
+          {
+            // Initialize variable use in the calculation.
```
blaisb, line 574:
```suggestion
            // Initialize variable used in the calculation.
```

include/core/shape.h
```diff
@@ -133,6 +134,950 @@ class Shape : public AutoDerivativeFunction<dim>
...
+                              previous_position[0] -
+                              ((diff_results[0] - diff_results[1]) / 2) /
+                                ((diff_results[0] + diff_results[1]) /
+                                 max_step);
+                            new_point[1] =
+                              previous_position[1] -
+                              ((diff_results[2] - diff_results[3]) / 2) /
+                                ((diff_results[2] + diff_results[3]) /
+                                 max_step);
+                          }
```
blaisb, line 686: Same comment as before. INstead of having two blocks of code, do what is common for both dim values and put the additional stuff in a const expr dim==3 block

include/core/shape.h
```diff
@@ -408,6 +1500,40 @@ class Sphere : public Shape<dim>
   get_shape_manifold() override;
 
 
+  std::tuple<double, Tensor<1, dim>, Point<dim>>
+  distance_to_shape_with_cell_guess(
+    Shape<dim>                                           &shape,
```
blaisb, line 1505: Is this function lacking a brief or it has been defined above?

include/core/shape.h
```diff
@@ -936,12 +2182,185 @@ class CompositeShape : public Shape<dim>
...
+                    x_max = std::max(x_max, new_point[0]);
+                    x_min = std::min(x_min, new_point[0]);
+                    y_max = std::max(y_max, new_point[1]);
+                    y_min = std::min(y_min, new_point[1]);
+                  }
+              }
+          }
+        else
+          {
+            std::vector<Point<3>> component_bounding_box_vertex(8);
```
blaisb, line 2254: I don't undertand why this part is not done in a loop. It just seems like copy pasted code

include/core/shape.h
```diff
@@ -936,12 +2182,185 @@ class CompositeShape : public Shape<dim>
...
+                    xyz_min_max_is_initialized = true;
+                  }
+                else
+                  {
+                    x_max = std::max(x_max, new_point[0]);
+                    x_min = std::min(x_min, new_point[0]);
+                    y_max = std::max(y_max, new_point[1]);
+                    y_min = std::min(y_min, new_point[1]);
+                    z_max = std::max(z_max, new_point[2]);
+                    z_min = std::min(z_min, new_point[2]);
```
blaisb, line 2351: Why have these xmax xmin etc. varibles. Can't you have an array directly instead and put this into a loop? It would be a lot more readable.

blaisb (changes requested):
Other batch of comments. I will continue this week. This is a chunky PR to review really...

include/core/parameters.h
blaisb: It is not.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -4330,68 +4597,141 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
...
+              if (volume == 0)
+                {
+                  if (dim == 2)
+                    {
+                      volume =
+                        PI * particles[p_i].radius * particles[p_i].radius;
+                    }
+                  else if (dim == 3)
+                    {
+                      volume = 4.0 / 3.0 * PI * particles[p_i].radius *
```
blaisb, line 4651: These could be constexpr no?

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -4401,120 +4741,117 @@ GLSSharpNavierStokesSolver<dim>::load_particles_from_file()
...
+                               particles[p_i].radius * particles[p_i].radius;
+                    }
```
blaisb, line 4804: Could be constexpr no?

blaisb (approved):
Two last comments then this is good go to on my side. Please rebase and make sure CI works ok :)

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1093,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                            0, *particles[p].shape->get_shape_manifold());
+
+                          local_face_dof_handler.reinit(
+                            local_face_projection_triangulation);
+                          local_face_dof_handler.distribute_dofs(local_face_fe);
+
+                          // Define the solution on IB surface cell by using
+                          // the IB stencil to extrapolate the fluid stress
+                          // tensor.
+
```
voferreira, line 1201:
```suggestion
```

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -1068,486 +1093,531 @@ GLSSharpNavierStokesSolver<dim>::force_on_ib()
...
+                          approximate_surface_cell_normal =
+                            approximate_surface_cell_normal / vertices_per_face;
+
+                          local_face_cell_data[0].material_id = 0;
+
+                          local_face_projection_triangulation =
+                            Triangulation<dim - 1, dim>();
+                          // Create a dof handler that contains the
+                          // triangulation of the projection of the face on the
+                          // IB. This create the surface cell on the IB
```
voferreira, line 1184:
```suggestion
                          // IB. This creates the surface cell on the IB
```

voferreira (changes requested):
Very last comments and suggestions. What a huge PR, @Luckabarbeau. I will give it a check after you apply (or reject them).

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -216,19 +226,29 @@ GLSSharpNavierStokesSolver<dim>::generate_cut_cells_map()
...
+                          if (particles_cutting_this_cell.size() == 0)
                             {
                               cell_is_cut                      = true;
                               particle_id_which_cuts_this_cell = p;
                               cell_is_inside                   = false;
                               particle_id_in_which_this_cell_is_embedded = 0;
                             }
-                          number_of_particles_cutting_this_cell += 1;
+                          particles_cutting_this_cell.push_back(p);
+                          if (particles_cutting_this_cell.size() > 1)
```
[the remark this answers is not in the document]
Luckabarbeau: I guess the check itself would cost as much as this loop or close to, but yeah, maybe it could help if we found a nice way to do it.

source/fem-dem/gls_sharp_navier_stokes.cc
```diff
@@ -2270,39 +2404,48 @@ GLSSharpNavierStokesSolver<dim>::integrate_particles()
...
                          DBL_MIN);
                 }
-              // Define the correction vector.
-              Tensor<1, 3> omega_correction_vector =
-                residual_omega * 1 / inverse_of_relaxation_coefficient_omega;
+              // Define the correction vector. And check if it is not NAN. NAN
+              // happen when the residual tends to zero in cases where the fluid
+              // density is 0.
+              Tensor<1, 3> omega_correction_vector;
+              if (inverse_of_relaxation_coefficient_omega ==
```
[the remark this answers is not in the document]
Luckabarbeau: No, this is the trick to catch NAN. NAN are not equal to NAN.

blaisb:
@Luckabarbeau can you fix the breaking tests and I will merge after that? I think this is ready for a merge

include/core/ib_particle.h
```diff
@@ -213,7 +211,9 @@ class IBParticle
 
   /**
    * @brief
-   * Sets the orientation of the particle and dependent members to the argument
+   * @brief Sets the orientation of the particle and dependent members to the argument
```
oguevremont, line 214: There are two "@brief"

oguevremont (approved):

blaisb:
@Luckabarbeau it seems the CI often fails for sharp edge test. Like some seg fault was introduced in the code. I'll try to take a look. I know you don't have much time, but if you find any, please also take a quick look at this so I can merge this can call it a day :)

voferreira (approved):
I won't be able to reread all files, but I have given my small comments a look and I think it is good to go.

source/fem-dem/ib_particles_dem.cc
```diff
@@ -464,175 +528,291 @@ IBParticlesDEM<dim>::calculate_pw_contact_force(
...
+                  particle_properties.object_radius =
+                    contact_radius_particle_one;
+                  particle_properties.object_restitution_coefficient =
+                    particle.restitution_coefficient;
+                  particle_properties.object_rolling_friction_coefficient =
+                    particle.rolling_friction_coefficient;
+                  particle_properties.object_youngs_modulus =
+                    particle.youngs_modulus;
+
+                  // Wall mass is 100 times higher than
```
PierreLaurentinCS, line 774: I'm just wondering why the wall properties should depend on the properties of the particle contacting it. And why the factor 100, is it an arbitrary choice or motivated by something?

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+              Point<3> point_on_boundary_3d = point_nd_to_3d(point_on_boundary);
 
+              Tensor<1, 3> point_to_particle_vector =
+                particle_position_3d - point_on_boundary_3d;
 
-          double gap = (projected_vector.norm()) - particle.radius;
+              // Finding the projected vector on the normal vector of the
+              // boundary. Here we have used the private function
+              // find_projection. Using this projected vector, the
+              // particle-wall distance is calculated
```
PierreLaurentinCS, line 922:
```suggestion
              // particle-wall distance is calculated.
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
-                  gap = h_min;
+                  // Limit the smallest gap calculated
+                  if (gap < h_min)
+                    {
+                      gap = h_min;
+                    }
+                  // Evaluate the force
+                  radial_velocity =
+                    scalar_product(-radial_vector, particle.velocity);
+                  f_lub = 3 / 2 * PI * mu * (particle.radius) *
```
PierreLaurentinCS, line 946:
```suggestion
                  f_lub = 3. / 2. * PI * mu * (particle.radius) *
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+                            (particle.radius) / gap * radial_velocity *
+                            radial_vector / radial_vector.norm() -
+                          3 / 2 * PI * mu * (particle.radius) *
```
PierreLaurentinCS, line 949:
```suggestion
                          3. / 2. * PI * mu * (particle.radius) *
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
                 {
-                  gap = h_min;
+                  // Limit the smallest gap calculated
+                  if (gap < h_min)
+                    {
+                      gap = h_min;
+                    }
+                  // Evaluate the force
+                  radial_velocity =
+                    scalar_product(-radial_vector, particle.velocity);
```
PierreLaurentinCS, line 945: shouldn't the radial_vector be divided by its norm here ? You do it three lines below (948) to project the lubrication force on the radial axis. So if you just want the radial component of the velocity there is a / radial_vector.norm() missing

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+
+  // Calculation of tangential relative velocity
+  // v_rt = v_ij - (v_ij⋅n_ij)*n_ij
+  tangential_relative_velocity =
+    contact_relative_velocity -
+    (normal_relative_velocity_value * normal_unit_vector);
+
+  // Calculation of new tangential_overlap, since this value is
+  // history-dependent it needs the value at previous time-step
+  // This variable is the main reason that we have iteration over
```
PierreLaurentinCS, line 1083:
```suggestion
  // This variable is the main reason that we have iterations over
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+  // two different vectors : tangential_overlap of the particles
+  // which were already in contact needs to
```
PierreLaurentinCS, line 1085:
```suggestion
  // which were already in contact needs to be
```

source/fem-dem/ib_particles_dem.cc
```diff
@@ -655,113 +835,323 @@ IBParticlesDEM<dim>::calculate_pw_lubrication_force(
...
+    tangential_spring_constant * contact_info.tangential_overlap;
+
+  double coulomb_threshold =
+    effective_coefficient_of_friction * normal_force.norm();
+
+  // Check for gross sliding
+  if (tangential_force.norm() > coulomb_threshold)
+    {
+      // Gross sliding occurs and the tangential overlap and tangential
+      // force are limited to Coulumb's criterion
```
PierreLaurentinCS, line 1136:
```suggestion
      // force are limited to Coulomb's criterion
```

PierreLaurentinCS (commented):
I'm arriving after the war, mostly typos and some questions about the implementation.

source/core/shape.cc
```diff
@@ -366,6 +370,148 @@ Sphere<dim>::set_position(const Point<dim> &position)
...
+  Tensor<1, dim> normal;
+  Point<dim>     contact_point;
+  if (typeid(shape) == typeid(Sphere<dim>))
+    {
+      Tensor<1, dim> center_to_center_vector =
+        shape.get_position() - this->position;
+      distance = (center_to_center_vector.norm() - this->effective_radius -
+                  shape.effective_radius) /
+                 2;
+      normal = center_to_center_vector / center_to_center_vector.norm();
```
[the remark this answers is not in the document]
Luckabarbeau: it is not possible

source/fem-dem/ib_particles_dem.cc
Luckabarbeau: the idea here is to make the wall way bigger than the particle while avoiding using inf. a 100 is an arbitrary choice but it is good enough.

Luckabarbeau:
I think this PR is now ready to merge.

PR #986, opened 2024-01-18
Dem output solid object option

OGaboriault:
# Description of the problem

- Solid object output files were always being outputted at every output time step. When using a large amount of solid object, the amount of output files can become astronomical pretty quickly. 

# Description of the solution

- I have added a new parameter to solid object. We can now specify if we want output files to be generated or not. By default, this parameter is at true, which keep compatibility with the previous parameter files (like the examples). 

# How Has This Been Tested?

- I used one of my prm file for my LPBF simulation and put solid object 0 and 2 to false. Ran the simulation with zero particle and no output file were generated for those two solid objects. 

# Documentation

- Small correction to the solid object documentation.
- Added the parameter to the doc.

# Future changes

- We could replace this new bool to a function like for the "translational velocity", so that vtus would be outputted during a certain period.

### DIFF
--- a/doc/source/parameters/dem/floating_mesh.rst
+++ b/doc/source/parameters/dem/floating_mesh.rst
@@ -14,10 +14,10 @@ This subsection explains the solid objects (floating meshes) information. First 
 
 .. code-block:: text
 
- subsection floating walls
+ subsection solid objects
   # Total number of floating walls
     set number of solids  = 1
-      subsection  solid object 0
+      subsection solid object 0
         subsection mesh
             set type                   = gmsh
             set file name              = none
@@ -45,6 +45,10 @@ This subsection explains the solid objects (floating meshes) information. First 
             # Z COR
             set z = 0
         end
+
+        subsection output
+            set output solid object = true
+        end
     end
  end
 
@@ -57,4 +61,6 @@ This subsection explains the solid objects (floating meshes) information. First 
 * In the subsection ``angular velocity``, we define the angular velocity of the floating mesh.
 
 * In the subsection ``center of rotation``, we define the center of rotation of the solid object (in the case of rotational motion).
+
+* In the subsection ``output``, we define if we want an output file to be generated for the solid object at every output time step.
 
--- a/doc/source/parameters/dem/floating_mesh.rst
+++ b/doc/source/parameters/dem/floating_mesh.rst
@@ -46,9 +46,8 @@ This subsection explains the solid objects (floating meshes) information. First 
             set z = 0
         end
 
-        subsection output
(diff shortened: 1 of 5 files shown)


doc/source/parameters/dem/floating_mesh.rst
```diff
@@ -45,6 +45,10 @@ This subsection explains the solid objects (floating meshes) information. First
             # Z COR
             set z = 0
         end
+
+        subsection output
+            set output solid object = true
+        end
```
blaisb, line 51: Why did you to create a subsection? Could it not be just a free parameter?

include/core/serial_solid.h
```diff
@@ -291,7 +291,8 @@ class SerialSolid
   Vector<double>                                displacement_since_mapped;
 
   // Output management
-  PVDHandler pvdhandler;
+  PVDHandler  pvdhandler;
+  const bool *output_bool;
```
blaisb, line 295: why a pointer???

source/core/serial_solid.cc
```diff
@@ -48,6 +45,7 @@ SerialSolid<dim, spacedim>::SerialSolid(
   , this_mpi_process(Utilities::MPI::this_mpi_process(mpi_communicator))
   , param(param)
   , id(id)
+  , output_bool(&param->output_bool)
```
blaisb, line 48: no need for a pointer with an adress, just pass the value itself. I don't see the point of having this pool pointer floating around

source/core/serial_solid.cc
```diff
@@ -469,52 +467,55 @@ void
...
-
-  write_vtu_and_pvd<dim, spacedim>(pvdhandler,
-                                   data_out,
-                                   folder,
-                                   solution_name,
-                                   time,
-                                   iter,
-                                   group_files,
-                                   this->mpi_communicator);
+  if (*output_bool)
```
blaisb, line 470: instead I would have something like this
```suggestion
  if (!*output_bool) return;
```
You just leave the function if you are not outputting. Easier to read

blaisb (changes requested):
Some comments to address, The pointer thing is very weird

doc/source/parameters/dem/floating_mesh.rst
OGaboriault: Yes it can be. It would be the only parameter not in a subsection. It's not a problem in itself. ]

acdaigneault (commented):
My comments are the same as Bruno (pointer & subsections)

blaisb (approved):
Much better, thanks!

PR #987, opened 2024-01-19
Calculate average velocities and Reynolds stresses for deal.II vectors

lpsaavedra:
# Description of the problem

The calculation of average velocities and Reynolds stresses was not supported for deal.II vectors yet and consequently not supported by the `lethe-fluid-matrix-free` application.

# Description of the solution

The `postprocessing_velocities.cc` file was updated with the corresponding way of calculating these quantities for deal.II vectors. 

# How Has This Been Tested?

Two new tests were added in the `tests/solvers` folder to verify the implementation:

- [X] `average_velocities_03.cc`
- [X] `reynolds_stress_03.cc`

# Comments

There is a lot of code duplication but this is a temporary solution until we figure out what to do with Trilinos and deal.II vectors. Also, the current implementation only supports equal-order elements, which is not desirable for the `lethe-fluid ` solver. Therefore, I will be working in a more general implementation that I started during the last Hackathon.


blaisb (approved):
Amazing, good job @lpsaavedra

PR #988, opened 2024-01-21
Fix heat flux postprocessor material_id processing

AmishgaAlphonius:
# Description of the problem

- In the previous implementation of the heat flux postprocessor, there was a confusion related to the `material_id` of a cell. As a cell's `material_id` is defined in the mesh, fluids moving from a cell to another did not have the right `material_id` when outputted. Futhermore, some dealii meshs have an intrinc way of numbering `material_ids` (e.g. colorized subdivided_hyper_rectangle) that were leading to inadequate comparison of the `material_ids`.

# Description of the solution

- As of this PR, in the heat flux postprocessor, the `material_id` is only used to indentify cell material when solids are present. When simulating with fluids and solids, cells of the mesh in which the fluids lie should have a `material_id` of `0`.
- For _fluids-only_ simulations, at the moment, their heat fluxes will be output on the entire domain, as the current `DataPostprocessor` can only hold 1 `dof_handler`. During postprocessing, the user could clip the domain of interest of each fluid using the `phase` or `filtered_phase` fields outputted.

# How Has This Been Tested?

- Visual inspection of outputted results for a:

- [x] single fluid simulation,
- [x] 1 fluid + 1 solid simulation, and
- [x] 2 fluids simulaiton.

# Future changes

- For multiple-fluids simulations, a `DataPostprocessor` handling multiple `dof_handlers` to allow postprocessing of quantities of interest of fluids in their respective subdomains will have to be added to the code.

### DIFF
--- a/include/solvers/postprocessors.h
+++ b/include/solvers/postprocessors.h
@@ -511,16 +511,20 @@ private:
 
 
 /**
- * @class Calculates heat flux (q = -k * grad T) of a material at every
- * quadrature point.
+ * @class Postprocessor that calculates local heat fluxes \f$(q = -k * \nabla
+ * T)\f$ of a material at every evaluation point of the domain.
+ *
  * @param p_thermal_conductivity_model Thermal conductivity model of the
  * material
  * @param p_material_string Type of material: "f" (fluid) or "s" (solid)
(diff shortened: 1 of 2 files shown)


blaisb (approved):
Nice fix, thank you :)!

PR #989, opened 2024-01-22
Quick fix

OGaboriault:
# Description of the problem

- Pyvista doesn't work with the new padding


blaisb (approved):
That's really a temporary fix though, because it means we are not reading the pvtu files

PR #992, opened 2024-01-23
Update Windows Lethe Installation

OresteMarquis:
# Added some clarification in the installation steps for Windows.

- No step to install WSL in the linked tutorial in step 1.
- The default gcc installation was higher than 10. on my computer, not sure if it is a problem?
- Rephrase the first sentence in step # 2 since it is weird to read if you don't pass by candi installation.
- apt-get numdiff was not working on my computer, I used sudo apt install numdiff

# More issues that I was not able to solve:
- When I tested the installation with : "ctest -j$numprocs". I had 15 failed tests : 88,90,92,137,232,246,281,287,289,290,291,292,293 and 294. I also had a massage when building saying "could not find Doxygen", maybe it is linked.

# Other suggestion:
- The documentation only show how to update deal.II using candi, maybe add an other way since some people will not use candi for installation.

### DIFF
--- a/doc/source/installation/windows_wsl.rst
+++ b/doc/source/installation/windows_wsl.rst
@@ -86,11 +86,11 @@ When prompted "do you want to continue?", proceed by typing ``y`` and hitting ``
 
 The following step is to install deal.II. This can be done through
 
-  1. Advanced Packaging Tool (apt) (**this is by far the easiest way to proceed**) : :ref:`install-deal.II-apt` (recommended for users)
+  1. Advanced Packaging Tool (apt) (**this is by far the easiest way to proceed**) : :ref:`install-deal.II apt` (recommended for users)
 
-  2. Candi shell script (`candi github page <https://github.com/dealii/candi>`_): :ref:`install-deal.II-candi` (recommended for developers)
+  2. Candi shell script (`candi github page <https://github.com/dealii/candi>`_): :ref:`install-deal.II candi` (recommended for developers)
 
-.. _install-deal.II-apt:
+.. _install-deal.II apt:
 
 Installing deal.II using apt (Step #1)
 -----------------------------------------
@@ -125,7 +125,7 @@ This should output several information about the installed version. Everything w
 
   If the installed version is other than ``deal.ii-9.5.1``, follow `this link <https://github.com/dealii/dealii/wiki/Getting-deal.II>`_.
 
-.. _install-deal.II-candi:
+.. _install-deal.II candi:
 
 Installing deal.II using Candi (Step #1)
 -----------------------------------------
--- a/doc/source/installation/windows_wsl.rst
+++ b/doc/source/installation/windows_wsl.rst
@@ -86,16 +86,18 @@ When prompted "do you want to continue?", proceed by typing ``y`` and hitting ``
 
 The following step is to install deal.II. This can be done through
 
-  1. Advanced Packaging Tool (apt): **this is by far the easiest way to proceed**
+  1. Advanced Packaging Tool (apt) (**this is by far the easiest way to proceed**) : :ref:`install-deal.II-apt` (recommended for users)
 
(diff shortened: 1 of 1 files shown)


blaisb (approved):

blaisb:
@OresteMarquis the tests that you have that are failing are most likely related to the usage of p4est. Did you install p4est 2.2 (manually through candi) or did you install the apt-get version? if you install the apt-get version, it's normal that these tests don't pass because the CI uses a previous version of p4est. Maybe adding a comment to the PR to say that if those test don't pass, it's not a big deal.

doc/source/installation/windows_wsl.rst
```diff
@@ -152,7 +159,7 @@ Installing deal.II using Candi (Step #1)
 
     gcc --version
 
-  should return ``gcc (Ubuntu 10.X.X...) 10.X.X``. If not, go to :ref:`change compiler default version`.
+  should return ``gcc (Ubuntu 10.X.X...) 10.X.X`` or higher. If not, go to :ref:`change compiler default version`.
```
voferreira, line 162: Good catch

doc/source/installation/windows_wsl.rst
```diff
@@ -328,6 +342,13 @@ Where ``$numprocs`` corresponds to the number of processors used for the compila
     
     apt-get numdiff
 
+  or
+
+  .. code-block:: text
+    :class: copy-button
+    
+    sudo apt install numdiff
```
voferreira, line 350: You don't need sudo privilege to any machine or user. I would keep it as a warning, but do not add it to the command itself. I suggest having a comment like "if you need superuser privilege, use sudo"

voferreira (approved):
Nice! Just added some comments, but to me it is good to go. Please, just make sure that apt and apt-get install commands are uniform in terms of privilege needed.

voferreira:
> @OresteMarquis the tests that you have that are failing are most likely related to the usage of p4est. Did you install p4est 2.2 (manually through candi) or did you install the apt-get version? if you install the apt-get version, it's normal that these tests don't pass because the CI uses a previous version of p4est. Maybe adding a comment to the PR to say that if those test don't pass, it's not a big deal.

Good call. When we do candi we change the version. Shouldn't this be uniform?

blaisb:
@OresteMarquis ping me when you have added this and this will be ready for a merge. Very good work man!

OresteMarquis:
@blaisb Perfect, I just want to make sure that the p4est I installed on my computer is indeed a version more recent than 2.2 to be sure that it is the cause of the failing tests. I will look into it when I come back home and add all your comments at this moment and let you know! Thank you everyone :)

OresteMarquis:
I added your comments and a little bit more clarifications let me know what you think. Also, I noticed that some steps are really similar to the regular linux installation once wsl is installed. Do you think it is possible to link the 2 installation procedure so in the future it will not be necessary to keep track that both documentation have all the informations?

blaisb:
> I added your comments and a little bit more clarifications let me know what you think. Also, I noticed that some steps are really similar to the regular linux installation once wsl is installed. Do you think it is possible to link the 2 installation procedure so in the future it will not be necessary to keep track that both documentation have all the informations?

It might be possible, but I am unsure how to do it. For the time let's keep them seperate

voferreira:
> > I added your comments and a little bit more clarifications let me know what you think. Also, I noticed that some steps are really similar to the regular linux installation once wsl is installed. Do you think it is possible to link the 2 installation procedure so in the future it will not be necessary to keep track that both documentation have all the informations?
> 
> It might be possible, but I am unsure how to do it. For the time let's keep them seperate

Agreed. But a way of doing it would be to have a separate section like "Dependencies with Candi", or something like this. For now I would just add this to the project so that we don't forget. I could make it eventually (maybe during the doxygen hackathon?), it is a quick refactoring of this small part of the documentation anyways.

blaisb (approved):

PR #993, opened 2024-01-24
Automate padding string length

AmishgaAlphonius:
# Description of the problem

- When trying to check an example for another PR, the postprocessing with `lethe_pyvista_tools` was displaying an error. Upon a closer look, it was identified that the error rooted from an incorrect padding of the `.vtu` files numbering.

# Description of the solution

- The padding length is now determined by reading the `.pvtu` of the list and extracting the length of the number.

# How Has This Been Tested?

- By postprocessing successfully the "Sloshing in a Rectangular Tank" example

### DIFF
--- a/contrib/postprocessing/lethe_pyvista_tools/__init__.py
+++ b/contrib/postprocessing/lethe_pyvista_tools/__init__.py
@@ -69,6 +69,8 @@ class lethe_pyvista_tools():
 
         self.list_vtu           -> Returns the list of names of .vtu files.
 
+        self.padding            -> Returns the padding of vtu file numbering.
+
         """
 
         self.path_case = case_path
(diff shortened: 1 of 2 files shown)


contrib/postprocessing/lethe_pyvista_tools/__init__.py
```diff
@@ -69,6 +69,8 @@ def __init__(self, case_path = '.', prm_file_name = '', pvd_name = '', prefix =
 
         self.list_vtu           -> Returns the list of names of .vtu files.
 
+        self.padding            -> Returns the padding of vtu file numbering.
```
voferreira, line 72: Nice

voferreira (approved):
Good job! Good to go.

PR #994, opened 2024-01-24
Add temperature extrapolation

AmishgaAlphonius:
# Description 
- This PR adds the extrapolation in time of temperature and temperature gradient value for source terms in the Navier-Stokes equations requiring them (Marangoni and evaporation recoil force terms). 

# How Has This Been Tested?
- All tests have passed

# Comment
- Velocity extrapolation has been moved to `include/solvers/vof_scratch_data.h`. Solution extrapolations are now located in `ScratchData` to avoid constraint on assemblers order.

### DIFF
--- a/include/solvers/navier_stokes_scratch_data.h
+++ b/include/solvers/navier_stokes_scratch_data.h
@@ -20,6 +20,7 @@
 #include <core/parameters.h>
 #include <core/physical_property_model.h>
 #include <core/rheological_model.h>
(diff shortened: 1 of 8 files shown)


blaisb (approved):
That's a very good change. Amazing work.
I don't have any comments. Good work, seriously good work :). that was a much needed feature :)
@hepap it will be interesting to see if this changes the results for the Marangoni case. It should not.

PR #997, opened 2024-01-25
Add uniform laser distribution for 1D lpbf problems

hepap:
# Description of the problem

- Previously, the only laser irradiation distribution available was a Gaussian distribution, which is not suitable for 1D approximation of the lpbf process.
- A sign mistake was found in the recoil pressure assembler while investigating the 1D case.

# Description of the solution

- A uniform distribution for the laser is now available
- Bug was corrected

# How Has This Been Tested?

- A new test with an analytical solution for the laser model was added
- The test for the recoil pressure was updated

- [x] applications_tests/lethe-fluid/heat_transfer_vof_interface_uniform_laser.prm
- [x] applications_tests/lethe-fluid/evaporation_static_droplet_constant_mass_flux.prm

# Documentation

- Laser documentation was updated

# Future changes

- The pre-exponential factor associated to the 2D gaussian laser model will be added in the next PR.


blaisb (approved):
Ver good PR. Great work @hepap .
quick question, the scaling factor (for the gaussian in 1D, 2D, 3D) you apply it manually in the prm file i suppose?

source/core/parameters.cc
```diff
@@ -1372,9 +1372,10 @@ namespace Parameters
       prm.declare_entry(
         "type",
         "exponential_decay",
```
AmishgaAlphonius, line 1374: This is not directly related to this PR, but I was wondering, since we will be using more often the `gaussian_heat_flux_vof_interface` laser `type`, why don't we make it the default?

Additionally, if we do change it, the order of the if/else if/else statements when parsing `(lines 1429-1434)` could also be changed in an order of decreasing likelihood of events (most probable to least). They don't match the ones of the assemblers either in `heat_transfer.cc`.

applications_tests/lethe-fluid/heat_transfer_vof_interface_uniform_laser.prm
```diff
@@ -0,0 +1,246 @@
...
+  set type = nodal
+  subsection uvwp
+    set Function expression = 0; 0; 0
+  end
+  subsection VOF
+    set Function expression = if (y<0.43, 1, 0)
+    subsection projection step
+      set enable           = false
+      set diffusion factor = 1
+    end
```
AmishgaAlphonius, line 98:
```suggestion
```

applications_tests/lethe-fluid/heat_transfer_vof_interface_uniform_laser.prm
```diff
@@ -0,0 +1,246 @@
...
+  set number of material interactions = 1
+  subsection material interaction 0
+    set type = fluid-fluid
+    subsection fluid-fluid interaction
+      set first fluid id              = 0
+      set second fluid id             = 1
+      set surface tension model       = constant
+      set surface tension coefficient = 0
+    end
+  end
```
AmishgaAlphonius, line 134: You don't use surface tension from what I understand ^^'
```suggestion
```

applications_tests/lethe-fluid/heat_transfer_vof_interface_uniform_laser.prm
```diff
@@ -0,0 +1,246 @@
...
+    set beta      = 30
+  end
+  subsection surface tension force
+    set enable                                   = true
+    set phase fraction gradient diffusion factor = 1
+    set curvature diffusion factor               = 1
+    set output auxiliary fields                  = true
+    set enable marangoni effect                  = false
+  end
+end
```
AmishgaAlphonius, line 172: Same comment as above 

```suggestion
subsection VOF
  subsection phase filtration
    set type = tanh
    set beta = 30
  end
end
```

AmishgaAlphonius (commented):
Great work! :D 
I've only got a few minor comments!

source/core/parameters.cc
blaisb: I would not worry about the order of the if, but for sure it can be good to put the default laser to the good one we want to use

blaisb:
Very good catches @AmishgaAlphonius , you have a better eye than I do :)
@hepap you can ping me when you are ok with all the changes (or add the reviewed and ready to merge label)
:)

voferreira:
Do you still need my review? I can do it tomorrow afternoon after a meeting or during the weekend.

source/core/parameters.cc
hepap: I will add the factor for the 2D laser in another PR today, and at the same time I'll switch the default laser model :)

hepap:
I'va apply your suggestions! @blaisb I'll add the scaling in my next PR (hopefully done by the end of the day)!

PR #998, opened 2024-01-25
Fix kelly estimator in parallel for deal.II vectors

lpsaavedra:
# Description of the problem

The `lethe-fluid-matrix-free` application was not working in parallel when the Kelly error estimator was used for adaptive mesh refinement. 

# Description of the solution

This PR fixes the issue by using a locally relevant solution with ghosts information in the Kelly error estimator `estimate(...) `call. 

# How Has This Been Tested?

- All the tests that use Kelly error estimator in the `lethe-fluid` application pass. 
- A new cylinder test case is added to the `lethe-fluid-matrix-free` application.


source/solvers/navier_stokes_base.cc
```diff
@@ -900,27 +900,83 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_kelly()
...
+                        std::is_same_v<VectorType,
+                                       TrilinosWrappers::MPI::BlockVector>)
+            {
+              KellyErrorEstimator<dim>::estimate(
+                *this->mapping,
+                this->dof_handler,
+                *this->face_quadrature,
+                typename std::map<types::boundary_id,
+                                  const Function<dim, double> *>(),
+                present_solution,
```
peterrum, line 954: What type of ghosting does `present_solution` have?

lpsaavedra: It is declared here:
https://github.com/lethe-cfd/lethe/blob/6dc4461ccfc72a4e2ba0ec572e467197a7614544/include/solvers/navier_stokes_base.h#L441

Initialized for the `lethe-fluid-matrix-free` application using the matrix-free operator here: 
https://github.com/lethe-cfd/lethe/blob/6dc4461ccfc72a4e2ba0ec572e467197a7614544/source/solvers/mf_navier_stokes.cc#L209

peterrum: Thanks! 

May I propose an other solution to the problem. Always copy the values from `present_solution` to a locally relevant version (AMR is not a performance-critical part of the code). This is done at other places in lethe.

peterrum: I don't think you need to use `copy_locally_owned_data_from()` but you can simply use the copy assignment operator.

peterrum (approved):
Thanks! Looks good to me!

peterrum:
```
An error occurred in line <2010> of file <../source/grid/grid_in.cc> in function
    void dealii::GridIn<dim, spacedim>::read_msh(std::istream&) [with int dim = 2; int spacedim = 2; std::istream = std::basic_istream<char>]
The violated condition was: 
    in.fail() == false
```

lpsaavedra:
@blaisb I am having some issues in the CI with the mesh of the new test... everything works in my machine. Do you spot a mistake in the CMakeLists.txt?

applications_tests/lethe-fluid-matrix-free/CMakeLists.txt
```diff
@@ -2,11 +2,13 @@ set(TEST_TARGET lethe-fluid-matrix-free)
 
 string(TOLOWER ${CMAKE_BUILD_TYPE} _build_type)
 
+file(COPY cylinder_structured.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/cylinder_kelly_estimator.${_build_type}")
+
```
blaisb, line 6: @lpsaavedra 
Generally to copy the files in the example we use:
file(COPY taylorCouette.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
and this would copy the file in which the case is running.
I am unsure in the present case where the file is being copied when you run it like this?

applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.prm
```diff
@@ -0,0 +1,186 @@
...
+  end
+end
+
+#---------------------------------------------------
+# Mesh
+#---------------------------------------------------
+
+subsection mesh
+  set type      = gmsh
+  set file name = ../cylinder_structured.msh
```
peterrum, line 83: Maybe:
```suggestion
  set file name = ../../cylinder_structured.msh
```

If this is not working, one needs to look at in which working directly we are and list all files there. I would say we are in `lethe/build/applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.debug/mpirun=2`.

lpsaavedra: Yes, I will try to solve this now

lpsaavedra: I have copied now the mesh also into this folder to ensure that we can use the same path in the parameter file for all the tests... let's see if it works now

PR #1004, opened 2024-01-28
Fix CI after introduction of serial folder in deal.II 9.6

blaisb:
# Description of the problem

- Deal.II introduced a serial folder for serial tests. This is a very good idea because it introduces uniformity when we have serial and mpitests. However, this is not maintainable if we have deal.II 9.5 and 9.6 working. The PR is this one: https://github.com/dealii/dealii/pull/16551

# Description of the solution

- For the time being, all tests that have files are given the label mpirun=1 before their output. As such, their behavior is identical between 9.5 and 9.6. This ensures that the tests are working correctly. Once we migrate away from deal.II 9.5 (this summer) this idea could be reverted to use the regular serial folders. This is an adequate fix for the time being.


lpsaavedra (approved):
All the tests are now passing on my machine using deal.II-master

PR #1007, opened 2024-01-29
Fix read checkpoint for the matrix-free application

lpsaavedra:
# Description of the problem

The restart option was not working in parallel for the matrix-free application. 

# Description of the solution

Use locally relevant vectors for the present and previous solutions regardless of the type of vector.  In addition, use a locally relevant vector for mesh refinement (in the `init_temporary_vector()` call).

# How Has This Been Tested?

- [X] All existing tests pass.
- [X] A new test `tgv_restart_bdf1` is added to the MF application in serial and in parallel to verify restart feature.


blaisb (approved):

blaisb:
Sorry about that, I should have caught that error before merging the other PR.
Thanks for adding a test.

PR #1021, opened 2024-02-09
Doxygen ib particles

oguevremont:
# Description of the problem

- Some comments needed to be added or completed for Doxygen.

# Description of the solution

- It is done for IBParticle class.

# Documentation

- Yes

### DIFF
--- a/include/core/ib_particle.h
+++ b/include/core/ib_particle.h
@@ -33,22 +33,25 @@
 using namespace dealii;
 
 /**
- * @brief This class defines values related to a particle used in the sharp interface IB. These values are used in the solver
+ * @brief Define behavior of a specific particle for the sharp interface IB solver.
  *
- * @tparam dim  An integer that denotes the number of spatial dimensions
+ * Each particle defined will have these value used in the solver.
+ *
+ * @tparam dim An integer that denotes the dimension of the space in which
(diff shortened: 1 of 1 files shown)


include/core/ib_particle.h
```diff
@@ -33,55 +33,62 @@
...
 
 /**
- * @brief This class defines values related to a particle used in the sharp interface IB. These values are used in the solver
+ * @brief Define the behavior of a specific particle for the sharp interface IB solver.
  *
- * @tparam dim  An integer that denotes the number of spatial dimensions
+ * Each particle defined will have these values used in the solver.
+ *
+ * @tparam dim An integer that denotes the dimension of the space in which
+ * the flow is solved.
```
blaisb, line 41: Is flow really the appropriate comment here for the tparam? Maybe we could not put the tparam at all or here you could say that dim is the number of spatial dimension.

blaisb (approved):
Just one comment. another thing, do we need to specify the brief for all the class member comments or not? @AmishgaAlphonius ?

include/core/ib_particle.h
oguevremont: You are right, I changed it to the number of spatial dimensions.

blaisb (approved):

PR #1023, opened 2024-02-09
Remove brief description

lpsaavedra:
# Description of the problem

In the current doxygen documentation we see the brief description of each function in the "Public Member Functions" tab.

# Description of the solution

The corresponding option in the doxygen tab is fixed, so that we only see the list and if more recommendation required one can go to the "Member Function Documentation" section.  Just like in deal.ii.


lpsaavedra:
@acdaigneault This PR fixes what you pointed out earlier.

PR #1024, opened 2024-02-09
Change title of doxygen main page

lpsaavedra:
# Description of the problem

The word "documentation" was redundant in the title of the index page of the doxygen documentation. 

# Description of the solution

This avoids repetition of the word and makes it clearer.


lpsaavedra:
@paulapatience This should avoid what you told me earlier, what do you think? Do you have any other suggestion?

PR #1029, opened 2024-02-10
Update comments in vof_filter.h

AmishgaAlphonius:


include/solvers/vof_filter.h
```diff
@@ -21,46 +21,61 @@
 #include <core/parameters_multiphysics.h>
 
 /**
- * @brief Class that allows phase fraction filtering
+ * @brief Abstract class for filtering phase fraction in volume of fluid (VOF)
+ * simulations.
```
blaisb, line 25: I would remove the abstract class. What do you think?

blaisb (approved):
One tiny change

include/solvers/vof_filter.h
AmishgaAlphonius: I think that the **brief** should say that you can't directly instantiate an object of the class; should it not?

blaisb: Agreed

PR #1030, opened 2024-02-11
Colorized cylinder shell

blaisb:
# Description of the problem

- The current cylinder shell in deal.II does not have boundary condition. 
- 
# Description of the solution

- I am writing a PR to address that in deal.II, but in the meantime, I added a colorized hypershell mesh in Lethe to ensure that we can proceed with the turbulent Taylor-Couette project

# How Has This Been Tested?

- An application test to test this was added

# Documentation

- Documentation for this mesh generation was added


source/core/parameters.cc
```diff
@@ -1922,12 +1922,13 @@ namespace Parameters
...
-                          "gmsh|dealii|periodic_hills|cylinder"),
-                        "Type of mesh "
-                        "Choices are <gmsh|dealii|periodic_hills|cylinder>.");
+      prm.declare_entry(
+        "type",
+        "dealii",
+        Patterns::Selection(
+          "gmsh|dealii|periodic_hills|cylinder|colorized_cylinder_shell"),
+        "Type of mesh "
+        "Choices are <gmsh|dealii|periodic_hills|cylinder>.");
```
lpsaavedra, line 1931:
```suggestion
        "Choices are <gmsh|dealii|periodic_hills|cylinder|colorized_cylinder_shell>.");
```

lpsaavedra (approved):
Only some minor comments... but otherwise good to go

PR #1033, opened 2024-02-12
Add time step limit for adaptive time stepping scheme

hepap:
# Description of the problem

- Adaptive time stepping was only based on the CFL, meaning that for problems with another time step limit (i.e. surface tension problems(, the latter could be not respected.

# Description of the solution

- A max_dt was added

# How Has This Been Tested?

- Default value of max_dt was added to unit test so they don't fail
- [x] core/simulation_control_03.cc
- [x] solvers/average_velocities_02.cc
- [x] solvers/reynolds_stress_02.cc

# Documentation
- Update simulation control doc
- [x] simulation_control.rst

### DIFF
--- a/doc/source/parameters/cfd/simulation_control.rst
+++ b/doc/source/parameters/cfd/simulation_control.rst
@@ -48,6 +48,9 @@ This subsection contains the general information of the simulation, including th
 
     # Maximum CFL value
     set max cfl                      = 1
+		
+		# Maximum time step value
+    set max time step                = 1e6
 
     # Adaptative time step scaling
     set adaptative time step scaling = 1.1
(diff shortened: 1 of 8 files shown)


doc/source/parameters/cfd/simulation_control.rst
```diff
@@ -117,6 +120,8 @@ This subsection contains the general information of the simulation, including th
 
 * ``max cfl``: maximal value of the :math:`\text{CFL}` condition that can be reached during the simulation. This parameter is only used when ``set adapt = true``.
 
+* ``max time step``: maximal value of the time step that can be reached during the simulation. This parameter is only used when ``set adapt = true``. It is useful when the problem of interest has an additional time step constraint such as the capillary time step limit described in :doc:`../../examples/multiphysics/capillary-wave/capillary-wave`.
```
blaisb, line 123: maximal or maximum?

blaisb (approved):
Minor question, otherwise all good. great work!

include/core/parameters.h
```diff
@@ -115,10 +115,13 @@ namespace Parameters
     // Max CFL
     double maxCFL;
 
+    // Max dt
```
AmishgaAlphonius, line 118: Just to be consistent with the comment in `simulation_control.h`
```suggestion
    // Max time step
```

doc/source/parameters/cfd/simulation_control.rst
AmishgaAlphonius, line 123:
```suggestion
* ``max time step``: maximum time step value that can be reached during simulation. This parameter is only used when ``set adapt = true``. It is useful when the problem of interest has an additional time step constraint such as the capillary time step limit described in :doc:`../../examples/multiphysics/capillary-wave/capillary-wave`.
```

AmishgaAlphonius (approved):
Great work as always!
I have two very minor comments. :)

PR #1037, opened 2024-02-20
Optimize heat transfer

blaisb:
# Description of the problem

- The heat transfer calls a lot of time the calculate_bdf_coefficients and this turns out to be significantly more expensive than I thought it was.

# Description of the solution

- we change the calculation of the bdf coefficients to occur only once per time step in the simulation control object and we just query the value. For phase change cases this has a tremendous impact on performance

### DIFF
--- a/include/core/bdf.h
+++ b/include/core/bdf.h
@@ -37,10 +37,6 @@ using namespace dealii;
  * integration based on the BDF for viscous flows." Journal of Computational
  * Physics 291 (2015): 151-176.
  *
- * For example, if the method is a BDF2, it uses three values for the time :
- * \f$t\f$, \f$t-\Delta t_1\f$ and \f$t-\Delta t_2\f$. Thus the @p time_steps
- * vector should contain \f$\Delta t_1\f$ and \f$\Delta t_2\f$.
- *
  * @param[in] method Time stepping method.
  *
  * @param[in] time_steps Vector containing all the time steps. The time steps
--- a/include/core/simulation_control.h
+++ b/include/core/simulation_control.h
@@ -139,7 +139,7 @@ protected:
   /// BDF coefficients used for time-stepping methods
   Vector<double> bdf_coefs;
 
-   /**
+  /**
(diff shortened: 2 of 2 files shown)


include/core/specific_heat_model.h
```diff
@@ -237,43 +240,47 @@ class PhaseChangeSpecificHeat : public SpecificHeatModel
     std::vector<double> time_steps_vector =
       simulation_control->get_time_steps_vector();
 
+    // If change between the temperature is insufficient, backtrack to the
+    // first order implementation
```
hepap, line 244:
```suggestion
```

include/core/utilities.h
```diff
@@ -207,6 +207,22 @@ sgn(T val)
   return (static_cast<T>(0) < val) - (val < static_cast<T>(0));
 }
 
+
+/**
+ * @brief Used to clip a properties between a lower and a higher value
+ * @tparam T val argument's type
```
hepap, line 213:
```suggestion
 * @tparam T argument's type
```

hepap (approved):
Nice optimization and it looks cleaner :)

include/core/bdf.h
```diff
@@ -82,8 +55,9 @@ bdf_coefficients(unsigned int order, const std::vector<double> time_steps);
  * @note At the moment, the highest implemented scheme is @p bdf3 (@p order=3).
  */
 Vector<double>
-bdf_coefficients(const Parameters::SimulationControl::TimeSteppingMethod method,
-                 const std::vector<double> time_steps);
+calculate_bdf_coefficients(
+  const Parameters::SimulationControl::TimeSteppingMethod method,
+  const std::vector<double>                               time_steps);
```
AmishgaAlphonius, line 61: lines 40-43 could be removed since it is repeated at lines 49-52. I think it is better placed after the `time_steps` parameter

AmishgaAlphonius (approved):
I'm excited to see the changes in runtime after this PR is merge! :D
(I only have a few minor comments:) )

PR #1043, opened 2024-02-26
Modify the periodic particles QCM test after modification in deal.II

acdaigneault:
# Description of the problem

- deal.II had an update on the tolerance for is_inside_unit_cell() check. It resulted in a slightly different results of the  Periodic Particles QCM test, but that made crash the test. (PR #16691 of deal.II)
- Only the position in z gave slightly different results that may be from the curvature on the cylinder.
- The QCM in periodic mechanism has not affected.

# Description of the solution
- The domain was changed to a rectangular pipe so the curvature will not disturb the particle path in z.

# Addition
- Some difficulties were encounter while modifying the new test with the insertion list. This was related to the mandatory diameter list. 
- I changed the Assert to AssertThrow for unequal lists (x positions vs diameters)
- I added a default diameter list building with the diameter in particle parameter when no diameter list are provided since most case are mono-disperse.


blaisb (approved):
Nothing to say here, very good work @acdaigneault . Thank you!

PR #1044, opened 2024-02-27
Fix dem memory growth

blaisb:
# Description of the problem

- The first level of the fine search candidates for particle-particle, particle-wall and particle-floating walls was never cleared even when the value of the unordered_map became an empty unordered_map. In simulations with load balancing or particles that were moving significantly, this could potentially lead to a scenario where the size of the first level of the fine search candidate became equal to the number of total particles in the simulation, potentially leading to a crash.

# Description of the solution

- This is fixed by clearing the first level of the fine search data structure when the number of potential contactee is zero.

# How Has This Been Tested?

- All unit tests should pass once @acdaigneault PR has been merged.

# Documentation

- Changelog has been modified.
-


OGaboriault (approved):
Excited to see how this affect performance!

PR #1049, opened 2024-02-29
Add matrix-free full GLS operator

lpsaavedra:
# Description of the problem

Only the SUPG/PSPG stabilization was supported by the `lethe-fluid-matrix-free`. This PR adds the GLS steady and transient operators for the matrix-free solver.

# Description of the solution

The operators were coded in a similar fashion as the SUPG/PSPG operators adding the new GLS terms, including the LSIC term. 

# How Has This Been Tested?

The results of several tests were compared with the matrix-based operator leading to the same results. For completeness a TGV test using the operator is added:

- [X] `applications_tests/lethe-fluid-matrix-free/tgv_fe1_ilu_gls.prm`

# Documentation

Only one file was updated:
- [X] `source/parameters/cfd/stabilization.rst`


peterrum (commented):
Looks good to me. However, I am wondering whether it is time to merge the classes. There are now 4 classes: `NavierStokesSUPGPSPGOperator`, `NavierStokesTransientSUPGPSPGOperator`, `NavierStokesGLSOperator`, and `NavierStokesTransientGLSOperator`. Wouldn't it be better to only have one class which can be configured via 2 parameters: `full_gls` and `transient`. The performance penalty would be minor. @lpsaavedra What do you think?

blaisb:
@lpsaavedra @peterrum . I think that's a good idea. It could even be a template parameter and there the cost would be zero...
Anyway right now the majority of the cost is in the set-up of the relaxation parameter and eigenvalue estsimation

lpsaavedra:
Yes, it is definitely a good idea. Let me see what I can do...

blaisb (approved):

PR #1051, opened 2024-02-29
Add compressibility to heat transfer

AmishgaAlphonius:
# Description of the problem

- The previously implemented weak compressibility () was not working with heat transfer physic.

# Description of the solution

- Missing pressure field was added to scratch data.

# How Has This Been Tested?

- The new feature has been tested on an impinging jet case.
- Two application tests were added to test this feature : 
  - `applications_tests/lethe-fluid/heat_transfer_weakly_compressible_flow.prm` (`.output`)
  - `applications_tests/lethe-fluid/heat_transfer_vof_weakly_compressible_flow.prm` (`.output`)

# Future changes

- Pressure field should be added to other scratch data using the density. To generalize, all `fields` should be checked if required in all scratch data that calculates physical properties.

# Comments

- Other small typo corrections were made.

### DIFF
--- a/include/solvers/heat_transfer_scratch_data.h
+++ b/include/solvers/heat_transfer_scratch_data.h
@@ -284,29 +284,39 @@ public:
       }
   }
 
-  /** @brief Reinitialize the velocity, calculated by the fluid dynamics while also taking into account ALE
+  /**
+   * @brief Reinitialize cell's velocity and pressure values with values
+   * calculated in the Fluid Dynamics (FD) physic while also taking into account
+   * ALE for the velocity.
    *
-   * @tparam VectorType The Vector type used for the solvers
+   * @tparam VectorType Type of vector used for storing fluid dynamics
+   * solutions.
    *
-   * @param cell The cell for which the velocity is reinitialized
-   * This cell must be compatible with the Fluid Dynamics FE
+   * @param[in] cell Pointer to an active cell for which the velocity and
+   * pressure are reinitialized.This cell must be compatible with the FD
+   * FiniteElement.
    *
-   * @param current_solution The present value of the solution for [u,p]
+   * @param[in] current_solution Solution vector from FD containing velocity
+   * components and pressure values.
    *
-   * @param ale The ALE parameters which include the ALE function
+   * @param[in] ale The ALE parameters which include the ALE function
    *
    */
 
   template <typename VectorType>
   void
-  reinit_velocity(const typename DoFHandler<dim>::active_cell_iterator &cell,
-                  const VectorType           &current_solution,
-                  const Parameters::ALE<dim> &ale)
+  reinit_fluid_dynamics(
+    const typename DoFHandler<dim>::active_cell_iterator &cell,
+    const VectorType                                     &current_solution,
+    const Parameters::ALE<dim>                           &ale)
   {
     this->fe_values_fd.reinit(cell);
 
     this->fe_values_fd[velocities].get_function_values(current_solution,
                                                        velocity_values);
+    if (!this->properties_manager.density_is_constant())
+      this->fe_values_fd[pressure].get_function_values(current_solution,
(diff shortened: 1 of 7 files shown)


source/solvers/heat_transfer.cc
```diff
@@ -454,8 +454,10 @@ HeatTransfer<dim>::assemble_local_system_matrix(
   const DoFHandler<dim> *dof_handler_fluid =
     multiphysics->get_dof_handler(PhysicsID::fluid_dynamics);
 
-  typename DoFHandler<dim>::active_cell_iterator velocity_cell(
-    &(*triangulation), cell->level(), cell->index(), dof_handler_fluid);
+  typename DoFHandler<dim>::active_cell_iterator fd_cell(&(*triangulation),
```
voferreira, line 457: Is fd meant to mean "fluid dynamics"? Maybe I would use a more meaningful name here.

voferreira (approved):
Very nice! The only thing I would change is the name of the dof handler. Maybe just use "fluid_dynamics_cell"?

include/solvers/heat_transfer_scratch_data.h
```diff
@@ -341,6 +345,29 @@ class HeatTransferScratchData
...
+   */
+  template <typename VectorType>
+  void
+  reinit_pressure(const typename DoFHandler<dim>::active_cell_iterator &cell,
+                  const VectorType &current_solution)
+  {
+    this->fe_values_fd.reinit(cell);
+    this->fe_values_fd[pressure].get_function_values(current_solution,
+                                                     pressure_values);
+  }
```
blaisb, line 369: Wait why is this a second FE_values reinit? This is very expensive as a call. It would be better to gather pressure when we gather the fluid dynamics instead of calling reinit when we need both...

source/solvers/heat_transfer.cc
```diff
@@ -486,20 +488,32 @@ HeatTransfer<dim>::assemble_local_system_matrix(
...
 
+  if (!this->simulation_parameters.physical_properties_manager
+         .density_is_constant())
+    {
+      if (!multiphysics->fluid_dynamics_is_block())
+        scratch_data.reinit_pressure(
+          fd_cell, *multiphysics->get_solution(PhysicsID::fluid_dynamics));
+      else
+        scratch_data.reinit_pressure(fd_cell,
+                                     *multiphysics->get_block_solution(
```
blaisb, line 513: It's better to  move all the reinits to a single reinit which would be reinit_fluid_dynamics,. Otherwise the call to fe_values reinit will get prohibitively expensive..

blaisb (changes requested):
I don't think making a second reinit function is a good pattern here. It would be a better idea to reinit everything related to a single physics in a single function. Otherwise you will have multiple calls to fe_values.reinit and that's expensive.

source/solvers/heat_transfer.cc
blaisb: I don't mind the fd here to be honest. All the fluid dynamics functions have fd, heat transfers have ht, etc. Right now this is a standard nomenclature in the code, so I would not change that unless we change it everywhere and that owuld be another PR in itself.

source/solvers/heat_transfer.cc
AmishgaAlphonius: You're right, I'll do so tomorrow!
I'll also check if it is more costly to add an `if` inside to get the pressure values or if the `get_function_values()` is not to bad. If it is not too bad (which I don't think is the case), I'll leave it without the `if` is that okay ?

blaisb: Yup sounds good

blaisb (approved):

blaisb:
Will merge after CI @AmishgaAlphonius

mivaia (approved):
Nice job Amish! Everything fine on my side.

PR #1059, opened 2024-03-08
Update lpbf benchmark example prm

hepap:
# Description of the problem

- Prm of the lpbf example needed to be updated

# Description of the solution

- Update the prm as it is now for the benchmark. However, it is not optimal.
- Add the 1D and 3D ones

# Comments

- If you want to simulate without marangoni, just put the temperature-driven surface tension gradient to zero.


blaisb (commented):
What is the regular benchmark file now that you have a 1D-2D-3D folder?

AmishgaAlphonius (approved):
same comment as Bruno otherwise I have nothing to add :)

hepap:
I added duplicate by mistake! It is fixed now.

The .prm in each folder are the regular one for the respective dimension.

PR #1064, opened 2024-03-13
Use the eigenvalue estimation of PreconditionRelaxation

lpsaavedra:
# Description of the problem

The eigenvalue estimation for the smoother of the multigrid preconditioners was done through a `PreconditionChebyshev`.

# Description of the solution

Due to recent changes in deal.II (https://github.com/dealii/dealii/pull/16742), now we are able to do it using the `PreconditionRelaxation` directly. This allows us to remove the estimate omega function since it is done internally by the smoother if the relaxation parameter is set to 1. This allows also to remove the `degree` parameter that was only used in the case of Chebyshev.


peterrum (approved):

blaisb (approved):
Wow I love this, this is so much cleaner and easier to read now.
@lpsaavedra Could you add a changelog entry?

peterrum:
I guess we need to drop deal.II 9.5 CI?

lpsaavedra:
I will check the deal.II version as we did here: https://github.com/lethe-cfd/lethe/blob/82095bab9286844d9f1cec0dab67cf3575c1791a/source/solvers/mf_navier_stokes.cc#L651

PR #1067, opened 2024-03-15
Remove parameter qmapping_all

peterrum:
since it is not used anyway according to https://www.dealii.org/developer/doxygen/deal.II/classMappingQ.html#a9b34a7f2ec00aa65e73402865ac2887d


blaisb (commented):
Most likely it is still present in some tests though, so we will have to fix the PRM files that break with respect to this.

Can you change the documentation in the ".rst" file accordingly?

blaisb (approved):

PR #1076, opened 2024-03-21
Add information linear and non-linear solver outputs

hepap:
# Description of the problem

- The console output of the non-linear solver in the NS solver was showing the the norms of the newton update without distinction between the velocity and pressure contributions.
- The final residual norm of the linear solver was not outputted, only the tolerance.

# Description of the solution

- Add the distinction between the velocity and pressure contributions in the non-linear solver output.
- Add the final value of the linear residual when verbosity is at verbose. If it is at extra verbose, the solver output the residual for each iteration (already implemented but not documented).

# How Has This Been Tested?

A unit test was added.

- [x] vector_problem.cc

# Documentation

Add a comment for the extra verbose verbosity of the linear solver.

- [x] linear_solver_control.rst


source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+               j++)
+            {
+              double dof_newton_update = newton_update[*j];
+
+              local_sum += dof_newton_update * dof_newton_update;
+
+              if (dof_newton_update > local_max)
+                {
+                  local_max = dof_newton_update;
+                }
```
blaisb, line 2522: maybe it's better to use std::max here no?
like local_max = std::max(local_max,dof_newton_update);

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+           j++)
+        {
+          double dof_newton_update = newton_update[*j];
+
+          local_sum += dof_newton_update * dof_newton_update;
+
+          if (dof_newton_update > local_max)
+            {
+              local_max = dof_newton_update;
+            }
```
blaisb, line 2545: same as above :)

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+                  << newton_update.block(0).l2_norm() << std::setw(6)
+                  << "\t||du||_Linfty = "
+                  << std::setprecision(display_precision)
+                  << newton_update.block(0).linfty_norm() << std::endl;
+      this->pcout << std::setprecision(display_precision)
+                  << "\t||dp||_L2 = " << std::setw(6)
+                  << newton_update.block(1).l2_norm() << std::setw(6)
+                  << "\t||dp||_Linfty = "
+                  << std::setprecision(display_precision)
+                  << newton_update.block(1).linfty_norm() << std::endl;
```
blaisb, line 2578: It's funny how much easier it is with the blocks heh ?%

tests/core/vector_problem.cc
```diff
@@ -0,0 +1,173 @@
...
+      for (auto j = index_set_velocity[d].begin();
+           j != index_set_velocity[d].end();
+           j++)
+        {
+          correction_norm += dummy_solution[*j] * dummy_solution[*j];
+
+          if (dummy_solution[*j] > max_correction)
+            {
+              max_correction = dummy_solution[*j];
+            }
```
blaisb, line 111: std::max here also

tests/core/vector_problem.cc
```diff
@@ -0,0 +1,173 @@
...
+  ComponentMask pressure_mask = fe.component_mask(pressure);
+
+
+  std::vector<IndexSet> index_set_velocity =
+    DoFTools::locally_owned_dofs_per_component(dof_handler, velocity_mask);
+  std::vector<IndexSet> index_set_pressure =
+    DoFTools::locally_owned_dofs_per_component(dof_handler, pressure_mask);
+
+  double correction_norm = 0.0;
+  double max_correction  = DBL_MIN;
```
blaisb, line 98: should it not be instead -DBL_MAX ?

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+      ComponentMask velocity_mask = fe->component_mask(velocities);
+      ComponentMask pressure_mask = fe->component_mask(pressure);
+
+      const std::vector<IndexSet> index_set_velocity =
+        DoFTools::locally_owned_dofs_per_component(dof_handler, velocity_mask);
+      const std::vector<IndexSet> index_set_pressure =
+        DoFTools::locally_owned_dofs_per_component(dof_handler, pressure_mask);
+
+      double local_sum = 0.0;
+      double local_max = DBL_MIN;
```
blaisb, line 2507:
```suggestion
      double local_max = -std::numeric_limits<double>::max();
```
is a better way to write this. The max you want it to start at -max, this way 0 can still be a max :)

tests/core/non_linear_test_system_01.h
```diff
@@ -126,6 +126,15 @@ class NonLinearProblemTestClass : public PhysicsSolver<Vector<double>>
...
     return newton_update;
   };
+  virtual void
+  output_newton_update_norms(const unsigned int display_precision) override
+  {
+    this->pcout << std::setprecision(display_precision)
+                << "\t||dx||_L2 = " << std::setw(6) << newton_update.l2_norm()
+                << std::setw(6)
+                << "\t||dx||_Linfty = " << std::setprecision(display_precision)
+                << newton_update.linfty_norm() << std::endl;
```
blaisb, line 136: Should the test output this also? Could be interesting if it is portable.

blaisb (changes requested):
Very few comments, amazing work @hepap

source/solvers/navier_stokes_base.cc
PierreLaurentinCS: One can even use std::numeric_limits<double>::lowest() to avoid the -max notation

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+            }
+        }
+
+      double global_velocity_l2_norm =
+        std::sqrt(Utilities::MPI::sum(local_sum, this->mpi_communicator));
+      double global_velocity_linfty_norm =
+        Utilities::MPI::max(local_max, this->mpi_communicator);
+
+      local_sum = 0.0;
+      local_max = DBL_MIN;
```
PierreLaurentinCS, line 2532:
```suggestion
      local_max = -std::numeric_limits<double>::max();
```
just in case

PierreLaurentinCS (commented):
Amazing feature, and very few comments. Good job !

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+            {
+              double dof_newton_update = newton_update[*j];
+
+              local_sum += dof_newton_update * dof_newton_update;
+
+              if (dof_newton_update > local_max)
+                {
+                  local_max = dof_newton_update;
+                }
+            }
```
AmishgaAlphonius, line 2523: These loops on index_sets could be inlined if you wish :)

source/solvers/navier_stokes_base.cc
```diff
@@ -2483,6 +2483,102 @@ NavierStokesBase<dim, VectorType, DofsType>::
...
+                  << "\t||du||_Linfty = "
+                  << std::setprecision(display_precision)
+                  << global_velocity_linfty_norm << std::endl;
+      this->pcout << std::setprecision(display_precision)
+                  << "\t||dp||_L2 = " << std::setw(6) << global_pressure_l2_norm
+                  << std::setw(6) << "\t||dp||_Linfty = "
+                  << std::setprecision(display_precision)
+                  << global_pressure_linfty_norm << std::endl;
+    }
+  if constexpr (std::is_same_v<VectorType, GlobalBlockVectorType>)
```
AmishgaAlphonius, line 2565:
```suggestion
  else if constexpr (std::is_same_v<VectorType, GlobalBlockVectorType>)
```

AmishgaAlphonius (approved):
Great work ! Thanks ! :)
I only have a few minor comments.

source/solvers/navier_stokes_base.cc
blaisb: I don't see the point of the else here.  it's actually better to have an undefined behavior than have an else taht could maybe not be an else (other than if you throw at the end something in an else)

hepap:
I think it is ready to merge if the tests pass! The output was add for CH as well :)

blaisb (approved):
Once CI is fixed, this is ready for a merge IMO

hepap:
Should be alright now!

PR #1074, opened 2024-03-19
Dem volume insertion box parameters

OGaboriault:
# Description of the problem

- The insertion box for the volume insertion was being defined by a large number of parameters (9 to be exact)  

# Description of the solution

- The box is defined by one parameter and the order of direction by an other. 

# Documentation

- The documentation has been updated accordingly.

# Comments

- The list insertion parameters are now parsed with "convert_string_to_vector", which is cleaner to look at. 
- I have made (with the help of Amishga) a bash script to parse every prm file and change the old parameters into the new ones. We could do something with it... maybe?

PR #1078, opened 2024-03-28
Dem volume insertion initial velocities

OGaboriault:
# Description of the problem

- This is a follow up from Victor's comment from PR [#1074

The initial velocity and initial angular velocity are now define with one parameter each. 

Really quick PR. 

No example or application test (except for  one) was using those parameters.

### DIFF
--- a/source/dem/insertion.cc
+++ b/source/dem/insertion.cc
@@ -79,12 +79,12 @@ Insertion<dim>::assign_particle_properties(
                    ->particle_sizes[particle_counter]);
       double density =
         physical_properties.density_particle[current_inserting_particle_type];
-      double vel_x        = dem_parameters.insertion_info.vel_x;
-      double vel_y        = dem_parameters.insertion_info.vel_y;
-      double vel_z        = dem_parameters.insertion_info.vel_z;
-      double omega_x      = dem_parameters.insertion_info.omega_x;
-      double omega_y      = dem_parameters.insertion_info.omega_y;
-      double omega_z      = dem_parameters.insertion_info.omega_z;
+      double vel_x        = dem_parameters.insertion_info.initial_vel[0];
+      double vel_y        = dem_parameters.insertion_info.initial_vel[1];
+      double vel_z        = dem_parameters.insertion_info.initial_vel[2];
+      double omega_x      = dem_parameters.insertion_info.initial_ang_vel[0];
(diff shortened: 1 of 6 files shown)


blaisb (approved):
Are you sure there are no examples that use this ? I am thinking about the particle bouncing on a wall one. Otherwise, all good for me.

source/dem/insertion.cc
```diff
@@ -79,12 +79,12 @@ Insertion<dim>::assign_particle_properties(
...
-      double vel_x        = dem_parameters.insertion_info.vel_x;
-      double vel_y        = dem_parameters.insertion_info.vel_y;
-      double vel_z        = dem_parameters.insertion_info.vel_z;
-      double omega_x      = dem_parameters.insertion_info.omega_x;
-      double omega_y      = dem_parameters.insertion_info.omega_y;
-      double omega_z      = dem_parameters.insertion_info.omega_z;
+      double vel_x        = dem_parameters.insertion_info.initial_vel[0];
+      double vel_y        = dem_parameters.insertion_info.initial_vel[1];
+      double vel_z        = dem_parameters.insertion_info.initial_vel[2];
+      double omega_x      = dem_parameters.insertion_info.initial_ang_vel[0];
```
voferreira, line 85: I like the new parameter name, but I think I would rather keep `dem_parameters.insertion_info.omega[0]`, because the rest of the code uses omega. Otherwise it is all clear to me :)

voferreira: Ah, and you don't need `initial`, since this is insertion info

voferreira (approved):
Good job! Only one comment about the insertion_info, but nothing special.

blaisb:
Ohh shoot, did not see the comments before the merge. You can fix them in a secondary quick PR @OGaboriault

source/dem/insertion.cc
OGaboriault: I think the <initial> is useful. This way, there's not confusion with the velocity at a given time step. 

I can do initial_omega. I agree, omega is more coherent with the rest of code.

What do you think?

voferreira: In general, I tend to prefer long meaningful names instead of confusing abbreviations, but in this case I found it slightly redundant, hence, not necessary. This is not a strong opinion, though. So this is really up to you :)

PR #1079, opened 2024-03-29
Dem insertion naming refactor

OGaboriault:
# Description of the problem

- The naming scheme for the insertion files were not coherent with the rest of the code. (i.e., `integrator_...` , `particle_wall_...` , etc.). This was annoying when trying to find a specific insertion file. Now they are next to each other.  

# Description of the solution

- The files are now named like this: `insertion_method.cc` or `insertion_method.h`

# How Has This Been Tested?

- Tests are running on my machine with these changes. 

# Comments

- Anything that you want to add up that didn't fit in any other category can be written here

### DIFF
--- a/include/dem/insertion_file.h
+++ b/include/dem/insertion_file.h
@@ -24,19 +24,19 @@
 
 #include <deal.II/particles/particle_handler.h>
 
-#ifndef lethe_file_insertion_h
-#  define lethe_file_insertion_h
+#ifndef lethe_insertion_file_h
+#  define lethe_insertion_file_h
 
 template <int dim>
-class FileInsertion : public Insertion<dim>
+class InsertionFile : public Insertion<dim>
 {
 public:
-  FileInsertion(const DEMSolverParameters<dim> &dem_parameters,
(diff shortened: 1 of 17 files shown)


blaisb (approved):
I had not noticed that, very cool change, makes thing so much cleaner now.

include/dem/insertion_file.h
```diff
@@ -24,19 +24,19 @@
...
 {
 public:
-  FileInsertion(const DEMSolverParameters<dim> &dem_parameters,
+  InsertionFile(const DEMSolverParameters<dim> &dem_parameters,
                 const std::vector<std::shared_ptr<Distribution>>
                   &distribution_object_container);
 
   /**
-   * @brief The FileInsertion class inserts particles using data stored in a file.
+   * @brief The InsertionFile class inserts particles using data stored in a file.
```
acdaigneault, line 39: Not specific for this PR, but we should avoid to write "The ClassName class do..." for the Doxygen doc. If you plan to apply some comment before merge of this PR, I would remove it (so just "Insert particles..."), so it will be done. :)

PR #1081, opened 2024-03-30
replace reduced pressure with kinematic pressure

AmishgaAlphonius:
# Documentation

- kinematic pressure: pressure/reference_state_density
- reduced pressure: pressure/critical_pressure

The "kinematic pressure" in the documentation on incompressible Navier-Stokes equations was referred to as "reduced pressure"; this has been replaced.


blaisb (approved):

blaisb:
Kinematic pressure is indeed the right way to say it

PR #1082, opened 2024-03-30
Dem fixing tangential damping

OGaboriault:
# Description of the problem

- The particle-wall nonlinear model wasn't following the same coulomb-limit as the particle-particle hertz-limit-overlap model

# Comments

- Tests outputs had to be updated since the position of the particles are slightly different.


blaisb (approved):
Well done. Will merge after CI

PR #1084, opened 2024-04-01
Update restart files, moving receptable 

OGaboriault:
# Description of the problem

- Dealii made some changes recently forcing us to regenerate our restart files for our test.
- The CI uses an older version of Boost, thus it's not every machine who can generate compatible file with the CI since Boost is partially backwards compatible (Files generated on an newer version won't work on an older one)  

# Comments

- Dealii was pull with master and compile before generating the restart files.
- I'm not sure why the output had to be changed... In the last few months, the insertion method and the contact force calculation of some model were changed, so maybe the old restart files were made before these two changes.


blaisb:
@OGaboriault Can you rebase the branch it has weird conflicts?

OGaboriault:
Rebase done, let's see if this works

blaisb:
I think it's really a question of which boost version is being used...

blaisb:
@OGaboriault can you try to remake the tests files by using a deal.II version compiled in which the boost was forced to be the included one? If you don't know how to do it it's ok, I will do it when I get back

blaisb (approved):

PR #1086, opened 2024-04-03
Quick fix, OMP_NUM_THREADS = 1

OGaboriault:
# Description of the problem

- On some clusters, it is recommend to specify OMP_NUM_THREADS = 1, otherwise Trilinos can start using multithreading.

# Description of the solution

- I added a line in the .dealii file with a small comment.


hepap (approved):

blaisb (approved):
On some clusters this is not mandatory but it is still a good practice

PR #1088, opened 2024-04-05
Explicitly set max num of threads

lpsaavedra:
# Description of the problem

The third argument of the [`MPI_InitFinalize(...)` ](https://www.dealii.org/developer/doxygen/deal.II/classUtilities_1_1MPI_1_1MPI__InitFinalize.html#a71954ff48c33b1101b8a4d64d9c685db) call in all applications in Lethe was set to `numbers::invalid_unsigned_int`; if this is the case, then the number of threads is determined automatically according to the system and set via `MultithreadInfo::set_thread_limit(...)`.

Due to confusions this week in the lab, I though that it would be a good idea to set it explicitly to 1, which guarantees that it will start as many MPI processes per node as there are cores on each node avoiding multi-threading.

For the main applications, there was a call `MultithreadInfo::set_thread_limit(1)` in the respective `solve()` function, however, according to deal.II, it only works if it is called before any threads are created, therefore, it is safer to call it at the beginning of `main()` instead.

# Description of the solution

Removed `MultithreadInfo::set_thread_limit(1)` from the applications and set the third argument of `MPI_InitFinalize(...)` to 1 in all cases across the library.

# How Has This Been Tested?

All the tests pass without any issue.

### DIFF
--- a/applications/initial_conditions/initial_conditions.cc
+++ b/applications/initial_conditions/initial_conditions.cc
@@ -118,8 +118,7 @@ main(int argc, char *argv[])
 
 
 
-      Utilities::MPI::MPI_InitFinalize mpi_initialization(
-        argc, argv, 1);
+      Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
 
       const Parameters::SizeOfSubsections size_of_subsections =
         Parameters::get_size_of_subsections(argv[1]);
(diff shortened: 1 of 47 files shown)


blaisb (approved):
Sure, good idea, this will be more unifrom. I hope this prevents Trilinos from starting threads also :)

PR #1089, opened 2024-04-06
Pressure constraint gmg

blaisb:
# Description of the problem

- When the mesh is not homogenous and the domain is closed (e.g. only dirichlet or periodic BCs), the coarse grid solver of the matrix-free GCMG or LSMG does not work. This is because the problem is ill-posed due to the pressure constant.

# Description of the solution

- On the coarse grid level, fix a pressure DOF to be zero by adding a single constraint. This ensures that the solution works.

# How Has This Been Tested?

- I changed the example to use this feature, but I also need to add an application test. Will do so in this PR.

# Documentation

- New parameter is updated and documented.

# Comments

- I am glad this is fixed.


blaisb:
@lpsaavedra we should merge this after your PR is merged. We just need to regenerate the output. I made it from my mac and it seems to lead to problems

lpsaavedra (approved):

PR #1090, opened 2024-04-09
Support flow dynamic control for matrix-free

lpsaavedra:
# Description of the problem

Flow dynamic control was not being used in the matrix-free operators, which was preventing us from using it to simulate cases like the periodic hills.

# Description of the solution

The operators now consider both the source term function and the beta term; the latter only if dynamic flow control is enabled.

# How Has This Been Tested?

- The results for the first couple of iterations of the periodic hills case on my machine are the same as in the matrix-based solver. 
- A simple case analogous to the one available for the MB application was included: `applications_tests/lethe-fluid-matrix-free/poiseuille3d_flow_control.prm`.

# Comments

Now we can simulate the periodic hills with `lethe-fluid-matrix-free` :)


blaisb (approved):
One minor comment, that's all. This is amaaaaazing :). Major periodic hills simulations incoming :)

blaisb (approved):

blaisb:
Will merge after CI :)

PR #1092, opened 2024-04-15
Flow rates postprocessing

PierreLaurentinCS:
# Description of the problem

- The previous flow rate post processing tool would add a line to the `.dat` file for each defined boundary (in the boundary condition subsection of the `.prm` file) for every time iteration. This is not the expected output as one would expect a single line to be added for each time step with the flow rates of every boundary.
- Also, the values of the flow rates were added with a 12 digits precision. Instead, a 4 digits precision in **scientific notation** was chosen, which seems like a reasonable precision. 

# Description of the solution

- The value of time is now added outside of the loop on the boundary conditions.

# Comments
- At the moment there is no unified output format : with/without scientific notation, different numbers of digits. This could be the work of a future PR.

### DIFF
--- a/source/solvers/navier_stokes_base.cc
+++ b/source/solvers/navier_stokes_base.cc
@@ -1386,7 +1386,18 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
   // Calculate flow rate at every boundary
   if (this->simulation_parameters.post_processing.calculate_flow_rate)
     {
+      this->flow_rate_table.add_value("time",
+                                      simulation_control->get_current_time());
+      this->flow_rate_table.set_scientific("time", true);
+
       TimerOutput::Scope t(this->computing_timer, "flow_rate_calculation");
+
+      if (this->simulation_parameters.post_processing.verbosity ==
+          Parameters::Verbosity::verbose)
+        {
+          announce_string(this->pcout, "Flow rates");
+        }
+
       for (unsigned int boundary_id = 0;
            boundary_id < simulation_parameters.boundary_conditions.size;
            ++boundary_id)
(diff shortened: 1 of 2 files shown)


source/solvers/navier_stokes_base.cc
```diff
@@ -1424,13 +1437,13 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
             simulation_parameters.post_processing.flow_rate_output_name +
             ".dat";
           std::ofstream output(filename.c_str());
-          flow_rate_table.set_precision("time", 12);
+          flow_rate_table.set_precision("time", 4);
```
blaisb, line 1440: I'm not sure this is a good idea. For the other variables we use 12 significant digits. Since this is always written to a data file, I think it's a much better idea to keep as many digits as possible. Using scientific notation is a good idea however.

source/solvers/navier_stokes_base.cc
```diff
@@ -1424,13 +1437,13 @@ NavierStokesBase<dim, VectorType, DofsType>::postprocess_fd(bool firstIter)
...
           for (unsigned int boundary_id = 0;
                boundary_id < simulation_parameters.boundary_conditions.size;
                ++boundary_id)
             flow_rate_table.set_precision("flow-rate-" +
                                             std::to_string(boundary_id),
```
blaisb, line 1445: Use Utilities::int_to_string instead
and set the padding manually to 2, because we will never have more than 99 boundary conditions (it makes no sense)

source/solvers/navier_stokes_base.cc
blaisb: I think I misunderstood. Here is it the one you are printing on the screen or the one your are printing on the data file? For the screen on e I think 4 digit scientific is ok. For the data file one, I think it's not a good idea to print just 4 digits.

blaisb (changes requested):
A few comments regarding formatting.

source/solvers/navier_stokes_base.cc
blaisb: For log outputted to the terminal you can also youse simulation_control->get_log_precision() which is meant exactly for that.

blaisb:
@PierreLaurentinCS please also suggest two other reviewers

source/solvers/navier_stokes_base.cc
PierreLaurentinCS: Agreed, will keep the scientific notation but with 12 digits for the data file. For the screen display I'll keep 4 digits in scientific for better readability

blaisb (approved):

blaisb:
@PierreLaurentinCS Can you rebase for the changelog? Then I will be able to merge

oguevremont (approved):

PR #1093, opened 2024-04-16
Fix restart transient with variable time step

blaisb:
# Description of the problem

- Transient simulation with adaptive time-stepping were not restarted correctly (their time-step was not fixed adequately)

# Description of the solution

- The issue was that the time-step value was not reset correctly when the time-step values vector was read in the checkpoint

# How Has This Been Tested?

- Added. a new test for this feature

# Documentation

- Changelog added

# Comments

-  This class uses an old checkpointing mechanism that is not boost serialize. I will need to fix it.  It will be a good opportunity to streamline the class afterwards also and make it simpler to understand. Maybe reduce the tree of possibilities in term of classes.


source/core/simulation_control.cc
```diff
@@ -167,6 +167,9 @@ SimulationControl::read(std::string prefix)
   input >> buffer >> CFL;
   input >> buffer >> current_time;
   input >> buffer >> iteration_number;
+
+  // Fix time step to be the last time_step that was used
+  time_step = time_step_vector[0];
```
PierreLaurentinCS, line 172: It is crazy how one line can fix this problem so easily

PierreLaurentinCS (commented):
Great job, I've got nothing to say expect for .prm formating

hepap (approved):
I'm still looking for why we have a difference in the residuals for NS-VOF cases, but I think this change can still be merged in the meanwhile :)

PR #1095, opened 2024-04-17
CHNS Free Energy Postprocessing

PierreLaurentinCS:
# Description of the problem

The Cahn-Hilliard equations are energy based equations for solving multiphase flows : the interface is the result of two energies in competition, bulk and interface energy. The system, which starts from a non-equilibrium state, lowers its total energy (bulk + interface) to reach equilibrium. In order to visualize this process I added the energy computation into the postprocessing section.
Very minor : added the computation of the volume of each phases.

# Description of the solution

- New functions `calculate_phase_energy()` and `write_phase_energy` were implemented as well as new parameter entries in the postprocess subsection : `compute phase energy` (boolean) and `phase energy name`

# How Has This Been Tested?

- A new test was added : `cahn_hilliard_navier_stokes_energy_postprocessing`

# Documentation

- Postprocessing documentation was updated to include this new feature

- File modified : `post_processing.rst`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Added
 
-- MINOR Added computation of the energies involved in the Cahn-Hilliard equations and updated related documentation. [#1095](https://github.com/lethe-cfd/lethe/pull/1095)
+- MINOR Added Cahn-Hilliard equations energy computation and output. The documentation was updated accordingly. [#1095](https://github.com/lethe-cfd/lethe/pull/1095)
 
 ## [Master] - 2024-04-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-04-17
+
+### Added
+
+- MINOR Added computation of the energies involved in the Cahn-Hilliard equations and updated related documentation. [#1094](future PR link)
+
 ## [Master] - 2024-04-16
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Added
 
-- MINOR Added computation of the energies involved in the Cahn-Hilliard equations and updated related documentation. [#1094](future PR link)
+- MINOR Added computation of the energies involved in the Cahn-Hilliard equations and updated related documentation. [#1095](https://github.com/lethe-cfd/lethe/pull/1095)
 
 ## [Master] - 2024-04-16
 
--- a/applications_tests/lethe-fluid/cahn_hilliard_navier_stokes_energy_postprocessing.prm
+++ b/applications_tests/lethe-fluid/cahn_hilliard_navier_stokes_energy_postprocessing.prm
@@ -1,0 +1,127 @@
+set dimension = 2
+
+subsection simulation control
+  set method           = bdf1
+  set time end         = 1
+  set time step        = 0.5
+  set output name      = out
+  set output frequency = 0
+end
+
+subsection initial conditions
+  set type = L2projection
+  subsection cahn hilliard
+    set Function constants  = nbr_refs=5
+    set Function expression = -tanh(sqrt(2*0.25)*(0.5 - sqrt(x*x+y*y))/(2^(-nbr_refs+1))); 0
+  end
+end
+
+subsection boundary conditions
+  set number         = 4
+  set time dependent = false
+  subsection bc 0
+    set id   = 0
+    set type = slip
+  end
+  subsection bc 1
+    set id   = 1
+    set type = slip
+  end
+  subsection bc 2
+    set id   = 2
+    set type = slip
+  end
+  subsection bc 3
+    set id   = 3
+    set type = slip
+  end
+end
+
+subsection boundary conditions cahn hilliard
+  set number         = 1
+  subsection bc 0
+    set id   = 0
+    set type = dirichlet
+    subsection phi
+      set Function constants  = nbr_refs=5
+      set Function expression = -tanh(sqrt(2*0.25)*(0.5 - sqrt(x*x+y*y))/(2^(-nbr_refs+1)))
+    end
+  end
+end
+
+subsection multiphysics
+  set fluid dynamics = true
+  set cahn hilliard  = true
+end
+
+subsection cahn hilliard
+  set potential smoothing coefficient = 0.0
+
+  subsection epsilon
+    set method = automatic
+  end
+end
+
+subsection physical properties
+  set number of fluids = 2
+  subsection fluid 0
+    set density             = 1000
+    set kinematic viscosity = 0.01
+  end
+  subsection fluid 1
+    set density             = 100
+    set kinematic viscosity = 0.01
+  end
+  set number of material interactions = 1
+  subsection material interaction 0
+    set type = fluid-fluid
+    subsection fluid-fluid interaction
+      set first fluid id              = 0
+      set second fluid id             = 1
+      set surface tension model       = constant
+      set surface tension coefficient = 24.5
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-04-17
+
+### Added
+
+- MINOR Added computation of the energies involved in the Cahn-Hilliard equations and updated related documentation. [#1095](https://github.com/lethe-cfd/lethe/pull/1095)
```
voferreira, line 10:
```suggestion
- MINOR Added Cahn-Hilliard equations energy computation and output. The documentation was updated accordingly. [#1095](https://github.com/lethe-cfd/lethe/pull/1095)
```

applications_tests/lethe-fluid/cahn_hilliard_navier_stokes_energy_postprocessing.prm
```diff
@@ -0,0 +1,127 @@
+set dimension = 2
+
```
voferreira, line 2: We don't add the subsection comments anymore? (Like, the ***************)

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -67,6 +67,8 @@ This subsection controls the post-processing other than the forces and torque on
     # Other Cahn-Hilliard postprocessing
     set calculate phase statistics       = false
     set phase statistics name            = phase_statistics
+    set calculate phase energy           = true
```
voferreira, line 70: True by default? It should be false, no?

voferreira: What do you think about removing the word "calculate" from the parameters? (can be another pr)

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -219,8 +221,28 @@ This subsection controls the post-processing other than the forces and torque on
 
 * ``mass conservation name``: name of the output file containing the mass of both fluids for VOF simulations. The default file name is ``mass_conservation_information``.
   
-* ``calculate phase statistics``: outputs phase statistics from the solution of the Cahn-Hilliard equations, including minimum, maximum, average, and standard deviation of the phase order parameter. This works only with the :doc:`cahn_hilliard` solver.
+* ``calculate phase statistics``: outputs phase statistics from the solution of the Cahn-Hilliard equations, including minimum, maximum, average, integral of the phase order parameter. Also includes the volume of each phase. This works only with the :doc:`cahn_hilliard` solver.
```
voferreira, line 224:
```suggestion
* ``calculate phase statistics``: outputs phase statistics from the solution of the Cahn-Hilliard equations, including minimum, maximum, average, and integral of the phase order parameter. The volume of each phase is also computed. This works only with the :doc:`cahn_hilliard` solver.
```

voferreira: " This works only with the :doc:`cahn_hilliard` solver." can be a warning

voferreira: I would rephrase it to: outputs Cahn-Hilliard phase statistics, including .............

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -219,8 +221,28 @@ This subsection controls the post-processing other than the forces and torque on
...
 
 * ``phase statistics name``: name of the output file containing phase order parameter statistics from Cahn-Hilliard simulations. The default file name is ``phase_statistics``.
 
+* ``calculate phase energy``:  outputs phase energies from the solution of the Cahn-Hilliard equations : bulk energy, interface energy and total energy. This works only with the :doc:`cahn_hilliard` solver. The energies are computed as follow:
```
voferreira, line 228:
```suggestion
* ``calculate phase energy``:  outputs phase energies from the solution of the Cahn-Hilliard equations, including: bulk energy, interface energy, and total energy. This works only with the :doc:`cahn_hilliard` solver. The energies are computed as:
```

voferreira: Same comment about the warning

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -219,8 +221,28 @@ This subsection controls the post-processing other than the forces and torque on
...
+      
+  .. math::
+
+     E_{interface} = \int_{\Omega} 0.5\epsilon^2|\nabla \phi |^2 \mathrm{d}\Omega 
+      
+  .. math::
+
+     E_{total} = E_{bulk} + E_{interface}  
+    
+  where :math:`\epsilon` is the numerical interface thickness. Note that these energies are not homogeneous to physical energies, but they give nonetheless a good insight on the system's evolution.
```
voferreira, line 242:
```suggestion
  where :math:`\epsilon` is the numerical interface thickness. Note that these energies are not homogeneous to physical energies. Nonetheless, they are a convenient way to track the system's evolution.
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -219,8 +221,28 @@ This subsection controls the post-processing other than the forces and torque on
...
+
+
+* ``phase energy name``: name of the output file containing phase energies from Cahn-Hilliard simulations. The default file name is ``phase_energy``.
```
voferreira, line 245: Is there an extension to this file?
Where is it stored? at the output folder?
Could you add this information here?

include/solvers/cahn_hilliard.h
```diff
@@ -402,11 +402,24 @@ class CahnHilliard : public AuxiliaryPhysics<dim, GlobalVectorType>
...
   /**
-   * @brief Writes the phase integral to an output file
+   * @brief Writes the phase statistics to an output file
    */
   void
   write_phase_statistics();
 
+  /**
+   *
+   * @return A pointer to an array of double containing the energies. First element is bulk energy, second is interface energy and third is total energy
```
voferreira, line 412: This is not a return function. It is void. Does it make sense to use return here?

source/core/parameters.cc
```diff
@@ -1828,6 +1828,18 @@ namespace Parameters
         "mass_conservation_information",
         Patterns::FileName(),
         "Name of mass conservation output file in VOF simulations");
+
+      prm.declare_entry(
+        "calculate phase energy",
+        "false",
```
voferreira, line 1834: So it is false by default. Can you please change it in the documentation?

source/core/parameters.cc
```diff
@@ -1828,6 +1828,18 @@ namespace Parameters
...
+        Patterns::Bool(),
+        "Enable calculation of phase energies : total energy, bulk energy and interface energy");
```
voferreira, line 1836:
```suggestion
        "Enable calculation of phase energies, including: total energy, bulk energy, and interface energy");
```

source/core/parameters.cc
```diff
@@ -1828,6 +1828,18 @@ namespace Parameters
...
+
+      prm.declare_entry(
+        "phase energy name",
+        "phase_energy",
+        Patterns::FileName(),
+        "Name of energy output file in Cahn-Hilliard simulations.");
```
voferreira, line 1842: Comment on the output path

voferreira (approved):
Good job :)
Very few and tiny suggestions but it looks good to me

applications_tests/lethe-fluid/cahn_hilliard_navier_stokes_energy_postprocessing.prm
PierreLaurentinCS: Agreed but some tests have it and some other don't, I would say that the test's .prm must be as minimalists as possible, hence the absence of subsection decorations. Can be discussed though

doc/source/parameters/cfd/post_processing.rst
PierreLaurentinCS: should definitely be false, you're right. For the calculate I think I'll leave it to another PR

blaisb (approved):
Nothing to add here, good work :).  I'm a bit alarmed by the growing number of post-processing parameters, but I presume this is normal as we increase the number of physics.

PR #1097, opened 2024-04-18
Partial fix for unresolved CFD-DEM restart

blaisb:
# Description of the problem

- Unresolved CFD-DEM restart had jump in the residual

# Description of the solution

- There was a bug where the percolation of the void fraction was taking place after the checkpointing which is wrong.

# How Has This Been Tested?

- The results of the tests won't be affected because the restart files have not been remade. 

# Comments

- The code needs major cleanup...


voferreira (approved):
No comments. It is pretty straightforward.

PR #1098, opened 2024-04-18
Refactor DEM to remove function pointers

blaisb:
# Description of the problem

- The DEM solver makes some use of function pointers something for which there is really no use for and which we do not use in the code.

# Description of the solution

- Remove the function pointer and replace them with straightforward if statements.

# Comments

- This is part of a long term effort to streamline even more the DEM and the CFD-DEM solver


include/dem/dem.h
```diff
@@ -142,6 +141,12 @@ class DEMSolver
 #  endif
 
 
+  /**
+   * Finds contact search steps for constant contact search method
```
voferreira:
```suggestion
   * Returns true if contact search time step
```

voferreira: The comment is just like the function below

source/dem/dem.cc
```diff
@@ -143,60 +143,6 @@ DEMSolver<dim>::DEMSolver(DEMSolverParameters<dim> dem_parameters)
     });
 #endif
 
-
```
voferreira, line 146: Shouldn't you call the is_contact_search_step() here?

source/dem/dem.cc
```diff
@@ -748,23 +732,40 @@ DEMSolver<dim>::check_load_balance_dynamic()
...
+    {
+      throw std::runtime_error(
+        "Specified contact detection method is not valid");
+      return false;
+    }
+}
+
 template <int dim>
 inline bool
 DEMSolver<dim>::check_contact_search_step_dynamic()
```
voferreira, line 756: Isn't this member not necessary? You could just have find_particle_contact_detection_step as a return function, couldn't you?

include/dem/dem.h
```diff
@@ -154,6 +156,12 @@ class DEMSolver
   inline bool
   check_contact_search_step_dynamic();
 
+  /**
+   * @brief Establish if this is a load balance step using the adequate method.
```
voferreira, line 160: This is defined first and then is_contact_search_step(). I would maybe put contact search below load balance as it is in dem.cc

include/dem/dem.h
```diff
@@ -142,6 +138,12 @@ class DEMSolver
 #  endif
 
 
+  /**
+   * @brief Establish if this is a contact detection step using the adequate contact search method
```
voferreira, line 142: I haven't done Doxygen so far, but as this is a return function, shouldn't it have a return instead (or with) a brief? Same question for the member is_load_balance_step().

source/dem/dem.cc
```diff
@@ -546,6 +492,44 @@ DEMSolver<dim>::load_balance()
...
+    {
+      return check_load_balance_dynamic();
+    }
+  else if (parameters.model_parameters.load_balance_method ==
+           Parameters::Lagrangian::ModelParameters::LoadBalanceMethod::
+             dynamic_with_disabling_contacts)
+    {
+      return check_load_balance_with_disabled_contacts();
+    }
+  else if (parameters.model_parameters.load_balance_method ==
```
voferreira, line 521: I think "none" should be the first to be tested.

source/dem/dem.cc
```diff
@@ -748,23 +732,40 @@ DEMSolver<dim>::check_load_balance_dynamic()
...
 }
 
+template <int dim>
+inline bool
+DEMSolver<dim>::is_contact_search_step()
+{
+  if (parameters.model_parameters.contact_detection_method ==
+      Parameters::Lagrangian::ModelParameters::ContactDetectionMethod::constant)
+    return check_contact_search_step_constant();
+  else if (parameters.model_parameters.contact_detection_method ==
```
voferreira, line 742: Should we be evaluating if the method is constant at all steps? This could be constant expression no?

voferreira (commented):
Way cleaner this way. Just some questions and small comments. Let me know when I can rereview it

source/dem/dem.cc
blaisb: No because this is the constructor of the class. What used to be done here was just establishing the right function pointer. This is why I wanted to remove that function pointer logic, kinda created a two-step logic abstraction that we don't need

source/dem/dem.cc
blaisb: Ok, that's aesthetics but I don't mind.

source/dem/dem.cc
voferreira: Ok. Got it. I thought I have deleted this comment actually :)

source/dem/dem.cc
```diff
@@ -548,44 +499,73 @@ DEMSolver<dim>::load_balance()
...
+  else if (parameters.model_parameters.load_balance_method ==
+           Parameters::Lagrangian::ModelParameters::LoadBalanceMethod::
+             dynamic_with_disabling_contacts)
+    {
+      return check_load_balance_with_disabled_contacts();
+    }
+  else
+    {
+      throw std::runtime_error("Specified load balance method is not valid");
+      return false;
```
OGaboriault, line 533:
```suggestion
```

source/dem/dem.cc
```diff
@@ -548,44 +499,73 @@ DEMSolver<dim>::load_balance()
 
 template <int dim>
 inline bool
-DEMSolver<dim>::no_load_balance()
+DEMSolver<dim>::is_load_balance_iteration()
 {
-  return false;
+  if (parameters.model_parameters.load_balance_method ==
```
OGaboriault, line 504: those if/else if are being check at every time step. 

+ or minus 5 if at every time step won't affect the code performance, so it is ok for this PR. 

I don't think it is clean. I already have a "clean" implementation thank to Paul. We could discuss this a your return. :)

OGaboriault (approved):
Just a tiny change. (The extra return)

I can help for the falling test is you want. 

Good work! :)

source/dem/dem.cc
voferreira: I like Olivier's new implementation. It is indeed way cleaner. But this can be merged as is as a checkpoint to his implementation.

voferreira (approved):
Good to go for me. Since Gabo's new implementation changes the architecture of the checks, IMO we should merge this and move to his in another PR.

blaisb:
The revert to using std::function brings us a bit closer to the original implementation that used function pointers, although in a cleaner fashion. If you both (@OGaboriault and @voferreira ) think this is a cleaner implementation, then we can continue with it. However, in order to remain coherent, we need to do the same approach for the contact detection And the load balance. Now only the load balance is done with std::function. @OGaboriault can you modify the PR to use the same mechanism for the contact detection ?

voferreira:
@blaisb To me it is a go. @OGaboriault Pin me and I will review it.

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
     const CellStatus status) const;
 #  endif
 
+  /**
+   * @brief Sets the right iteration check function according to the chosen contact detection method.
```
voferreira, line 141:
```suggestion
   * @brief Sets the iteration check function according to the chosen contact detection method.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   *
+   * @return Return a function. This function returns a bool indicating if the contact search should be carried out in the current iteration calculated using the appropriate method.
```
voferreira, line 143:
```suggestion
   * @return Return a function. This function returns a bool indicating if the contact search should be carried out in the current iteration.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   */
+  inline std::function<bool()>
+  set_contact_search_iteration_function();
 
   /**
-   * Finds contact search steps for constant contact search method
+   * @brief Establish if this is a contact detection iteration using the constant contact detection frequency.
```
voferreira, line 149:
```suggestion
   * @brief For a constant detection frequency, determines whether the present step is a contact detection step.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   *
+   * @return bool indicating if the contact search should be carried out in the current iteration.
    */
   inline bool
-  check_contact_search_step_constant();
+  check_contact_search_iteration_constant();
 
   /**
-   * Finds contact search steps for dynamic contact search method
+   * @brief Establish if this is a contact detection iteration using the maximal displacement of the particles.
```
voferreira, line 159:
```suggestion
   * @brief Determines whether the present step is a contact detection step using the maximal displacement of the particles.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * If this particle displacement surpasses a threshold, this iteration is a
```
voferreira, line 160:
```suggestion
   * If the particle displacement surpasses a threshold, this iteration is a
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * @return Return a function. This function returns a bool indicating if the current time step is a load balance iteration.
    */
-  inline bool
-  check_load_balance_once();
+  inline std::function<bool()>
+  set_load_balance_iteration_check_function();
 
   /**
-   * For cases where load balance method is equal to none
+   * @brief Establish if this is a load-balance step when load balance is
```
voferreira, line 177:
```suggestion
   * @brief For `load balance method = once`, determines whether the present is the load balance step.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * established once at a user-defined iteration number.
```
voferreira, line 178:
```suggestion
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   *
+   * @return bool indicating if this is a load balance iteration.
    */
   inline bool
-  no_load_balance();
+  check_load_balance_once();
 
   /**
-   * Finds load-balance step for frequent load-balance
+   * @brief Establish if this is a load-balance step when load balance is
```
voferreira, line 186:
```suggestion
   * @brief Determine whether the present is a load-balance step given a user-defined frequency.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * established at a user-defined frequency.
```
voferreira, line 187:
```suggestion
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
    */
   inline bool
   check_load_balance_frequent();
 
   /**
-   * Finds load-balance step for dynamic load-balance
+   * @brief Establish if this is a load-balance step using the dynamic method. The dynamic method
+   * uses the load imbalance between the core as a load balancing criteria.
+   *
+   * @return bool indicating if this is a load balance iteration.
```
voferreira, line 198:
```suggestion
   * @return bool indicating if the present is a load balance iteration.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * established at a user-defined frequency.
+   *
+   * @return bool indicating if this is a load balance iteration.
    */
   inline bool
   check_load_balance_frequent();
 
   /**
-   * Finds load-balance step for dynamic load-balance
+   * @brief Establish if this is a load-balance step using the dynamic method. The dynamic method
```
voferreira, line 195:
```suggestion
   * @brief Determine whether the present is a load balance step for the dynamic load balancing method. The dynamic method uses the load imbalance between cores as the criterium.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * uses the load imbalance between the core as a load balancing criteria.
+   *
+   * @return bool indicating if this is a load balance iteration.
    */
   inline bool
   check_load_balance_dynamic();
 
   /**
-   * Finds load-balance step for dynamic load-balance with disabled contact
+   * @brief Establish if this is a load-balance step using the dynamic method when the disabled contacts mechanism is enabled.
```
voferreira, line 204:
```suggestion
   * @brief Determine whether the present is a load-balance step using dynamic load-balancing for the case of disabled contact mechanism.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
+   * The dynamic method uses the load imbalance between the core as a load
+   * balancing criteria.
+   *
+   * @return bool indicating if this is a load balance iteration.
```
voferreira, line 208:
```suggestion
   * @return bool indicating if the present is a load balance iteration.
```

include/dem/dem.h
```diff
@@ -141,53 +137,82 @@ class DEMSolver
...
    */
   inline bool
   check_load_balance_with_disabled_contacts();
 
   /**
-   * @brief Manages the call to the load balancing. Returns true if
-   * load balancing is performed
-   *
+   * @brief Manages the call to the load balance by first identifying if
```
voferreira, line 214:
```suggestion
   * @brief Call load-balance step checking functions and execute it upon finding a load-balancing step.
```

voferreira (approved):
All clear. I have some suggestions on the doxygen it is good to go IMO.

include/dem/dem.h
```diff
@@ -141,53 +137,80 @@ class DEMSolver
...
+   *
+   * @return bool indicating if the contact search should be carried out in the current iteration.
    */
   inline bool
-  check_contact_search_step_dynamic();
+  check_contact_search_iteration_dynamic();
 
   /**
-   * Finds load-balance step for single-step load-balance
+   * @brief Sets the right iteration check function according to the chosen load balancing method.
```
blaisb, line 169:
```suggestion
   * @brief Sets the right contact iteration check function according to the chosen load balancing method.
```

blaisb:
I will rebase then merge after CI has passed :)

PR #1102, opened 2024-04-22
Allow to reuse the preconditioner in the Newton solver

lpsaavedra:
# Description of the problem

The multigrid preconditioners were being setup at every non-linear iteration, however, we want to be able to reuse them to reduce the setup time. 

# Description of the solution

- The `MFNavierStokesPreconditionGMG` class now sets the level constraints, operators and transfers in the constructor. Then, depending on the MG being used, there is an initialize function in charge of setting up the smoother and the coarse-grid solver. 
-  Modifications were made in the Newton nonlinear solver to allow to reuse the preconditioner along non-linear newton iterations. A new parameter `reuse preconditioner` was introduced for this purpose. When this change was introduced, the `assemble_matrix` and `setup_preconditioner` functions of all the NavierStokesBase solvers had to be slightly modified. This change now allows to reuse the preconditioner in both `lethe-fluid` and `lethe-fluid-matrix-free.`  Due to this change, also the auxiliary physics needs a `setup_preconditioner` function, however, it was explicitly written that is not used for those applications.
- The constant modes are now also being used for the local smoothing approach using the recently introduced `DoFTools::extract_level_constant_modes(...)`.
- `this->` was also included several times in the `MFNavierStokesPreconditionGMG` class to be able to differentiate between this class member variables and others. 
- The output of the coarse-grid iterations was finally corrected, and now prints all the information.

# How Has This Been Tested?

- All the existent test pass.
- Ran several tests on my machine and observed a benefit in certain cases (e.g., the TGV) when reusing the preconditioner across all Newton iterations of a transient iteration. 

# Documentation

- The new `reuse preconditioner`  parameter was included in the non linear solver documentation page.

# Future changes

- This is the second step towards reusing parts of multigrid to improve the performance of the matrix-free solver.


include/core/newton_non_linear_solver.h
```diff
@@ -95,6 +95,12 @@ NewtonNonLinearSolver<VectorType>::solve(const bool is_initial_step)
       evaluation_point = present_solution;
 
       solver->assemble_system_matrix();
+
+      if (this->params.reuse_preconditioner && this->outer_iteration == 0)
+        solver->setup_preconditioner();
+      else if (!this->params.reuse_preconditioner)
+        solver->setup_preconditioner();
```
peterrum, line 102:
```suggestion
      if (this->outer_iteration == 0 || !this->params.reuse_preconditioner)
        solver->setup_preconditioner();
```

source/solvers/mf_navier_stokes.cc
```diff
@@ -1539,53 +1633,33 @@ template <int dim>
 void
 MFNavierStokesSolver<dim>::setup_GMG()
 {
-  gmg_preconditioner = std::make_shared<MFNavierStokesPreconditionGMG<dim>>();
+  this->computing_timer.enter_subsection("Setup GMG");
+
+  gmg_preconditioner = std::make_shared<MFNavierStokesPreconditionGMG<dim>>(
```
peterrum, line 1638:
```suggestion
  if(!gmg_preconditioner)
  gmg_preconditioner = std::make_shared<MFNavierStokesPreconditionGMG<dim>>(
```

source/solvers/mf_navier_stokes.cc
```diff
@@ -1539,53 +1633,33 @@ template <int dim>
...
-                                      this->fe,
-                                      this->mg_computing_timer,
-                                      this->cell_quadrature,
-                                      this->forcing_function,
-                                      this->present_solution,
-                                      this->time_derivative_previous_solutions,
-                                      this->pcout,
-                                      this->simulation_control,
-                                      this->flow_control);
+    gmg_preconditioner->initialize_ls(this->mg_computing_timer);
```
peterrum, line 1653: I guess `initialize_ls()` and `initialize_gc()` can be merged, since `simulation_parameters` is passed to the constructor, which can be used to determine `lsmg` vs. `gcmg`.

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,244 +44,580 @@
...
+      this->mg_transfer_ls->build(this->dof_handler, partitioners);
+      this->mg_transfer_ls->interpolate_to_mg(this->dof_handler,
+                                              mg_solution,
+                                              present_solution);
+
+      if (is_bdf(simulation_control->get_assembly_method()))
+        this->mg_transfer_ls->interpolate_to_mg(
+          this->dof_handler,
+          mg_time_derivative_previous_solutions,
+          time_derivative_previous_solutions);
```
peterrum, line 275: I think this needs to go to `initialize()`.

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,244 +44,580 @@
...
-      this->mg_operators[level]->evaluate_non_linear_term(mg_solution[level]);
+      this->mg_transfer_gc->interpolate_to_mg(this->dof_handler,
+                                              mg_solution,
+                                              present_solution);
 
       if (is_bdf(simulation_control->get_assembly_method()))
+        this->mg_transfer_gc->interpolate_to_mg(
+          this->dof_handler,
+          mg_time_derivative_previous_solutions,
+          time_derivative_previous_solutions);
```
peterrum, line 585: This as well.

source/solvers/mf_navier_stokes.cc
blaisb: True

include/solvers/mf_navier_stokes.h
```diff
@@ -135,7 +107,24 @@ class MFNavierStokesPreconditionGMG
...
+   */
+  void
+  print_relevant_info() const;
+
 private:
+  /// Min level
+  unsigned int minlevel;
+
+  /// Max level
+  unsigned int maxlevel;
```
blaisb, line 122: maybe Min level of the multigrid preconditioner ?

include/solvers/mf_navier_stokes.h
```diff
@@ -373,6 +374,13 @@ class MFNavierStokesSolver
   void
   setup_ILU();
 
+  /**
+   * @brief Prints the setup times for the gemoetric multigrid preconditioner.
```
blaisb, line 378:
```suggestion
   * @brief Prints the setup times for the geometric multigrid preconditioner.
```

doc/source/parameters/cfd/non-linear_solver_control.rst
```diff
@@ -87,3 +90,4 @@ The Navier-Stokes equations (and other) are non-linear equations. The parameters
 * The ``residual precision`` parameter enables to change the number of digits displayed when showing residuals (with ``set verbosity = verbose``).
 * The ``force_rhs_calculation``: Force RHS recalculation at the beginning of every non-linear steps, This is required if there is a fixed point component to the non-linear solver that is changed at the beginning of every newton iteration. This is notably the case of the sharp edge method. The default value of this parameter is false.
 * The ``abort at convergence failure`` allows the user to stop the simulation and throw an error if the non-linear solver has failed to converge. Setting ``abort at convergence failure = true`` will enable this feature. This is generally useful when launching a large batch of simulation to quickly identify which one have failed.
+* The ``reuse preconditioner = true`` allows the simulation to use the same preconditioner between Newton iterations when using the Newton solver. This can reduce the overall time depending on the problem, and it is especially useful for the ``lethe-fluid-matrix-free`` application.
```
blaisb, line 93: So in this case you are recreating the preconditioner between iterations time steps for example right?

blaisb (changes requested):
Some small changes. I have one more file to review, will do it tomorrow :)

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,1083 +44,1229 @@
...
+      int mg_level_min_cells =
+        this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
+          .mg_level_min_cells;
 
-      if (is_bdf(simulation_control->get_assembly_method()))
-        {
-          mg_time_derivative_previous_solutions[level].update_ghost_values();
-          mg_operators[level]->evaluate_time_derivative_previous_solutions(
-            mg_time_derivative_previous_solutions[level]);
+      AssertThrow(
```
blaisb, line 288: Is the best approach here to throw or would it be to just print a warning and use the max level as the min level (e.g. no GMG)

blaisb (approved):
One last comment. These are all minor things. ping me when everything has been incorporated then we can merge. Great work @lpsaavedra . Can you add a changelog entry?

doc/source/parameters/cfd/non-linear_solver_control.rst
lpsaavedra: Exactly... it reuses it for Newton iterations within one transient iteration...

applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.with_dealii.geq.9.6.mpirun=1.output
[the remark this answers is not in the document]
lpsaavedra: Yes, I was also not sure about this. It changed when I added the if to reuse the operators, transfer and constraints that are set in the constructor:
```
  if (!gmg_preconditioner)
    gmg_preconditioner = std::make_shared<MFNavierStokesPreconditionGMG<dim>>(
      this->simulation_parameters,
      this->dof_handler,
      this->mapping,
      this->cell_quadrature,
      this->forcing_function,
      this->simulation_control,
      this->mg_computing_timer,
      this->fe);
```
However, the number of newton iterations, linear solver iterations and coarse-grid solver iterations remain the same. It only change the residual in the 3rd/4th decimal place which caused these changes in the value of the forces. In my opinion, the change is minimal so it should not be an issue.

source/solvers/mf_navier_stokes.cc
lpsaavedra: I am not sure about this... I will think about it and then maybe we can do this in the next MG PR

peterrum (approved):

PR #1103, opened 2024-04-25
Correct DCDD stabilization in VOF

AmishgaAlphonius:
# Description of the problem

- The DCDD VOF stabilization term taken form Tezduyar (2003) had minor a mistake in it. A square was missing.

# Description of the solution

- The missing square was added to the stabilization term.

# How Has This Been Tested?

The following tests were updated due to the small changes in computed values :
- lethe-fluid/gls_droplet_marangoni_effect
- lethe-fluid/gls_vof_1_isothermal_compressible_fluid
- lethe-fluid/gls_vof_2_isothermal_compressible_fluids
- lethe-fluid/gls_vof_dirichlet_boundary_condition
- lethe-fluid/gls_vof_ht_adapt-dissipation
- lethe-fluid/gls_vof_hydrostat_initial_refine
- lethe-fluid/gls_vof_hydrostat_mesh-adapt
- lethe-fluid/gls_vof_periodic_boundary_condition
- lethe-fluid/gls_vof_static_droplet_surface_tension_force_with_tanh_filter
- lethe-fluid/gls_vof_tanh_filter
- lethe-fluid/heat_transfer_vof_phase_change_constrain_solid_domain
- lethe-fluid/heat_transfer_vof_weakly_compressible_flow
- lethe-fluid/time_dependent_boundaries_vof

- When tested with the rising bubble case, there seems to be no noticeable difference from before the correction.

# Documentation

Documentation on VOF's theory was updated.
- doc/source/theory/multiphysics/vof.rst

# Comments

- It would be interresting to study the impact of this DCDD stabilization term and in which case it really has a great influence.

# Reference

T. E. Tezduyar, “Computation of moving boundaries and interfaces and stabilization parameters,” Int. J. Numer. Methods Fluids, vol. 43, no. 5, pp. 555–575, 2003, doi: 10.1002/fld.505.


source/solvers/vof_assemblers.cc
```diff
@@ -77,7 +77,7 @@ VOFAssemblerCore<dim>::assemble_matrix(VOFScratchData<dim>       &scratch_data,
       Tensor<1, dim> velocity_unit_vector =
         velocity / (velocity.norm() + 1e-12);
       const Tensor<2, dim> k_corr =
-        (gradient_unit_vector * velocity_unit_vector) *
+        std::pow(gradient_unit_vector * velocity_unit_vector, 2) *
```
blaisb, line 80: Can you use instead https://www.dealii.org/current/doxygen/deal.II/namespaceUtilities.html#a834c6b595ada6f8e73c78d9bbd8ec551


This way this is done at compile time since you know the power at compile time. This is always better :)

source/solvers/vof_assemblers.cc
```diff
@@ -222,7 +222,7 @@ VOFAssemblerCore<dim>::assemble_rhs(VOFScratchData<dim>       &scratch_data,
...
```
blaisb, line 225: Same as above

blaisb (approved):
1 minor thing to fix, ping me when this is done and this will be good to merge

source/solvers/vof_assemblers.cc
AmishgaAlphonius: I'll change all the ones in the file ! :)

hepap (approved):

source/solvers/vof_assemblers.cc
AmishgaAlphonius: @blaisb It's done ! There's only the CI left ! :)

blaisb: Thanks for going the extra mile and cleaning things up
:)

Issue #275, opened 2021-10-31
Nitsche model restart does not work with simplices


blaisb:
The current code does not write the simplex mesh for the immersed boundary and thus, when restarting will crash.
Depending on the difficulty of serializing meshes that are made of simplices, this may be hard or difficult to do :)

anicusan:
The main advantage of the IB in our cases was that we can avoid the hexahedral meshing of very complex geometries; it's great that we can use a refined, simple hexahedral background mesh and a tetrahedralization of the complex bodies inside.

Writing simplex Nitsche meshes when checkpointing simulations would be very useful when running long jobs on shared HPCs that limit the maximum walltime. I don't see any special handling of hex over tet meshes in `source/core/solid_base.cc`, just:

```cpp
  if (auto tria = dynamic_cast<parallel::distributed::Triangulation<dim> *>(
        this->solid_tria.get()))
    {
      std::string triangulationName = prefix + ".triangulation";
      tria->save(prefix + ".triangulation");
    }
```

Does the mesh serialization need to be done in deal.II? If not, would you have any pointers as to where to start with this? Apologies if my issues sound more like asking for guidance than concise problem-solving - I can follow Lethe's architecture, but using deal.II is a bigger step...

blaisb:
Dear @anicusan that's a very good question. I am not sure what's the exact issue with the checkpoint of simplex meshes right now as I have not tried it in a while. I know at the time we encountered challenges (hence this issue). Give me a day or two and I can try it on my side of things this week to see if there's any bug left and if they can be fixed. I'll message you after that :). I think I can come up with a fix since my understanding of these classes has improved over time :)!

PR #1106, opened 2024-04-29
Fix nitsche restart

blaisb:
# Description of the problem

- Nitsche does not restart when the mesh was generated with simplices. See #275 

# Description of the solution

- Fixes the issue by restarting the fully distributed triangulation for the solid correctly.

# How Has This Been Tested?

- Still pending a unit test, WIP

# Comments

- While running this test case I managed to find another bug, which is that the Nitsche model is unable of mesh adaptation once it has been restarted. This bug was there all the time. I will try to figure out something, in the meantime I will try to m erge this PR anyway.

### DIFF
--- a/source/core/solid_base.cc
+++ b/source/core/solid_base.cc
@@ -752,36 +752,24 @@ SolidBase<dim, spacedim>::write_checkpoint(std::string prefix)
 
   system_trans_vectors.prepare_for_serialization(sol_set_transfer);
 
-  // if (auto tria = dynamic_cast<parallel::distributed::Triangulation<dim> *>(
-  //       this->solid_tria.get()))
-  //   {
-      std::string triangulationName = prefix + ".triangulation";
-      this->solid_tria.get()->save(prefix + ".triangulation");
-    // }
+  std::string triangulationName = prefix + ".triangulation";
+  this->solid_tria.get()->save(prefix + ".triangulation");
 }
 
 template <int dim, int spacedim>
 void
 SolidBase<dim, spacedim>::read_checkpoint(std::string prefix)
 {
-  // Setup an un-refined triangulation before loading
-  setup_triangulation(true);
+  // Setup an un-refined triangulation before loading if we have a fully distributed triangulation.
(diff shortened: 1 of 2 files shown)


source/core/solid_base.cc
```diff
@@ -285,23 +285,24 @@ SolidBase<dim, spacedim>::load_triangulation(const std::string filename_tria)
...
-                    "but the restart file <") +
-                  filename_tria + "> does not appear to exist!"));
+  // std::ifstream in_folder(filename_tria.c_str());
+  // if (!in_folder)
+  //   AssertThrow(false,
+  //               ExcMessage(
+  //                 std::string(
+  //                   "You are trying to restart a previous computation, "
+  //                   "but the restart file <") +
+  //                 filename_tria + "> does not appear to exist!"));
```
lpsaavedra, line 295: Do you want to leave this comments there? Or should we remove them?

source/core/solid_base.cc
```diff
@@ -285,23 +285,24 @@ SolidBase<dim, spacedim>::load_triangulation(const std::string filename_tria)
...
-            dynamic_cast<parallel::distributed::Triangulation<dim, spacedim> *>(
-              get_solid_triangulation().get()))
-        {
-          solid_tria->load(filename_tria.c_str());
-        }
+      // if (auto solid_tria =
+      //       dynamic_cast<parallel::distributed::Triangulation<dim, spacedim>
+      //       *>(
+      //         get_solid_triangulation().get()))
+      //   {
```
lpsaavedra, line 303: Same as above...

lpsaavedra (commented):
So the mixer example that we are testing is still not working, am I correct? Even with mesh adaptation set to none?

blaisb:
> So the mixer example that we are testing is still not working, am I correct? Even with mesh adaptation set to none?
With the new deal.II version coming up it will work without restart
With restart it will take more work, but this will be a follow up PR

source/core/solid_base.cc
blaisb: I will remove those

PR #1107, opened 2024-04-30
Dem fixing tangential overlap

OGaboriault:
# Description of the problem

- The normal overlap and tangential overlap vectors needs to be perpendicular through out a collision, which is not the case in the current implementation. 

# Description of the solution
- We subtract the projection of the tangential overlap on the normal vector from the tangential overlap at every time step.
- This is not done for particle-wall contacts, since the normal vector remains the same through out a collision with a flat wall (which is the only type of wall currently implemented).

# How Has This Been Tested?

- With the application test. Some results have slightly changed


blaisb:
Why do you have so many changed files that are unrelated to DEM????

blaisb (changes requested):
Please clean the PR by adding the files of the test that actually need modifications, not all the tests...

blaisb (approved):

blaisb (approved):

PR #1110, opened 2024-05-02
Think it is fixed, test are still passing, need to make a new test

OGaboriault:
# Description of the problem

- Periodic boundary conditions were not working when they were not declare as the last BC in the parameter file.
- BC_type was changing value at every BC being parsed. "BC_type" was being checked later in the code (in dem.cc) to set "has_periodic_boundaries" to true or false.  If the periodic BC was being declared before the an other BC, "BC_type" was not set to "BoundaryType::periodic", thus "has_periodic_boundaries = false".

# Description of the solution

- We store every BC type in a vector (BC_types). 
- We check each component of that vector, if it's equal to "BoundaryType::periodic", then "has_periodic_boundaries = true"

# How Has This Been Tested?

- Every test is passing.
- I can modify one of our current test to add a dummy BC at the end of a parameter file, so that it check for this kind of problem.

# Comments

- Not gonna lie, it was not hard a bug spot in the code... wish I had spot this mistake earlier in my results....

### DIFF
--- a/source/core/parameters_lagrangian.cc
+++ b/source/core/parameters_lagrangian.cc
@@ -1118,12 +1118,12 @@ namespace Parameters
 
       if (boundary_type == "outlet")
         {
-          BC_type = BoundaryType::outlet;
+          bc_types.push_back(BoundaryType::outlet);
           this->outlet_boundaries.push_back(boundary_id);
         }
       else if (boundary_type == "translational")
         {
-          BC_type = BoundaryType::translational;
+          bc_types.push_back(BoundaryType::translational);
           Tensor<1, 3> translational_velocity;
           translational_velocity[0] = prm.get_double("speed x");
           translational_velocity[1] = prm.get_double("speed y");
@@ -1134,7 +1134,7 @@ namespace Parameters
         }
       else if (boundary_type == "rotational")
         {
-          BC_type                 = BoundaryType::rotational;
+          bc_types.push_back(BoundaryType::rotational);
           double rotational_speed = prm.get_double("rotational speed");
 
           // Read the rotational vector from a list of doubles
@@ -1157,11 +1157,11 @@ namespace Parameters
         }
       else if (boundary_type == "fixed_wall")
         {
-          BC_type = BoundaryType::fixed_wall;
+          bc_types.push_back(BoundaryType::fixed_wall);
         }
       else if (boundary_type == "periodic")
         {
-          BC_type                   = BoundaryType::periodic;
+          bc_types.push_back(BoundaryType::periodic);
           this->periodic_boundary_0 = prm.get_integer("periodic id 0");
           this->periodic_boundary_1 = prm.get_integer("periodic id 1");
           this->periodic_direction  = prm.get_integer("periodic direction");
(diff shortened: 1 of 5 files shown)


source/core/parameters_lagrangian.cc
```diff
@@ -1240,7 +1242,9 @@ namespace Parameters
           point_on_rotation_axis.insert({counter, Point<3>(zero_tensor)});
         }
 
+      // NOTE This first vector should not be initialized this big.
       outlet_boundaries.reserve(DEM_BC_number);
+      bc_types.reserve(DEM_BC_number);
```
blaisb, line 1247: Yeah but how big is DEM_BC_number? Here it is reserved and not initialized btw, so this is not really a big deal...

source/dem/read_mesh.cc
```diff
@@ -15,10 +15,12 @@ read_mesh(const Parameters::Mesh   &mesh_parameters,
 
   attach_grid_to_triangulation(triangulation, mesh_parameters);
 
-  if (bc_params.BC_type ==
-      Parameters::Lagrangian::BCDEM::BoundaryType::periodic)
+  // Check if there's periodic boundaries
```
blaisb, line 18:
```suggestion
  // Check if there are periodic boundaries
```

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1470,10 +1470,16 @@ CFDDEMSolver<dim>::dem_setup_contact_parameters()
   load_balance_step      = false;
 
   // Check if there's periodic boundaries
```
blaisb, line 1472:
```suggestion
  // Check if there are periodic boundaries
```

blaisb (approved):
Just tiny comments

blaisb:
@OGaboriault I think this needs a CHANGELOG entry

source/core/parameters_lagrangian.cc
OGaboriault: You are right, DEM_BC_number is really small

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-05-02
+
+### Fixed
+
+- MAJOR Periodic boundary conditions were not working if they weren't the last boundary condition being declare in the parameter file. Now, every boundary condition work in which even order they are being declared. [#1110](https://github.com/chaos-polymtl/lethe/pull/1110)
```
acdaigneault, line 10: Specify that this is related the periodic boundaries for DEM only

applications_tests/lethe-particles/periodic_boundary_box.prm
```diff
@@ -93,9 +93,14 @@ end
 #---------------------------------------------------
 
 subsection DEM boundary conditions
-  set number of boundary conditions = 1
+  set number of boundary conditions = 2
 
-  subsection boundary condition 0
+   subsection boundary condition 0
+    # Choices are fixed_wall|outlet|rotational|translational|periodic
```
acdaigneault, line 99: What is fixed_wall actually? 
Also, maybe (in another pr) we should have a test also for rotational/translational and PBC (like a rotating drum with PBC on flat walls) and make sure all PB match well.

acdaigneault (changes requested):
This fix is good, sorry about that haha.
Just change the log to specify that the bug was only related to PBC in DEM since there are PBC for fluid that were not affected by this bug.

acdaigneault (approved):

acdaigneault (commented):
Btw, change the name of the PR haha

PR #1111, opened 2024-05-03
Fix NavierStokesOperatorBase::get_system_matrix()

peterrum:
At the moment, the matrix is cleared during `reinit()`, and only setup once although the parameters have changed.


blaisb (approved):

blaisb:
This will change the steady-state results significantly. @lpsaavedra I think some stuff will need to be rerun. I iwll merge after CI

lpsaavedra:
I have added a small correction to avoid re-calculating the system matrix of the operator when we use AMG in the coarse-grid solver... after CI, we can merge this

PR #1114, opened 2024-05-04
Added runtime errors when reading files.

OGaboriault:
# Description of the problem

- The code doesn't crash when it tries to open a file that doesn't exist in the current directory. In my opinion, this is not a feature...
- i.e., tried to insert particles with the file insertion and the simulation ran for many hours with no particles in it because it didn't crash.

# Comments

- I have put Lucka as a reviewer, because it may be a "feature" for sharp.

### DIFF
--- a/source/core/utilities.cc
+++ b/source/core/utilities.cc
@@ -258,7 +258,9 @@ fill_table_from_file(TableHandler     &table,
       myfile.close();
     }
   else
-    std::cout << "Unable to open file";
+    {
+      throw std::runtime_error("Unable to open file " + file_name + ".");
+    }
 }
 
 void
@@ -316,7 +318,9 @@ fill_vectors_from_file(std::map<std::string, std::vector<double>> &map,
       myfile.close();
     }
   else
-    std::cout << "Unable to open file";
+    {
+      throw std::runtime_error("Unable to open file " + file + ".");
+    }
 }
 
 void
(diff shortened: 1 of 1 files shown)


blaisb:
I don't think that's a feature anywhere. It's just a lack of sanity checking.

source/core/utilities.cc
```diff
@@ -258,7 +258,9 @@ fill_table_from_file(TableHandler     &table,
       myfile.close();
     }
   else
-    std::cout << "Unable to open file";
+    {
+      throw std::runtime_error("Unable to open file " + file_name + ".");
```
blaisb, line 262: Can you use the deal.II exception instead?
https://www.dealii.org/current/doxygen/deal.II/group__Exceptions.html#ga2f31207e832b937688f75c04e939cc21

blaisb (changes requested):
A quick modification

source/core/utilities.cc
```diff
@@ -657,8 +644,8 @@ get_dimension(const std::string &file_name)
       // a carriage-return without a newline -- so the error message looks
       // like this:
       //
-      //    >.  While reading the dimension from the input file, ASPECT found a
-      //    string that can not be converted to an integer: <2
+      //    >.  While reading the dimension from the input file, ASPECT found
+      //    a string that can not be converted to an integer: <2
```
blaisb, line 648:
```suggestion
      //    >.  While reading the dimension from the input file, Lethe found
      //    a string that can not be converted to an integer: <2
```

blaisb (approved):
Good to merge after CI.

Luckabarbeau:
Nothing to add here ! seems good to me.

PR #1115, opened 2024-05-05
Improve Scratch data for CFD-DEM simulations

blaisb:
# Description of the problem

- The scratch data for CFD-DEM simulation was very much all over the place. Memory was being reallocated and interpolation was sometimes done manually whereas some other times it was done using a quadrature defined on the fly

# Description of the solution

- Unify interpolation
- Prevent memory reallocation

# How Has This Been Tested?

- All tests pass with no changed results (as expected)

# Comments

- This does not lead to a large speed boost (like 3-4% on what I tested), but it really makes the code more readable and more uniform. Next PR will be on improving the assemblers. Since this was really limited to the scratch data I feel this PR can stand on its own.

### DIFF
--- a/include/solvers/navier_stokes_scratch_data.h
+++ b/include/solvers/navier_stokes_scratch_data.h
@@ -674,7 +674,7 @@ public:
   template <typename VectorType>
   void
   reinit_particle_fluid_interactions(
-    const typename DoFHandler<dim>::active_cell_iterator &cell,
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
     const VectorType /*current_solution*/,
     const VectorType                       previous_solution,
     const VectorType                       void_fraction_solution,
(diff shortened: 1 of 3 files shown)


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -724,11 +724,7 @@ class NavierStokesScratchData
       number_of_particles = i_particle;
     }
 
-    // If there were particles, calculate the average particle velocity within
-    // the cell
-    if (number_of_particles > 0)
-      average_particle_velocity =
-        average_particle_velocity / number_of_particles;
+
```
voferreira, line 727:
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -674,44 +674,24 @@ class NavierStokesScratchData
...
-
-    const auto &void_fraction_dh_cell = typename DoFHandler<dim>::cell_iterator(
-      *this->fe_values_void_fraction->get_cell(), &void_fraction_dof_handler);
-    void_fraction_dh_cell->get_dof_indices(void_fraction_dof_indices);
-
     // Loop over particles in cell
     double total_particle_volume = 0;
     {
-      unsigned int particle_no = 0;
+      unsigned int i_particle = 0;
```
voferreira, line 694: why not particle id?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -723,57 +703,30 @@ class NavierStokesScratchData
...
-            particle_velocity[particle_no][2] =
+          if constexpr (dim == 3)
+            particle_velocity[i_particle][2] =
               particle_properties[DEM::PropertiesIndex::v_z];
 
-          // Interpolate velocity and void fraction at particle position
-          // Reference location of the particle
-          auto reference_location = particle.get_reference_location();
+          cell_void_fraction[i_particle] = 0;
+          if (!interpolated_void_fraction)
```
voferreira, line 716: Do we still use this? I think we could remove interpolated void fraction = false.

voferreira: I have implemented this for testing purposes, but I don't think it is necessary

voferreira (commented):
Good refactoring, especially for readability purposes.
I find it weird that we don't call the particle counter "particle_id", but this is a minor thing.
Do you think we still need the non-interpolated void fraction option? I think this could be removed and have interpolated used always.
I have no further comments, and I will approve it after finishing this discussion.

include/solvers/navier_stokes_scratch_data.h
blaisb: Well it's not the particle_id right, particles they have an id, they store it as a property, so this would be even more confusing because this particle_i or i_particle is a loop index.

As for removing the feature, this should be done in a seperate PR standing on its own with a CHANGELOG entry

include/solvers/navier_stokes_scratch_data.h
blaisb: Because it is not a particle id

voferreira (approved):

include/solvers/navier_stokes_scratch_data.h
blaisb: You or I can make another PR to remove the feature afterwards. I just feel that removing features require PR on its own because of the documentation change and the CHANGELOG. This way it is clearer.

PR #1118, opened 2024-05-06
Improve CFD-DEM assemblers performance

blaisb:
# Description of the problem

- The CFD-DEM assemblers were not really optimized in the way components were managed
- We did not have any VANS modelA manufactured solution tests

# Description of the solution

- Optimize the way the assemblers are built
- Add case0, case1 and case2 MMS for the model A of VANS

# How Has This Been Tested?

- All tests are unchanged and 3 new tests are added for modelA

# Comments

- Model A is clearly the better model, we should add more test for it.
-


source/fem-dem/vans_assemblers.cc
```diff
@@ -483,41 +507,55 @@ GLSVansAssemblerCoreModelA<dim>::assemble_matrix(
 
           for (unsigned int j = 0; j < n_dofs; ++j)
             {
+              const unsigned int component_j = scratch_data.components[j];
+
```
voferreira, line 511:
```suggestion
```

voferreira (approved):
All clear for me.

PR #1122, opened 2024-05-08
Refactor headers of DEM and CFD-DEM

acdaigneault:
# Description of the problem

- Update the year of the copyright of Lethe
- Remove Lethe headers in .cc
- Remove names of authors
- Try to uniformize the header

### DIFF
--- a/include/dem/boundary_cells_info_struct.h
+++ b/include/dem/boundary_cells_info_struct.h
@@ -1,6 +1,6 @@
 /* ---------------------------------------------------------------------
  *
- * Copyright (C) 2019 - 2019 by the Lethe authors
+ * Copyright (C) 2019 - 2024 by the Lethe authors
  *
  * This file is part of the Lethe library
  *
@@ -12,9 +12,7 @@
  * the top level of the Lethe distribution.
  *
  * ---------------------------------------------------------------------
-
  *
- * Author: Shahab Golshan, Polytechnique Montreal, 2019
  */
 #include <deal.II/base/point.h>
 #include <deal.II/base/tensor.h>
--- a/include/fem-dem/gls_sharp_navier_stokes.h
+++ b/include/fem-dem/gls_sharp_navier_stokes.h
@@ -1,6 +1,6 @@
 /* ---------------------------------------------------------------------
  *
- * Copyright (C) 2019 - 2019 by the Lethe authors
+ * Copyright (C) 2019 - 2024 by the Lethe authors
  *
  * This file is part of the Lethe library
  *
@@ -12,10 +12,8 @@
  * the top level of the Lethe distribution.
  *
  * ---------------------------------------------------------------------
+ *
  */
-
-#ifndef LETHE_GLSSHARPNS_H
-#define LETHE_GLSSHARPNS_H
 
 #include <core/ib_particle.h>
 #include <core/ib_stencil.h>
@@ -29,6 +27,9 @@
 
 #include <deal.II/dofs/dof_tools.h>
 
+#ifndef LETHE_GLSSHARPNS_H
+#  define LETHE_GLSSHARPNS_H
+
 using namespace dealii;
 
 /**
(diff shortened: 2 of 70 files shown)


include/fem-dem/gls_sharp_navier_stokes.h
```diff
@@ -29,6 +27,9 @@
 
 #include <deal.II/dofs/dof_tools.h>
 
+#ifndef LETHE_GLSSHARPNS_H
+#  define LETHE_GLSSHARPNS_H
+
```
lpsaavedra, line 32: Why was this moved after the #includes? Isn't it generally better to have it before the includes?

include/fem-dem/gls_vans.h
```diff
@@ -45,7 +40,8 @@
 #include <boost/archive/text_iarchive.hpp>
 #include <boost/archive/text_oarchive.hpp>
 
-
+#ifndef lethe_gls_vans_h
+#  define lethe_gls_vans_h
```
lpsaavedra, line 44: Same as above

include/fem-dem/ib_particles_dem.h
```diff
@@ -35,8 +29,10 @@
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/dofs/dof_tools.h>
 
-using namespace dealii;
+#ifndef lethe_ib_particles_dem_h
+#  define lethe_ib_particles_dem_h
```
lpsaavedra, line 34: Same as above

include/fem-dem/parameters_cfd_dem.h
```diff
@@ -12,20 +12,17 @@
...
-#define lethe_parameters_cfd_dem_h
-
 #include <core/parameters.h>
 #include <core/parameters_lagrangian.h>
 
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/base/parsed_function.h>
 
+#ifndef lethe_parameters_cfd_dem_h
+#  define lethe_parameters_cfd_dem_h
```
lpsaavedra, line 25: Same as above

include/dem/boundary_cells_info_struct.h
```diff
@@ -1,6 +1,6 @@
 /* ---------------------------------------------------------------------
  *
- * Copyright (C) 2019 - 2019 by the Lethe authors
+ * Copyright (C) 2019 - 2024 by the Lethe authors
```
lpsaavedra, line 3: There are some files in Lethe that do not have a second date, and are like these:
`Copyright (C) 2019 - by the Lethe authors`
What is the right approach for this? Let's decide on something so we can have the same across all files.

lpsaavedra (approved):
Nice clean up. We should do this with the whole code... since it is not really uniform.

include/fem-dem/gls_sharp_navier_stokes.h
acdaigneault: I looked into it online and there are no general consensus about it. I think it should be better (with Bruno we looked in deal.II after I did the PR, and yes the includes are in the include guards). I did what was mostly done in Lethe, but I can do the opposite. It could be something in our in-house guidelines.

include/dem/boundary_cells_info_struct.h
acdaigneault: I agree

acdaigneault: I see deal.II uses the same format (year change with the update of the year). 
But I think this is not mandatory, and only the publication date is necessary (legally) https://liferay.dev/blogs/-/blogs/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change.
What do you think?

lpsaavedra: In that case I guess we can do it in the same way as deal.II

blaisb (approved):

PR #1124, opened 2024-05-09
Allow to disable Navier-Stokes source term in MF

lpsaavedra:
# Description of the problem

The evaluation of the source term function, even if it was zero, was affecting the performance of the matrix-free application, as the evaluation needed to be executed every time we performed a vmult using the matrix-free operators.

# Description of the solution

This PR adds a new `enable` parameter that allows to disable the source term explicitly in the matrix free application. It allows to reset the pointer of the forcing function in `mf_navier_stokes.cc` and consequently to check within the matrix-free operators if the pointer exists or not, to evaluate the source term function. 

# How Has This Been Tested?

- All the tests pass and the` enable = false` parameter was added to all the matrix-free tests that do not require a source term. 

# Documentation

- The parameter was added to the source term page in the documentation with a tip that specifies when it is useful to explicitly disable the source term. 

# Future changes

- The Navier-Stokes source term parameter subsection needs to be renamed from `xyz` to `navier stokes`. Since this changes more than 100 files (tests, examples, documentation), I will do this in the next PR.

### DIFF
--- a/applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.prm
+++ b/applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.prm
@@ -115,6 +115,16 @@ subsection boundary conditions
 end
 
 #---------------------------------------------------
+# Source term
+#---------------------------------------------------
+
+subsection source term
+  subsection xyz
+    set enable = false
+  end
+end
+
+#---------------------------------------------------
 # Mesh Adaptation Control
 #---------------------------------------------------
 
--- a/applications_tests/lethe-fluid-matrix-free/poiseuille3d_flow_control.prm
+++ b/applications_tests/lethe-fluid-matrix-free/poiseuille3d_flow_control.prm
@@ -54,6 +54,16 @@ subsection boundary conditions
 end
 
 #---------------------------------------------------
+# Source term
+#---------------------------------------------------
+
+subsection source term
+  subsection xyz
+    set enable = false
+  end
+end
+
+#---------------------------------------------------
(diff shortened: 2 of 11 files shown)


applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.prm
```diff
@@ -114,6 +114,16 @@ subsection boundary conditions
   end
 end
 
+#---------------------------------------------------
+# Source term
+#---------------------------------------------------
+
+subsection source term
+  subsection xyz
```
peterrum, line 122: What is `subsection xyz` standing for?

peterrum (approved):

applications_tests/lethe-fluid-matrix-free/cylinder_kelly_estimator.prm
lpsaavedra: It is actually the subsection of the source term for the navier-stokes. Not a great name, which is why, I will rename it in a follow PR, because it changes more than 100 files (tests, examples, documentation).

blaisb: The name sucks
It's my mistake from 5 years ago
I take the blame for that

blaisb (approved):

blaisb:
@lpsaavedra you have a merge conflict now since I merged the other PR. Can you rebase?

lpsaavedra:
I have rebased now @blaisb

PR #1125, opened 2024-05-09
Allow to use FE_Q_iso_Q1 elements for the MG coarse level

lpsaavedra:
# Description 

Adds a new feature for the GMG preconditioners of the matrix-free application. It allows to use FE_Q_iso_Q1 elements for the coarsest level if the new parameter `set mg coarse grid use fe q iso q1 = true`. This might be useful for certain problems when high order elements are used.

# How Has This Been Tested?

The flow around a sphere case was tested and everything seems to work as it should.  We will test it more in the following weeks with different benchmarks.

# Documentation

- The new parameter was added to the description of the MG linear solver parameters.


peterrum (approved):

blaisb (approved):
Very cool, looking forward to the results

PR #1127, opened 2024-05-10
Add functionality to disable hessians in matrix-free

lpsaavedra:
# Description of the problem

Evaluating terms involving the hessian is expensive. In the matrix-free application this is done way more times than in the matrix-based application, and is affecting performance significantly. 

# Description of the solution

In this PR we include 3 new parameters that allow to enable or disable all the terms involving the hessian. The first two allow to do so for the jacobian and the residual of the linear system. The other one, allows to do the same for the jacobian of the multigrid preconditioner operators.

# How Has This Been Tested?

- By default all terms are enabled, so all existent test pass.
- I have tested two problems in my machine and this seems promising in terms of performance. Further testing will be done in the following days.

# Documentation

- The three new parameters and a brief description have been added to the linear solver parameters documentation: `lethe/doc/source/parameters/cfd/linear_solver_control.rst`. It is emphasized that these parameters only work for `lethe-fluid-matrix-free` at the moment.

### DIFF
--- a/doc/source/parameters/cfd/linear_solver_control.rst
+++ b/doc/source/parameters/cfd/linear_solver_control.rst
@@ -117,6 +117,16 @@ When set to ``extra verbose``, the residual at each iteration of the linear solv
 .. caution:: 
 		Be aware that the setup of the ``amg`` preconditioner is very expensive and does not scale linearly with the size of the matrix. As such, it is generally preferable to minimize the number of assembly of such preconditioner. This can be achieved by using the ``inexact newton`` for the nonlinear solver (see :doc:`non-linear_solver_control`).
 
+* There are two additional parameters that can be used in this subsection that only work for the ``lethe-fluid-matrix-free`` application at the moment. They allow to turn on or off the hessian terms present in the Jacobian and the residual (or right-hand side) of the Navier-Stokes problem:
+
+.. code-block:: text
+
+    set enable hessians in jacobian = true
+    set enable hessians in rhs      = true
+
+.. caution::
+   This is useful for performance reasons, however, it highly depends on the problem being solver and must be used carefully.
+
 In addition to the method parameters, one can also set specific parameters for each of the preconditioners by adding specific lines inside of the specific physics subsection:
 
 -------------------
@@ -189,9 +199,11 @@ Different parameters for the main components of the two geometric multigrid algo
 .. code-block:: text
 
     # General MG parameters
-    set mg verbosity       = quiet
-    set mg min level       = -1
-    set mg level min cells = -1
+    set mg verbosity                   = quiet
+    set mg min level                   = -1
+    set mg level min cells             = -1
+    set mg enable hessians in jacobian = true
+    set mg enable hessians in rhs      = true
 
     # Relaxation smoother parameters
     set mg smoother iterations     = 10
@@ -238,3 +250,6 @@ Different parameters for the main components of the two geometric multigrid algo
 
 .. tip::
   If your coarse-grid level is small enough, it might be worth it for some problems to set ``mg amg use default parameters = true`` to use a direct solver. On the other hand, if high order elements are used, it might be useful to set ``set mg coarse grid use fe q iso q1 = true`` to solve the coarse grid problem using `FE_Q_iso_Q1 elements <https://www.dealii.org/developer/doxygen/deal.II/classFE__Q__iso__Q1.html>`_.
+
+.. tip::
+  Evaluating terms involving the hessian is expensive. Therefore, one can turn on or off those terms in the mg level operators to improve performance by setting ``mg enable hessians in jacobian`` and ``mg enable hessians in rhs`` equal to ``false``. This is useful for certain problems and must be used carefully.
--- a/doc/source/parameters/cfd/linear_solver_control.rst
+++ b/doc/source/parameters/cfd/linear_solver_control.rst
@@ -203,7 +203,6 @@ Different parameters for the main components of the two geometric multigrid algo
     set mg min level                   = -1
     set mg level min cells             = -1
     set mg enable hessians in jacobian = true
-    set mg enable hessians in residual = true
 
     # Relaxation smoother parameters
     set mg smoother iterations     = 10
@@ -252,4 +251,4 @@ Different parameters for the main components of the two geometric multigrid algo
   If your coarse-grid level is small enough, it might be worth it for some problems to set ``mg amg use default parameters = true`` to use a direct solver. On the other hand, if high order elements are used, it might be useful to set ``set mg coarse grid use fe q iso q1 = true`` to solve the coarse grid problem using `FE_Q_iso_Q1 elements <https://www.dealii.org/developer/doxygen/deal.II/classFE__Q__iso__Q1.html>`_.
 
 .. tip::
-  Evaluating terms involving the hessian is expensive. Therefore, one can turn on or off those terms in the mg level operators to improve performance by setting ``mg enable hessians in jacobian`` and ``mg enable hessians in residual`` equal to ``false``. This is useful for certain problems and must be used carefully.
+  Evaluating terms involving the hessian is expensive. Therefore, one can turn on or off those terms in the mg level operators to improve performance by setting ``mg enable hessians in jacobian`` to ``false``. This is useful for certain problems and must be used carefully.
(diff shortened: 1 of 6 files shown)


source/core/parameters.cc
```diff
@@ -2469,10 +2488,12 @@ namespace Parameters
...
-        relative_residual  = prm.get_double("relative residual");
-        minimum_residual   = prm.get_double("minimum residual");
-        max_iterations     = prm.get_integer("max iters");
-        max_krylov_vectors = prm.get_integer("max krylov vectors");
+        relative_residual        = prm.get_double("relative residual");
+        minimum_residual         = prm.get_double("minimum residual");
+        max_iterations           = prm.get_integer("max iters");
+        max_krylov_vectors       = prm.get_integer("max krylov vectors");
+        enable_hessians_jacobian = prm.get_bool("enable hessians in jacobian");
+        enable_hessians_residual = prm.get_bool("enable hessians in residual");
```
peterrum, line 2496:
```suggestion
        enable_hessians_jacobian = prm.get_bool("enable hessians in jacobian");
        enable_hessians_residual = prm.get_bool("enable hessians in residual");
        
        Assert(enable_hessians_residual || !enable_hessians_jacobian);
```

source/core/parameters.cc
```diff
@@ -2509,6 +2531,8 @@ namespace Parameters
 
         mg_min_level       = prm.get_integer("mg min level");
         mg_level_min_cells = prm.get_integer("mg level min cells");
+        mg_enable_hessians_jacobian =
+          prm.get_bool("mg enable hessians in jacobian");
```
peterrum, line 2535:
```suggestion
        mg_enable_hessians_jacobian =
          prm.get_bool("mg enable hessians in jacobian");
        Assert(enable_hessians_jacobian || !mg_enable_hessians_jacobian);
```

peterrum (approved):

doc/source/parameters/cfd/linear_solver_control.rst
```diff
@@ -117,6 +117,16 @@ When set to ``extra verbose``, the residual at each iteration of the linear solv
 .. caution:: 
 		Be aware that the setup of the ``amg`` preconditioner is very expensive and does not scale linearly with the size of the matrix. As such, it is generally preferable to minimize the number of assembly of such preconditioner. This can be achieved by using the ``inexact newton`` for the nonlinear solver (see :doc:`non-linear_solver_control`).
 
+* There are two additional parameters that can be used in this subsection that only work for the ``lethe-fluid-matrix-free`` application at the moment. They allow to turn on or off the hessian terms present in the Jacobian and the residual (or right-hand side) of the Navier-Stokes problem:
+
+.. code-block:: text
+
+    set enable hessians in jacobian = true
+    set enable hessians in residual = true
```
blaisb, line 125: Is it not highly problematic to remove the hessian in the residual ? I would fear that this would quickly lead to a variational inconsistency.

doc/source/parameters/cfd/linear_solver_control.rst
```diff
@@ -189,9 +199,10 @@ Different parameters for the main components of the two geometric multigrid algo
 .. code-block:: text
 
     # General MG parameters
-    set mg verbosity       = quiet
-    set mg min level       = -1
-    set mg level min cells = -1
+    set mg verbosity                   = quiet
+    set mg min level                   = -1
+    set mg level min cells             = -1
+    set mg enable hessians in jacobian = true
```
blaisb, line 205: Here the other parameter default value should be specified
set enable hessians in residual = true

doc/source/parameters/cfd/linear_solver_control.rst
```diff
@@ -237,4 +248,7 @@ Different parameters for the main components of the two geometric multigrid algo
...
 
 .. tip::
-  If your coarse-grid level is small enough, it might be worth it for some problems to set ``mg amg use default parameters = true`` to use a direct solver. On the other hand, if high order elements are used, it might be useful to set ``set mg coarse grid use fe q iso q1 = true`` to solve the coarse grid problem using `FE_Q_iso_Q1 elements <https://www.dealii.org/developer/doxygen/deal.II/classFE__Q__iso__Q1.html>`_.
\ No newline at end of file
+  If your coarse-grid level is small enough, it might be worth it for some problems to set ``mg amg use default parameters = true`` to use a direct solver. On the other hand, if high order elements are used, it might be useful to set ``set mg coarse grid use fe q iso q1 = true`` to solve the coarse grid problem using `FE_Q_iso_Q1 elements <https://www.dealii.org/developer/doxygen/deal.II/classFE__Q__iso__Q1.html>`_.
+
+.. tip::
+  Evaluating terms involving the hessian is expensive. Therefore, one can turn on or off those terms in the mg level operators to improve performance by setting ``mg enable hessians in jacobian`` to ``false``. This is useful for certain problems and must be used carefully.
```
blaisb, line 254: I feel like this tip is a bit a repetition of the above one, but if you feel this is necessary I don,t mind :)

include/core/parameters.h
```diff
@@ -1064,6 +1063,12 @@ namespace Parameters
     /// Maximum number of krylov vectors
     int max_krylov_vectors;
 
+    /// Enable hessians in jacobian
+    bool enable_hessians_jacobian;
```
blaisb, line 1067: why is the parameter repeated twice?

include/solvers/mf_navier_stokes_operators.h
```diff
@@ -410,6 +423,18 @@ class NavierStokesOperatorBase : public Subscriptor
...
+   * @brief Flag to turn hessian terms from jacobian on or off.
+   *
+   */
+  bool enable_hessians_jacobian;
+
+  /**
+   * @brief Flag to turn hessian terms from residual on or off.
+   *
+   */
+  bool enable_hessians_residual;
```
blaisb, line 436: You could use /// instead here. I don't think  you need a brief because this is not a function

source/solvers/mf_navier_stokes_operators.cc
```diff
@@ -945,26 +968,29 @@ NavierStokesStabilizedOperator<dim, number>::do_cell_integral_local(
             {
               for (unsigned int k = 0; k < dim; ++k)
                 {
-                  for (unsigned int l = 0; l < dim; ++l)
+                  if (this->enable_hessians_jacobian)
```
blaisb, line 971: Clearly this must have a significant impact on performance...

blaisb (changes requested):
I'm not 100% sure what you are neglecting when you remove the hessian from the RHS. You neglect it in the vmult on the coarser levels, but keep it on the finer level? Is it still variationnaly consistent ?

include/solvers/mf_navier_stokes_operators.h
lpsaavedra: That's true... but all this class was documented before we knew that we could use ///. So, if you are okay, I will leave it like this and then update all the files that have member variables documented in this fashion in one PR.

doc/source/parameters/cfd/linear_solver_control.rst
lpsaavedra: There is no residual parameter in this case. So, there are three parameters, enable hessians in jacobian, enable hessians in residual and mg enable hessians in jacobian. The first two can enable them for the system operator, while this one only considers the operators of the mg preconditioner.

doc/source/parameters/cfd/linear_solver_control.rst
lpsaavedra: For the same reason as above, this is a different parameter as the parameter above.

include/core/parameters.h
lpsaavedra: Same explanation as in the comments above.

doc/source/parameters/cfd/linear_solver_control.rst
lpsaavedra: Well, in principle, we could leave the hessians in both the jacobian and the residual of the system operator and disable them completely for the mg operator. Here, we would be solving the exact same problem with a slighlty different preconditioner, which can lead to more iterations but if they are not that expensive, that should not be an issue. As for removing the hessians in the system operator, we would have to test more problems and check. I find it nice to have the option to turn them on and off to experiment a bit. We could maybe not write them in the documentation and test them further until we are really sure about their behavior

include/solvers/mf_navier_stokes_operators.h
blaisb: Sure thing.  I'm ok with this

blaisb (approved):

PR #1000, opened 2024-01-26
Neumann BC for heat

voferreira:
# Description of the problem

- Lethe did not have a constant heat flux (Neumann) BC

# Description of the solution

- Implemented a heat flux boundary condition

# How Has This Been Tested?

- 2D jet case
- [x] Add application test

# Documentation

- Added the parameters to the documentation

# Future changes

- Make it space dependent as well as emissivity, h, and Tinf
- Needs some cleaning
- Make an example

PR #1128, opened 2024-05-11
Robin/Neumann BC parameters were not space dependent

voferreira:
# Description of the problem

- The convection-radiation-flux boundary condition had function expression parameters that were not space dependent.

# Description of the solution

- Made it space-dependent using the function evaluation at the quadrature point position.

# How Has This Been Tested?

- Simple 2D heat flux case. The test was added to the application tests.

# Documentation

- Remove disclaimer from parameters guide regarding the no longer existent limitation.

# Comments

- This PR is a follow up of PR #1000 .
- There is no way to have ``set number of fluids = 0``. This was added to the maintenance project and I will fix it in another PR.


applications_tests/lethe-fluid/non-uniform_heat_flux_bc.output
```diff
@@ -0,0 +1,24 @@
...
+	 Std-Dev : 10.1926
+
+*****************************
+Steady iteration:        1/1
+*****************************
+--------------
+Heat Transfer
+--------------
+  -Tolerance of iterative solver is : 9.00917e-05
+  -Iterative solver took : 16 steps to reach a residual norm of 4.55329e-05
```
blaisb, line 19: Tolerances and linear solver verbosity should not be in the test because these are not portable :)

applications_tests/lethe-fluid/non-uniform_heat_flux_bc.prm
```diff
@@ -0,0 +1,131 @@
...
+    set type = convection-radiation-flux
+  end
+end
+
+subsection post-processing
+  set verbosity                        = verbose
+  set output frequency                 = 100000
+  set calculate temperature statistics = true
+  set temperature statistics name = non_uniform_heat_flux
+end
```
blaisb, line 87: Can you run prm indent, seems like it needs to be run :)

applications_tests/lethe-fluid/non-uniform_heat_flux_bc.prm
```diff
@@ -0,0 +1,131 @@
...
+  subsection heat transfer
+    set verbosity                             = verbose
+    set method                                = gmres
+    set max iters                             = 5000
+    set relative residual                     = 1e-7
+    set minimum residual                      = 1e-8
+    set preconditioner                        = ilu
+    set ilu preconditioner fill               = 0
+    set ilu preconditioner absolute tolerance = 1e-9
+    set ilu preconditioner relative tolerance = 1.00
```
blaisb, line 121:
```suggestion

```
default parameters are ok here I think

applications_tests/lethe-fluid/non-uniform_heat_flux_bc.prm
```diff
@@ -0,0 +1,131 @@
...
+end
+
+#---------------------------------------------------
+# Non-Linear Solver Control
+#---------------------------------------------------
+
+subsection non-linear solver
+  subsection heat transfer
+    set verbosity      = quiet
+    set tolerance      = 1e-1
```
blaisb, line 96: This is very big for a tolerance, why is it so big? I would lower it :)

applications_tests/lethe-fluid/non-uniform_heat_flux_bc.prm
blaisb: Maybe a good idea to check some other sections to keep it to the minimal, at least as best as you can

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -399,9 +402,17 @@ HeatTransferAssemblerRobinBC<dim>::assemble_matrix(
                   for (unsigned int q = 0; q < scratch_data.n_faces_q_points;
                        ++q)
                     {
+                      Point<dim> quadrature_point_on_face;
+                      for (unsigned int d = 0; d < dim; d++)
+                        quadrature_point_on_face[d] =
+                          scratch_data.quadrature_points[q][d];
```
blaisb, line 408: Why do you need to make a local copy of the points? why not just juse teh points directly?

blaisb (changes requested):
Very cool work. Since it'S not a lot of work and while you are at it, cna you also ensure that the set-time is called also to set up the current value of the time?

voferreira:
@blaisb Thanks! I thought set_time was called on update_boundary_conditions, but the if in the integration was preventing it from being called. This is fixed now. Plus I have placed the assertion for the emissivity at the right place, I believe. I also added a transient test for heat_flux = 10*t*y.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -471,6 +484,16 @@ HeatTransferAssemblerRobinBC<dim>::assemble_rhs(
...
                         scratch_data.temperature_face_value[f][q];
                       const double JxW = scratch_data.face_JxW[f][q];
+                      const double h =
+                        h_function.value(scratch_data.quadrature_points[q]);
+                      const double T_inf =
+                        T_inf_function.value(scratch_data.quadrature_points[q]);
+                      const double emissivity = emissivity_function.value(
+                        scratch_data.quadrature_points[q]);
+                      Assert(emissivity <= 1.0 && emissivity >= 0.0,
+                             EmissivityError(emissivity));
```
blaisb, line 494: Since  you are declaring an exception don't you want to call AssertThrow instead of just Assert?

blaisb (approved):
One last comment and then ready for merge

blaisb:
I'll let @mivaia take a look at it and then if he's ok with it I'll merge

mivaia (approved):
All good on my side! Thanks for giving me some time to look into it :)

PR #1130, opened 2024-05-11
Rename source term NS subsection

lpsaavedra:
# Description of the problem

The subsection's name to specify the source term of the Navier-Stokes equations in the parameter file was set to be `xyz`.

# Description of the solution

The subsection was renamed to `navier stokes`, since there are source terms for other physics where the subsection has the physics name, for example, `subsection heat transfer`. The change had to be made in all the parameter files of the examples, the tests and the documentation.


blaisb:
@lpsaavedra I think it's a good idea to change the name, but the physics name used elsewhere for the fluid dynamics is "fluid dynamics" and not navier-stokes. I think it would be a better idea to used fluid dynamics because this source term is also used in the VANS equations for example

lpsaavedra:
@blaisb that makes a lot of sense. I have now updated the name of the section to fluid dynamics. I also found some parts in the documentation (multiphysics examples only) were the "enable" parameter was mentioned in the source term section. However, this parameter was deleted a while ago, so I have corrected that as well.

blaisb (approved):

PR #1139, opened 2024-05-15
One way coupling MF and heat transfer

lpsaavedra:
# Description of the problem

The `lethe-fluid-matrix-free` application was not coupled to any other physics. 

# Description of the solution

This PR enables the one way coupling between fluid dynamics and heat transfer using the existent multiphysics interface.

# How Has This Been Tested?
One test was added to verify the coupling. It is the same one that is used in the matrix-based application:
- [X] `applications_tests/lethe-fluid-matrix-free/heat_transfer_high_pe_stab_gls.prm`

### DIFF
--- a/applications_tests/lethe-fluid-matrix-free/heat_transfer_high_pe_stab_gls.prm
+++ b/applications_tests/lethe-fluid-matrix-free/heat_transfer_high_pe_stab_gls.prm
@@ -1,0 +1,194 @@
+# Listing of Parameters
+#----------------------
+
+set dimension = 2
+
+#---------------------------------------------------
+# Simulation Control
+#---------------------------------------------------
+
+subsection simulation control
+  set method            = steady
+  set number mesh adapt = 2
+  set output name       = stab_adv_diff
+  set output frequency  = 0
+  set subdivision       = 1
+end
+
+#---------------------------------------------------
+# FEM
+#---------------------------------------------------
+
+subsection FEM
+  set velocity order    = 1
+  set pressure order    = 1
+  set temperature order = 1
+end
+
+#---------------------------------------------------
+# Source term
+#---------------------------------------------------
+
+subsection source term
+  subsection fluid dynamics
+    set enable = false
+  end
+end
+
+#---------------------------------------------------
+# Timer
+#---------------------------------------------------
+
+subsection timer
+  set type = none # <none|iteration|end>
+end
+
+#---------------------------------------------------
+# Initial condition
+#---------------------------------------------------
+
+subsection initial conditions
+  set type = nodal
+  subsection uvwp
+    set Function expression = 1; 0; 0
+  end
+end
+
+#---------------------------------------------------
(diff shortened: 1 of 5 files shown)


include/solvers/mf_navier_stokes.h
```diff
@@ -420,6 +419,13 @@ class MFNavierStokesSolver
   void
   print_mg_setup_times();
 
+
```
voferreira, line 422:
```suggestion
```

voferreira (approved):
All clear. I don't think I have anything to add. Thank you, Laura and Mikael, for the great work.

applications_tests/lethe-fluid-matrix-free/heat_transfer_high_pe_stab_gls.prm
```diff
@@ -0,0 +1,194 @@
+# Listing of Parameters
+#----------------------
+
+set dimension = 2
+
+#---------------------------------------------------
```
blaisb, line 6: I think it was a good idea to make a unit test with this. Great thinking.

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +44,24 @@
...
 #include <deal.II/numerics/vector_tools.h>
 
+/**
+ * @brief Helper function that allows to convert deal.II vectors to Trilinos vectors.
+ *
+ * @tparam number Abstract type for number across the class (i.e., double).
+ * @param out Destination TrilinosWrappers::MPI::Vector vector.
+ * @param in Source LinearAlgebra::distributed::Vector<number> vector.
+ */
+template <typename number>
```
blaisb, line 54: Could this be moved outside of the MF files? I feel like this could be useful in many placed

blaisb (approved):
1-2 minor comments to address. Great work !!!

source/solvers/mf_navier_stokes.cc
lpsaavedra: We could move it but I think we should leave it here for now... this is the only place where it is used and eventually we will remove it once multiphysics support also `LinearAlgebra::distributed::Vector`

include/solvers/mf_navier_stokes.h
```diff
@@ -420,6 +419,12 @@ class MFNavierStokesSolver
   void
   print_mg_setup_times();
 
+  /**
+   * @brief  Provide relevant solutions to multiphysics interface.
```
mivaia, line 423:
```suggestion
   * @brief  Provide present and previous flow solutions to the multiphysics interface. These solutions can be accessed through the multiphysics interface by the different physics. 
```

mivaia (approved):
Nice job! Only a single comment :)

source/solvers/mf_navier_stokes.cc
blaisb: OKidoke

PR #1152, opened 2024-05-21
Increase core count to 4 for CI

blaisb:
# Description of the problem

- Public github actions now have 4 cores instead of 2 like in the past. They should be leveraged

# Description of the solution

- Increase the core count to 4 for serial jobs and compilation

# How Has This Been Tested?

- Will it work? we shall see

EDIT
It seems to remove 10-15 minutes from the CI. I like this

### DIFF
--- a/.github/workflows/main_parameter_files.yml
+++ b/.github/workflows/main_parameter_files.yml
@@ -17,7 +17,7 @@ concurrency:
   cancel-in-progress: true
   
 env:
-  COMPILE_JOBS: 2
+  COMPILE_JOBS: 4
   MULTI_CORE_TESTS_REGEX: "mpirun=2"
 
 jobs:
--- a/.github/workflows/main_release.yml
+++ b/.github/workflows/main_release.yml
@@ -18,7 +18,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  COMPILE_JOBS: 2
+  COMPILE_JOBS: 4
   MULTI_CORE_TESTS_REGEX: "mpirun=2"
 
 jobs:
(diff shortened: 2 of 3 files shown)


OGaboriault:
So it compile with 4 cores instead of 2? Just for Release part?

With Maxwell, the bottle neck is still the Debug part, right?

Is it possible to run the multicore tests first?

OGaboriault (approved):

blaisb:
> So it compile with 4 cores instead of 2? Just for Release part?
> 
> With Maxwell, the bottle neck is still the Debug part, right?
> 
> Is it possible to run the multicore tests first?



> So it compile with 4 cores instead of 2? Just for Release part?

All jobs compile with 4 cores now other than the Debug which runs on our own machines


> 
> With Maxwell, the bottle neck is still the Debug part, right?

Yes and no. The issue is we only have one CI machine for the Debug. If we find a second machine as efficient as Maxwell to dedicate to the CI, this won,t be an issue.


> 
> Is it possible to run the multicore tests first?

Yes, but I am not sure if there will be any benefit. Feel free to make a PR to try it out though, no worries.

OresteMarquis:
I don't remember how many cores Maxwell have? Should the github actions use the maximum number of core or only a part of it?

blaisb:
> I don't remember how many cores Maxwell have? Should the github actions use the maximum number of core or only a part of it?

Maxwell has 12 cores but 24 threads. In general compiling with threads is sometimes better, but running with threads is not. In the case of Maxwell I already optimized and tested all that.
@OresteMarquis  Accept the pr (with the green checkmark) when you are ok with it :)

OresteMarquis (approved):

PR #1163, opened 2024-05-29
Add refinement at frequency tests

oguevremont:
# Description of the problem

- The "refinement_at_frequency" member is set by setting the frequency to 0.
- There was not test for this. 

# Description of the solution

- 2 tests were added.

# How Has This Been Tested?

- [application_tests/lethe-fluid/refinement_at_frequency_on] With frequency set to 1
- [application_tests/lethe-fluid/refinement_at_frequency_on] With frequency set to 0

# Comments

- I would have preferred to have a true unit test for this instead of an application test, because with these changes the other physics are not tested. I'll keep looking for a way to test the parser itself.


OGaboriault:
Regarding the test instead of an application test, what is the problem?

blaisb (changes requested):
It's allright to have an application test for now. Can youi apply prmindent to the prm files?

oguevremont:
> Regarding the test instead of an application test, what is the problem?

The problem is that the boolean variable "refinement_at_frequency" is set in the parsing function when "frequency" is set to 0, so we cannot test that it is properly set unless we do the parsing.

OGaboriault:
> > Regarding the test instead of an application test, what is the problem?
> 
> The problem is that the boolean variable "refinement_at_frequency" is set in the parsing function when "frequency" is set to 0, so we cannot test that it is properly set unless we do the parsing.

You can create a fake prm object in a test a think, that could solve your problem (but I'm not sure). But like Bruno said it is okay for now. :)

applications_tests/lethe-fluid/refinement_at_frequency_on.prm
```diff
@@ -0,0 +1,83 @@
...
+    set id   = 0
+    set type = noslip
+  end
+  subsection bc 1
+    set id   = 1
+    set type = noslip
+  end
+  subsection bc 2
+    set id   = 2
+    set type = noslip
```
OGaboriault, line 37: Not related to this PR specifically (can coming from someone who never uses the CFD solvers) but "noslip" shouldn't be written "no_slip" ?

OGaboriault (approved):
No real comments. Good job

applications_tests/lethe-fluid/refinement_at_frequency_on.prm
oguevremont: Yeah or even "no slip". I'm not sure if there is a reason why it is like this, but it could be changed

blaisb: could be no_slip indeed. It's an historic thing

blaisb (approved):

PR #1164, opened 2024-05-30
Dem force chains

cleodeletre:
# Description of the problem

Need for a post-processing feature to visualize force chains between particles.This code enables the visualization of force chains for each time step with a vtu file. It is noted that the parallelization is not completely done. Ghost particles are not taken into account for the moment. 

# Description of the solution

Modification of parameters_lagrangian(.cc & .h) : Add a force_chains entry variable for users to enable the post-processing feature in the parameter file (.prm)
Creation of force_chains_visualization(.cc & .h) : New class and its methods to calculate the normal force between particles and write vtu files to display force chains.
Modification of dem.cc : Call force_chains_visualization methods to calculate normal forces and write vtu files.

# How Has This Been Tested?

Tested on two dem examples: Small Scale Rotating Drum & Rectangular Hopper

# Future changes

- Parallelize the feature.

# Comments

- source/dem/test_cd_fc.cc should not be in this PR.


include/core/parameters_lagrangian.h
```diff
@@ -474,6 +474,8 @@ namespace Parameters
     {
       // A bool variable which sets-up the Lagrangian post-processing
       bool Lagrangian_post_processing;
+      // A bool variable which sets-up the force chains visualization
```
blaisb, line 477:
```suggestion
      /// A bool variable which sets-up the force chains visualization
```

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <boost/range/adaptor/map.hpp>
+
+#include <vector>
+
+using namespace dealii;
+using namespace DEM;
+
+#ifndef particles_force_chains_v2_h
+#  define particles_force_chains__v2_h
```
blaisb, line 21: Fix the name to remove the _v2

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,61 @@
...
+template <
+  int                                                       dim,
+  Parameters::Lagrangian::ParticleParticleContactForceModel force_model,
+  Parameters::Lagrangian::RollingResistanceMethod rolling_friction_model>
+class ParticlesForceChains
+  : public ParticleParticleContactForce<dim,
+                                        force_model,
+                                        rolling_friction_model>
+{
+public:
```
blaisb, line 34: You do not know it yet so that's not a big deal, but try to add a @brief and a description of all the [in] arguments. You can look at other files to see how this is done :)

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,61 @@
...
+
+  void
+  calculate_force_chains(DEMContactManager<dim> &container_manager,
+                         const double            dt);
+
+
+  void
+  write_force_chains(MPI_Comm           mpi_communicator,
+                     const std::string  folder,
+                     const unsigned int iter);
```
blaisb, line 52: The MPI_Comm should be const

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,61 @@
...
+{
+public:
+  ParticlesForceChains(const DEMSolverParameters<dim> &dem_parameters);
+
+  virtual ~ParticlesForceChains()
+  {}
+
+  void
+  multi_general_cell(Triangulation<1, 3>         &tria,
+                     const std::vector<Point<3>> &vertices);
```
blaisb, line 42: Reference to the triangulation could be const :)

include/dem/preliminary_force_chain_visualization.h
blaisb, line 1: You can remove this file i believe :)

source/core/parameters_lagrangian.cc
```diff
@@ -1395,6 +1395,11 @@ namespace Parameters
           "false",
           Patterns::Bool(),
           "State whether Lagrangian post-processing should be performed.");
+        prm.declare_entry(
+          "force chains post-processing",
```
blaisb, line 1399: no need for the post-processing because this is already in a post-processing section
I would just write force chains

source/dem/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ add_library(lethe-dem
   find_boundary_cells_information.cc
   find_cell_neighbors.cc
   find_contact_detection_step.cc
+  force_chains_visualization.cc
```
blaisb, line 13: Thanks for putting this in alphabetical order, appreciate it :)

source/dem/dem.cc
```diff
@@ -1057,6 +1058,21 @@ DEMSolver<dim>::write_output_results()
...
+        dim,
+        Parameters::Lagrangian::ParticleParticleContactForceModel::hertz,
+        Parameters::Lagrangian::RollingResistanceMethod::no_resistance>
+        particles_force_chains_object(parameters);
+      particles_force_chains_object.calculate_force_chains(
+        contact_manager, simulation_control->get_time_step());
+
+      particles_force_chains_object.write_force_chains(this->mpi_communicator,
+                                                       folder,
+                                                       iter);
```
blaisb, line 1074: Right now this is hardcoded to use only a single collision model. It is important to refactor this to enable all variants of the force model. Feel free to ask questions regarding this if you need help.

source/dem/dem.cc
```diff
@@ -1448,6 +1464,8 @@ DEMSolver<dim>::solve()
           force,
           periodic_offset);
 
+
+
```
blaisb, line 1468: Both lines can be removed :)!

source/dem/preliminary_force_chain_visualization.cc
```diff
@@ -0,0 +1,95 @@
+
+#include <core/manifolds.h>
+#include <core/solutions_output.h>
+
+#include <dem/data_containers.h>
```
blaisb, line 5: Is this file still necessary? You are not using this code, in that case, I would remove it :)

source/dem/test_cd_fc.cc
```diff
@@ -0,0 +1,66 @@
...
+  triangulation.refine_global(4);
+
+  std::ofstream out("hyper_rectangle.svg");
+  GridOut       grid_out;
+  grid_out.write_svg(triangulation, out);
+  std::cout << "Grid written to hyper_rectangle.svg" << std::endl;
+}
+
+void
+generalcell_grid()
```
blaisb, line 47:
```suggestion
general_cell_grid()
```

source/dem/test_cd_fc.cc
```diff
@@ -0,0 +1,66 @@
...
+{
+  Triangulation<1>            triangulation;
+  const std::vector<Point<1>> vertice<>;
```
blaisb, line 50: why do you have <> at the end of vertice?
Should also be vertices

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+  Parameters::Lagrangian::RollingResistanceMethod rolling_friction_model>
+ParticlesForceChains<dim, contact_model, rolling_friction_model>::
+  ParticlesForceChains(const DEMSolverParameters<dim> &dem_parameters_in)
+  : ParticleParticleContactForce<dim, contact_model, rolling_friction_model>(
+      dem_parameters_in)
+  , dem_parameters(dem_parameters_in)
+{
+  force_normal.push_back(0);
+  vertices.push_back(Point<3>(0, 0, 0));
+  vertices.push_back(Point<3>(0, 0, 0));
```
blaisb, line 50: it's allright to initialize to make sure it's never empty, but is this 100% sure necessary?

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+
+
+
+template <
+  int                                                       dim,
+  Parameters::Lagrangian::ParticleParticleContactForceModel contact_model,
+  Parameters::Lagrangian::RollingResistanceMethod rolling_friction_model>
+void
+ParticlesForceChains<dim, contact_model, rolling_friction_model>::
+  multi_general_cell(Triangulation<1, 3>         &tria,
```
blaisb, line 61: Forget by comment, this cannot be const.

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+{
+  const unsigned int       n_cells = vertices.size() / 2;
+  std::vector<CellData<1>> cells(n_cells, CellData<1>());
+  for (unsigned int i = 0; i < n_cells; ++i)
+    {
+      for (unsigned int j = 0; j < 2; ++j)
+        {
+          cells[i].vertices[j] = 2 * i + j;
+          cells[i].material_id = 0;
+        }
```
blaisb, line 72: since this loop is always 2, do we really need it? You could replace it with just the 3 lines.

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+
+  auto &local_adjacent_particles = container_manager.local_adjacent_particles;
+  // auto &ghost_adjacent_particles =
+  // container_manager.ghost_adjacent_particles; auto
+  // &local_periodic_adjacent_particles =
+  //  container_manager.local_periodic_adjacent_particles;
+  // auto &ghost_periodic_adjacent_particles =
+  //  container_manager.ghost_periodic_adjacent_particles;
+  // auto &ghost_local_periodic_adjacent_particles =
+  //  container_manager.ghost_local_periodic_adjacent_particles;
```
blaisb, line 100: If this is not necessary, you can remove those :)

blaisb (changes requested):
First batch of comments, will continue tomorrow :)!

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,61 @@
...
+
+  void
+  write_force_chains(MPI_Comm           mpi_communicator,
+                     const std::string  folder,
+                     const unsigned int iter);
+
+private:
+  std::vector<double>             force_normal;
+  std::vector<Point<3>>           vertices;
+  const DEMSolverParameters<dim> &dem_parameters;
```
OGaboriault, line 57: Add a small description of every attribute in the class. Doesn't need to be long.

Also, do you really need to have `dem_parameters` as an attribute of the class? I think the `ParticleParticleContactForce `object could be an attribute build in the constructor (but maybe I'm wrong)

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+#include <deal.II/numerics/data_out.h>
+
+#include <cmath>
+#include <fstream>
+#include <iostream>
+
+using namespace DEM;
+
+
+
```
OGaboriault, line 37:
```suggestion
```

OGaboriault: Remove extra lines

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+  : ParticleParticleContactForce<dim, contact_model, rolling_friction_model>(
+      dem_parameters_in)
+  , dem_parameters(dem_parameters_in)
+{
+  force_normal.push_back(0);
+  vertices.push_back(Point<3>(0, 0, 0));
+  vertices.push_back(Point<3>(0, 0, 0));
+}
+
+
```
OGaboriault, line 53: Remove extra lines

source/dem/force_chains_visualization.cc
OGaboriault: I told her to keep those since she soon be working on the ghost cells... my fault

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+          Point<3> particle_one_location = [&] {
+            if constexpr (dim == 3)
+              {
+                return particle_one->get_location();
+              }
+            if constexpr (dim == 2)
+              {
+                return (point_nd_to_3d(particle_one->get_location()));
+              }
+          }();
```
OGaboriault, line 141: Nice uses of the lambda function, but I don't think it is necessary. You can just use point_nd_to_3d even when dim == 3

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,511 @@
...
+              Point<3> particle_two_location = [&] {
+                if constexpr (dim == 3)
+                  {
+                    return particle_two->get_location();
+                  }
+                if constexpr (dim == 2)
+                  {
+                    return (point_nd_to_3d(particle_two->get_location()));
+                  }
+              }();
```
OGaboriault, line 161: Same here

source/dem/test_cd_fc.cc
```diff
@@ -0,0 +1,66 @@
...
+#include <deal.II/grid/tria.h>
+
+#include <cmath>
+#include <fstream>
+#include <iostream>
+
+
+using namespace dealii;
+
+
```
OGaboriault, line 12: extra lines

OGaboriault (changes requested):
Good work for your first PR!

include/dem/force_chains_visualization.h
blaisb: actually it cannot be const, that's my bad.

source/dem/force_chains_visualization.cc
blaisb: No it's allright, we can keep it. Just add a comment asto why we are keeping this for now.

source/dem/force_chains_visualization.cc
blaisb: no let's leave it this way. We do this everywhere in the contact code, let's not change it please.

source/dem/force_chains_visualization.cc
blaisb: Please don't leave it like it is now, this is the right way.

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,500 @@
...
+
+              if (normal_overlap > 0.0)
+                {
+                  // This means that the adjacent particles are in contact
+
+
+
+                  // Since the normal overlap is already calculated, we update
+                  // this element of the container here. The rest of information
+                  // are updated using the following function
```
blaisb, line 167: there seems to be a lot of empty lines here for no reason.

source/dem/force_chains_visualization.cc
```diff
@@ -0,0 +1,500 @@
...
+                  vertices.push_back(particle_two_location);
+                  force_normal.push_back(sqrt(normal_force.norm()));
+                }
+
+              else
+                {
+                  // if the adjacent pair is not in contact anymore, only the
+                  // tangential overlap is set to zero
+                  contact_info.tangential_overlap.clear();
+                }
```
blaisb, line 304: this is not important in this case because you don't have a tangential overlap .

blaisb (approved):
just 2-3 last comments. Nothing big. Could you add an application test even if it won't deplay anything? Just modify an existing test and add a force chain output. This will enable us to ensure that at least the feature does not scratch. @OGaboriault can give you a hand with this. It can be something very small / minor.

blaisb:
Will merge after @OGaboriault accepts

include/core/parameters_lagrangian.h
```diff
@@ -474,6 +474,8 @@ namespace Parameters
     {
       // A bool variable which sets-up the Lagrangian post-processing
       bool Lagrangian_post_processing;
+      /// A bool variable which sets-up the force chains visualization
```
OGaboriault, line 477:
```suggestion
      // A bool variable which sets-up the force chains visualization
```

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,132 @@
...
+   * vectors.
+   *
+   * @param container_manager The container manager object that contains
+   * containers to modify of contact pair periodic candidates with other
+   * containers with periodic neighbors lists
+   * @param dt DEM time step
+   */
+  virtual void
+  calculate_force_chains(DEMContactManager<dim> &container_manager,
+                         const double            dt) = 0;
```
OGaboriault, line 47: I think this function doesn't need the `dt` as a parameter. 

I would delete this parameter. You use `dt` at line 165 for `update_contact_information` function which is used to update the tangential overlap. The tangential forces are not being used for the forces chains.

 I would just replace the `dt` by `0.` when calling `update_contact_information`.

OGaboriault (changes requested):
Tiny changes

OGaboriault (approved):

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,130 @@
...
+ * ParticlesForceChains class which is templated by the contact model
+ * type.
+ */
+template <int dim>
+class ParticlesForceChainsBase
+{
+public:
+  /**
+   * @brief Use a ParticleParticleContactForce object to calculate normal forces between
+   * all touching particles. Stock normal forces and particles position in
```
blaisb, line 37:
```suggestion
   * all touching particles. Store normal forces and particles position in
```

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,129 @@
...
+ * @brief Class that carries out the calculation of
+ * particle-particle contact force and the visualization of force chains
+ * by writing vtu files. Instead of using a inheritance hiearchy to distinguish
+ * between the contact model, the class is templated with the type of force
+ * model and rolling friction model. Consequently, the code for each combination
+ * of force model is generated at compile time.
+ *
+ * @tparam dim The dimension of the problem
+ * @tparam force_model The particle-particle contact force model
+ * @tparam rolling_friction_model The rolling resistance model used
```
blaisb, line 69:
```suggestion
 * @tparam rolling_friction_model The rolling resistance model
```

include/dem/force_chains_visualization.h
```diff
@@ -0,0 +1,129 @@
...
+  void
+  write_force_chains(const MPI_Comm     mpi_communicator,
+                     const std::string  folder,
+                     const unsigned int iter) override;
+
+private:
+  // vector of normal forces between each touching particles.
+  std::vector<double> force_normal;
+  // vector of positions of touching particles.
+  std::vector<Point<3>> vertices;
```
blaisb, line 125:
```suggestion
  /// vector of normal forces between each touching particles.
  std::vector<double> force_normal;
  /// vector of positions of touching particles.
  std::vector<Point<3>> vertices;
```

blaisb (approved):
I fixed some last minute typos, will merge after CI

PR #1168, opened 2024-06-10
Distribute drag forces cfd-dem simulation

yashuuzi:
### Description

In current code of cfd-dem solver, Each quadrature point have the same value, and only consider the particle inside the cell. 
To solve this problem, I propose to evaluate the drag force at each quadrature point using reference sphere.

1. When calculate the particle-fluid-interaction, stored the beta_drag as each particle's propertied.
2. When assemble the drag force, evaluate drag force at each quadrature point using reference sphere. To do so, we need to consider the neighbor cell like QCM method.

### Documentation

I add the parameter "distribute drag force" to enable and disable this future easily.

I modified only Rong drag model

Code related list:
- [ ] New feature has unit test(s) (preferred) or application test(s)

Pull request related list:
- [ ] Labels are applied


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1208,6 +1209,15 @@ class NavierStokesScratchData
...
 
+  // If distribute force enabled
+  // These are used to get neighbor cells when assemble drag force
+  std::map<unsigned int,
+           std::set<typename DoFHandler<dim>::active_cell_iterator>>
+    vertices_to_cell;
+  std::map<unsigned int,
+           std::set<typename DoFHandler<dim>::active_cell_iterator>>
+    vertices_to_periodic_cell;
+
```
yashuuzi, line 1220: These are unnecessary parameter, should be deleted

blaisb (commented):
We can talk about it more when I get back, but i do not understand why you need a new set of functions for this. You can calculate the drag on the particles themselves and only redistribute them once  you are calculating the action of the forces on the cell. I'm not sure why you need more than just adding an if in the assembly of the equations ?

PR #1175, opened 2024-06-19
Fix partial slip documentation

voferreira:
Quick fix on documentation of `partial slip` boundary condition.


voferreira:
@blaisb I had a problem with the PR comment template, but I figured since it is a small fix it wouldn't be a problem. Let me know it that's an issue and I can try more to fix it.

blaisb (approved):
Good, thanks for the unification

PR #1180, opened 2024-06-26
Fix NS-VOF stong residual formulation

hepap:
### Description

- The term associated with the viscosity jump in the strong residual for the NS-VOF assembler added in PR https://github.com/chaos-polymtl/lethe/pull/1149 led to a ill-posed formulation in phase change cases where the solid is represented by a highly viscous fluid. Hence, we don't consider this term in the current formulation.
- There was no appropriate application test for lpbf/phase change with vof cases that could have allowed us to identify this ill-posed formulation prior to the merge of PR https://github.com/chaos-polymtl/lethe/pull/1149.
- The LPBF benchmark example prm files were almost up to date, but did not exactly matched the one we used to generate the benchmark results. The mesh files were missing from the repo.

### Solution

- This PR removes the term associated with the viscosity jump.
- A new test is now included to avoid future mistake in lpbf/phase change with vof cases.
- LPBF benchmark example prm files were modified and are now up to date with the current working version. The mesh files are now available in the repo.

### Testing

- The new test simulate the lpbf benchmark case using the extra-extra-coarse mesh. Temperature statistics and mass conservation monitoring are the outputted metrics (to track both the change in HT and NS-VOF solvers).
- Other NS-VOF tests were updated, and results for those tests returned to their previous values (or almost with small differences).

### Documentation

No modification in the documentation.

I guess the tests will fail because of the CI issue. I have still opened the PR since the formulation won't change and I will update the restart of the new test when I have updated my P4est version :)

Code related list:
- [x] Fix has unit test(s) (preferred) or application test(s)

Pull request related list:
- [x] Labels are applied


blaisb:
The debug test still seems to crash?

AmishgaAlphonius:
> The debug test still seems to crash?

It's a bit weird, the load function from a checkpoint for the triangulation has an assertion that checks if all cells are at the same level of refinement. In Release there is no issue, but in Debug, it crashes because of this. I tried it locally also and I had the same issue.

blaisb:
> > The debug test still seems to crash?
> 
> It's a bit weird, the `load` from a checkpoint function for the triangulation has an assertion that checks if all cells are at the same level of refinement. In Release there is no issue, but in Debug, it crashes because of this. I tried it locally also and I had the same issue. Is it that assertions are only checked on Debug?

You need to make sure the restart files does NOT refine the triangulation. Generally this is the case for all the solvers (we do not do mesh refinement when we are reading a restart file), however we don't have any restart tests for VOF I guess,so maybe this issue was there all along. You can check a restart with any VOF cse in the debug to see if that's the issue.
But otherwise, the triangulation/ mesh must have the same properties as the one used in the restart (aka the coarse grid has to be the same)

applications_tests/lethe-fluid/heat_transfer_vof_lpbf_benchmark.mpirun=1.output
```diff
@@ -0,0 +1,199 @@
...
+   Number of degrees of freedom: 339
+   Volume of triangulation:      0.36
+   Number of thermal degrees of freedom: 113
+   Number of VOF degrees of freedom: 113
+Initial refinement in box - Step  3 of 3
+   Number of active cells:       268
+   Number of degrees of freedom: 942
+   Volume of triangulation:      0.36
+   Number of thermal degrees of freedom: 314
+   Number of VOF degrees of freedom: 314
```
blaisb, line 24: This part should not be happening in a restart file. Maybe this option just has neverbeen tested with a restart. IF we are restarting, we should not do any mesh adaptation :)

hepap:
So I removed the box refinement for the new test in this PR, and I fixed the restart problem with box ref in PR #1184.

This PR (#1180) can be merge before PR #1184 (I added a test in PR #1184 to verify restart with box ref).

blaisb:
@hepap can you rebase? Then this will be ready to go :)

AmishgaAlphonius (approved):
Great work Hepap! Everything looks good to me! Thanks for the fix! :)

applications_tests/lethe-fluid/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ file(COPY cylinder_structured.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY taylorCouette.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY box.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY cylinder_manifold.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY 2d-benchmark.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
blaisb, line 11: it would be maybe better to give this msh a more approriate title. Like mesh_lpbf or something similar to that. Just so in the future we know what it is used for.

applications_tests/lethe-fluid/CMakeLists.txt
```diff
@@ -53,6 +54,20 @@ file(COPY poiseuille_restart_files/poiseuille_restart.triangulation DESTINATION
...
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/restart.triangulation DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/restart.triangulation.info DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/restart.triangulation_fixed.data DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/mass_conservation_information.checkpoint DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/temperature_statistics.checkpoint DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
+
+file(COPY heat_transfer_vof_lpbf_benchmark_restart_files/restart.pvdhandler DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/heat_transfer_vof_lpbf_benchmark.${_build_type}/mpirun=1/")
```
blaisb, line 65: I feel like you are copying the file twice here. am I wrong?

blaisb (approved):
two minor things after rebase and this is ready to merge :)

blaisb (approved):

Issue #1198, opened 2024-07-17
Insertion info documentation must be updated


cleodeletre:
"set insertion order of direction" should be "set insertion direction sequence" https://chaos-polymtl.github.io/lethe/documentation/parameters/dem/insertion_info.html

PR #1202, opened 2024-07-19
Rearrange the insertion documentation for DEM

acdaigneault:
### Description

There are now 4 insertion methods in DEM that some parameters are used by one method and some are used by many methods. @OGaboriault did a nice job by splitting the documentation into sections and specifiying what parameters are used with which methods. However, it seems still confusing from a novice's perspective because of the specificities on the parameters according to the method.
I introduce all the parameters that can be used for each method, so we can clearly know if I wanna use method X I can use parameter a, b, c, ... 
It also closes issue #1198

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/doc/source/parameters/dem/insertion_info.rst
+++ b/doc/source/parameters/dem/insertion_info.rst
@@ -2,7 +2,7 @@
 Insertion Info
 ==============
 
-In this subsection, insertion methods which are ``volume``, ``plane``, ``list`` and ``file`` are defined.
+In this subsection, insertion methods ``volume``, ``plane``, ``list`` and ``file`` are defined.
 
 .. note::
     Insertion in Lethe starts inserting particles from type 0 and proceeds to the next type when all the particles from the previous type are inserted.
@@ -14,30 +14,25 @@ In this subsection, insertion methods which are ``volume``, ``plane``, ``list`` 
     set insertion method                               = volume
 
     # Every method
-    set insertion frequency                            = 20000
-    set remove particles                               = false
-    set removal box points coordinates                 = 0., 0., 0.: 1., 1., 1.
+    set insertion frequency                            = 1
 
     # If method = volume
-    set inserted number of particles at each time step = 100
-
-    set insertion box points coordinates               = -0.05, -0.05, -0.03 : 0.05, 0.05, 0.07
-
-    set insertion order of direction                   = 0, 1, 2
-
-    set initial velocity                               = 0.0, 0.0, 0.0
-    set initial angular velocity                       = 0.0, 0.0, 0.0
-
-    set insertion distance threshold                   = 2
+    set inserted number of particles at each time step = 1
+    set insertion box points coordinates               = 0., 0., 0. : 1., 1., 1.
+    set insertion insertion direction sequence         = 0, 1, 2
+    set insertion distance threshold                   = 1.
 
     # If method = plane
     set insertion method                               = plane
     set insertion plane point                          = 0, 0, 0
     set insertion plane normal vector                  = 1, 0, 0
+    set insertion plane threshold distance             = 0.
 
     # If method = volume or plane
-    set insertion maximum offset                       = 0.75
-    set insertion prn seed                             = 19
+    set initial velocity                               = 0.0, 0.0, 0.0
+    set initial angular velocity                       = 0.0, 0.0, 0.0
+    set insertion maximum offset                       = 1.
+    set insertion prn seed                             = 1
 
     # If method = list
     set list x                                         = 0.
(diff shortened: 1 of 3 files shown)


doc/source/parameters/dem/insertion_info.rst
```diff
@@ -53,38 +48,38 @@ In this subsection, insertion methods which are ``volume``, ``plane``, ``list``
...
 
-* The ``insertion order of direction`` parameter defines the directions of insertion. For example, if the parameter is equal to ``0, 1, 2``, the particles are inserted in priority in the x, in y, and then in z directions. This is the default configuration. This is useful to specify the insertion directions to cover a specific area of the insertion box with the first and second direction parameters.
+* ``insertion box points coordinates`` defines the insertion box dimensions using two points: ``x1, y1, z1 : x2, y2, z2``. It is based on the same principle as for the rectangular mesh `CFD <../../parameters/cfd/mesh.html>`_ triangulation.
 
-* The ``initial velocity`` determine the initial translational velocity (in :math:`\frac{m}{s}`) at which particles are inserted in the x, y, and z directions.
+.. note::
+    We recommend that the defined insertion box has at least a distance of :math:`{d^{max}_p}` (maximum diameter of particles) from the triangulation boundaries. Otherwise, particles may have an overlap with the triangulation walls in the insertion.
```
OGaboriault, line 75: maximum diameter of particle plus the maximum offset maybe?

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -94,31 +89,51 @@ The distance between the inserted particles is equal to:
...
+The ``plane`` insertion method inserts particles at the centroid of insertion cells. These cells are defined as intersected by a mathematical plane. This plane is defined by an ``insertion plane point`` and an ``insertion plane normal vector``. A cell is considered as intersected by the plane if at least one of its vertex is on each side of the plane of if at least one of its vertex is directly on the plane (the normal distance between the vertex and the plane is zero). At each insertion step, a particle will be inserted in a insertion cell if that cell is empty (no particle is present inside it). This guarantee the absence of big overlap with the particles already inserted. This method of inserting is useful when dealing with a domain dense with particles.
```
OGaboriault, line 101:
```suggestion
The ``plane`` insertion method inserts particles at the centroid of insertion cells. These cells are defined as intersected by a mathematical plane. This plane is defined by an ``insertion plane point`` and an ``insertion plane normal vector``. A cell is considered as intersected by the plane if at least one of its vertex is on each side of the plane or if at least one of its vertex is directly on the plane (the normal distance between the vertex and the plane is zero). At each insertion step, a particle will be inserted in a insertion cell if that cell is empty (no particle is present inside it). This guarantee the absence of big overlap with the particles already inserted. This method of inserting is useful when dealing with a domain dense with particles.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -133,8 +148,21 @@ The ``file`` insertion method inserts particles in a similar way to the ``list``
 
 Each line is associated with a particle and its properties. The ``fem_force`` and ``fem_torque`` properties are only used in the CFD-DEM solver, but must be specified in all cases. The main advantage of using the ``file`` method over the ``list`` method is that the number of inserted particles is not limited to the maximum number of characters on a single line of parameter files. To generate an insertion file, particle positions and properties can be generated manually or with any script. An other option is to use the python code ``extract-particles-properties-from-vtu.py`` in ``lethe/contrib/preprocessing/`` directory. This code extracts particle properties from the last vtu file from a given simulation.
 
+* ``insertion frequency`` defines the frequency of the insertion of particles based on the list in the file(s)
+
+* ``list of input files`` defines the list of files to be used for the insertion. The default value is ``particles.input``.
```
OGaboriault, line 153: You should add something regarding the sequential order of files. Like, for the first insertion, the first file will be used, second insertion the second file will be used, etc. 

If the end of the list is reached, next insertion will use the first file, etc.

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -133,8 +148,21 @@ The ``file`` insertion method inserts particles in a similar way to the ``list``
...
     The critical Rayleigh time step is computed from the parameters in the ``particle type`` subsections, not the ``insertion info`` subsection. It is the user's responsibility to fill the ``particle type`` subsections correctly according to the diameter values stored in the insertion input file, otherwise Rayleigh time percentage displayed at the start of every DEM simulation may not be accurate.
+
+--------------------
+Removal
+--------------------
+With all insertion methods, it is possible to define a removal box where particles will be removed from the triangulation just before the insertion of new particles.
+
+* ``remove particles`` enables (true) or disables (false) the particle removal.
+
+* ``removal box points coordinates`` defines a removal box where particles will be removed.
```
OGaboriault, line 168: It uses the same principal as the insertion box.

OGaboriault (approved):
Small comments. Good work, it is way clearer this way.

doc/source/parameters/dem/insertion_info.rst
acdaigneault: Yeah good addition!

doc/source/parameters/dem/insertion_info.rst
acdaigneault: Good point. I have a question related to that. I think when the insertion method were uniform vs non uniform, the discretization in the box taking that into account, but I did not managed to find it in the current code. Was it removed? If so, would it be a good idea to add it since we have to do it?

acdaigneault: Ok I managed to find where the code does take that into account. Actually, it works if there are no offset (according to function `find_insertion_location_volume()` in insertion_volume.cc), the minimal position will be the max distance threshold * radius. Since the threshold should be always > 1, we should not take into account the Dp,max. The code should automatically take the threshold into account tho!
I will add the maths in the doc

image

doc/source/parameters/dem/insertion_info.rst
acdaigneault: Actually it is said in the intro of the method.

doc/source/parameters/dem/insertion_info.rst
acdaigneault: I tested the parameters and find out a few criteria to avoid overlap. I did add them in the doc.

doc/source/parameters/dem/insertion_info.rst
OGaboriault: Fogot about it then

blaisb (approved):

PR #1210, opened 2024-07-27
Add ASM as preconditioner of smoother

peterrum:


source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +45,209 @@
 
 #include <deal.II/numerics/vector_tools.h>
 
+template <typename VectorType>
+class PreconditionBase
```
blaisb, line 49: Is there a reason why you want to declare the class just in the .cc and not in the .h upstream? 
Also, can you add @brief and documentation for the base class following doxygen syntax.

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +45,209 @@
...
+{
+public:
+  virtual void
+  vmult(VectorType &dst, const VectorType &src) const = 0;
+};
```
blaisb, line 54: Can you please add doxygen documentation for member function.

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +45,209 @@
...
+
+template <typename VectorType>
+class MyDiagonalMatrix : public PreconditionBase<VectorType>
```
blaisb, line 57: Same as above, won't repeat it for the other classes.
MyDiagonalMatrix seems a weird class name choice. Maybe PreconditionDiagonalMatrix?

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +45,209 @@
...
+    right,
+    symm
+  };
+
+public:
+  using Number = typename VectorType::value_type;
+
+  PreconditionASM()
+    : weighting_type(WeightingType::left)
+  {}
```
blaisb, line 91: So right now this is still a prototype class I presume right (because the weighting type of hardcoded).

blaisb (commented):
First batch of comments, will continue reviewing later

source/solvers/mf_navier_stokes.cc
peterrum: > Is there a reason why you want to declare the class just in the .cc and not in the .h upstream?

I would like to keep this local here as long as it is not needed anywhere else.

> Also, can you add @brief and documentation for the base class following doxygen syntax.

Done!

source/solvers/mf_navier_stokes.cc
peterrum: > So right now this is still a prototype class I presume right (because the weighting type of hardcoded).

Yes its still a prototype. I need to make some clean up; but the version here should be enough to conduct some first experiments.

For the weighting type, one probably does not need any parameter, since the others don't really work in the NS context (I wrote the original implementation for a Poisson operator where the other variants worked better).

blaisb: Perfect, good for me.

blaisb (approved):

include/core/parameters.h
```diff
@@ -1190,6 +1190,14 @@ namespace Parameters
     /// MG smoother relaxation parameter
     double mg_smoother_relaxation;
 
+    /// Type of multigrid
```
lpsaavedra, line 1193:
```suggestion
    /// Type of preconditioner for the MG smoother
```

source/solvers/mf_navier_stokes.cc
```diff
@@ -44,6 +45,251 @@
...
+   * @brief Apply preconditioner.
+   */
+  void
+  vmult(VectorType &dst, const VectorType &src) const override
+  {
+    diagonal_matrix.vmult(dst, src);
+  }
+
+private:
+  // DiagonalMatrix class from deal.II.
```
lpsaavedra, line 86:
```suggestion
  /// DiagonalMatrix class from deal.II.
```

lpsaavedra (approved):
Seems good to me! I am excited to try it out. I have only two minor comments... should we add this to the documentation? Or since it is a prototype, should we test it first? What do you think?

blaisb:
@lpsaavedra this is very prototype-ish. Let's test it first

PR #1215, opened 2024-07-30
Average velocity as initial condition

mivaia:
The average velocity profile of an old simulation can now be used as an initial condition to a new simulation. The base mesh needs to stay the same, but auxiliary physics can now be added a posteriori without setting them as true in the initial simulation. The time stepping scheme can also be modified.

### Description

The refactored restart mechanism had some problems:
- The auxiliary physics needed to be set as true while the average velocity profile was calculated. The matrix for the auxiliary physic was therefore assembled and could lead to useless memory and simulation time usage.
- The user could not choose to add an auxiliary physics to an already calculated average velocity profile if it was not specified in the initial simulation. This is because the restart mechanism would always search for a solution to the auxiliary physic in the checkpoint files even if it was non existent.
- In my opinion, the location of the "use average velocity profile = true" in the multiphysics subsection was somewhat awkward as the average velocity profile is not really an auxiliary physic.

To solve these problems, the already implemented checkpoint mechanism is used. The simulation solving for the average velocity profile therefore needs to have checkpointing activated as well as the set calculate average velocities set to true (as before). Parameters in the initial condition section have been added as well as some conditions in the read_checkpoint() function.

Conditions like:
- Do not search for a solution to the auxiliary physics if the average velocity is used as an initial condition.
- Do search for a solution to the auxiliary physics if the restart is normally used. 
- Use the checkpoint file specified by the user in the appropriate initial condition section and not the simulation control section if the average velocity profile initial condition is used.

### Testing

A test using the average velocity profile was modified. 

### Documentation

Documentation will be added in a subsequent PR if all the modifications are accepted. 

An initial condition using the VTU files was attempted, but it is way easier to use the already existing checkpoint mechanism. No known function to retrieve the triangulation information from the vtu have been found in deal.II. The read_vtu() function is only oriented towards reading vtu meshes generated with deal.II. 

Code related list:

Pull request related list:
- [x] Labels are applied


Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
blaisb, line 1: You added files here which should not be included in a PR. This file should be removed. (you can use git rm).
If I can give you a general advice, try to double check what files you are adding and avoid using git commit -a to prevent adding files which are not necessary.

Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jul 29 09:59 EDT
```
blaisb, line 1: Same as previous

applications_tests/lethe-fluid/cavity_ht_average_velocity_as_initial_condition.mpirun=1.output
blaisb, line 1: This is a good example of a test with results that will be reproducable, that's nicely done.

applications_tests/lethe-fluid/cavity_ht_average_velocity_as_initial_condition.prm
```diff
@@ -20,7 +20,8 @@ end
 #---------------------------------------------------
 
 subsection initial conditions
-  set type = nodal
+  set type = average_velocity_profile
```
blaisb, line 23: Nice

include/core/simulation_control.h
```diff
@@ -318,6 +318,29 @@ class SimulationControl
     CFL = p_CFL;
   }
 
+  /**
+   * @brief Set the value of the simulation time.
+   *
+   * @param p_current_time Desired time at the current time step.
```
blaisb, line 324:
```suggestion
   * @param p_current_time Value of the current time.
```

include/core/simulation_control.h
```diff
@@ -318,6 +318,29 @@ class SimulationControl
     CFL = p_CFL;
   }
 
+  /**
+   * @brief Set the value of the simulation time.
```
blaisb, line 322: So for your feature you need to "reset" the time?

include/solvers/initial_conditions.h
```diff
@@ -34,7 +34,8 @@ namespace Parameters
     L2projection,
     viscous,
     nodal,
-    ramp
+    ramp,
+    average_velocity_profile
```
blaisb, line 38: I would add the documentation for this in the current PR in the rst file. It is generally not a good habit to add features without documenting them from the get go.
Quick question, this average_velocity_profile needs to reside on the same mesh as the simulation am I right? In the future, we could also have it reside on any mesh.

include/solvers/initial_conditions.h
```diff
@@ -108,6 +109,10 @@ namespace Parameters
     Functions::ParsedFunction<dim> cahn_hilliard =
       Functions::ParsedFunction<dim>(2);
 
+    // Path to the checkpointed average velocity profile
+    std::string checkpoint_folder;
+    std::string checkpoint_file_name;
```
blaisb, line 114:
```suggestion
    // Path to the checkpointed average velocity profile
    std::string checkpoint_folder;
    std::string checkpoint_file_name;
```

I would maybe use more descriptive variable names like average_velocity_folder and average_velocity_file_name ?

source/solvers/cahn_hilliard.cc
```diff
@@ -730,8 +730,8 @@ CahnHilliard<dim>::postprocess(bool first_iteration)
 
       if (multiphysics->fluid_dynamics_is_block())
         {
-          if (this->simulation_parameters.multiphysics
-                .use_time_average_velocity_field &&
+          if (this->simulation_parameters.initial_condition->type ==
+                Parameters::InitialConditionType::average_velocity_profile &&
```
blaisb, line 734: I am unsure I understand what this check is for, but I feel also maybe it's because I am not seeing the big picture. COuld be worth it to add a comment here (and in the other auxiliary physics also)

source/solvers/gls_navier_stokes.cc
```diff
@@ -137,7 +138,12 @@ GLSNavierStokesSolver<dim>::setup_dofs_fd()
                        dsp,
                        this->mpi_communicator);
 
-  if (this->simulation_parameters.post_processing.calculate_average_velocities)
+  if (this->simulation_parameters.post_processing
+        .calculate_average_velocities ||
+      this->simulation_parameters.initial_condition->type ==
+        Parameters::InitialConditionType::average_velocity_profile ||
+      this->simulation_parameters.initial_condition->type ==
+        Parameters::InitialConditionType::average_velocity_profile)
```
blaisb, line 146: You have a repeated || statement here.
```suggestion
        .calculate_average_velocities ||
      this->simulation_parameters.initial_condition->type ==
        Parameters::InitialConditionType::average_velocity_profile)
```

source/solvers/gls_navier_stokes.cc
```diff
@@ -1269,6 +1278,33 @@ GLSNavierStokesSolver<dim>::set_initial_condition_fd(
...
+  else if (initial_condition_type ==
+           Parameters::InitialConditionType::average_velocity_profile)
+    {
+      this->pcout << "*********************************************************"
+                  << std::endl;
+      this->pcout
+        << " Initial condition using average velocity from checkpoint "
+        << std::endl;
+      this->pcout << "*********************************************************"
+                  << std::endl;
```
blaisb, line 1290: There is a function in utilities called announce string
https://chaos-polymtl.github.io/lethe/doxygen/utilities_8h.html#a56a967052620929f4db03cee351d6890

You can use that to announce strings

source/solvers/gls_navier_stokes.cc
```diff
@@ -1269,6 +1278,33 @@ GLSNavierStokesSolver<dim>::set_initial_condition_fd(
...
+
+      // The average velocity profile needs to come from the results of a
+      // simulation. The checkpoint/restart mechanism is used to obtain all the
+      // simulations information necessary to restart the simulation. The same
+      // mesh needs to be used, but new physics can be added to the simulation.
+      this->read_checkpoint();
```
blaisb, line 1296: This is a limitation we can also remove in the future :)

source/solvers/gls_navier_stokes.cc
```diff
@@ -1269,6 +1278,33 @@ GLSNavierStokesSolver<dim>::set_initial_condition_fd(
...
+
+      // This is an initial condition, not a restart. Therefore, the current
+      // time is set to 0, the iteration number as well and the pvd is cleared
+      // from previous results
+      this->simulation_control->set_current_time(0.0);
+      this->simulation_control->set_iteration_number(0);
+      this->simulation_control->set_current_time_step(
+        this->simulation_parameters.simulation_control.dt);
+
+      this->pvdhandler.times_and_names.clear();
```
blaisb, line 1306: I am not sure, but maybe a better approach would be to have a simulation_control->reset_time() function which  would reset the time, iteration and dt to parameter it has read instead.
This would prevent this manual modification of the simulation control object from outside of it and make it clear inside of the simulation_control object what is being modified. I think this would be a better approach.

source/solvers/heat_transfer.cc
```diff
@@ -465,8 +465,8 @@ HeatTransfer<dim>::assemble_local_system_matrix(
 
   if (multiphysics->fluid_dynamics_is_block())
     {
-      if (this->simulation_parameters.multiphysics
-            .use_time_average_velocity_field &&
+      if (this->simulation_parameters.initial_condition->type ==
```
blaisb, line 468: Add a comment on what this check does exactly.

source/solvers/heat_transfer_assemblers.cc
```diff
@@ -92,7 +92,7 @@ HeatTransferAssemblerCore<dim>::assemble_matrix(
 
       // Calculate the artificial viscosity of the shock capture as in equation
       // (79) of Tezduyar 2003
-      const double nu_dcdd = 0.5 * h * h * u_mag *
+      const double nu_dcdd = is_steady(method) ? 0 : 0.5 * h * h * u_mag *
                              dcdd_temperature_gradient.norm() /
```
blaisb, line 96: So you remove DCDD for steady simulations?

source/solvers/navier_stokes_base.cc
```diff
@@ -1720,9 +1736,31 @@ void
...
+      checkpoint_file_name =
+        this->simulation_parameters.initial_condition->checkpoint_file_name;
+    }
+  else
+    {
+      checkpoint_folder =
+        this->simulation_parameters.simulation_control.output_folder;
+      checkpoint_file_name =
+        this->simulation_parameters.restart_parameters.filename;
+    }
```
blaisb, line 1760: Wait what?
Here you don't mean the output path but the checkpoint path for the average_velocity right?

source/solvers/navier_stokes_base.cc
```diff
@@ -1808,7 +1850,22 @@ NavierStokesBase<dim, VectorType, DofsType>::read_checkpoint()
...
+  // new simulation, the code would try to get an inexistant solution.
+  // Therefore, the multiphysics initial condition should not be set from the
+  // checkpoint.
+  if (this->simulation_parameters.initial_condition->type !=
+        Parameters::InitialConditionType::average_velocity_profile ||
+      (this->simulation_parameters.initial_condition->type ==
+         Parameters::InitialConditionType::average_velocity_profile &&
+       this->simulation_parameters.restart_parameters.restart))
+    {
+      // std::cout << "Entering normally" << std::endl;
```
blaisb, line 1865:
```suggestion
```

source/solvers/navier_stokes_base.cc
```diff
@@ -1808,7 +1850,22 @@ NavierStokesBase<dim, VectorType, DofsType>::read_checkpoint()
...
+  // mechanism to fetch a previous simulation solution. If an auxiliary physics
+  // was set to false in this previous simulation and is now set as true in a
+  // new simulation, the code would try to get an inexistant solution.
+  // Therefore, the multiphysics initial condition should not be set from the
+  // checkpoint.
+  if (this->simulation_parameters.initial_condition->type !=
+        Parameters::InitialConditionType::average_velocity_profile ||
+      (this->simulation_parameters.initial_condition->type ==
+         Parameters::InitialConditionType::average_velocity_profile &&
+       this->simulation_parameters.restart_parameters.restart))
```
blaisb, line 1863: Ok I understand what you mean here. It's not the most elegant thing, but I see whereyou are trying to get at.
Maybe you could write it instead as != average_velocity_field || !=restart ?
This way you would not have this double if condition.

blaisb (changes requested):
Here's a first batch of comments. I think it would be wise to add the documentation within this PR. Good job btw

source/solvers/navier_stokes_base.cc
oguevremont: I'm also a bit confused here, but I think it's due to variable names. As suggested by @blaisb, it could be worth renaming `checkpoint_folder `and `checkpoint_file_name`

source/solvers/cahn_hilliard.cc
oguevremont: Also, if we reset time to 0, is this condition true when it should be?

source/solvers/gls_navier_stokes.cc
```diff
@@ -1269,6 +1278,38 @@ GLSNavierStokesSolver<dim>::set_initial_condition_fd(
           this->pcout << "*********************************" << std::endl;
         }
     }
+  else if (initial_condition_type ==
+           Parameters::InitialConditionType::average_velocity_profile)
+    {
+      announce_string(this->pcout, "Initial condition using average velocity profile");
+      // this->pcout << "*********************************************************"
```
oguevremont, line 1285: These comments should be removed

source/solvers/navier_stokes_base.cc
```diff
@@ -1720,88 +1736,39 @@ void
...
+  //     checkpoint_file_name =
+  //       this->simulation_parameters.initial_condition->checkpoint_file_name;
+  //   }
+  // else
+  //   {
+  //     checkpoint_folder =
+  //       this->simulation_parameters.simulation_control.output_folder;
+  //     checkpoint_file_name =
+  //       this->simulation_parameters.restart_parameters.filename;
+  //   }
```
oguevremont, line 1764: These comments should be removed

oguevremont (commented):
Very nice feature!

oguevremont:
So now if I understand correctly, we can run simulation that solve FD only while checkpointing, then run subsequent simulations where auxiliary physics are activated by using the proper type of initial condition?

include/solvers/initial_conditions.h
mivaia: Yes, for the moment it needs to reside on the same mesh

source/solvers/heat_transfer_assemblers.cc
mivaia: Yep! It is too non-linear without this.

source/core/simulation_control.cc
```diff
@@ -177,6 +177,37 @@ SimulationControl::read(std::string prefix)
...
+    input >> buffer >> time_steps[i];
+  input >> buffer >> buffer;
+  input >> buffer >> last_checkpointed_time;
+  input >> buffer >> buffer;
+
+  // Fix the current time as final checkpointed time
+  simulation_control_info[0] = last_checkpointed_time;
+
+  // Fix time step to be the last time_step that was used
+  simulation_control_info[1] = time_steps[0];
```
blaisb, line 207: You really want to have the new simulation start at the time the previous one finished? why not restart it from zero? does that really pose a problem? I'm not sure what's the goal her (to me this is not intuitive, because if you set the initial condition, shouldnt the time be at zero still ? )

blaisb (commented):
One last comment, but this is very good. There's just this "setting the time" that I am not sure about. Can you explain why you need to have the simulation control start at the new time? :)

source/core/simulation_control.cc
mivaia: This is only to get end time for the average velocity calculation! The simulation time will be restarted from zero in the second simulation. This is necessary because of the way we calculate the average velocity.

blaisb: :)perfect

PR #1246, opened 2024-08-16
Remove several lines in cmake files

lpsaavedra:
### Description

With a few lines we do need to include the .cc and .h files in the CMakeLists.cc manually.


blaisb:
@lpsaavedra There is a reason why we do not use GLOB. You can ask @paulapatience, but we switched away from this. It is not a good idea to revert this change.

lpsaavedra:
Oh okay.. I just thought it was annoying to always add sth everytime we add a file, but okay, I will close it then

PR #1255, opened 2024-08-21
Print parameters (lethe-fluid-matrix-free)

peterrum:
The output is like this:
```
set dimension = 3
subsection simulation control
  set time step                    = 0.05
  set adapt                        = true
  set max cfl                      = 1000
  set stop tolerance               = 1e-5
  set adaptative time step scaling = 1.2
  set output path                  = output/
  set output frequency             = 0
end
  subsection fluid 0
    set kinematic viscosity = 0.006666667
  end
end
subsection mesh
  set type               = gmsh
  set file name          = ./sphere.msh
  set initial refinement = 1
end
subsection boundary conditions
  subsection bc 0
    set type = noslip
  end
  subsection bc 1
    set type = function
    subsection u
      set Function expression = 1
    end
  end
  subsection bc 2
    set type = slip
  end
  subsection bc 3
    set type = outlet
  end
end
subsection initial conditions
  set type = ramp
  subsection ramp
    subsection kinematic viscosity
      set initial kinematic viscosity = 0.1
      set iterations                  = 6
    end
  end
end
subsection FEM
  set velocity order = 3
  set pressure order = 3
end
subsection timer
  set type = end
end
subsection forces
  set verbosity       = verbose
  set calculate force = true
end
subsection linear solver
  subsection fluid dynamics
    set verbosity                                 = extra verbose
    set relative residual                         = 1e-4
    set minimum residual                          = 1e-5
    set max iters                                 = 200
    set max krylov vectors                        = 200
    set preconditioner                            = gcmg
    set ilu preconditioner fill                   = 1
    set ilu preconditioner absolute tolerance     = 1e-6
    set ilu preconditioner relative tolerance     = 1
    set amg preconditioner ilu fill               = 1.00
    set amg preconditioner ilu absolute tolerance = 1e-6
    set mg smoother iterations                    = 2
    set mg smoother relaxation                    = 0.15
    set mg smoother preconditioner type           = additive schwarz method
    set mg smoother eig estimation                = true
    set eig estimation smoothing range            = 5
    set eig estimation cg n iterations            = 4
    set mg coarse grid solver                     = direct
    set mg coarsening type                        = hp
    set mg gmres max iterations                   = 500
    set mg gmres tolerance                        = 1e-5
    set mg gmres max krylov vectors               = 500
    set mg gmres preconditioner                   = ilu
    set mg verbosity                              = extra verbose
  end
end
subsection non-linear solver
  subsection fluid dynamics
    set tolerance = 1e-4
  end
end
subsection manifolds
  set number = 1
  subsection manifold 0
    set type              = spherical
    set id                = 0
    set point coordinates = 0, 0, 0
  end
end
subsection source term
  subsection fluid dynamics
    set enable = false
  end
end
```
Only the changed parameters are printed (Lethe just has too many parameters).

depends on https://github.com/dealii/dealii/pull/17593

### DIFF
--- a/applications/lethe-fluid-matrix-free/fluid_dynamics_matrix_free.cc
+++ b/applications/lethe-fluid-matrix-free/fluid_dynamics_matrix_free.cc
@@ -22,18 +22,25 @@ main(int argc, char *argv[])
     {
       Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
 
-      if (argc != 2)
+      if (argc == 1)
         {
           std::cout << "Usage:" << argv[0] << " input_file" << std::endl;
           std::exit(1);
         }
 
-      const unsigned int                  dim = get_dimension(argv[1]);
-      const Parameters::SizeOfSubsections size_of_subsections =
-        Parameters::get_size_of_subsections(argv[1]);
 
-      ConditionalOStream pcout(
-        std::cout, Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0);
+      const std::string file_name(argv[argc - 1]);
+      const bool        print_parameters =
+        (argc == 2) ? false : (std::string(argv[1]) == "--print-parameters");
+
+      const unsigned int                  dim = get_dimension(file_name);
+      const Parameters::SizeOfSubsections size_of_subsections =
+        Parameters::get_size_of_subsections(file_name);
+
+      ConditionalOStream pcout(std::cout,
+                               (Utilities::MPI::this_mpi_process(
+                                  MPI_COMM_WORLD) == 0) &&
+                                 print_parameters);
 
       if (dim == 2)
         {
@@ -41,7 +48,7 @@ main(int argc, char *argv[])
           SimulationParameters<2> NSparam;
           NSparam.declare(prm, size_of_subsections);
           // Parsing of the file
-          prm.parse_input(argv[1]);
+          prm.parse_input(file_name);
           NSparam.parse(prm);
 
           if (pcout.is_active())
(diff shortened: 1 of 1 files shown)


blaisb:
I am not sure what is your goal here. All the parameters in Lethe are documented in the following webpage
https://chaos-polymtl.github.io/lethe/documentation/parameters/parameters.html

I would refrain from duplicating information sources / locations. I don,t want to get to a point where we print the prm file  whenever we are launching a job. There is already a parameter template generator executable. I know there are many parameters, but there is no way around this.

peterrum:
The idea is to print all relevant parameters at the end of the simulation. We can add a parameter to enable/disable this, but having the simulation output and the chosen parameters next to each other/in the same file is quite useful in my opinion. If you run too many parameter studies you loose the overview of the modified parameters.

blaisb:
> The idea is to print all relevant parameters at the end of the simulation. We can add a parameter to enable/disable this, but having the simulation output and the chosen parameters next to each other/in the same file is quite useful in my opinion. If you run too many parameter studies you loose the overview of the modified parameters.



> The idea is to print all relevant parameters at the end of the simulation. We can add a parameter to enable/disable this, but having the simulation output and the chosen parameters next to each other/in the same file is quite useful in my opinion. If you run too many parameter studies you loose the overview of the modified parameters.

Sure that's a good idea. However it would be relevant to add the switch right now, otherwise all tests will explode for the moment being.

Maybe add a runtime option --print-parameters to the executable. In that case it would print the parameters, otherwise it would be the regular behavior. Then we could implement the same behavior for all solvers because I think it would be very useful. It's a cool feature

peterrum:
@blaisb What do you think about the last commit?

blaisb:
> @blaisb What do you think about the last commit?

I love the last commit. That's a very nice way to do it.
Could you just put include guards around it to ensure we can still compile under 9.6? I think this is  a very nice way to do this. I will port this feature to the other solvers.

PR #1263, opened 2024-08-27
Print intermediate mg iterations

peterrum:


lpsaavedra (approved):
I just tested it and it works as expected... I will change the parameters in a follow PR

PR #1265, opened 2024-08-27
Refactor cell diameter computation to utilities

AmishgaAlphonius:
### Description

The cell diameter calculation was repeated in every scratch data. This PR refactors it in an inline function in the utilities.

### Testing

No impact on current tests.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/core/utilities.h
+++ b/include/core/utilities.h
@@ -605,11 +605,13 @@ compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
   if constexpr (dim == 2)
     h = std::sqrt(4. * cell_measure / numbers::PI) / fe_degree;
   else if constexpr (dim == 3)
-    h = std::pow(6. * cell_measure / numbers::PI, 1. / 3.) / fe_degree;
+    h = std::cbrt(6. * cell_measure / numbers::PI) / fe_degree;
   else
-    throw std::invalid_argument(
-      "'dim' should have a value of either 2 or 3. Only 2D and 3D simulations "
-      "are supported.");
+    Assert(
+      false,
+      ExcMessage(std::string(
+        "'dim' should have a value of either 2 or 3. Only 2D and 3D simulations "
+        "are supported.")));
   return h;
 }
 
--- a/include/core/utilities.h
+++ b/include/core/utilities.h
@@ -585,4 +585,28 @@ value_string_to_tensor(const std::string &value_string_0,
   return output_tensor;
 }
 
+/**
+ * @brief Computes equivalent cell diameter by comparing the area to a disk (2D)
+ * or the volume to a sphere (3D).
+ *
+ * @tparam dim Number of spatial dimensions (2D or 3D).
+ *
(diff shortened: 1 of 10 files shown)


blaisb (approved):

include/core/utilities.h
```diff
@@ -585,4 +585,28 @@ value_string_to_tensor(const std::string &value_string_0,
...
+template <int dim>
+inline double
+compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
+{
+  double h;
+  if constexpr (dim == 2)
+    h = std::sqrt(4. * cell_measure / numbers::PI) / fe_degree;
+  else if constexpr (dim == 3)
+    h = std::pow(6. * cell_measure / numbers::PI, 1. / 3.) / fe_degree;
+  return h;
```
blaisb, line 609:
```suggestion
  if constexpr (dim == 2)
    return std::sqrt(4. * cell_measure / numbers::PI) / fe_degree;
  else if constexpr (dim == 3)
    return std::cbrt(6. * cell_measure / numbers::PI, 1. / 3.) / fe_degree;
    else
    Assert("....")
```
The code does not support 1D anyway right now so might as well assert if the wrong number is called instead of returning something without an initial value.

blaisb (changes requested):
One small thing to change

include/core/utilities.h
```diff
@@ -585,4 +585,32 @@ value_string_to_tensor(const std::string &value_string_0,
...
+ */
+template <int dim>
+inline double
+compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
+{
+  double h;
+  if constexpr (dim == 2)
+    h = std::sqrt(4. * cell_measure / numbers::PI) / fe_degree;
+  else if constexpr (dim == 3)
+    h = std::pow(6. * cell_measure / numbers::PI, 1. / 3.) / fe_degree;
```
blaisb, line 608:
```suggestion
    h = std::cbrt(6. * cell_measure / numbers::PI, 1. / 3.) / fe_degree;
```
std::cbrt is used for a cubic root, not std::pow. It's faster to use std::cbrt (this was in my previous suggestion)

include/core/utilities.h
```diff
@@ -585,4 +585,32 @@ value_string_to_tensor(const std::string &value_string_0,
...
+  else
+    throw std::invalid_argument(
+      "'dim' should have a value of either 2 or 3. Only 2D and 3D simulations "
+      "are supported.");
```
blaisb, line 612: The Assert() macro here should be used and not a throw statement. This I also suggested in my previous comments.
```suggestion
    Assert(false, ExcMessage(std::string(
      "'dim' should have a value of either 2 or 3. Only 2D and 3D simulations "
      "are supported."));
```
Something like this should be the syntax. There's nothing to throw here, we won't recover from that situation.

AmishgaAlphonius: I wasn't sure of what to put as a condition and was confused, but this totally makes sens!

blaisb (approved):

PR #1279, opened 2024-09-11
Modify cell diameter function

AmishgaAlphonius:
### Description

The previous implementation to calculate cell diameter used the `measure` function which assumed a linear mapping.

### Solution

This PR adds a function that calculates the volume by integrating 1 over the cell, by summing JxW values returned by the FEValues object. 

### Testing

Tests that were outdated have been updated.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/utilities.h
```diff
@@ -615,4 +617,26 @@ compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
...
+ *
+ * @return Volume (3D) or area (2D) of the cell.
+ */
+template <int dim>
+inline double
+compute_volume_with_JxW(const FEValues<dim> &fe_values, unsigned int n_q_points)
+{
+  double cell_volume = 0;
+  for (unsigned int q = 0; q < n_q_points; ++q)
+    cell_volume += fe_values.JxW(q);
```
oguevremont, line 638: Maybe using get_JxW_values() outside of the loop could be faster than calling JxW multiple times, but I'm not sure about the performance gains here

include/core/utilities.h
```diff
@@ -615,4 +617,26 @@ compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
   return h;
 }
 
+/**
+ * @brief Computes the volume of the cell by integrating after applying
```
oguevremont, line 621: I see what you mean, but the sentence is a bit difficult to follow. Maybe "Computes the volume of the cell by integrating 1 using its quadrature formula"

include/core/utilities.h
blaisb: Why not instead of providing the FeValues you could provide instead the vector of JxW. I don,t like the fact that you need to provide the loop limit (n_q_points) AND the fe_values. This is a bit weird.
You could just provide the vector of JxW and then just pass the vector to the function. The size of the vector would also give you the number of quadrature points.

AmishgaAlphonius: You're both right! I kind of just forgot about the vector. I'll change that! :)

include/core/utilities.h
AmishgaAlphonius: Yes, I'll write something more similar to the PR description.

oguevremont (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-09-11
+
+### Fixed
+
+- MINOR The previous implementation to calculate cell diameter used the `measure` function which assumed a linear mapping. A function that calculates the volume by integrating 1 over the cell, by summing JxW values returned by the FEValues object now replaces the `measure` function. [#1279](https://github.com/chaos-polymtl/lethe/pull/1279)
```
blaisb, line 10:
```suggestion
- MINOR The previous implementation to calculate cell diameter used the `measure` function which assumed a linear mapping. A function that calculates the volume by  summing JxW values returned by the FEValues object now replaces the `measure` function. [#1279](https://github.com/chaos-polymtl/lethe/pull/1279)
```

include/core/utilities.h
```diff
@@ -591,21 +593,21 @@ value_string_to_tensor(const std::string &value_string_0,
...
+ * @param[in] cell_volume Area (2D) or volume (3D) of the cell.
  *
  * @param[in] fe_degree Polynomial degree of the shape function.
  *
  * @return Cell diameter value.
  */
 template <int dim>
 inline double
-compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
+compute_cell_diameter(const double cell_volume, const unsigned int fe_degree)
```
blaisb, line 604: the variable cell_volume is not a very good choice. I prefer the word measure because the measure is the area or the volume depending on dim

blaisb (changes requested):
Some changes to address

include/core/utilities.h
```diff
@@ -615,4 +617,24 @@ compute_cell_diameter(const double cell_measure, const unsigned int fe_degree)
...
+ * FEValues object.
+ *
+ * @tparam dim Number of spatial dimensions (2D or 3D).
+ *
+ * @param[in] JxW_values Vector of mapped quadrature weights.
+ *
+ * @return Area (2D) volume (3D) of the cell.
+ */
+inline double
+compute_volume_with_JxW(const std::vector<double> &JxW_values)
```
blaisb, line 632:
```suggestion
compute_cell_measure_with_JxW(const std::vector<double> &JxW_values)
```

would be a better name IMO

blaisb (changes requested):
Sorry about that, one last comment to change IMO

include/core/utilities.h
AmishgaAlphonius: You're right, I missed it

blaisb:
is this ready to review @AmishgaAlphonius ?

AmishgaAlphonius:
> is this ready to review @AmishgaAlphonius ?
Yes :)

include/core/utilities.h
```diff
@@ -23,6 +23,8 @@
 
 #include <deal.II/dofs/dof_handler.h>
 
+#include <deal.II/grid/grid_tools.h>
```
blaisb, line 26: Do you need this include?

blaisb (approved):
One last question then good to go

include/core/utilities.h
AmishgaAlphonius: Not anymore indeed, I'll remove it !

AmishgaAlphonius: Done! Let's wait for the CI :)

PR #1283, opened 2024-09-16
Clean up weak boundary function evaluation

peterrum:
follow-up to https://github.com/chaos-polymtl/lethe/pull/1275


lpsaavedra (approved):
Thanks for the clean up! I will merge this PR

PR #1294, opened 2024-09-25
Correct update of solutions for multiphysics in matrix-free application

lpsaavedra:
### Description

The matrix free application was updating the results outside of the iterate loop. This works only if the other physics are solved before the fluid dynamics, but we have other physics that are solved right after the fluid dynamics and in that case, they need the most recent fluid information within the iterate loop. 

### Solution

In order to fix this, we have to update the solutions right after solving the problem for the fluid dynamics, i.e., within the iterate loop right after `solve_non_linear_system` is called. For this, a function had to be created in the parent class `NavierStokesBase`.

### Testing

No test available at the moment, the behaviour of the matrix-free application with the other physics is still a work in progress.

Code related list:

Pull request related list:
- [X] Labels are applied

### DIFF
--- a/include/solvers/navier_stokes_base.h
+++ b/include/solvers/navier_stokes_base.h
@@ -298,6 +298,12 @@ protected:
   virtual void
   update_multiphysics_time_average_solution() = 0;
 
+  /**
+   * @brief  Required only for the matrix-free solver to update solutions within iterate function.
+   */
+  virtual void
+  update_solutions_for_multiphysics() = 0;
+
   virtual void
   set_initial_condition_fd(
     Parameters::InitialConditionType initial_condition_type,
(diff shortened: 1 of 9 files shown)


include/solvers/navier_stokes_base.h
```diff
@@ -298,6 +298,12 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   virtual void
   update_multiphysics_time_average_solution() = 0;
 
+  /**
+   * @brief  Required only for the matrix-free solver to update solutions within iterate function.
+   */
+  virtual void
+  update_solutions_for_multiphysics() = 0;
```
blaisb, line 305: Would it be not better to have a default version of this virtual functoin that is not a pure virtual function but that does nothing and then just override it in the matrix free?
This would be less invasive I think

lpsaavedra: Yes, this makes a lot of sense. I will do the changes.

PR #1300, opened 2024-09-30
Fix simulation restart when using averaging

blaisb:
### Description

There was an issue where simulation restarts would crash on large parallel simulation when the time-averaging of the velocity was enabled. This issue was caused by the fact that the vectors that were checkpointed were the locally_relevant solutions but the restart was reading into the locally_owned solutions.

### Solution

I have fixed this by reading the restart into the locally_relevant solutions. However, I had to add a function to manually sanitize the time_averaging operator because the locally_owned vectors were now empty.

### Testing

This should not affect any of the current test results. I do not know if it is necessary to add a new tests since restart tests are cumbersome. @lpsaavedra what do you think?

### Documentation

No changes to documentation.

Code related list:

Pull request related list:
- [X] Labels are applied
- [X] There are at least 2 reviewers (or 1 if small feature) excluding the responsible for the merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-09-30
+
+### Fixed
+
+- MINOR Simulations with time-averaging of the velocity fields could be unable to restart when the domain was very large due to the fact that the restart vectors were read into the wrong vectors (locally_owned instead of locally_relevant). This PR fixes this. This also ensures that Lethe is able to restart with a different core-count than what was used to generate the restart file. [#1300](https://github.com/chaos-polymtl/lethe/pull/1300)
+
 ## [Master] - 2024-09-26
 
 ### Changed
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2024-09-30
+
+### Fixed
+
+- MINOR Simulations with time-averaging of the velocity fields could be unable to restart when the domain was very large due to the fact that the restart vectors were read into the wrong vectors (locally_owned instead of locally_relevant). This PR fixes this. This also ensures that Lethe is able to restart with a different core-count than what was used to generate the restart file. [#1300](https://github.com/chaos-polymtl/lethe/pull/1300)
```
lpsaavedra, line 10:
```suggestion
- MINOR Simulations with time-averaging of the velocity fields were unable to restart when the domain was very large due to the fact that the restart vectors were read into the wrong vectors (locally_owned instead of locally_relevant). This PR fixes this. This also ensures that Lethe is able to restart with a different core-count than what was used to generate the restart file. [#1300](https://github.com/chaos-polymtl/lethe/pull/1300)
```

lpsaavedra (approved):
Good that you found this bug. I don't think we need to add more tests since we already have restart tests and it is not easy to reproduce with small cases.

PR #1327, opened 2024-10-18
Double restart 

OGaboriault:
### Description

Sometimes, the file system on a cluster can be slow. In the last few month, we had many simulations where the allocated time expired while the checkpoint files were being written. This made the restart files corrupted and those simulations had to be restart from the beginning, which is annoying.

This feature make make code alternates between two different version of the restart files (_0. and _1. ) this way, there is always at least one version of restart file that is not corrupted.

Currently, it only works for DEM simulation, but will be implemented for every solver. Also, we could add a new parameter to specify which restart version to use.

### Testing

I don't think it's necessary to add new tests, since the existing tests using restart had to be modified for this feature.

Code related list:
- [X] New feature has unit test(s) (preferred) or application test(s), and restart files are in the generator folder

Pull request related list:
- [X] Labels are applied
- [X] There are at least 2 reviewers (or 1 if small feature) excluding the responsible for the merge


source/dem/read_checkpoint.cc
```diff
@@ -23,13 +25,32 @@ read_checkpoint(
...
+  boost::archive::text_iarchive ia_checkpoint_controller_obj(
+    iss_checkpoint_controller_obj, boost::archive::no_header);
+  checkpoint_controller.deserialize(ia_checkpoint_controller_obj, 0);
+
+  // Update predix with the checkpoint id
+  prefix =
+    prefix + "_" +
+    Utilities::int_to_string(checkpoint_controller.get_next_checkpoint_id());
+
+  // Need to increment the next checkpoint id for the next checkpoint
```
PierreLaurentinCS, line 51:
```suggestion
  // Increment the next checkpoint id for the next checkpoint
```

source/dem/read_checkpoint.cc
```diff
@@ -23,13 +25,32 @@ read_checkpoint(
...
+  checkpoint_controller.increment_checkpoint_id();
```
PierreLaurentinCS, line 52: Why do we need to increment the checkpoint_id after reading?

OGaboriault: If you read 0, this means that the last checkpoint was a 0, thus the next one needs to be 1.

source/dem/write_checkpoint.cc
```diff
@@ -67,6 +73,21 @@ write_checkpoint(
...
+  std::string checkpoint_controller_object_filename =
+    parameters.restart.filename + ".checkpoint_controller";
+  std::ofstream oss_checkpoint_controller_obj(
+    checkpoint_controller_object_filename);
+  boost::archive::text_oarchive oa_checkpoint_controller_obj(
+    oss_checkpoint_controller_obj, boost::archive::no_header);
+  checkpoint_controller.serialize(oa_checkpoint_controller_obj, 0);
+
+  // Increment for the next checkpoint
+  checkpoint_controller.increment_checkpoint_id();
```
blaisb, line 90: Why not increment the id after you have written?

blaisb (changes requested):
Few things to change I think.
1- The incrementation of the CheckpointController should be done from within the controller, and not from outside of the controller. The same way the simulation control does the incrementation inside of itself, it should be like that.
2- Knowing when you write a checkpoint could also be decided by the CheckpointController now. You could pass the restart parameters to the simulation control and have a function that decides if it will be a restart iteratoin yes or no.
We can talk about it tomorrow

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,64 @@
...
+  {
+    return next_checkpoint_id;
+  }
+
+  /**
+   * @brief Serialize the checkpoint controller object to an output archive.
+   * @param ar Output archive where the attributes are stored.
+   */
+  void
+  serialize(boost::archive::text_oarchive &ar, const unsigned int)
```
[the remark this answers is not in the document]
PierreLaurentinCS: Apparently, this is for retro compatibility with other versions of the code. (see Class Versioning on the boost documentation https://www.boost.org/doc/libs/1_86_0/libs/serialization/doc/index.html). I don't think it is useful here though.

include/dem/read_checkpoint.h
```diff
@@ -41,6 +41,7 @@ read_checkpoint(
   parallel::distributed::Triangulation<dim>               &triangulation,
   Particles::ParticleHandler<dim>                         &particle_handler,
   std::shared_ptr<Insertion<dim>>                         &insertion_object,
-  std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solid_surfaces);
+  std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solid_surfaces,
+  CheckpointControl &checkpoint_controller);
```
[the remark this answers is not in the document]
OGaboriault: We need to pass the checkpoint controller as a reference in the read and write since we increment the <next_checkpoint_id> in both function. (Tell me if I'm wrong)

source/dem/read_checkpoint.cc
OGaboriault: We just increment after reading so that the next write doesn't erase the last most advance checkpoint in the simulation. 

I'm not sure what you mean... currently the read function is incrementing the checkpoint id.

source/dem/write_checkpoint.cc
OGaboriault: That is what is being done at line 90. The last line of the function, thus after the writing.

blaisb: Yes but why not do it wihtin the class?
Let's talk about it tomorrow

include/dem/read_checkpoint.h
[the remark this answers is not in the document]
blaisb: Yeah, that's true, or you could just pass the checkpoint id and let the checkpoint controller increment.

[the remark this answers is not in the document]
OGaboriault: With this new implementation, it is now const. Is it ok?

include/core/checkpoint_control.h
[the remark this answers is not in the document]
OGaboriault: For the serialize and deserialize fonction, I used the same thing we used for the insertion objects... 

I don't know what is the uses of the <const unsigned int> here .

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,91 @@
...
+   * @brief Return the next checkpoint id.
+   */
+  unsigned int
+  get_next_checkpoint_id() const
+  {
+    return next_checkpoint_id;
+  }
+
+  /**
+   * @brief Return de prefix for the restart files.
```
PierreLaurentinCS, line 57:
```suggestion
   * @brief Return the prefix for the restart files.
```

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * @brief Serialize the checkpoint controller object to an output archive.
+   * @param ar Output archive where the attributes are stored.
+   */
+  void
+  serialize(boost::archive::text_oarchive &ar, const unsigned int) const
+  {
+    ar &next_checkpoint_id;
```
PierreLaurentinCS, line 72: I still find this notation (and the other below at line 82) confusing and I would prefer it to be << in this case ( and >> at line 82) for clarity.
In the end, it is up to you.

source/dem/write_checkpoint.cc
```diff
@@ -67,6 +73,18 @@ write_checkpoint(
     {
       solid_objects[i]->write_checkpoint(prefix);
     }
+
+  // Prepare the checkpoint controller for checkpointing
+  // We don't use the same prefix, since this file needs to have the same name
+  // regardless of the checkpoint id being used. This file is giving the
+  // information of which checkpoint id to use when restarting
```
PierreLaurentinCS, line 80:
```suggestion
  // information of which checkpoint id to use when restarting.
```

source/dem/dem.cc
```diff
@@ -1065,10 +1066,11 @@ DEMSolver<dim>::solve()
 
       // Write checkpoint if needed
       if (parameters.restart.checkpoint &&
-          simulation_control->get_step_number() %
-              parameters.restart.frequency ==
-            0)
+          checkpoint_controller.is_checkpoint_time_step(
```
PierreLaurentinCS, line 1069: It is nice to have everything embedded in one object !

PierreLaurentinCS (changes requested):
Very good job, can you please fix the few remaining typos and edit the changelog? Then it'll be ready to merge

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @param checkpoint_freq Checkpoint frequency.
+   * @param restart_filename Filename used for the prefix during checkpointing.
+   */
+  CheckpointControl(const unsigned int checkpoint_freq,
+                    const std::string &restart_filename)
+    : max_checkpoint_id(2)
+    , next_checkpoint_id(max_checkpoint_id)
+    , checkpointing_frequency(checkpoint_freq)
+    , filename(restart_filename)
```
blaisb, line 25: I'm just curious, butinstead of passing just two parameters like that, why not pass the entire structure of checkpoint parameters ?

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,91 @@
...
+  {}
+
+  /**
+   * @brief Return true if a checkpoint should be made at the current time step.
+   * @param time_step_number Current time step.
+   */
+  bool
+  is_checkpoint_time_step(const unsigned int time_step_number)
```
blaisb, line 33:
```suggestion
  is_checkpoint_time_step(const unsigned int time_step_number) const
```

include/dem/read_checkpoint.h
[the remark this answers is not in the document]
blaisb: Yeah if it's const now it's fine

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * @brief Calculate the next checkpoint id.
+   */
+  void
+  increment_checkpoint_id()
+  {
+    next_checkpoint_id = (next_checkpoint_id + 1) % max_checkpoint_id;
+  }
```
blaisb, line 45: Where is this function called? It does not seem to be called from the checkpoint controller. Don,t forget that in the end we can make sure that the checkpoint controller is the only one that increments the checkpoint id.

source/dem/dem.cc
```diff
@@ -1065,10 +1066,11 @@ DEMSolver<dim>::solve()
...
+            simulation_control->get_step_number()))
         {
+          checkpoint_controller.increment_checkpoint_id();
+
```
blaisb, line 1073: I don't this this is a good design.
Why not increment the checkpoint id from within the is_checkpoint_time_step?
In the end you should only check ONCE per time-step taht this is a checkpoint time step and so you could do the incrementation from within the check.

I don't like the idea of controlling the CheckpointController from the outside. The checkpoint controller should be independent in terms of logic from teh simulation.

include/core/checkpoint_control.h
blaisb: I'M not sure the << and >> notation works for boost archives @PierreLaurentinCS . Everywhere in the code we used this notation.

blaisb (changes requested):
few things left to address

include/core/checkpoint_control.h
OGaboriault: It's called in dem.cc when you enter the of statement

include/core/checkpoint_control.h
OGaboriault: Passing dem_parameter as an input requires the class to be templated no?

blaisb: No I meant the structure that contains the restart parameters
in a similar way the simulation control gets the strucutre that contains all of the simulation control parameters

include/core/checkpoint_control.h
OGaboriault: it works

PierreLaurentinCS: I was referring to the boost documentation (https://www.boost.org/doc/libs/1_39_0/libs/serialization/doc/archives.html) but it's true I did not test it. 
I think the notation is strictly equivalent. Though if & is used everywhere else in the code, then I see no reason not to use it here. It's not a big deal in the end.

include/core/checkpoint_control.h
PierreLaurentinCS: I think it can't be const because it may call increment_checkpoint_id() which is not const (and not meant to be). I could be mistaken

PierreLaurentinCS (approved):
Looks fine to me. Very good job!

include/core/checkpoint_control.h
blaisb: Yeah now that it increments it won't be const. This is fine

include/core/checkpoint_control.h
blaisb: I'll let you decide what is best :)

blaisb (approved):

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,99 @@
...
+
+class CheckpointControl
+{
+public:
+  /**
+   * @brief The CheckpointControl is responsible for keeping track of the
+   * next checkpoint id to use.
+   *
+   * @param checkpoint_freq Checkpoint frequency.
+   * @param restart_filename Filename used for the prefix during checkpointing.
```
blaisb, line 20: There are parameters missing here

include/core/checkpoint_control.h
```diff
@@ -0,0 +1,99 @@
...
+  std::string
+  get_filename() const
+  {
+    return filename;
+  }
+
+  /**
+   * @brief Serialize the checkpoint controller object to an output archive.
+   * @param ar Output archive where the attributes are stored.
+   */
```
blaisb, line 66: What is the other parameter (unsigned int) that is not used? Can you remove it or is it something that comes from BOOST%

blaisb (approved):
One last thing to address and then good for merge

include/core/checkpoint_control.h
OGaboriault: I'll keep this implementation to stay coherent with the rest of the code. When I'll fix the extra parameter in every serialize and deserialize function in the issue I have just opened, I will put the << and >> notation since I find in more intuitive. (the thing goes in or out of the archive.)

PR #1340, opened 2024-11-03
Improve cavity example

blaisb:
### Description

The lid-driven cavity example was just fine, but there was a small issue with the Reynolds 7500 case. The tolerance was so low that it took forever to converge. So I increased it, the results don't change. However, this made me realize that the python script was hardcoded to use a given file number. I fixed this by using a regular expression to find the last file no matter what.

Code related list:
- [x] Links are added to parent .rst files
- [x] The example is following the [standard format](https://chaos-polymtl.github.io/lethe/documentation/contributing.html#general-rules-and-format)

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/doc/source/examples/incompressible-flow/2d-lid-driven-cavity-flow/lid-driven-cavity-flow.rst
+++ b/doc/source/examples/incompressible-flow/2d-lid-driven-cavity-flow/lid-driven-cavity-flow.rst
@@ -20,12 +20,12 @@ Files Used in This Example
 
 All files mentioned below are located in the example's folder (``examples/incompressible-flow/2d-lid-driven-cavity``).
 
-- Base case parameter file (:math:`Re=400`): ``cavity.prm``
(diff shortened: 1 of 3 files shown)


lpsaavedra (changes requested):
Good changes, I reread the example and found a few typos that I will commit directly, I hope you don't mind. However, I think something is wrong with the script as this is what I get for the case of Re 400 when running the simulation and directly using the script: 

image

blaisb:
> Good changes, I reread the example and found a few typos that I will commit directly, I hope you don't mind. However, I think something is wrong with the script as this is what I get for the case of Re 400 when running the simulation and directly using the script:
> 
> image

You are not running the right script, that script is for the higher Reynolds number. There are two scripts for the two different RE values
 :)

lpsaavedra (approved):
You were right, weird stuff going on on my machine. The example seems good to me now!

PR #1341, opened 2024-11-04
Refactor output of paraview files and time step calculation

lpsaavedra:
### Description

The time step was being modified for Paraview outputs at the end or during a simulation. This was dangerous since abrupt changes of the time step might lead to instabilities, wrong results when calculating derivatives of quantities outputted in the simulation (e.g., kinetic energy rate) and useless iterations when performing controlled restarts with adaptive time stepping (i.e., setting the end time of the simulation and then modifying it and restarting it from the last iteration). 

This PR aims at avoiding all these issues by applying the following changes:
-  It does not change the time step in a simulation (to reach the `time end` in the last iteration) unless a new parameter in the simulation control section is set explicitly to false: `set time step independent of end time = false`.  The default is true. 
- It refactors the Paraview control for transient simulations. There are still two options for output control: iteration output control and time output control. The first one outputs paraview files according to an iteration output frequency. The second one outputs Paraview files according to time depending on an output frequency or one or more times specified by the `output times` variable. 
- For specific Paraview time outputs, the time step was being modified. Since this is no longer the case, we always output the file before and after an specified time. For example, if we want to output Paraview files at 5 seconds, but we only have one iteration with 4.8 seconds and another one with 5.2, both Paraview files are written, to ensure that the user can recover the time at exactly 5 seconds via Paraview. 

### Testing

- The output of four application tests corresponding to the `lethe-fluid `application was updated for the last time step, since it is no longer modified by default.
- The tests in core and solvers that needed a simulation control object were modified to include new default parameters. 
- The `simulation_control_03` test was extended to test all the output possibilities for both constant and adaptive time step. 
- The `simulation_control_06` tested the output in certain time interval. This was moved to the `simulation_control_03` test that considers all output possibilities.

### Documentation

- The simulation control section in the documentation was updated. It has the new parameters and explains in detail how to use the output control parameters.  
- Whenever the `output time` variable was being used in the examples, it was replaced by its new name `output time frequency` in both the prm file and the doc file.

Code related list:
- [X] If parameters are modified, the tests and the documentation of examples are up to date

Pull request related list:
- [x] Labels are applied


include/core/simulation_control.h
```diff
@@ -499,19 +492,41 @@ class SimulationControlTransient : public SimulationControl
...
+
+  // Specific output times for time output control
+  std::vector<double> output_times_vector;
+
+  // Counter to move between output times given in previous vector
+  unsigned int output_times_counter = 0;
+
+  // Variable to check whether we still have specific times to check in output
+  // times vector
+  bool no_more_output_times = false;
```
blaisb, line 509: Are you sure that it's a good idea to declare default value here instead of in the constructor?

source/core/parameters.cc
```diff
@@ -225,9 +238,11 @@ namespace Parameters
         {
           std::runtime_error("Invalid output control scheme");
         }
-      dt             = prm.get_double("time step");
-      timeEnd        = prm.get_double("time end");
-      adapt          = prm.get_bool("adapt");
+      dt      = prm.get_double("time step");
+      timeEnd = prm.get_double("time end");
```
blaisb, line 242: time to refactor this parameter. 
```suggestion
      time_end = prm.get_double("time end");
```

blaisb (approved):
I barely have any comments. Thank you for this very good work.

mivaia (approved):
Clean implementation! I actually have nothing new to add. Nice work!

PR #1342, opened 2024-11-05
Force chains, add the arbitrary rule

OGaboriault:
### Description

Force chains between local-ghost particle were being duplicated since the arbitrary rule with particle ids was not implemented. 

### Solution

The `execute_contact_calculation` function in the `force_chain_visualization.h` file is templated with the contact type. Weh nwe are dealing with `ghost_particle_particle` contacts, we apply the rule.

### Testing

I could add a test, but I'm not sure what is the best way to do it. It would require two particles, being in two different subdomains. 

### Documentation
N/A

Code related list:

Pull request related list:
- [X] Labels are applied

### DIFF
--- a/include/dem/force_chains_visualization.h
+++ b/include/dem/force_chains_visualization.h
@@ -47,13 +47,19 @@ public:
    * @param[in] ghost_adjacent_particles Container of the contact pair
    * candidates information for calculation of the local-ghost particle-particle
    * contact forces.
+   * @param[out] vertices the vector of positions of touching particles.
+   * @param[out] normal_forces_vector the vector of normal forces between each
+   * touching particles.
    */
   virtual void
   calculate_force_chains(
     typename dem_data_structures<dim>::adjacent_particle_pairs
       &local_adjacent_particles,
     typename dem_data_structures<dim>::adjacent_particle_pairs
-      &ghost_adjacent_particles) = 0;
+                          &ghost_adjacent_particles,
+    std::vector<Point<3>> &vertices,
+    std::vector<double>   &normal_forces_vector) = 0;
+
   /**
    * @brief Output the force chains in VTU and PVTU files for each iteration and
    * a PVD file.
@@ -73,7 +79,11 @@ public:
                      const MPI_Comm                  mpi_communicator,
                      const std::string               folder,
                      const unsigned int              iter,
-                     const double                    time) = 0;
+                     const double                    time,
+                     typename dem_data_structures<dim>::adjacent_particle_pairs
+                       &local_adjacent_particles,
+                     typename dem_data_structures<dim>::adjacent_particle_pairs
+                       &ghost_adjacent_particles) = 0;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


source/dem/force_chains_visualization.cc
```diff
@@ -91,8 +87,21 @@ ParticlesForceChains<dim, contact_model, rolling_friction_model>::
...
-                     const double                    time)
+                     const double                    time,
+                     typename dem_data_structures<dim>::adjacent_particle_pairs
+                       &local_adjacent_particles,
+                     typename dem_data_structures<dim>::adjacent_particle_pairs
+                       &ghost_adjacent_particles)
 {
+  // Creating containers
+  std::vector<Point<3>> vertices;
+  std::vector<double>   normal_forces_vector;
```
blaisb, line 98: I think you need to give those a defaul value otherwise if you have a processor with nothing won't you get an error ? LIke if the vectors are empty

source/dem/force_chains_visualization.cc
```diff
@@ -102,7 +111,7 @@ ParticlesForceChains<dim, contact_model, rolling_friction_model>::
   Vector<float> force_values(triangulation.n_active_cells());
```
blaisb, line 111: Wait why is this a Vector of float? This will be in single precision... why do that?

blaisb (changes requested):
Smalls changes nothing big

include/dem/force_chains_visualization.h
```diff
@@ -136,30 +129,71 @@ class ParticlesForceChains
...
    * @param[in] adjacent_particles_list Container of the adjacent particles of a
    * particles.
+   * @param[out] vertices the vector of positions of touching particles.
+   * @param[out] normal_forces_vector the vector of normal forces between each
+   * touching particles.
    */
+  template <ContactType contact_type>
   inline void
   execute_contact_calculation(
     typename DEM::dem_data_structures<dim>::particle_contact_info
```
blaisb, line 193: This could be const

include/dem/force_chains_visualization.h
```diff
@@ -136,30 +129,71 @@ class ParticlesForceChains
...
+   * @param[out] vertices the vector of positions of touching particles.
+   * @param[out] normal_forces_vector the vector of normal forces between each
+   * touching particles.
+   */
+  void
+  calculate_force_chains(
+    typename dem_data_structures<dim>::adjacent_particle_pairs
+      &local_adjacent_particles,
+    typename dem_data_structures<dim>::adjacent_particle_pairs
+                          &ghost_adjacent_particles,
```
blaisb, line 172: Both of these could be const no?

blaisb (approved):
Last changes and ready for merge.

include/dem/force_chains_visualization.h
OGaboriault: No it cannot. It has something to do with the particle_particle_contact_force.h file. 

There's an other function called `execute_contact_calculation` with the same input parameters. In that function, we clear the tangential overlap, thus it is not const. 

I'm not sure if the ParticlesForceChains class needs to inherit from ParticleParticleContactForce, but this is way it cannot be const.

blaisb: Perfect

blaisb (approved):

PR #1375, opened 2024-11-20
NACA0012 example review

mivaia:
Tested:
1. The .geo file to generate .msh files -> it works perfectly. Different NACA angles can be generated by changing a single parameter.
2. The .prm file. The simulation is REALLY expensive (10% of a single simulation in 15h on 15 cores). Results match perfectly after 10% of the simulation.
3. The post-processing file. It has been modified to add the reference correlation in the fast Fourier transform analysis.


blaisb (changes requested):
One question, then it will be good to go.

mivaia:
You're too fast haha. Thought I had more time before the review

blaisb (approved):

PR #1381, opened 2024-11-21
3D ribbon mixer example review

mivaia:
- Modified mesh section in the doc.
- Modified the gather-torques because of the new boundary convention.
- ALSO, the experimental.dat file is missing and the reference is not in the doc.


blaisb:
Hmm damn. I Hope we have not lost the experimental results

PR #1421, opened 2025-02-13
Add laser angle

hepap:
### Description

Add the possibility to rotate the laser around one axis when using the `gaussian_heat_flux_vof_interface model`. 

### Testing

One unit test is added to test the distance computation from the laser axis to a given point. This function (`LetheGridTools::find_point_line_distance()`) is used to compute the position of the quadrature points in the laser referential and the resulting gaussian heat flux.

### Documentation

Two parameters (`beam rotation angle` and `beam rotation axis`) were added to the documentation.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/lethe_grid_tools.h
```diff
@@ -316,6 +316,33 @@ namespace LetheGridTools
...
+   * @param[in] line_origin Origin of the line (any point on the line)
+   * @param[in] line_direction Vector tangent to the line
+   * @param[in] point A point for which we want to find the distance to the
+   * triangle
+   *
+   * return distance
+   *
+   */
+  template <int dim>
+  double
```
blaisb, line 341: This function is called a lot I believe. I would declare it in the.h and inline it. It is a relatively short function anyway so inlining won't have a big cost.

source/core/lethe_grid_tools.cc
```diff
@@ -1683,3 +1683,28 @@ template double
...
+}
+
+template double
+LetheGridTools::find_point_line_distance(const Point<2>     &line_origin,
+                                         const Tensor<1, 2> &line_direction,
+                                         const Point<2>     &point);
+template double
+LetheGridTools::find_point_line_distance(const Point<3>     &line_origin,
+                                         const Tensor<1, 3> &line_direction,
+                                         const Point<3>     &point);
```
blaisb, line 1710: Yeah clearly it is better to inline this function. It's very tiny

tests/core/lethe_grid_tools_point_line_distance.output
```diff
@@ -0,0 +1,37 @@
...
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
+DEAL::The distance is 1.00000
```
blaisb, line 37: It's a bit funny that the distance is 1 for your entire test. Is that what you expected? seems a bit like a simple test no?

blaisb (changes requested):
1-2 small changes, nothing big.
I think maybe your test is overly simple? don't you want a test that would actually have a non-1 answer?

tests/core/lethe_grid_tools_point_line_distance.output
hepap: It is what I expect, because the test is placing point on a cylinder of radius 1 using the equation of a cylinder. Then it checks if the `find_point_line_distance` returns the right distance between the cylinder axis and the point, which always the radius of 1.

I can make it more complex by varying the radius, and maybe randomize the point positions?

mivaia (changes requested):
Nice job hepap :) here are my comments!

mivaia (commented):
Also, maybe add a simple application test in case someone messes up the parameters

blaisb (approved):
The timeout is just because the test is long i guess?

blaisb:
@mivaia I'll just need y our OK ot merge

mivaia (approved):
Nice job! Approved :)

PR #1451, opened 2025-03-10
Vof geometric reinitialization architecture

hepap:
### Description

This PR adds the (almost) empty architecture of the SignedDistanceSolver and its coupling in the VOF solver (geometric reinitialization method ) , following the structure of the already existing vof_advection prototype. The parameters associated to the geometric reinitialization method for the VOF solver are also added.

Some methods of the SignedDistanceSolver are implemented to be able to test this new architecture.
### Testing

One unit is implemented:

`interface_tools_signed_distance`: This test checks the instanciation of the InterfaceTools::SignedDistanceSolver class from within an abitrary structure that replicates the usual physical solver structure. This is a dummy test intended to check the architecture, and not the SignedDistanceSolver performances yet! When the implementation of the SignedDistanceSolver will be completed, this test will be extended to a verification of the signed distance computation for a sphere.

### Documentation

The new parameters are not documented yet because they are not currently used. They will be in the next PR implementing the solver itself.

Following this PR, the complete solver will be implemented.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/interface_tools.h
```diff
@@ -286,6 +304,414 @@ namespace InterfaceTools
                                       &interface_reconstruction_cells,
     std::set<types::global_dof_index> &intersected_dofs);
 
+  /**
+   * @brief Solver to compute the signed distance from a given level of a level-set field. It is based on the method proposed by Ausas (2010).
+   *
```
blaisb, line 309: Can you put the DOI of the reference here. The year / Author is often not enough to find the pape.r

include/core/interface_tools.h
```diff
@@ -286,6 +304,414 @@ namespace InterfaceTools
...
+
+    /**
+     * @brief Attach the solution vector to the DataOut provided. This function
+     * enable the auxiliary physics to output their solution via the background
+     * solver.
+     *
+     * @param[in,out] data_out DataOut reponsible for solution output
+     */
+    void
+    attach_solution_to_output(DataOut<dim> &data_out);
```
blaisb, line 423: Good idea to have both an attach or an individual output. I like this a lot.

include/core/interface_tools.h
```diff
@@ -286,6 +304,414 @@ namespace InterfaceTools
...
+
+    /// DoFHandler describing the problem
+    DoFHandler<dim> dof_handler;
```
blaisb, line 426: Is there a reason why the dof_handler is public? You use it elsewhere I guess?

include/core/interface_tools.h
```diff
@@ -286,6 +304,414 @@ namespace InterfaceTools
...
+    compute_first_neighbors_distance();
+
+    /**
+     * @brief Compute the geometric distance (marching method) between the
+     * interface reconstruction and the rest of the dofs (second neighbors)
+     *
+     * @param[in] mpi_communicator MPI communicator
+     */
+    void
+    compute_second_neighbors_distance(const MPI_Comm &mpi_communicator);
```
blaisb, line 470: Do you need to pass the communicator here since this is a member function of the class and the class has the dof handler?
You can always get the communicator from the dof handler (see https://www.dealii.org/current/doxygen/deal.II/classDoFHandler.html#acd999cd3bf8d91a1157d16b01872664c)

include/core/interface_tools.h
```diff
@@ -286,6 +304,414 @@ namespace InterfaceTools
...
+    compute_cell_wise_volume_correction();
+
+    /**
+     * @brief Correct the global volume to match the volume englobed by the given level
+     * of the level_set field
+     *
+     * @param[in] mpi_communicator MPI communicator
+     */
+    void
+    conserve_global_volume(const MPI_Comm &mpi_communicator);
```
blaisb, line 493: Same as above, why not use communicator from dof handler?

source/core/parameters_multiphysics.cc
```diff
@@ -541,6 +551,76 @@ Parameters::VOF_AlgebraicInterfaceReinitialization::parse_parameters(
...
+  dealii::ParameterHandler &prm)
+{
+  prm.enter_subsection("geometric interface reinitialization");
+  {
+    prm.declare_entry(
+      "enable",
+      "false",
+      Patterns::Bool(),
+      "Enables the interface reinitialization with the geometric method "
+      "<true|false>");
```
blaisb, line 565: Would it not be better to  instead a parameter which would be:
retinitialization method = "none, pde, projection, geometric" and then enter that section?
It's not the best idea to have a ton of enable / everywhere and to rely on checking. This way we would have less parameters.

include/core/interface_tools.h
hepap: It is to use it when we will interpolate the signed distance to the main solver (VOF) dofhandler

hepap: I can add a get_dof_handler method also :)

include/core/interface_tools.h
hepap: I changed it everywhere!

source/core/parameters_multiphysics.cc
hepap: I agree. Do I do it in this PR or I open a refactoring one? @blaisb

blaisb: I'll merge the PR as is now, this could be in another PR?

PR #1475, opened 2025-03-28
Add options to print version and parameters

lpsaavedra:
### Description

This PR adds two new features:

1. Allows to print the deal.II and Lethe versions used to run an application. An example of the usage would be as follows:
` lethe-fluid -V`
yields:
```
Running: lethe-fluid -V
lethe/1.0-77-g9fe541d99 deal.II/9.6.2-2966-g7a74e69540
```
2. Allows to print the parameters of the input file by specifying an additional parameter as follows:
```
#---------------------------------------------
# Print non-default parameters to output file
#---------------------------------------------

set print parameters = only changed
```
Options are none (default), all, or only changed (only supported with deal.II versions > 9.7).

### Testing

The features were tested for the examples of all the applications. 

### Documentation

I have not introduced this in the documentation yet as there is no section for such parameters. We can discuss whether such a section is really needed.

The idea is to be able to obtain both the version and the parameters when running simulations on the clusters. This will require two calls in the launch files. An example looks as follows:
```
lethe-fluid -V
lethe-fluid parameters.prm
```

Code related list:

Pull request related list:
- [ ] Labels are applied

### DIFF
--- a/source/rpt/rpt_calculating_parameters.cc
+++ b/source/rpt/rpt_calculating_parameters.cc
@@ -6,6 +6,12 @@
 void
 RPTCalculatingParameters::declare(ParameterHandler &prm)
 {
+  prm.declare_entry("print parameters",
+                    "none",
+                    Patterns::Selection("none|only changed|all"),
+                    "Print all the parameters, or only"
+                    "the changed parameters or none");
+
   Parameters::RPTParameters::declare_parameters(prm);
   Parameters::RPTTuningParameters::declare_parameters(prm);
   Parameters::DetectorParameters::declare_parameters(prm);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2025-03-28
+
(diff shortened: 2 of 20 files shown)


source/rpt/rpt_calculating_parameters.cc
```diff
@@ -6,6 +6,12 @@
 void
 RPTCalculatingParameters::declare(ParameterHandler &prm)
 {
+  prm.declare_entry("print parameters",
+                    "none",
+                    Patterns::Selection("none|only changed|all"),
```
blaisb, line 11: i don't know if only changed or changed is the best choice of a keyword here, that would be my only question. Otherwise dope PR

blaisb (approved):

source/rpt/rpt_calculating_parameters.cc
lpsaavedra: I do not have any preference. I called it "only changed" because in the parameter handler the option is called "KeepOnlyChanged" but if you prefer "changed" that would be also possible

blaisb: Then let's keep only changed if that is coherent with the documentation

blaisb:
Can you update the changelog? Otherwise this would be good for me.

PR #1478, opened 2025-03-31
Particle particle heat transfer

marineharel:
### Description

This PR adds the functions that will be used for thermal DEM including the calculation of the thermal resistances, the total conductance  and two functions to apply heat transfer to particles for local contacts or ghost-local contacts, in the same way it is done for the contact force.

### Testing

A test of the calculation of the thermal resistances is added.

For now, these functions can only be used directly in tests. 
I plan on using them in the particle_particle_contact_force files next.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/dem/particle_particle_heat_transfer.h
+++ b/include/dem/particle_particle_heat_transfer.h
@@ -24,7 +24,7 @@
  * particles.
  * @return Contact radius of the two particles.
  */
-inline double
+double
 calculate_corrected_contact_radius(const double effective_radius,
                                    const double effective_youngs_modulus,
                                    const double effective_real_youngs_modulus,
@@ -38,7 +38,7 @@ calculate_corrected_contact_radius(const double effective_radius,
  * @param[in] contact_radius Contact radius of the two particles.
  * @return Macrocontact thermal resistance between the two particles.
  */
-inline double
+double
 calculate_macrocontact_resistance(const double harmonic_particle_conductivity,
                                   const double contact_radius);
 
@@ -57,7 +57,7 @@ calculate_macrocontact_resistance(const double harmonic_particle_conductivity,
  * @param[in] maximum_pressure Maximum pressure for hertzian contacts.
  * @return Microcontact thermal resistance between the two particles.
  */
-inline double
+double
 calculate_microcontact_resistance(const double equivalent_surface_slope,
                                   const double equivalent_surface_roughness,
                                   const double effective_microhardness,
@@ -75,7 +75,7 @@ calculate_microcontact_resistance(const double equivalent_surface_slope,
  * @param[in] contact_radius Contact radius of the two particles.
  * @return Solid macrogap thermal resistance between the two particles.
  */
-inline double
+double
 calculate_solid_macrogap_resistance(const double radius_one,
                                     const double radius_two,
                                     const double thermal_conductivity_one,
@@ -88,7 +88,7 @@ calculate_solid_macrogap_resistance(const double radius_one,
  * @param[in] x Value to which apply the function.
  * @return erfc^-1(x)
  */
-inline double
+double
 erfc_inverse_approximation(const double x);
 
 /**
@@ -106,7 +106,7 @@ erfc_inverse_approximation(const double x);
  *
  * @return Interstitial gas microgap thermal resistance between the two particles.
  */
-inline double
+double
 calculate_interstitial_gas_microgap_resistance(
   const double equivalent_surface_roughness,
   const double contact_radius,
@@ -126,7 +126,7 @@ calculate_interstitial_gas_microgap_resistance(
  *
  * @return Interstitial gas microgap thermal resistance between the two particles.
  */
-inline double
+double
 calculate_interstitial_gas_macrogap_resistance(
   const double harmonic_radius,
   const double thermal_conductivity_gas,
@@ -159,7 +159,7 @@ calculate_interstitial_gas_macrogap_resistance(
  * @param[out] thermal_conductance Total thermal conductance between the two
  * particles.
  */
-inline void
+void
 calculate_contact_thermal_conductance(
   const double radius_one,
   const double radius_two,
@@ -189,7 +189,7 @@ calculate_contact_thermal_conductance(
  * applied to particle two.
  *
  */
-inline void
+void
 apply_heat_transfer_on_local_particles(const double temperature_one,
                                        const double temperature_two,
                                        const double thermal_conductance,
@@ -209,7 +209,7 @@ apply_heat_transfer_on_local_particles(const double temperature_one,
  * applied to particle one.
  *
  */
-inline void
+void
 apply_heat_transfer_on_single_local_particle(
   const double temperature_one,
   const double temperature_two,
--- a/include/dem/particle_particle_heat_transfer.h
+++ b/include/dem/particle_particle_heat_transfer.h
@@ -28,17 +28,7 @@ inline double
 calculate_corrected_contact_radius(const double effective_radius,
                                    const double effective_youngs_modulus,
                                    const double effective_real_youngs_modulus,
-                                   const double normal_force_norm)
-{
-  double contact_radius = pow((3 * normal_force_norm * effective_radius) /
-                                (4 * effective_youngs_modulus),
-                              (1.0 / 3.0));
-
-  // apply correction
-  return contact_radius *
-         pow(effective_youngs_modulus / effective_real_youngs_modulus,
-             1.0 / 5.0);
-}
+                                   const double normal_force_norm);
 
 /**
  * @brief Calculate the macrocontact thermal resistance between two particles.
@@ -50,10 +40,7 @@ calculate_corrected_contact_radius(const double effective_radius,
  */
 inline double
 calculate_macrocontact_resistance(const double harmonic_particle_conductivity,
-                                  const double contact_radius)
-{
-  return 1 / (2.0 * contact_radius * harmonic_particle_conductivity + DBL_MIN);
-}
+                                  const double contact_radius);
 
 /**
  * @brief Calculate the microcontact thermal resistance between two particles.
@@ -76,14 +63,7 @@ calculate_microcontact_resistance(const double equivalent_surface_slope,
                                   const double effective_microhardness,
                                   const double contact_radius,
                                   const double harmonic_particle_conductivity,
-                                  const double maximum_pressure)
-{
-  return 1.184 /
-         (M_PI * harmonic_particle_conductivity * contact_radius *
-          contact_radius) *
-         (equivalent_surface_roughness / equivalent_surface_slope) *
-         pow(effective_microhardness / (maximum_pressure + DBL_MIN), 0.96);
-}
+                                  const double maximum_pressure);
 
 /**
  * @brief Calculate the solid macrogap thermal resistance between two particles.
@@ -100,22 +80,7 @@ calculate_solid_macrogap_resistance(const double radius_one,
                                     const double radius_two,
                                     const double thermal_conductivity_one,
                                     const double thermal_conductivity_two,
-                                    const double contact_radius)
-{
-  // resistance_i = characteristic length parallel to heat flux /
-  // (thermal_conductivity
-  // * characteristic area perpendicular to the heat flux)
-  const double resistance_one =
-    (M_PI * radius_one / 4.0) * 1 /
-    (M_PI * (radius_one * radius_one - contact_radius * contact_radius) *
-     thermal_conductivity_one);
-  const double resistance_two =
-    (M_PI * radius_two / 4.0) * 1 /
-    (M_PI * (radius_two * radius_two - contact_radius * contact_radius) *
-     thermal_conductivity_two);
-
-  return resistance_one + resistance_two;
-}
+                                    const double contact_radius);
 
 /**
  * @brief Define an approximation of the erfc^-1 function for values between 1e-9 and 1.9.
@@ -124,31 +89,7 @@ calculate_solid_macrogap_resistance(const double radius_one,
  * @return erfc^-1(x)
  */
 inline double
-erfc_inverse_approximation(const double x)
-{
-  if (x < 1e-9 || x > 1.9)
-    {
-      std::cerr
-        << "Error. Input out of range (10^-9 <= x <= 1.9) for erfc-1 function in thermal resistance calculation."
-        << std::endl;
-      return NAN;
-    }
-
-  if (x <= 0.02)
-    {
-      return 1.0 / (0.218 + 0.735 * pow(x, 0.173));
-    }
-  else if (x <= 0.5)
-    {
-      return (1.05 * pow(0.175, x)) / pow(x, 0.12);
-    }
-  else if (x <= 1.9)
-    {
-      return (1 - x) / (0.707 + 0.862 * x - 0.431 * x * x);
-    }
-
-  return NAN;
-}
+erfc_inverse_approximation(const double x);
 
 /**
  * @brief Calculate the interstitial gas microgap thermal resistance between two particles.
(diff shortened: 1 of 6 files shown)


source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,236 @@
...
+      return 1.0 / (0.218 + 0.735 * pow(x, 0.173));
+    }
+  else if (x <= 0.5)
+    {
+      return (1.05 * pow(0.175, x)) / pow(x, 0.12);
+    }
+  else if (x <= 1.9)
+    {
+      return (1 - x) / (0.707 + 0.862 * x - 0.431 * x * x);
+    }
```
Luckabarbeau, line 88: A sequence of conditions like that (that will be evaluated extremely frequently), ideally, if you know what is solved and the general order of magnitude of the terms, ensures that the first condition is the most likely one to be true.

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,236 @@
...
+  const double a_1 =
+    erfc_inverse_approximation(2 * maximum_pressure / effective_microhardness);
+  const double a_2 = erfc_inverse_approximation(0.03 * maximum_pressure /
+                                                effective_microhardness) -
+                     a_1;
+  // If the values are out of the range of the erfc^-1 function
+  if (a_1 == NAN || a_2 == NAN)
+    {
+      return INFINITY;
+    }
```
Luckabarbeau, line 111: If you code as a defined behavior when this function returns infinity, this is fine if either it crashes and it is the desired behavior or you catch this situation adequately; otherwise, you could use a saturation and just evaluate the erfc_inverse with the value of the closest bound.

Luckabarbeau (commented):
This is a nice new functionality. I just saw it and was curious. I added a few comments, mostly asking for a bit more comments here and there in the function. Nice work !

source/dem/particle_particle_heat_transfer.cc
blaisb: In this case since it's a < , < , < condition this current order is fine, you can then let the branch predictor do the trick :)

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,236 @@
...
+                                  const double contact_radius,
+                                  const double harmonic_particle_conductivity,
+                                  const double maximum_pressure)
+{
+  return 1.184 /
+         (M_PI * harmonic_particle_conductivity * contact_radius *
+          contact_radius) *
+         (equivalent_surface_roughness / equivalent_surface_slope) *
+         pow(effective_microhardness / (maximum_pressure + DBL_MIN), 0.96);
+}
```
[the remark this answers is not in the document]
marineharel: @Luckabarbeau I planned to add it to the documentation once thermal DEM is fully implemented, do you think I should still put it as a comment?

source/dem/particle_particle_heat_transfer.cc
marineharel: @Luckabarbeau This resistance is in parallel with another one, so when we are out of the range of erfc^-1, I just consider that this resistance has no physical meaning (particles are too close or not close enough for a microgap between particles to make sense) and putting it to infinity ignores it in the total resistance.
But I should put a comment to explain this.

include/dem/particle_particle_heat_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * @brief Calculate the macrocontact thermal resistance between two particles.
+ *
+ * @param[in] harmonic_particle_conductivity Harmonic mean of the two particles'
+ * thermal conductivities.
+ * @param[in] contact_radius Contact radius of the two particles.
+ * @return Macrocontact thermal resistance between the two particles.
+ */
+double
+calculate_macrocontact_resistance(const double harmonic_particle_conductivity,
```
blaisb, line 44: macrocontact or macro_contact? If it's in a single word in the reference material it is fine, I am just a bit surprised, I would presume it's maybe macro-contact, in that case I would call it macro_contact for enhanced readability :)

blaisb (commented):
Just starting, will finish tomorrow

blaisb (approved):
Great work @marineharel  :)

include/dem/particle_particle_heat_transfer.h
marineharel: @blaisb It is written macrocontact etc in the references I have used but I think both forms are used in literature. But maybe it would still be better to write macro_contact for readability, at least in the function names.

include/dem/particle_particle_heat_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+void
+apply_heat_transfer_on_local_particles(const double temperature_one,
+                                       const double temperature_two,
+                                       const double thermal_conductance,
+                                       double      &particle_one_heat_transfer,
+                                       double      &particle_two_heat_transfer);
+
+/**
+ * @brief Apply the heat transfer to the local-ghost particle pair in contact.
+ * The heat transfer is only applied to the local particle, so particle one
```
OGaboriault, line 194:
```suggestion
 * The heat transfer is only applied to the local particle, thus particle one
```

include/dem/particle_particle_heat_transfer.h
```diff
@@ -0,0 +1,212 @@
...
+ * applied to particle one.
+ *
+ */
+void
+apply_heat_transfer_on_single_local_particle(
+  const double temperature_one,
+  const double temperature_two,
+  const double thermal_conductance,
+  double      &particle_one_heat_transfer);
+
```
OGaboriault, line 211: what about local-ghost? 

In another PR?

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+  return contact_radius *
+         pow(effective_youngs_modulus / effective_real_youngs_modulus,
+             1.0 / 5.0);
+}
+
+double
+calculate_macrocontact_resistance(const double harmonic_particle_conductivity,
+                                  const double contact_radius)
+{
+  return 1 / (2.0 * contact_radius * harmonic_particle_conductivity + DBL_MIN);
```
OGaboriault, line 30:
```suggestion
  return 0.5 / (contact_radius * harmonic_particle_conductivity + DBL_MIN);
```

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+
+double
+calculate_microcontact_resistance(const double equivalent_surface_slope,
+                                  const double equivalent_surface_roughness,
+                                  const double effective_microhardness,
+                                  const double contact_radius,
+                                  const double harmonic_particle_conductivity,
+                                  const double maximum_pressure)
+{
+  return 1.184 /
```
OGaboriault, line 41: 1.184 comes from a simplification of the equation? If so: 

Could you write what 1.184 equal to in a comment? 
e.g.
```suggestion
  // 1.184 = PI / 3   
  return 1.184 
```

I know PI/3 is not equal to 1.184. but I think you get the point.

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+                                    const double thermal_conductivity_two,
+                                    const double contact_radius)
+{
+  // resistance_i = characteristic length parallel to heat flux /
+  // (thermal_conductivity
+  // * characteristic area perpendicular to the heat flux)
+  const double resistance_one =
+    (M_PI * radius_one / 4.0) * 1 /
+    (M_PI * (radius_one * radius_one - contact_radius * contact_radius) *
+     thermal_conductivity_one);
```
OGaboriault, line 66:
```suggestion
  const double resistance_one =
    0.25 * M_PI * radius_one /
    (M_PI * (radius_one * radius_one - contact_radius * contact_radius) *
     thermal_conductivity_one);
```

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+  const double resistance_two =
+    (M_PI * radius_two / 4.0) * 1 /
+    (M_PI * (radius_two * radius_two - contact_radius * contact_radius) *
+     thermal_conductivity_two);
+
```
OGaboriault, line 71:
```suggestion
  const double resistance_two =
    0.25 * M_PI * radius_two /
    (M_PI * (radius_two * radius_two - contact_radius * contact_radius) *
     thermal_conductivity_two);

```

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+}
+
+double
+calculate_solid_macrogap_resistance(const double radius_one,
+                                    const double radius_two,
+                                    const double thermal_conductivity_one,
+                                    const double thermal_conductivity_two,
+                                    const double contact_radius)
+{
+  // resistance_i = characteristic length parallel to heat flux /
```
OGaboriault, line 60: In this function, since you are using contact_radius more than once, you could assign a variable to it at the start of the function and use it instead.

OGaboriault: this comment also applies to the following functions.

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+  const double harmonic_radius,
+  const double thermal_conductivity_gas,
+  const double contact_radius,
+  const double gas_parameter_m)
+{
+  const double A = 2 * sqrt(harmonic_radius * harmonic_radius -
+                            contact_radius * contact_radius);
+  const double S = 2 * (harmonic_radius - (contact_radius * contact_radius) /
+                                            (2 * harmonic_radius)) +
+                   gas_parameter_m;
```
OGaboriault, line 118: contact_radius squared

OGaboriault, line 118:
```suggestion
  const double A = 2. * sqrt(harmonic_radius * harmonic_radius -
                            contact_radius * contact_radius);
  const double S = 2. * (harmonic_radius - (contact_radius * contact_radius) /
                                            (2. * harmonic_radius)) +
                   gas_parameter_m;
```

If it's a double, please put the dot.

tests/dem/particle_particle_thermal_resistances.cc
```diff
@@ -0,0 +1,293 @@
...
+  auto     particle_two          = std::next(particle_one);
+  Point<3> particle_one_location = particle_one->get_location();
+  Point<3> particle_two_location = particle_two->get_location();
+
+  const double radius_one = particle_diameter * 0.5;
+  const double radius_two = particle_diameter * 0.5;
+  const double effective_youngs_modulus =
+    youngs_modulus / (2.0 * (1 - poisson_ratio * poisson_ratio));
+  const double effective_real_youngs_modulus =
+    real_youngs_modulus / (2.0 * (1 - poisson_ratio * poisson_ratio));
```
OGaboriault, line 171:
```suggestion
    youngs_modulus / (2.0 * (1. - poisson_ratio * poisson_ratio));
  const double effective_real_youngs_modulus =
    real_youngs_modulus / (2.0 * (1. - poisson_ratio * poisson_ratio));
```

tests/dem/particle_particle_thermal_resistances.cc
```diff
@@ -0,0 +1,293 @@
...
+  const double normal_overlap =
+    (radius_one + radius_two) -
+    particle_one_location.distance(particle_two_location);
+  const double normal_force_norm = force[particle_one->get_id()].norm();
+  const double prandtl_gas =
+    dynamic_viscosity_gas * specific_heat_gas / thermal_conductivity_gas;
+  const double gas_parameter_m =
+    2 * (2 - thermal_accommodation) / thermal_accommodation *
+    (2 * specific_heats_ratio_gas) / (1 + specific_heats_ratio_gas) *
+    molecular_mean_free_path_gas / prandtl_gas;
```
OGaboriault, line 184:
```suggestion
    2. * (2. - thermal_accommodation) / thermal_accommodation *
    (2. * specific_heats_ratio_gas) / (1. + specific_heats_ratio_gas) *
    molecular_mean_free_path_gas / prandtl_gas;
```

include/dem/particle_particle_heat_transfer.h
OGaboriault, line 1: I think the declared functions in this file should be "inline", but I'm not certain. 

They are not use yet, so it doesn't really make a difference.

tests/dem/particle_particle_thermal_resistances.cc
```diff
@@ -0,0 +1,293 @@
...
+                                                   thermal_conductivity_gas,
+                                                   contact_radius,
+                                                   gas_parameter_m);
+
+  // Calculating total conductance and total resistance
+  const double total_conductance =
+    1 / (resistance_macrocontact +
+         1 / (1 / resistance_microcontact + 1 / resistance_gas_microgap)) +
+    1 / (resistance_solid_macrogap + resistance_gas_macrogap);
+  const double total_resistance = 1 / total_conductance;
```
OGaboriault, line 239:
```suggestion
  const double total_conductance =
    1. / (resistance_macrocontact +
         1. / (1. / resistance_microcontact + 1. / resistance_gas_microgap)) +
    1. / (resistance_solid_macrogap + resistance_gas_macrogap);
  const double total_resistance = 1. / total_conductance;
```

OGaboriault (changes requested):
Seems good to me. Just the inline thing that I'm not sure. 

Good job.

include/dem/particle_particle_heat_transfer.h
OGaboriault: ahh sorry you are right.  I miss read the description.

source/dem/particle_particle_heat_transfer.cc
OGaboriault: Long term, when the functions will actually be used, contact_radius_squared should be an input of the function (maybe).

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+void
+apply_heat_transfer_on_local_particles(const double temperature_one,
+                                       const double temperature_two,
+                                       const double thermal_conductance,
+                                       double      &particle_one_heat_transfer,
+                                       double      &particle_two_heat_transfer)
+{
+  const double heat_transfer =
+    thermal_conductance * (temperature_two - temperature_one);
+  particle_one_heat_transfer += heat_transfer;
```
OresteMarquis, line 213: Should this be divided by the heat capacity?

include/dem/particle_particle_heat_transfer.h
OGaboriault: The function can be inline and defined in a .cc if that function is only used in that same .cc  .

Since you have a test that uses those functions, it doesn't work. 

Maybe we should define the function in the .h ? 

(This is maybe related to the last question of reviewer 2 in the article btw)

OresteMarquis (approved):
Very nice Marine :)  Nothing much to say, except awesome work! On a side note, I was wondering Bruno if we also want a test that integrates in time/ a test for ghost particles contact/ a test for particles with  different properties  in the future? Those were elements that came up in my mind when looking at the PR.

source/dem/particle_particle_heat_transfer.cc
OresteMarquis: Actually I think is a factor coming from the heat transfer model

OGaboriault: Then it's fine!

source/dem/particle_particle_heat_transfer.cc
```diff
@@ -0,0 +1,225 @@
...
+      return INFINITY;
+    }
+
+  const double a_1 = boost::math::erfc_inv(x_1);
+  const double a_2 = boost::math::erfc_inv(x_2) - a_1;
+
+  return (2 * std::sqrt(2) * equivalent_surface_roughness * a_2) /
+         (M_PI * thermal_conductivity_gas * contact_radius * contact_radius *
+          std::log(1 + a_2 / (a_1 + gas_parameter_m /
+                                      (2 * std::sqrt(2) *
```
OGaboriault, line 103:
```suggestion
                                      (2.82842712475 *
```

source/dem/particle_particle_heat_transfer.cc
marineharel: @OresteMarquis It is divided by the heat capacity during temperature integration

include/dem/particle_particle_heat_transfer.h
marineharel: I could very well leave only the function that calculates the total conductance in the .cc and put the rest in the .h but it might decrease readability

blaisb: Yup, @Luckabarbeau is right. Inlining it's not sure if it has a good impact or not. When the function is only used locally i don't think it has any impact, it'S mostly for small functions that are used in many places in the code, but even then. Compilers have changed a lot in the last decade and many things I use to know are not true anymore.

OGaboriault (approved):

PR #1479, opened 2025-04-02
Fix refinement frequency bug in Sharp IB solver

oguevremont:
### Description

Before this PR, `lethe-fluid-sharp` would attempt to refine the cells close to the immersed boundary at every time step. This behavior is desired if solids moved since the last time step, since it ensures that the interface is properly captured regardless of the regular mesh adaptation frequency.

When solids are stationary, this behavior becomes costly if the mesh is large or if solids are numerous or complex.

### Solution

An additional condition is checked at the beginning of the function that refines immersed boundaries.

### Testing

All tests still pass. 

Costly simulations on computing clusters now run at a lower cost.

### Documentation

Not applicable.

Code related list:

Pull request related list:
- [ ] Labels are applied


blaisb (approved):
Good change. In the long term now that we can do "initial refinement" we could deprecate this completely and just use the initial refinement to refine around the immersed boundary. Maybe in a future PR @oguevremont ?

oguevremont:
> Good change. In the long term now that we can do "initial refinement" we could deprecate this completely and just use the initial refinement to refine around the immersed boundary. Maybe in a future PR @oguevremont ?

Sure! So we would include refine_ib as part of the initial refinement function, and just call initial refinement when there is movement? Since we do many cycles of flag/refine when initializing the mesh in sharp, we could find a way to only do 1 instead when not at time step 0.

PR #1462, opened 2025-03-18
Add mortar classes

peterrum:
### Description

Add classes useful for doing mortar FEM on sliding interfaces.

### Testing

A couple of unit tests. Some of them solve Poisson and Stokes equations and use multigrid.

### Documentation

--

### Miscellaneous (will be removed when merged)

--

### Checklist (will be removed when merged)
See [this page](https://chaos-polymtl.github.io/lethe/documentation/contributing.html#pull-requests) for more information about the pull request process.

Code related list:
- [x] All in-code documentation related to this PR is up to date (Doxygen format)
- [x] Copyright headers are present and up to date
- [x] Lethe documentation is up to date
- [x] New feature has unit test(s) (preferred) or application test(s), and restart files are in the generator folder
- [x] The branch is rebased onto master
- [x] Changelog (CHANGELOG.md) is up to date
- [x] Code is indented with indent-all and .prm files (examples and tests) with prm-indent

Pull request related list:
- [x] Labels are applied
- [x] There are at least 2 reviewers (or 1 if small feature) excluding the responsible for the merge
- [x] If this PR closes an issue or is related to a project, it is linked in the "Projects" or "Development" section
- [x] If the fix is temporary, an issue is opened
- [x] The PR description is cleaned and ready for merge

PR #1481, opened 2025-04-02
Read and merge meshes in mortar control

brunaccampos:
### Description

A mortar control subsection has been added to the code, based on the features from pull request #1462. 
The goal of using the mortar control is to simulate the interaction between two domains, i.e. a rotor and a stator part. In this sense, this PR introduces the mortar parameters to be read in the prm file, as well as defines a function able to read and merge two meshes. Currently this feature is accessible through the lethe-fluid solver, and is restricted to dealii mesh types only. This will be adjusted in future implementations.

### Testing

Two new tests were added considering different geometries. The goal is solely to test the function read_mesh_and_manifolds when the mortar control is enabled.

### Documentation

In the new subsection Mortar Control, the added parameters are:
- Secondary mesh: while the standard mesh read in the prm file is called the stator (outer) domain, a new (rotor) mesh is now read;
- Boundary IDs of the rotor/stator matching interfaces;
- Center of rotation for the rotor domain.

Code related list:

Pull request related list:
- [x] Labels are applied


source/core/grids.cc
```diff
@@ -394,6 +394,99 @@ read_mesh_and_manifolds(
...
+  // Check manifold IDs at rotor boundary
+  std::set<unsigned int> rotor_boundary_manifold_ids;
+  // Shift rotor boundary IDs #
+  for (const auto &face : rotor_temp_tria.active_face_iterators())
+    if (face->at_boundary())
+      {
+        face->set_boundary_id(face->boundary_id() +
+                              stator_temp_tria.get_boundary_ids().size());
+        if (face->manifold_id() != numbers::flat_manifold_id)
+          rotor_boundary_manifold_ids.insert(face->manifold_id());
```
peterrum, line 429: Why not use the same logic as for boundary_id? You can use `Triangulation::get_manifold_ids()`.

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -1625,12 +1625,19 @@ FluidDynamicsMatrixBased<dim>::solve()
...
+      this->simulation_parameters.mesh,
+      this->simulation_parameters.restart_parameters.restart,
+      this->simulation_parameters.mortar);
+  else
+    read_mesh_and_manifolds(
+      *this->triangulation,
+      this->simulation_parameters.mesh,
+      this->simulation_parameters.manifolds_parameters,
+      this->simulation_parameters.restart_parameters.restart,
+      this->simulation_parameters.boundary_conditions);
```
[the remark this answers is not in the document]
brunaccampos: This is just because the regular read_mesh_and_manifolds is setting up periodic boundary conditions, and I haven't added that to the mortar control yet (will do it soon though)

source/core/grids.cc
```diff
@@ -394,6 +394,99 @@ read_mesh_and_manifolds(
...
+
+  // Stator triangulation
+  Triangulation<dim> stator_temp_tria;
+  attach_grid_to_triangulation(stator_temp_tria, mesh_parameters);
+
+  // Rotor triangulation
+  Triangulation<dim> rotor_temp_tria;
+  attach_grid_to_triangulation(rotor_temp_tria, *mortar_parameters.rotor_mesh);
+  GridTools::rotate(mortar_parameters.rotor_mesh->rotation_angle,
+                    rotor_temp_tria);
```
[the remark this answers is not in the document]
brunaccampos: Sure, I can change that!
As it is the rotation is being done before initial refinement. When using the mapping the rotation happens after it; I'm not sure this is going to make a difference later on?

source/core/grids.cc
brunaccampos: I tried this way because I noticed that for the hyper_shell only the boundary manifold ids were stored in get_manifold_ids. For the hyper_cube_with_cylindrical_hole and the hyper_ball_balanced, though, numbers::flat_manifold_id was also stored. I added this extra check just to access ids of manifolds in the boundary only so that I can store them in the merged triangulation

source/core/grids.cc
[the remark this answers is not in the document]
blaisb: Peter is right, we should use a mapping here instead of "rotating the vertices" of the triangulation. This will lead to a more robust result and will be coherent when we are at high-order (e.g. if we have like a Q2-Q2 mesh with high-order mapping, might as well have mapping with the rotation so taht everything is rotated correctly :)

doc/source/parameters/cfd/mortar_control.rst
```diff
@@ -0,0 +1,31 @@
+==============
+Mortar Control
```
blaisb, line 2: The subsection is called mortar, so maybe this could just be called Mortar instead of mortar control? I'm not sure

doc/source/parameters/cfd/mortar_control.rst
```diff
@@ -0,0 +1,31 @@
...
+    end
+    set rotor boundary id  = 1
+    set stator boundary id = 2
+    set center of rotation = 0, 0
+  end
+
+* The mesh parameters in the :doc:`../cfd/mesh` subsection refer to the stator domain. The ``mesh`` subsection herein mentioned contains the parameters of the rotor domain; nonetheless, the input format is the same as in :doc:`../cfd/mesh`.
+
+.. note::
+  The initial number of cells at the rotor-stator interface has to be the same; the simulation will be aborted if that is not imposed. This restriction will be automatically constrained throughout the simulation if the mesh is refined.
```
blaisb, line 26:
```suggestion
  The initial number of cells at the rotor-stator interface has to be the same; the simulation will be aborted if that is not respected. This restriction will be automatically constrained throughout the simulation if the mesh is refined.
```

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
   const bool                                            &restart,
   const BoundaryConditions::BoundaryConditions          &boundary_conditions);
 
+/**
+ * @brief Completely set-up a mesh and its manifold for rotor-stator domains
+ *
+ * @param triangulation [stator] The triangulation to which a grid is attached. It will be substituted by the merged rotor-stator triangulation
+ *
+ * @param mesh_parameters [stator] The mesh parameters used to decide what type of mesh or primitive is  used
```
blaisb, line 66:
```suggestion
 * @param[in] mesh_parameters [stator] The mesh parameters used to decide what type of mesh or primitive is  used
```

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
   const bool                                            &restart,
   const BoundaryConditions::BoundaryConditions          &boundary_conditions);
 
+/**
+ * @brief Completely set-up a mesh and its manifold for rotor-stator domains
+ *
+ * @param triangulation [stator] The triangulation to which a grid is attached. It will be substituted by the merged rotor-stator triangulation
```
blaisb, line 64:
```suggestion
 * @param[in,out] triangulation [stator] The triangulation to which a grid is attached. It will be substituted by the merged rotor-stator triangulation
```

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
...
+ *
+ * @param mesh_parameters [stator] The mesh parameters used to decide what type of mesh or primitive is  used
+ *
+ * @param manifolds_parameters [stator] The information about the type of manifolds attached to the boundary conditions
```
blaisb, line 68:
```suggestion
 * @param[in] manifolds_parameters [stator] The information about the type of manifolds attached to the boundary conditions
```

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
...
+ *
+ * @param boundary_conditions [stator] The information about the boundary conditions id. This is used to set-up the periodicity of the domain
```
blaisb, line 70:
```suggestion
 * @param[in] boundary_conditions [stator] The information about the boundary conditions id. This is used to set-up the periodicity of the domain
```

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
...
+ *
+ * @param mortar [rotor] The information about the mortar method control, including the rotor mesh parameters
```
blaisb, line 72:
```suggestion
 * @param[in] mortar [rotor] The information about the mortar method control, including the rotor mesh parameters
```

include/core/parameters.h
```diff
@@ -1781,6 +1781,30 @@ namespace Parameters
...
+    // Boundary ID # of the stator at the rotor-stator interface
+    unsigned int stator_boundary_id;
+    // Center of rotation of the rotor domain
+    Point<dim> center_of_rotation;
+
+    void
+    declare_parameters(ParameterHandler &prm);
+
+    void
+    parse_parameters(ParameterHandler &prm);
```
blaisb, line 1805: Can you add the quick briefs and params here?

include/core/grids.h
```diff
@@ -59,6 +58,27 @@ read_mesh_and_manifolds(
   const bool                                            &restart,
   const BoundaryConditions::BoundaryConditions          &boundary_conditions);
 
+/**
+ * @brief Completely set-up a mesh and its manifold for rotor-stator domains
```
peterrum, line 62:
```suggestion
 * @brief Completely set up a mesh and its manifold for rotor-stator domains
```

blaisb (approved):
Just a few comments there and there. Very good work.

peterrum (commented):
@brunaccampos Let me know when the PR is ready from your side.

source/core/grids.cc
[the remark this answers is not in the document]
brunaccampos: That makes sense! I will change this on the next PR

source/core/grids.cc
brunaccampos: I tried this way just to make sure get_manifold_ids() wasn't including flat_manifold_id, which happens for some geometries but not in others

brunaccampos:
> @brunaccampos Let me know when the PR is ready from your side.

I think it's ready for now (unless any of my answers to the comments don't make sense); I will change the rotation to use the mapping in a following PR

source/core/grids.cc
```diff
@@ -394,6 +394,99 @@ read_mesh_and_manifolds(
...
+
+  // Check manifold IDs at rotor boundary
+  std::set<unsigned int> rotor_boundary_manifold_ids;
```
peterrum, line 421: I must admit I am not 100% sure if this function is doing what it should do.

First of all the question who is responsible for shifting boundary/manifold indices!? Let's discuss this issue in person, since it will be relevant once we want to specify via parameter files which face should be coupled to which.

Second, why do we only consider boundary manifolds? Manifolds can also be attached to cells and edges and these would be lost with the current implementation.

Generally, we could consider to extend `GridGenerator::merge_triangulations()` in deal.II.

peterrum (commented):
Sorry for being late.

PR #1488, opened 2025-04-09
Mortar: test oversampling

peterrum:


brunaccampos:
Is this oversampling to assign quadrature points at both interface boundaries? Or it has nothing to do with it?

peterrum:
Oversampling (in our case) means that you take more quadrature points for the coupling than for other integrals. This is for example done in preCICE (a coupling library). See: https://link.springer.com/chapter/10.1007/978-3-030-47956-5_10, where they used a factor of 3. Not sure if we will need this, but being able to easily test this does not harm.

brunaccampos (approved):

PR #1490, opened 2025-04-10
Mortar update merged triangulation data

brunaccampos:
### Description

The mortar subsection previously added in pull request #1462 is revised in this new PR. the main change is the storage of manifolds in the rotor/stator and merged triangulations, which has been fixed to account for all faces and cells. Also, a sanity check has been added to verify if the number of faces at the rotor and stator matching interfaces is the same.

### Testing

The two tests previously added for checking the `read_mesh_and_manifolds_for_stator_and_rotor` still pass, as well as all other tests.

### Documentation

The documentation has been updated to better explain the identification of the rotor/stator matching interfaces.

An additional test has been performed to check the parsing parameters function and output information once the mortar feature is enabled. Considering two separate meshes (no coupling added), a Taylor-Couette flow was simulated in each domain, and results were as expected. The test considered MPI communication, which seems to be working as well.

Code related list:

Pull request related list:
- [x] Labels are applied


source/core/parameters.cc
```diff
@@ -4152,11 +4152,6 @@ namespace Parameters
         "2",
         Patterns::Integer(),
         "Stator boundary ID # of the mortar matching interface");
-      std::string default_entry_point = (dim == 2) ? "0., 0." : "0., 0., 0.";
-      prm.declare_entry("center of rotation",
-                        default_entry_point,
-                        Patterns::List(Patterns::Double()),
-                        "Center of rotation coordinates of rotor domain");
```
peterrum, line 4159: Why did you remove this?

source/core/grids.cc
```diff
@@ -413,30 +413,78 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
+    {
+      if (face->at_boundary())
+        {
+          face->set_boundary_id(face->boundary_id() +
+                                stator_temp_tria.get_boundary_ids().size());
+          if (face->boundary_id() == mortar.rotor_boundary_id)
+            n_faces_rotor_interface++;
+        }
+      if (face->manifold_id() != numbers::flat_manifold_id)
+        face->set_manifold_id(face->manifold_id() + stator_ids_no_flat);
```
peterrum, line 452: We need to check if boundary and manifold ids of the rotor are contiguous.

include/core/parameters.h
```diff
@@ -1795,8 +1795,10 @@ namespace Parameters
     unsigned int rotor_boundary_id;
     // Boundary ID # of the stator at the rotor-stator interface
     unsigned int stator_boundary_id;
-    // Center of rotation of the rotor domain
-    Point<dim> center_of_rotation;
+    // Number of faces at the rotor side of the mortar interface
+    unsigned int n_faces_rotor_interface;
+    // Number of faces at the stator side of the mortar interface
+    unsigned int n_faces_stator_interface;
```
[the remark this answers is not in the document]
brunaccampos: I thought about storing them but then changed, and forgot to remove from here! will do it

source/core/grids.cc
```diff
@@ -400,7 +400,7 @@ read_mesh_and_manifolds_for_stator_and_rotor(
   parallel::DistributedTriangulationBase<dim, spacedim> &triangulation,
   const Parameters::Mesh                                &mesh_parameters,
   const bool                                            &restart,
-  const Parameters::Mortar<dim>                         &mortar_parameters)
+  const Parameters::Mortar<dim>                         &mortar)
```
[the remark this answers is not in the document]
brunaccampos: I wasn't sure about the naming, I can use mortar_parameters

source/core/grids.cc
```diff
@@ -413,30 +413,78 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
+        }
+      if (face->manifold_id() != numbers::flat_manifold_id)
+        face->set_manifold_id(face->manifold_id() + stator_ids_no_flat);
+    }
+
+  // Shift rotor cell manifold IDs #
+  for (const auto &cell : rotor_temp_tria.active_cell_iterators())
+    {
+      if (cell->manifold_id() != numbers::flat_manifold_id)
+        cell->set_manifold_id(cell->manifold_id() + stator_ids_no_flat);
```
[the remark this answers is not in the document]
brunaccampos: Oh sure, will add that

source/core/parameters.cc
brunaccampos: I think I thought about using the center of rotation that can be prescribed in the boundary conditions section, but I can add it again here

source/core/grids.cc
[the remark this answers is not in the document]
peterrum: If it's a parameter struct, let's keep it like this!

source/core/parameters.cc
peterrum: Let's not overthink this now. We will need to think about how to rotate the rotor in a nice way once we actually want to simulate rotating parts.

source/core/grids.cc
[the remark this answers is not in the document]
brunaccampos: If I loop over the cells and use set_all_manifold_ids, would I recursively set manifolds for faces and lines as well?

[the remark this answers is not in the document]
peterrum: Yes, you would. But you would override the values of the lines and faces, which might have a different value than the cell.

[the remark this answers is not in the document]
peterrum: The logic should be:

```
for(const auto & cell : tria.active_cell_iterators())
  {
    //  do something

    for(const auto & face : cell->face_iterators())
      //  do something

    if (dim==3)
      for(const auto line_index : cell->line_indices())
        // do something with cell->line(line_index)
  }
```

What you need to check is that you only modify the value of a face/line once. For that you can use `face->index()`/`line->index()`, to (access a vector of bools and to) check if a particular face/line has been already been modified before.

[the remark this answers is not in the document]
brunaccampos: Makes sense! I've updated the shifting IDs to loop over faces and lines inside the cell iterator

source/core/grids.cc
```diff
@@ -417,26 +417,100 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
-          rotor_boundary_manifold_ids.insert(face->manifold_id());
-      }
-
-  // Check manifold IDs at stator boundary
-  std::set<unsigned int> stator_boundary_manifold_ids;
+    {
+      if (face->at_boundary())
+        {
+          face->set_boundary_id(face->boundary_id() +
+                                stator_temp_tria.get_boundary_ids().size());
```
peterrum, line 447: Could you move `get_boundary_ids()` outside of the loop? In parallel this will probably deadlock, since the function does a global communication in each call.

source/core/grids.cc
```diff
@@ -417,26 +417,100 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
-  // Check manifold IDs at rotor boundary
-  std::set<unsigned int> rotor_boundary_manifold_ids;
+  // Get stator manifold ids without flat id
+  unsigned int stator_ids_no_flat = 0;
+  for (const auto &id : stator_temp_tria.get_manifold_ids())
+    {
+      if (id != numbers::flat_manifold_id)
+        stator_ids_no_flat++;
+    }
+
```
peterrum, line 427: I think this would be the place to check `stator_ids_no_flat == len(stator_temp_tria.get_manifold_ids() without flat_manifold_id)`. Similar a check for the boundary ids would be good. Better be safe!

peterrum (approved):

include/core/mortar_coupling_manager.h
```diff
@@ -253,12 +263,15 @@ class MortarManager
 
   /**
    * @brief Returns the weights of the quadrature points at both sides of the interface
+   * @param[in] rad Angular coordinate of cell center
+   * @param[out] points Angular weights of quadrature points of the cell
```
blaisb, line 267:
```suggestion
   *
   * @return points Angular weights of quadrature points of the cell
```

include/core/mortar_coupling_manager.h
```diff
@@ -303,10 +316,13 @@ class MortarManager
 
   /**
    * @brief Returns the normal vector for the quadrature points
+   * @param[in] rad Angular coordinate of cell center
+   * @param[out] result Normal vectors of the cell quadrature points
```
blaisb, line 320:
```suggestion
   *
   * @return result Normal vectors of the cell quadrature points
```

param[out] is better suited when you pass a variable (say a reference) and the function modifies the content of that references. If you are creating something and returning something, it is better to use the @return flag here :)

include/core/mortar_coupling_manager.h
```diff
@@ -319,12 +335,13 @@ class MortarManager
...
+   * @brief Returns the mesh alignement type and cell index
+   * @param[in] rad Angular coordinate of cell center
    * @param[out] type Cell configuration type at the interface
    *                  type = 0: mesh aligned
-   *                  type = 1: mesh not aligned, points are closer than a
-   * tolerance type = 2: mesh not aligned, points are in rotated configuration
-   * @param[out] id Point index
+   *                  type = 1: mesh not aligned, inner domain (allows rotation)
+   *                  type = 2: mesh not aligned, outer domain (fixed)
+   * @param[out] id Index of the cell in which lies the rotated cell center
```
blaisb, line 344: I would modify this part also to have return :). Especially since you are returning a pair :)

include/core/utilities.h
```diff
@@ -806,7 +806,9 @@ print_parameters_to_output_file(const ConditionalOStream &pcout,
                                 const std::string        &file_name);
 
 /**
- * @brief Converts point to radius (angle with the x axis in the x-y plane)
+ * @brief Converts point to radius
+ * @param[in] point Point in cartesian coordinates (x-y plane)
+ * @param[in] center_of_rotation Reference point
```
blaisb, line 811:
```suggestion
 * @brief Converts point to radius
 *
 * @tparam dim Number of dimension in space
 * @param[in] point Point in cartesian coordinates (x-y plane)
 * @param[in] center_of_rotation Reference point
```

source/core/grids.cc
blaisb: Agreed, this is the kind of thing that can prevent a lot of problems down the road

source/core/grids.cc
blaisb: Good catch

source/core/grids.cc
```diff
@@ -417,26 +417,100 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
+        cell->set_manifold_id(cell->manifold_id() + stator_ids_no_flat);
+      for (const auto &face : cell->face_iterators())
+        {
+          if (face->manifold_id() != numbers::flat_manifold_id &&
+              !changed_manifold_ids_faces[face->index()])
+            {
+              face->set_manifold_id(face->manifold_id() + stator_ids_no_flat);
+              changed_manifold_ids_faces[face->index()] = true;
+            }
+          if (dim == 3)
```
blaisb, line 472:
```suggestion
          if constexpr (dim == 3)
```

blaisb (approved):
1-2 small things, great work.

source/core/grids.cc
blaisb: And for GMSH grid they will all be set at once in a single grid, so this should be fine.

tests/mortar/read_mortar_data_01.cc
```diff
@@ -47,43 +49,45 @@ test()
...
-  stator_mesh.simplex                  = false;
-  stator_mesh.initial_refinement       = n_global_refinements;
-  stator_mesh.refine_until_target_size = false;
+  mesh_parameters.type                     = Parameters::Mesh::Type::dealii;
+  mesh_parameters.grid_type                = "hyper_shell";
+  mesh_parameters.grid_arguments           = "0, 0 : 0.5 : 1.0 : 6 : true";
+  mesh_parameters.scale                    = 1;
+  mesh_parameters.simplex                  = false;
+  mesh_parameters.initial_refinement       = n_global_refinements;
+  mesh_parameters.refine_until_target_size = false;
```
blaisb, line 63: don't forget to add the default parameters we talked about yesterday :)

blaisb:
@brunaccampos do you think you could add a CHANGELOG?
:) Then this would be good to go I think

PR #1493, opened 2025-04-12
Precompile functions related to revision

peterrum:
... else all files have to be recompiled after each commit 

follow-up PR to #1475


lpsaavedra (approved):

blaisb (approved):
Sounds good to me :)!

PR #1496, opened 2025-04-14
Fix rayleigh benard example

blaisb:
### Description

The Rayleigh-Bénard example would not yield the correct profile on my machine at home for the Ra=10k case and would instead lead to a quasi-static velocity field. This PR ensures that an instability is always generated by setting a linear initial condition. Furthermore, using this approach, the steady_bdf time-stepping scheme can be used to avoid going to very fine meshes and/or long running time. Using this, the simulation is much faster and the stopping criteria is independent of the end time. This actually saves a lot of time!

Code related list:
- [x] Links are added to parent .rst files
- [x] The example is following the [standard format](https://chaos-polymtl.github.io/lethe/documentation/contributing.html#general-rules-and-format)

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 - MINOR In the sharp immersed boundaries solver, added a condition that clears the combined_shape cache if a particle has moved. This fixes a bug that would lead to an incorrect level set function when particles would move accross subdomains. This only affected the visualization of the particles. [#1479](https://github.com/chaos-polymtl/lethe/pull/1495)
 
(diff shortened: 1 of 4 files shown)


OresteMarquis (approved):
I ran the problem with those changes with my home computer for the Ra10k and it works! Ithinks its a good fix.

Maybe remove the warning at the end that says : "The orientation of the vortex rotation obtained with the simulation may differ from the one above due to machine precision that generates the initial instability."

lpsaavedra (approved):

blaisb:
> I ran the problem with those changes with my home computer for the Ra10k and it works! I think its a good fix.
> 
> 
> 
> Maybe remove the warning at the end that says : "The orientation of the vortex rotation obtained with the simulation may differ from the one above due to machine precision that generates the initial instability."
> 
> 
> 
> Side note, the video doesn't correspond to the simulated flow anymore, should we mention it ?

I think the video is still fine. I will remove the warning though it's a good idea.

PR #1515, opened 2025-05-01
Mortar: Read gmsh files and compute rotor radius

brunaccampos:
### Description

Continuing the implementation of the mortar element method (#1481 and #1490), some new features have been added to the code:

- Read GMSH files in the matrix-based solver
- Read mortar section (function `read_mesh_and_manifolds_for_stator_and_rotor`) in the matrix-free solver
- New constructor for `CouplingOperator`, taking into account the Lethe parameters structure
- Compute number of subdivisions at the interface and rotor radius inside `mortar_coupling_manager`

### Testing

One test has been added to verify the computation of the rotor radius and the number of subdivisions of the rotor mesh at the rotor-stator interface. 

Furthermore, two other tests have been added to check the mortar manager (alignment/rotation of interface points) using the `read_mesh_and_manifolds_for_stator_and_rotor` and `compute_n_subdivisions_and_radius` functions.

Reading a GMSH file in the matrix-based solver and reading rotor-stator meshes in the matrix-free solver were tested with a temporary prm file (which will later become an example after interface coupling is properly added).

### Documentation

Two new simulation parameters are added: the `penalty factor` and the `oversampling factor`. Both are used in the weak imposition of the mortar coupling at the rotor-stator interface. Their descriptions have been added to the documentation.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/mortar_coupling_manager.h
```diff
@@ -213,6 +217,12 @@ class MortarManager
       }
   }
 
+  /**
+   * @brief Returns the coordinates of the quadrature points at the interface
+   * @param[in] rad Angular coordinate of cell center
```
blaisb, line 222: rad here is the radius or an angle? This is a bit ambiguous to read.

source/core/lethe_grid_tools.cc
```diff
@@ -1785,3 +1786,34 @@ LetheGridTools::find_line_sphere_intersection(
...
+    [&](const auto &cell, const auto &point) {
+      if (cell->center().norm() > radius)
+        return point;
+
+      return static_cast<Point<dim>>(
+        Physics::Transformations::Rotations::rotation_matrix_2d(
+          rotation_angle) *
+        point);
+    },
+    false);
```
blaisb, line 1812: Are you sure this work? mappingQCache is declared locally, won't it go out of scope and be de-allocated after the function?.
Have you tested that function in anyway?

tests/mortar/manager_03.cc
```diff
@@ -0,0 +1,171 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+/**
+ * @brief Mortar: check aligned mesh and output points using implemented function to
+ * read mesh and manifolds, and computing number of faces at the rotor-stator
+ * interface and the rotor radius.
+ */
```
blaisb, line 8: You don't seem to be printing the number of faces and etc. within the test, but just to output the grid and print the angle?. I would update the test description for it to match.

tests/mortar/manager_03.cc
```diff
@@ -0,0 +1,171 @@
...
+
+#include <deal.II/distributed/tria.h>
+
+#include <deal.II/fe/mapping_q.h>
```
blaisb, line 12: Might be worth it to add an output in parallel also just to test that everything is ok on 2 cores at least.

tests/mortar/rotor_radius_computation.cc
```diff
@@ -0,0 +1,217 @@
...
+                          double radius_current =
+                            v.distance(mortar_parameters.center_of_rotation);
+                          radius_min = std::min(radius_min, radius_current);
+                          radius_max = std::max(radius_max, radius_current);
+                        }
+                    }
+                }
+            }
+        }
+    }
```
blaisb, line 125: Ideally this calculation of the radius could be done within the mortar class and then you would just check that the radius you got from the mortar class is what you expected. I am not sure you need to duplicate that code here. Just make a "get_radius" function in the mortar class and use the radius calculated in the mortar class.

blaisb (changes requested):
Some comments. Ideally I would make sure that the tests are really testing the mortar interface however. Some codes seem to be duplicated from the interface to the tests.

include/core/mortar_coupling_manager.h
brunaccampos: indeed; I could create a small separate PR to refactor this name? since it appears in many functions

tests/mortar/manager_03.cc
brunaccampos: sure! will do that

tests/mortar/rotor_radius_computation.cc
brunaccampos: I used the complete code here just because I wanted to print the data separately from each mpi process, before merging info - which is done inside the function. If you think there's no need to do that I can call the function directly and just print the final result :)

tests/mortar/poisson_matrix_based_01.cc
```diff
@@ -0,0 +1,225 @@
...
+    // DoFTools::make_zero_boundary_constraints(dof_handler, 2, constraints);
+
+
+
+    coupling = std::make_shared<CouplingOperator<dim, 1, double>>(
+      mapping,
+      dof_handler,
+      constraints,
+      quadrature,
+      4 * Utilities::pow(2, n_global_refinements + 1),
```
peterrum, line 82:
```suggestion
      6 * Utilities::pow(2, n_global_refinements),
```

like in the other tests

tests/mortar/manager_03.cc
brunaccampos: I think I can delete tests manager_03 and manager_04. They are checking the mortar_manager output using the n_subdivisions and radius computed with the new function compute_n_subdivisions_and_radius. But its scope is already tested in rotor_radius_computation, so these two manager tests I have added look unnecessary

brunaccampos:
Content here transferred to merged PR #1526

PR #1523, opened 2025-05-09
Non-Newtonian Matrix-Free Operator

brunaccampos:
### Description

This PR couples the implementation of non-Newtonian models with the matrix-free solver. 
Instead of accessing solely the kinematic viscosity, the `NavierStokesOperatorBase` now access the physical properties manager, which allows the update of viscosity according to the rheological model.

### Testing

Two new tests have been added, using the Method of Manufactured Solutions with FE approximations of first and second order. The tests were adapted from the `lethe-fluid` application tests, with the only difference that now the multigrid preconditioner is chosen.

### Documentation

No parameters were introduced.

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/fluid_dynamics_matrix_free_operators.h
```diff
@@ -785,4 +825,51 @@ class NavierStokesStabilizedOperator
     const std::pair<unsigned int, unsigned int> &range) const override;
 };
 
+/**
+ * @brief Implements the matrix-free operator to solve steady/transient Non-newtonian Navier-Stokes equations using stabilization.
```
voferreira, line 829: We could add a small disclaimer saying we only have SUPG-PSPG for the time being

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -1215,6 +1215,7 @@ template <int dim>
 void
 FluidDynamicsMatrixBased<dim>::setup_preconditioner()
 {
+  // system_matrix.print(std::cout);
```
voferreira, line 1218:
```suggestion
```

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2743,7 +2737,7 @@ FluidDynamicsMatrixFree<dim>::setup_preconditioner()
 
   this->system_operator->evaluate_non_linear_term_and_calculate_tau(
     this->present_solution);
-
+  // this->system_operator->get_system_matrix().print(std::cout);
```
voferreira, line 2740:
```suggestion
```

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -785,11 +794,148 @@ NavierStokesOperatorBase<dim, number>::
...
+              //             grad_shear_rate_d += shear_rate[i][j] *
+              //             dGamma_ij_dd;
+              //           }
+              //       }
+              //     grad_shear_rate[d] = // 1.0;
+              //       (0.5 / shear_rate_magnitude) * grad_shear_rate_d;
+              //   }
+
+              // Store the shear rate magnitude in the appropriate data
+              // structure needed for the set field vector function
```
voferreira, line 898:
```suggestion
```

voferreira (approved):

source/solvers/fluid_dynamics_matrix_free_operators.cc
lpsaavedra: I think we should use this implementation instead of the other one. It is more readable and the optimizations done with the previous implementation do not really applied to matrix free. What do you think?

lpsaavedra: I just tested my implementation and it does not give the same result... weird, then I still do not understand this term correctly

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2287,19 +2289,12 @@ FluidDynamicsMatrixFree<dim>::set_initial_condition_fd(
...
-      std::shared_ptr<RheologicalModel> viscosity_model =
-        this->simulation_parameters.physical_properties_manager.get_rheology();
-
-      // Keep in memory the initial viscosity
-      const double viscosity_end = viscosity_model->get_kinematic_viscosity();
-
-      // Set it to the initial condition viscosity
-      viscosity_model->set_kinematic_viscosity(
-        this->simulation_parameters.initial_condition->kinematic_viscosity);
+      const double viscosity_end =
```
blaisb, line 2292: What is viscosity end ? Could you explain what it is or maybe give a more precise variable name ? End of what?

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2287,19 +2289,12 @@ FluidDynamicsMatrixFree<dim>::set_initial_condition_fd(
...
+        this->simulation_parameters.physical_properties_manager.get_rheology()
+          ->get_kinematic_viscosity();
 
-      // Set the kinematic viscosity in the system operator to be the
-      // temporary viscosity
+      // Set the kinematic viscosity in the system operator to be the temporary
+      // viscosity
       this->system_operator->set_kinematic_viscosity(
         this->simulation_parameters.physical_properties_manager
           .get_kinematic_viscosity_scale());
```
blaisb, line 2300: Why do you need to do that here? You could just set the viscosity in the rheology model and you would not need to specify it no? I am not sure you need to specify it in the viscosity model thorugh here? There is only one physical property manager after all.

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2503,6 +2491,12 @@ FluidDynamicsMatrixFree<dim>::assemble_system_rhs()
 {
   TimerOutput::Scope t(this->computing_timer, "Assemble RHS");
 
+  // Update the precomputed values needed for the evaluation of the residual.
+  // This is needed due to the alpha procedure of the non linear solver.
```
blaisb, line 2495:
```suggestion
  // This is needed, otherwise the line-search mechanism used in the Newton method might fail even though the Newton step should have been accepted due to a wrong evaluation of the residual and, consequently, a wrong evaluation of the step length.
```

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -785,11 +794,148 @@ NavierStokesOperatorBase<dim, number>::
...
                            9. * Utilities::fixed_power<2>(
-                                  4. * this->kinematic_viscosity / (h * h)));
+                                  4. * kinematic_viscosity / (h * h)));
 
           stabilization_parameter_lsic(cell, q) =
             std::sqrt(u_mag_squared) * h * 0.5;
         }
+
+      // Get the kinematic viscosity and its gradient from the physical
+      // properties
```
blaisb, line 804: This comment is unrelated no?

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -785,11 +794,148 @@ NavierStokesOperatorBase<dim, number>::
...
+                shear_rate_magnitude;
+
+              // Get kinematic viscosity gradient which consists of two things:
+              // 1. Grad shear rate
+              // 2. grad_kinematic_viscosity_shear_rate
+              for (unsigned int d = 0; d < dim; ++d)
+                {
+                  grad_shear_rate[d] = 0.;
+                  if constexpr (dim == 2)
+                    {
```
blaisb, line 848: In the following lines could you add the comments that explains what are the term being added like in the other matrix-free operators?

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -1323,7 +1476,7 @@ NavierStokesStabilizedOperator<dim, number>::do_cell_integral_local(
             {
               // (-ν∆δu + (u·∇)δu + (δu·∇)u)·τ∇q
```
blaisb, line 1477: Could you update those comments or add the corresponding one for the new operator?

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -1649,3 +1801,427 @@ template class NavierStokesStabilizedOperator<2, double>;
...
+NavierStokesNonNewtonianStabilizedOperator<dim, number>::
+  NavierStokesNonNewtonianStabilizedOperator() = default;
+
+/**
+ * The expressions calculated in this cell integral are:
+ * (q,∇δu) + (v,∂t δu) + (v,(u·∇)δu) + (v,(δu·∇)u) - (∇·v,δp) + ν(∇v,(∇δu +
+ * ∇δuT)) + (∇v, 0.5/γ_dot (∂ν/∂γ_dot)(∇u + ∇uT)(∇δu + ∇δuT)(∇u + ∇uT)) (Weak
+ * form Jacobian), plus three additional terms in the case of SUPG-PSPG
+ * stabilization:
+ * \+ (∂t δu +(u·∇)δu + (δu·∇)u + ∇δp - ν∆δu - (∇ν)(∇δu + ∇δuT))τ·∇q (PSPG
```
blaisb, line 1816: Very cool!

blaisb (changes requested):
Just a few comments, nothing too major. Great work :)!

source/solvers/fluid_dynamics_matrix_free_operators.cc
brunaccampos: I think just the [d] index was missplaced, but it should work! I have changed it and tests seem still ok...

include/solvers/fluid_dynamics_matrix_free_operators.h
```diff
@@ -388,7 +390,8 @@ class NavierStokesOperatorBase : public Subscriptor
   void
   set_kinematic_viscosity(const double p_kinematic_viscosity)
   {
-    kinematic_viscosity = p_kinematic_viscosity;
+    properties_manager->get_rheology()->set_kinematic_viscosity(
+      p_kinematic_viscosity);
   }
```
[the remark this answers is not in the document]
lpsaavedra: It is not necessary, we have removed it now... this was only being used since the MF operators did not have access to the physical properties, but now that they do, this is not needed.

source/solvers/fluid_dynamics_matrix_free.cc
lpsaavedra: This part of the code has now being updated. Before, we needed to set the kinematic viscosity of the operators when using viscous or ramp initial conditions, now they look very similar to the matrix-based code and use the rheological model of the physical properties.

source/solvers/fluid_dynamics_matrix_free.cc
lpsaavedra: This has been removed and updated

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2462,7 +2458,8 @@ FluidDynamicsMatrixFree<dim>::set_initial_condition_fd(
                level <= mg_operators.max_level();
                level++)
             {
-              mg_operators[level]->set_kinematic_viscosity(viscosity_end);
+              mg_operators[level]->properties_manager->set_rheology(
+                original_viscosity_model);
```
blaisb, line 2462: You don't need to do this.
Every operator has the same rheology model since they all have a shared_ptr to the physical property manager. There is only one physical property manager so you don't need to reset the properties inside the manager
Actually, you should not even access the properties manager from inside the operator.  You should just modify it outside and in the end, everybody is pointing to the same manager.

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2440,14 +2439,11 @@ FluidDynamicsMatrixFree<dim>::set_initial_condition_fd(
...
-      // Set the kinematic viscosity in the system operator to be the original
-      // viscosity
-      viscosity_model->set_kinematic_viscosity(viscosity_end);
-
-      // Reset kinematic viscosity to simulation parameters
-      viscosity_model->set_kinematic_viscosity(viscosity_end);
-      this->system_operator->set_kinematic_viscosity(viscosity_end);
+      // Reset original rheology for the system operator
+      this->system_operator->properties_manager->set_rheology(
+        original_viscosity_model);
```
blaisb, line 2444: Why not set just the property manager of the fluid dynamics class, I'm not sure i understand why you are accessing the one in the operator to do that?

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2390,22 +2395,17 @@ FluidDynamicsMatrixFree<dim>::set_initial_condition_fd(
...
+        std::make_shared<Newtonian>(
+          this->simulation_parameters.initial_condition->kinematic_viscosity);
 
-      // Set the kinematic viscosity in the system operator to be the
-      // temporary viscosity
-      this->system_operator->set_kinematic_viscosity(
-        this->simulation_parameters.physical_properties_manager
-          .get_kinematic_viscosity_scale());
+      this->system_operator->properties_manager->set_rheology(
+        temporary_rheology);
```
blaisb, line 2408: Why not set just the property manager of the fluid dynamics class, I'm not sure i understand why you are accessing the one in the operator to do that?
See below for more discussion

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -2111,16 +2111,23 @@ FluidDynamicsMatrixFree<dim>::FluidDynamicsMatrixFree(
...
+  if (this->simulation_parameters.physical_properties_manager
+        .is_non_newtonian())
+    {
+      system_operator = std::make_shared<
+        NavierStokesNonNewtonianStabilizedOperator<dim, double>>();
+      AssertThrow(
+        this->simulation_parameters.stabilization.stabilization ==
+          Parameters::Stabilization::NavierStokesStabilization::pspg_supg,
+        dealii::ExcMessage(
+          "Matrix free Non-Newtonian Navier-Stokes does not support GLS stabilization."));
```
blaisb, line 2126:
```suggestion
          "Matrix free Non-Newtonian Navier-Stokes only supports SUPG/PSPG stabilization."));
```

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -856,8 +856,7 @@ MFNavierStokesPreconditionGMGBase<dim>::reinit(
             level_constraints[level],
             quadrature_mg,
             forcing_function,
-            this->simulation_parameters.physical_properties_manager
-              .get_kinematic_viscosity_scale(),
+            this->simulation_parameters.physical_properties_manager,
```
blaisb, line 859: Since you are passing the pointer to it, which is just fine really, every level has the same physical property manager, however, in the code that follows you don't seem to be aware of that.

blaisb (changes requested):
I'll try to finish reviewing this today but i have found an issue in the way you use the physlcal property manager
Don't forget there is only one physical property manager, you set a rheology model once, no need to set it across levels.

lpsaavedra:
@blaisb Regarding the issue with the physical properties: we need to be able to change the kinematic viscosity for the viscous and the ramp initial conditions. For this, we need to change the rheological model, by getting the viscosity model and setting the kinematic viscosity to the different kinematic viscosity values required by the solver. Now, we agree that we should be able to do it through the physical properties object as in the matrix-based case, however, we are not able to do so now, because we are passing a const ref of the physical properties and using it to initialize a pointer in the operators class. Therefore, the local physical properties does not really point to the physical properties object of the matrix free class directly. We agree that we need to fix this, so that the properties within the operator are the same as the one outside of the solver (by creating a pointer to the physical properties object and using this pointer throughout both the matrix free and the operators code). We thought we would do it in a different PR since this is already large enough, but if you think this is critical now, then I can do it in the same PR.

lpsaavedra (approved):
The pointer to the physical properties manager has now been fixed. In my opinion this is now ready to be merged.

blaisb (approved):

include/solvers/fluid_dynamics_matrix_free_operators.h
```diff
@@ -693,6 +683,43 @@ class NavierStokesOperatorBase : public Subscriptor
...
+   */
+  Table<2, Tensor<1, dim + 1, Tensor<1, dim, VectorizedArray<number>>>>
+    previous_shear_rate;
+
+  /**
+   * @brief Table with correct alignment for vectorization to store the values
+   * of the shear_rate_magnitude.
+   *
+   */
+  Table<2, VectorizedArray<number>> previous_shear_rate_magnitude;
```
peterrum, line 721: We might want to move these at some time from base!?

PR #1527, opened 2025-05-12
Remove template argument from CouplingOperatorBase

peterrum:
Remove `DataType` template argument from `CouplingOperatorBase`. `CouplingOperatorBase` now only works on sets of `double`s. The `double`s need to be converted to/from the actual data types during evaluation/integration, e.g., with the help of `BufferRW`.

This PR is needed to be able to move the implementation to a source file.


blaisb (approved):

blaisb:
I'll merge now since it all looks good and I won't have my laptop with me today :). @brunaccampos please still do take a look and ask @peterrum if you have any questions :)

PR #1535, opened 2025-05-21
Add generate triangles flag when extracting slice from vtu

lpsaavedra:
### Description

Minor addition to the postprocessing script that allows to get a slice from vtu files. It adds a "generate_triangles" flag to the slice function to account correctly for vtu files with subdivisions.


mivaia (approved):
Nice addition! :)

PR #1542, opened 2025-05-28
Add particle-wall heat transfer 3/3

marineharel:
### Description

This PR finalizes the multiphysic DEM code by adding the calculation of particle-wall heat transfer and updating the temperature of solid objects when needed.
A parameter ``real_youngs_modulus`` is added, to be used for dem_mp when the Young's modulus is underestimated in the simulation.
A parameter ``disable_position_integration`` is also added to be able to freeze the position of particles. For example, this allows to have a higher time step for temperature variation in a previously loaded packed bed.

### Testing

Multiphysic DEM test cases will be added in a future PR.

### Documentation

Documentation for the new parameters is added.

Code related list:

Pull request related list:
- [x] Labels are applied


doc/source/parameters/dem/lagrangian_physical_properties.rst
```diff
@@ -151,6 +153,8 @@ In this subsection, gravitational acceleration, and the physical properties of t
 
 * The ``thermal accommodation particles`` defines the thermal accommodation coefficient of particles for each type. The thermal accommodation coefficient characterizes the quality of thermal energy exchange between gas molecules and a solid surface.
 
+* The ``real young modulus particles`` defines the real Young's modulus of particles for each type. It is used in multiphysic DEM when the Young's modulus in the simulation is underestimated. An underestimated Young's modulus would lead to an overestimated thermal contact radius. The real Young's modulus is only used when higher than the Young's modulus.
```
blaisb:
```suggestion
* The ``real young modulus particles`` defines the real Young's modulus of particles for each type. It is used in multiphysic DEM to correc the thermal contact radius. This is useful when the Young's modulus in the simulation is lowered to increase the time-step. An artificially low Young's modulus would lead to an overestimated thermal contact radius. The real Young's modulus can only be given a value that is higher than the Young's modulus. Otherwise, the regular Young's modulus will be used in the calculations.
```

include/dem/particle_wall_contact_force.h
```diff
@@ -1143,10 +1144,88 @@ class ParticleWallContactForce
...
+  /**
+   * @brief Set every containers needed to carry the heat transfer rate
+   * calculation.
+   *
+   * @param[in] dem_parameters DEM parameters declared in the .prm
+   * file.
+   */
+  void
+  set_multiphysic_properties(const DEMSolverParameters<dim> &dem_parameters)
+  {
```
blaisb: I would move the implementation of that function to the  .cc. No need in having so much stuff defined in .h :)

source/dem/dem.cc
```diff
@@ -577,6 +581,28 @@ DEMSolver<dim, PropertiesIndex>::move_solid_objects()
       simulation_control->get_previous_time());
 }
 
+template <int dim, typename PropertiesIndex>
+void
+DEMSolver<dim, PropertiesIndex>::update_temperature_solid_objects()
+{
+  // Update temperature, previous time must be used here
+  // instead of current time.
```
blaisb: Any idea why? I'm not sure I understand? I would add a more thorough comment on this :).

blaisb (changes requested):
Very few changes, very good work @marineharel

source/dem/dem.cc
marineharel: I am not really sure either but it was like this for the function move_solid_objects, so I assumed it should be the same here

marineharel: I think when we are doing the the calculation for t+dt, the wall should still be in the state of t (as are the particles since position is integrated at the end). Inside ``while (simulation_control->integrate())``, the time step is the time step that is currently calculated. I think it is because the wall is moved before calculating the force, as opposed to the particles.

blaisb: Yeah makes sense. I think that's it.

blaisb:
@OresteMarquis do you think you give it a check?

blaisb (approved):
No comments on my side. I think I will merge this @marineharel  :)

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
 
+## [Master] - 2025-06-05
+
+### Added
+
+- MAJOR Added the calculation of particle-wall heat transfer and the updating of the temperature of solid objects when needed.
+A parameter real_youngs_modulus is added, to be used for Multiphysic DEM when the Young's modulus is underestimated in the simulation.
+A parameter disable_position_integration is also added to be able to freeze the position of particles. [#1542](https://github.com/chaos-polymtl/lethe/pull/1542)
+
 ## Release of Lethe v1.0.1 - 2024-05-28
```
OresteMarquis: Not the right date for the release, maybe update it at the same time as this PR?

OresteMarquis (commented):
Great PR Marine, nothing to add! It can be merge according to me!

OresteMarquis (approved):

PR #1550, opened 2025-06-10
Mortar: some renaming

peterrum:
... and fix some comments.

### DIFF
--- a/source/core/mortar_coupling_manager.cc
+++ b/source/core/mortar_coupling_manager.cc
@@ -858,17 +858,16 @@ CouplingOperator<dim, Number>::vmult_add(VectorType       &dst,
 
   Vector<Number> buffer;
 
-  std::vector<Number> all_value_m(data.all_normals.size() * q_data_size);
-  std::vector<Number> all_value_p(data.all_normals.size() * q_data_size);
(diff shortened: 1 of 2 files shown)


source/core/mortar_coupling_manager.cc
```diff
@@ -1222,7 +1234,7 @@ CouplingEvaluationSIPG<dim, n_components, Number>::local_evaluate(
   const Vector<Number>                      &buffer,
   const unsigned int                         ptr_q,
   const unsigned int                         q_stride,
-  Number                                    *all_value_m) const
+  Number                                    *all_values_m) const
```
brunaccampos, line 1237: Would it also make sense to have this as all_values_local? And phi_m = phi_local?

blaisb: I think so yeah. Do you want to do that change directly @brunaccampos ?

peterrum: Let's keep it as it is. These are the evaluation functions. They don't know anything about ghosting but simply work on a face of a cell (m) and the one of a neighbor (p).

brunaccampos: Sure, sounds good :)

peterrum:
@blaisb :)

PR #1560, opened 2025-06-17
Cfd dem vof - variation of fluid properties with vof

antonellasuccar:
### Description

This PR adds the capability of taking into account the variations of the density and viscosity of the fluid with the volume phase fraction computed with the VOF scheme. These properties are then interpolated at the particles locations when calculating the forces involved in the VANS equations.

### Testing
 This feature has been tested on the example of  the single particle sedimentation with buoyancy and drag forces included. A prm file is used where the VOF phase fraction is initialized. A high density and a low density fluid are also used. The VOF phase fraction is set to 0 in one fluid, and to 1 in the other. The results obtained with and without VOF are identical for the same fluid.

The prm file was not added to the present branch but can be provided if needed. Future changes may include creating a new function calculate_physical_properties_at_particle_location which would involve cases where the properties in one fluid vary with parameters such as the temperature, pressure and shear rate.

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -820,9 +822,15 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-      properties_manager.get_kinematic_viscosity_scale();
+    // TODO, get the real viscosity at the particle location
+    double kinematic_viscosity_temp =
+      std::accumulate(this->kinematic_viscosity.begin(),
+                      this->kinematic_viscosity.end(),
+                      0.0) /
+      this->kinematic_viscosity.size();
+    double density_temp =
+      std::accumulate(this->density.begin(), this->density.end(), 0.0) /
+      this->density.size();
```
blaisb: I would call these variables cell_averaged since they are cell_averaged and I would comment that you are using the cell_averaged viscosity and density

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
+    const VectorType & /*current_solution*/,
+    const VectorType                      &previous_solution,
+    const VectorType                      &void_fraction_solution,
+    const Particles::ParticleHandler<dim> &particle_handler,
+    DoFHandler<dim>                       &dof_handler,
+    DoFHandler<dim>                       &void_fraction_dof_handler,
+    const DoFHandler<dim>                 &dof_handler_vof,
+    const VectorType                      &current_filtered_solution)
```
blaisb: I think it might be a better idea to provide a different function name.
Furthermore, I am not sure it's a good idea to copy paste the entire code of the reinit particle_fluid_interaction if we do any modifications it will be difficult to debug.
What I would suggest is to give this function another name (for example reinit_Particle_fluid_interactions_with_vof and to call the regular function from this function, and then on top of it do the additional VOF stuff.  Otherwise this is a lot of code duplication that could be avoided.

source/fem-dem/fluid_dynamics_vans.cc
```diff
@@ -434,16 +482,34 @@ FluidDynamicsVANS<dim>::assemble_local_system_matrix(
...
+  else
+    {
+      scratch_data.reinit_particle_fluid_interactions(
+        cell,
+        this->evaluation_point,
+        this->previous_solutions[0],
+        this->void_fraction_manager.void_fraction_locally_relevant,
+        particle_handler,
+        this->dof_handler,
+        void_fraction_manager.dof_handler);
```
blaisb: Goes back to my previous comment. Either two functions with different names or at least make it so the other function calls the other one to reduce code duplication.

source/fem-dem/vans_assemblers.cc
```diff
@@ -997,7 +996,13 @@ VANSAssemblerRong<dim>::calculate_particle_fluid_interactions(
...
          RequiresConstantDensity(
            "VANSAssemblerRong<dim>::calculate_particle_fluid_interactions"));
-  const double density = scratch_data.properties_manager.get_density_scale();
+
+  double density;
+
+  density = std::accumulate(scratch_data.density.begin(),
+                            scratch_data.density.end(),
+                            0.0) /
+            scratch_data.density.size();
```
blaisb: I don't think this is required anymore

source/fem-dem/vans_assemblers.cc
```diff
@@ -1820,7 +1824,9 @@ VANSAssemblerBuoyancy<dim>::calculate_particle_fluid_interactions(
     RequiresConstantDensity(
       "VANSAssemblerBuoyancy<dim>::calculate_particle_fluid_interactions"));
 
-  const double density = scratch_data.properties_manager.get_density_scale();
+  const std::vector<double> density = scratch_data.density_at_particle_location;
```
blaisb:
```suggestion
  const std::vector<double>& density = scratch_data.density_at_particle_location;
```
This should be a reference not a copy. Otherwise you are copying the array for every time.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1767,9 +1771,9 @@ VANSAssemblerVorticalTorque<dim>::calculate_particle_fluid_interactions(
     RequiresConstantDensity(
       "VANSAssemblerVorticalTorque<dim>::calculate_particle_fluid_interactions"));
 
-  const double density = scratch_data.properties_manager.get_density_scale();
-  const double kinematic_viscosity =
-    scratch_data.properties_manager.get_kinematic_viscosity_scale();
+  const std::vector<double> density = scratch_data.density_at_particle_location;
+  const std::vector<double> kinematic_viscosity =
```
blaisb:
```suggestion
  const std::vector<double>& density = scratch_data.density_at_particle_location;
  const std::vector<double>& kinematic_viscosity =
```
Both should be references to avoid copying.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1717,10 +1722,9 @@ VANSAssemblerViscousTorque<dim>::calculate_particle_fluid_interactions(
     RequiresConstantDensity(
       "VANSAssemblerViscousTorque<dim>::calculate_particle_fluid_interactions"));
 
-  const double density = scratch_data.properties_manager.get_density_scale();
-  const double kinematic_viscosity =
-    scratch_data.properties_manager.get_kinematic_viscosity_scale();
-
+  const std::vector<double> density = scratch_data.density_at_particle_location;
+  const std::vector<double> kinematic_viscosity =
```
blaisb:
```suggestion
  const std::vector<double>& density = scratch_data.density_at_particle_location;
  const std::vector<double>& kinematic_viscosity =
```
Should be references to avoid copies.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1866,7 +1873,8 @@ VANSAssemblerPressureForce<dim>::calculate_particle_fluid_interactions(
       "VANSAssemblerPressureForce<dim>::calculate_particle_fluid_interactions"));
 
 
-  const double density = scratch_data.properties_manager.get_density_scale();
+  const std::vector<double> density = scratch_data.density_at_particle_location;
```
blaisb:
```suggestion
  const std::vector<double>& density = scratch_data.density_at_particle_location;
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -1931,7 +1939,7 @@ VANSAssemblerShearForce<dim>::calculate_particle_fluid_interactions(
     scratch_data.properties_manager.density_is_constant(),
     RequiresConstantDensity(
       "VANSAssemblerShearForce<dim>::calculate_particle_fluid_interactions"));
-  const double density = scratch_data.properties_manager.get_density_scale();
+  const std::vector<double> density = scratch_data.density_at_particle_location;
```
blaisb:
```suggestion
  const std::vector<double>& density = scratch_data.density_at_particle_location;
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
   }
 
 
+  // Version of the function when VOF is used with CFD-DEM
```
voferreira: Could you add brief and parameters as the other functions?

source/fem-dem/vans_assemblers.cc
```diff
@@ -1790,7 +1794,7 @@ VANSAssemblerVorticalTorque<dim>::calculate_particle_fluid_interactions(
         M_PI *
         Utilities::fixed_power<3, double>(
           particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp]) *
-        kinematic_viscosity * density * 0.5;
+        kinematic_viscosity[i_particle] * density[i_particle] * 0.5;
```
voferreira, line 1781: Can you change it to particle_number as well?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+    fe_values_local_particles[velocities].get_function_values(
+      previous_solution, fluid_velocity_at_particle_location);
+
+    fe_values_local_particles[velocities].get_function_laplacians(
+      previous_solution, fluid_velocity_laplacian_at_particle_location);
+
+    // Calculate the values of the volume fraction at the particles locations
+    // when 2 fluids are present and VOF is used. These values are used to
+    // calculate the properties of the fluid at the particle location
+
```
voferreira:
```suggestion
```

voferreira (commented):
Great job! I will rereview it when you apply what was discussed about refactoring the reinit function.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -820,9 +822,15 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
     unsigned int particle_i                  = 0;
     average_fluid_particle_relative_velocity = 0;
 
-    // TODO, get the real viscosity at the particle localtion
-    double kinematic_viscosity =
-      properties_manager.get_kinematic_viscosity_scale();
+    // TODO, get the real viscosity at the particle location
+    double kinematic_viscosity_temp =
```
marineharel:
```suggestion
    const double kinematic_viscosity_temp =
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -820,9 +822,15 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+      std::accumulate(this->kinematic_viscosity.begin(),
+                      this->kinematic_viscosity.end(),
+                      0.0) /
+      this->kinematic_viscosity.size();
+    double density_temp =
```
marineharel:
```suggestion
    const double density_temp =
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
   }
 
 
+  // Version of the function when VOF is used with CFD-DEM
+  template <typename VectorType>
+  void
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
+    const VectorType & /*current_solution*/,
```
marineharel: Even in the original function, I don't understand why this is necessary ?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+                dim) /
+              (2 * dim);
+
+          average_particle_velocity += particle_velocity[particle_i];
+          particle_i++;
+        }
+      number_of_particles = particle_i;
+    }
+
+
```
marineharel:
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -848,6 +860,228 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+          (cell_volume - total_particle_volume) / cell_volume;
+
+        for (unsigned int j = 0; j < number_of_particles; ++j)
+          cell_void_fraction[j] = cell_void_fraction_bulk;
+      }
+
+    if (number_of_particles == 0)
+      return;
+
+
```
marineharel, line 1052:
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
blaisb: Since the variable is not used, you will get  a warning at compilation time :)

marineharel: I get that it is to avoid the warning but why not just remove it from the function parameters ?

antonellasuccar:
The comments have been implemented

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,67 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+        // Stock the values of particle velocity in a tensor
+        particle_velocity[particle_i][0] =
+          particle_properties[DEM::CFDDEMProperties::PropertiesIndex::v_x];
+        particle_velocity[particle_i][1] =
+          particle_properties[DEM::CFDDEMProperties::PropertiesIndex::v_y];
+        if constexpr (dim == 3)
+          particle_velocity[particle_i][2] =
+            particle_properties[DEM::CFDDEMProperties::PropertiesIndex::v_z];
+
+        cell_void_fraction[particle_i] = 0;
```
marineharel: Is this still needed ?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,67 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-          average_particle_velocity += particle_velocity[particle_i];
-          particle_i++;
-        }
-      number_of_particles = particle_i;
-    }
+        average_particle_velocity += particle_velocity[particle_i];
+        particle_i++;
+      }
+    number_of_particles = particle_i;
+    // Calculate the average particle velocity within the cell
```
marineharel:
```suggestion
    // Calculate the average particle velocity within the cell
    average_particle_velocity = average_particle_velocity / number_of_particles;
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -754,7 +736,31 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
     // Create a quadrature for the Navier-Stokes equations that is based on the
     // particle reference location
     Quadrature<dim> q_local(particle_reference_location, particle_weights);
-    FEValues<dim>   fe_values_local_particles(this->fe_values.get_fe(),
+    return q_local;
+  }
+
+  /** @brief Interpolates the velocity and pressure of the fluid, as well as the
+   * pressure gradient, and the laplacian, curl and gradient of the velocity, at
+   * at the locations of the particles.
```
marineharel:
```suggestion
   * the locations of the particles.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -754,7 +736,31 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   *
+   * @param previous_solution The solution (velocity and pressure) at the previous time step,
+   * which is used to interpolate the velocity and pressure at the particles
+   * locations.
+   *
+   * @param q_local Quadrature type object that contains the location of the particles
+   * relative the the cell's frame of reference.
```
marineharel:
```suggestion
   * relative to the cell's frame of reference.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -797,36 +803,106 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-    // Create a quadrature for the void fraction that is based on the particle
-    // reference location
-    if (this->interpolated_void_fraction)
-      {
-        FEValues<dim> fe_values_particles_void_fraction(
-          this->fe_values_void_fraction->get_fe(), q_local, update_values);
+  /** @brief Interpolates the void fraction at the locations of the particles.
+   *
+   * @param q_local Quadrature type object that contains the location of the particles
+   * relative the the cell's frame of reference.
```
marineharel:
```suggestion
   * relative to the cell's frame of reference.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +915,172 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
     average_fluid_particle_relative_velocity =
       average_fluid_particle_relative_velocity / particle_i;
   }
 
+  /** @brief Interpolates the filtered phase fraction value at the location of the particles.
+   * The latter values are used in calculating the density and viscosity of the
+   * fluid at the particles' locations when VOF is used.
+   *
+   * @param q_local Quadrature type object that contains the location of the particles
+   * relative the the cell's frame of reference.
```
marineharel:
```suggestion
   * relative to the cell's frame of reference.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +915,172 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+    // Calculate the particles properties
+    double total_particle_volume = 0;
+    total_particle_volume        = calculate_particle_properties();
+
+    calculate_cell_void_fraction(total_particle_volume);
+
+    if (number_of_particles == 0)
+      return;
+
+    average_particle_velocity = average_particle_velocity / number_of_particles;
```
marineharel: I think it makes more sense to put this in the function before

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +915,172 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+
+    // Calculate the particles properties
+    double total_particle_volume = 0;
+    total_particle_volume        = calculate_particle_properties();
+    calculate_cell_void_fraction(total_particle_volume);
+
+    if (number_of_particles == 0)
+      return;
+
+    average_particle_velocity = average_particle_velocity / number_of_particles;
```
marineharel: same here

marineharel (commented):
Just a few comments about the changes

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,67 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-  reinit_particle_fluid_interactions(
-    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
-    const VectorType & /*current_solution*/,
-    const VectorType                      &previous_solution,
-    const VectorType                      &void_fraction_solution,
-    const Particles::ParticleHandler<dim> &particle_handler,
-    DoFHandler<dim>                       &dof_handler,
-    DoFHandler<dim>                       &void_fraction_dof_handler)
+  double
+  calculate_particle_properties()
```
blaisb: What is calculate particle properties? If it extracts the properties from the particle I would just call it extract particle properties

include/solvers/navier_stokes_scratch_data.h
blaisb: Good question. Is this array used elsewhere?

blaisb: I don't think this is needed anymore, because an equality is used in the following function. No need to reset it to zero then.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,67 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+      }
+    number_of_particles = particle_i;
+    // Calculate the average particle velocity within the cell
+    return total_particle_volume;
+  }
 
+  /** @brief Computes the void fraction used for calculating the drag force as the
+   * cell void fraction
+   *
+   * @param total_particle_volume The total volume of the particles in the cell
```
blaisb:
```suggestion
  /** @brief Computes the cell void fraction.
   *
   * @param total_particle_volume The total volume of the particles in the cell
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -754,7 +736,31 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
     // Create a quadrature for the Navier-Stokes equations that is based on the
     // particle reference location
     Quadrature<dim> q_local(particle_reference_location, particle_weights);
-    FEValues<dim>   fe_values_local_particles(this->fe_values.get_fe(),
+    return q_local;
```
blaisb: You should return directly the call to the constructor. There's no point in creating the variable and then passing it by copy. You could also use a move constructor for that.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -728,17 +709,18 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   */
 
+  Quadrature<dim>
+  calculate_particle_location_quadrature_points()
+  {
     // Create local vector that will be used to spawn an in-situ quadrature to
-    // interpolate at the location of the particles
+    // interpolate at the locations of the particles
     std::vector<Point<dim>> particle_reference_location;
     std::vector<double>     particle_weights(number_of_particles, 1);
```
blaisb: I would use the opportunity to give the right size to the vector from the get go and then setting the elements one by one instead of pushing back.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -754,7 +736,31 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * at the locations of the particles.
+   *
+   * @param previous_solution The solution (velocity and pressure) at the previous time step,
+   * which is used to interpolate the velocity and pressure at the particles
+   * locations.
+   *
+   * @param q_local Quadrature type object that contains the location of the particles
+   * relative the the cell's frame of reference.
+   *
+   * @param dof_handler The DoFHandler of the Navier-Stokes problem
```
blaisb: I think it would be better to put the dof handler as the first argument, then the q local then the solution, this would make more sense.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -797,36 +803,106 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * @param void_fraction_dof_handler The DoFHandler of the void fraction equations
+   *
+   * @param void_fraction_solution The void fraction calculated with one of the methods of
+   * the VoidFractionBase class.
+   */
+
+  template <typename VectorType>
+  void
+  calculate_void_fraction_at_particle_location(
+    const Quadrature<dim> &q_local,
```
blaisb: I think calling it q_particles would be a better name no?

source/fem-dem/fluid_dynamics_vans.cc
```diff
@@ -789,6 +896,18 @@ FluidDynamicsVANS<dim>::solve()
...
+  Assert(
+    !this->simulation_parameters.physical_properties_manager.is_non_newtonian(),
+    RequiresConstantViscosity(
+      "VANSAssemblerDiFelice<dim>::calculate_particle_fluid_interactions"));
+
+  Assert(
+    this->simulation_parameters.physical_properties_manager
+      .density_is_constant(),
+    RequiresConstantDensity(
+      "VANSAssemblerDiFelice<dim>::calculate_particle_fluid_interactions"));
```
blaisb: why the Difelice assembler? I don't understand, right now all of the VANS require constnat viscosity per phase

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: I think it is needed in the case where ``interpolated_void_fraction`` is true, in the vans_assemblers.cc file, in ``cell_void_fraction =
        std::min(scratch_data.cell_void_fraction[particle_number], 1.0);``

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: I moved the comment to the ``reinit_particle_fluid_interactions function``.

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: The ``average_particle_velocity`` can only be calculated if ``number_of_particles==0``. In this case, ``reinit_particle_fluid_interactions`` should return. I can duplicate the condition, once inside the ``calculate_particle_properties``, and once in ``reinit_particle_fluid_interactions`` if you think it would make the code clearer

source/fem-dem/fluid_dynamics_vans.cc
antonellasuccar: I forgot to delete these lines. Right now, the constant properties are verified in each force assembler. Do you think it would be better to remove all of them and do the assert in ``fluid_dynamics_vans.cc`` before iterating?

blaisb: Yeah I would do the assertion once in the scratch data or something similar instead of repeating it. However, it can be ok to keep it in every assembler so that the day we change that, we do it in a well-thought out way.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -630,8 +624,7 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
       }
   }
 
-  /**
-   * @brief enable_particle_fluid_interactions Enables the calculation of the drag force by the scratch
+  /** @brief enable_particle_fluid_interactions Enables the calculation of the drag force by the scratch
```
blaisb:
```suggestion
  /** @brief enable_particle_fluid_interactions Enables the calculation of the fluid information at the particle location for the scratch.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,65 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-   * This cell must be compatible with the void fraction FE and not the
-   * Navier-Stokes FE
-   *
-   * @param current_solution The present value of the solution for [epsilon]
-   *
-   * @param previous_solutions The solutions at the previous time steps for
-   * [epsilon]
-   *
+  /** @brief Extracts the velocity of the particles and calculates their total volume
+   * in the cell
```
blaisb, line 684: Can you add a @return also with what is returned.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,65 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+        for (int d = 0; d < dim; ++d)
+          {
+            particle_properties
+              [DEM::CFDDEMProperties::PropertiesIndex::fem_force_x + d] = 0.;
+            particle_properties
+              [DEM::CFDDEMProperties::PropertiesIndex::fem_torque_x + d] = 0.;
+            undisturbed_flow_force[d]                                    = 0.;
+          }
+
+        // Stock the values of particle velocity in a tensor
```
blaisb:
```suggestion
        // Store the values of particle velocity in a tensor
```

include/solvers/navier_stokes_scratch_data.h
blaisb: yeah but before that it has been initialized with an equality no?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -728,34 +707,60 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-    if (number_of_particles == 0)
-      return;
-
-
-    // Calculate the average particle velocity within
-    // the cell
-    average_particle_velocity = average_particle_velocity / number_of_particles;
+  /** @brief Creates an object of type Quadrature<dim> that contains the
+   * particle reference location. This object can be used to interpolate any
+   * function known at the dofs at the location of the particles.
```
blaisb, line 758: Can you add a @return here like in the other function that return something.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -728,34 +707,60 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   *
+   * @param dof_handler The DoFHandler of the Navier-Stokes problem
+   */
+
+  template <typename VectorType>
+  void
+  calculate_fluid_fields_at_particle_location(
+    const DoFHandler<dim> &dof_handler,
+    const Quadrature<dim> &q_particles,
+    const VectorType      &previous_solution)
```
blaisb: is it really a previous solution or just a solution? I feel the function is agnostic of the solution being previous or not. I would just call the Vector solution.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -797,36 +802,106 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+  {
+    FEValues<dim> fe_values_particles_void_fraction(
+      this->fe_values_void_fraction->get_fe(), q_particles, update_values);
 
-        const auto &void_fraction_dh_cell =
-          typename DoFHandler<dim>::cell_iterator(
-            *this->fe_values_void_fraction->get_cell(),
-            &void_fraction_dof_handler);
+    const auto &void_fraction_dh_cell = typename DoFHandler<dim>::cell_iterator(
+      *this->fe_values_void_fraction->get_cell(), &void_fraction_dof_handler);
```
blaisb: I think this is ok for now, but this FEValues should really be a FEPointEvaluator instead of a regular FEValues. I would open an issue for this so that we keep track of this

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -797,36 +802,106 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-          void_fraction_solution, cell_void_fraction);
+    fe_values_particles_void_fraction.get_function_values(
+      void_fraction_solution, cell_void_fraction);
+  }
+
+  /** @brief Calculates the properties of the fluid at the locations of the particles. At the moment,
+   * only constant properties within the same fluid are supported. When two
+   * fluids are present and VOF is used, the properties are calculated based on
+   * the filtered phase fraction interpolated at the location of the particles.
+   * These properties are used in the forces calculations in the vans equations.
```
blaisb:
```suggestion
   * These properties are used in the forces calculations in the VANS equations.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +914,175 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   *
+   * @param dof_handler_vof The DoFHandler of the VOF equations
+   */
+
+  template <typename VectorType>
+  void
+  calculate_filtered_vof_at_particle_location(
+    const Quadrature<dim> &q_particles,
+    const VectorType      &current_filtered_solution,
+    const DoFHandler<dim> &dof_handler_vof)
```
blaisb: I think I would just call the function calculate_vof_at_particle_location and then whether we provide the filtered or not to the function would be given by the way we call it. 

Also, could you change the order of the function arguments to be dof_handler, quadrature, then solution. It is more logical this way.

include/solvers/navier_stokes_scratch_data.h
blaisb: Yeah I would duplicate it to at least make the code easier to understand.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +914,175 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+  template <typename VectorType>
+  void
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
+    const VectorType & /*current_solution*/,
+    const VectorType                      &previous_solution,
+    const VectorType                      &void_fraction_solution,
+    const Particles::ParticleHandler<dim> &particle_handler,
+    const DoFHandler<dim>                 &dof_handler,
+    const DoFHandler<dim>                 &void_fraction_dof_handler)
```
blaisb: I think it's a good opportunity to refactor the call of this scratch data function. Is it really necessary to pass the dof_handlers or can they be gathered from the member already owned? I'll double check.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +914,175 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * @param dof_handler_vof The DoFHandler object for the VOF equation
+   *
+   * @param current_filtered_solution The present value of the phase fraction at the dofs of the
+   * cell
+   */
+  template <typename VectorType>
+  void
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
+    const VectorType & /*current_solution*/,
```
blaisb: If cell is never used, I would consider removing it as an argument. to this function then.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +914,175 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * @param dof_handler The DoFHandler object for the Navier-Stokes problem
+   *
+   * @param void_fraction_dof_handler The DoFHandler object for the void fraction calculation
+   */
+
+  template <typename VectorType>
+  void
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator & /*cell*/,
+    const VectorType & /*current_solution*/,
```
blaisb: Maybe i misunderstand this, but rihgt now the particle_fluid_interaction is only calculated using the previous solution, not the current one? That's a bit of a surprise no?

source/fem-dem/fluid_dynamics_vans.cc
```diff
@@ -434,16 +480,33 @@ FluidDynamicsVANS<dim>::assemble_local_system_matrix(
...
-    this->dof_handler,
-    void_fraction_manager.dof_handler);
-
   scratch_data.calculate_physical_properties();
+
+  if (this->simulation_parameters.multiphysics.VOF)
+    {
+      scratch_data.reinit_particle_fluid_interactions(
+        cell,
+        this->evaluation_point,
```
blaisb, line 490: Hmm, that's something i'll have to doubel check. I thought we supported fully implicit coupling by evaluating everythign at the particle location using the current time-step values. We can leave it like that for now, but I'll have to revisit that

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -728,34 +707,60 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-    // Calculate the average particle velocity within
-    // the cell
-    average_particle_velocity = average_particle_velocity / number_of_particles;
+  /** @brief Creates an object of type Quadrature<dim> that contains the
+   * particle reference location. This object can be used to interpolate any
+   * function known at the dofs at the location of the particles.
+   */
 
+  Quadrature<dim>
+  calculate_particle_location_quadrature_points()
```
blaisb: It's not a calculation but more of a gathering step no?
```suggestion
  gather_particles_reference_location()
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -1289,34 +1282,32 @@ void
...
     scratch_data.properties_manager.density_is_constant(),
     RequiresConstantDensity(
       "VANSAssemblerGidaspow<dim>::calculate_particle_fluid_interactions"));
-  const double density = scratch_data.properties_manager.get_density_scale();
 
+  // particle_number is an increment that goes from 0 to n_particles_in_cell.
+  // It is incremented at the end of the loop over particles and is used to
+  // point to the element of the vectors relative_velocity and
+  // fluid_velocity_at_particle_location corresponding to the particle being
+  // looped over.
```
blaisb: there is no variable named particle number near here.... I would consider moving this comment to a more appropriate location.

source/fem-dem/vans_assemblers.cc
```diff
@@ -1717,10 +1708,12 @@ VANSAssemblerViscousTorque<dim>::calculate_particle_fluid_interactions(
     RequiresConstantDensity(
       "VANSAssemblerViscousTorque<dim>::calculate_particle_fluid_interactions"));
 
-  const double density = scratch_data.properties_manager.get_density_scale();
-  const double kinematic_viscosity =
-    scratch_data.properties_manager.get_kinematic_viscosity_scale();
+  // particle_number is an increment that goes from 0 to n_particles_in_cell.
+  // It is incremented at the end of the loop over particles
```
blaisb: Same here regarding the particle-number. The declaration of the variable is 7 lines below.

blaisb (changes requested):
Some comments to address. I'll retake a look at the code as a whole to double check some stuff tomorrow.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -645,79 +638,65 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+        auto particle_properties = particle.get_properties();
+        // Set the particle_fluid_interactions properties and vectors to 0
+        for (int d = 0; d < dim; ++d)
+          {
+            particle_properties
+              [DEM::CFDDEMProperties::PropertiesIndex::fem_force_x + d] = 0.;
+            particle_properties
+              [DEM::CFDDEMProperties::PropertiesIndex::fem_torque_x + d] = 0.;
+            undisturbed_flow_force[d]                                    = 0.;
+          }
```
marineharel, line 678: This makes sense in ``reinit_particle_fluid_interactions``  but considering this function is about getting the velocities and the particle volume, this feels a bit random here, could it be put somewhere else ?

include/solvers/navier_stokes_scratch_data.h
marineharel: Either way, I don't think it should be left here since this function is just to extract velocities and volume now

antonellasuccar: It is initialized in `navier-stokes_scratch_data.h`, line 1390, as `std::vector<double>         cell_void_fraction` and it is given a size in `navier-stokes_scratch_data.cc`, line 345, in `enable_particle_fluid_interaction` as `cell_void_fraction = std::vector<double>(n_global_max_particles_per_cell);` but I do not think a value is assigned to it in the case where `interpolated_void_fraction` is `true`. To make the code clearer, I will move the initialization to `calculate_cell_coid_fraction`

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: I agree that the function just takes a vector solution and does not care whether it is previous or not. It is determined when it is called. So I changed `previous_solution` to `vector_solution` in this function.

source/fem-dem/vans_assemblers.cc
antonellasuccar: I moved the comment to before the declaration of `particle_number` only in  `VANSAssemblerDiFelice<dim>` since it is where it first appears

source/fem-dem/vans_assemblers.cc
antonellasuccar: I removed the comments except for the first time `particle_number` is used which is in `VANSAssemblerDiFelice<dim>`. Also, the variable was incremented outside of the for loop in two instances, so i put it in the loop too.

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: You are right, but I was avoiding more loops (better performance?). Not sure what is the practice to follow in our case @blaisb

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: Will open an issue

antonellasuccar (commented):
I implemented most of the comments. I will also open an issue for the FEValues and PointEvaluator conversion.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -769,64 +783,132 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+              dynamic_viscosity_at_particle_location /
+              density_at_particle_location[particle_i];
+          }
       }
+    else
+      {
+        for (unsigned int particle_i = 0; particle_i < number_of_particles;
+             ++particle_i)
+          {
+            // Calculate the kinematic viscosity and density at the particle
```
voferreira: "Get" or "calculate"?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -769,64 +783,132 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+
+  { // Relative velocity and particle Reynolds
     unsigned int particle_i                  = 0;
     average_fluid_particle_relative_velocity = 0;
 
-    // TODO, get the real viscosity at the particle localtion
-    double kinematic_viscosity =
-      properties_manager.get_kinematic_viscosity_scale();
-
+    // At the moment, the kinematic viscosity and density are taken to be
```
voferreira: Maybe I would add this in the "brief". What do you think?

source/fem-dem/vans_assemblers.cc
```diff
@@ -1173,8 +1165,8 @@ VANSAssemblerKochHill<dim>::calculate_particle_fluid_interactions(
                   0.0232 / pow(cell_void_fraction, 5);
 
       double momentum_transfer_coefficient =
-        ((18 * kinematic_viscosity * pow(cell_void_fraction, 2) *
-          (1 - cell_void_fraction)) /
+        ((18 * kinematic_viscosity[particle_number] *
+          pow(cell_void_fraction, 2) * (1 - cell_void_fraction)) /
```
voferreira: Can you use fixed pow here?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -769,64 +783,132 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
     fluid_velocity_curls_at_particle_location_3d.resize(number_of_particles);
     fluid_pressure_gradients_at_particle_location.resize(number_of_particles);
 
-    // Evaluate the relevant information at the
-    // quadrature points to do the interpolation.
-    const auto &velocity_cell =
-      typename DoFHandler<dim>::cell_iterator(*this->fe_values.get_cell(),
-                                              &dof_handler);
+    // // Evaluate the relevant information at the
```
marineharel: This commented code can be removed I think

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
             fluid_particle_relative_velocity_at_particle_location[particle_i]
               .norm() *
             particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp] /
-            (kinematic_viscosity + DBL_MIN);
+            (kinematic_viscosity_at_particle_location[particle_i] + DBL_MIN);
         particle_i++;
       }
 
     average_fluid_particle_relative_velocity =
```
marineharel, line 968: Should we check here that particle_i !=0 ?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+    filtered_phase_values_at_particle_location.resize(number_of_particles);
+
+    fe_values_vof_local_particles.reinit(phase_cell);
+
+    fe_values_vof_local_particles.get_function_values(
+      current_filtered_solution, filtered_phase_values_at_particle_location);
+  }
+
+  /** @brief Calculates the variables needed to compute the particle fluid interactions
+   * in the vans equations.
```
marineharel:
```suggestion
   * in the VANS equations.
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   *
+   * @param previous_solution The solution at the previous time step for the fluid's velocity
```
marineharel: The params should be updated here

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+                                                     void_fraction_cell,
+                                                     void_fraction_solution);
+      }
+    calculate_fluid_properties_at_particle_location();
+    calculate_force_parameters_at_particle_location();
+  }
+
+  // Version of the function when VOF is used with CFD-DEM
+  /** @brief Calculates the variables needed to compute the particle fluid interactions
+   * in the vans equations.This version of the function is used when VOF is used
```
marineharel:
```suggestion
   * in the VANS equations.This version of the function is used when VOF is used
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * with CFD-DEM.
+   *
+   * @param previous_solution The solution at the previous time step for the fluid's velocity
```
marineharel: The params should be updated here

include/solvers/navier_stokes_scratch_data.h
marineharel: You should be able to remove this here now

marineharel (commented):
Some additional comments

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: You are right. I removed it altogether since it does not belong in this function. It is already added in the " brief" of the appropriate function: `calculate_fluid_properties_at_particle_location`

source/fem-dem/vans_assemblers.cc
blaisb: +1 to that comment, please use the fixed_pow utilities function instead of pow we should never use pow.

include/solvers/navier_stokes_scratch_data.h
blaisb: I would put it somewhere else. These type of micro-optimization won't help performance. Focus on readability for now. @marineharel is right.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -728,34 +712,64 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * @param vector_solution The solution (velocity and pressure) that is used to
+   * interpolate the velocity and pressure at the particles locations.
+   */
+
+  template <typename VectorType>
+  void
+  calculate_fluid_fields_at_particle_location(
+    const Quadrature<dim>                                &q_particles,
+    const typename DoFHandler<dim>::active_cell_iterator &velocity_cell,
+    const VectorType                                     &vector_solution)
```
blaisb:
```suggestion
    const VectorType                                     &solution)
```

blaisb: I think solution would be better than vector_solution.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -769,64 +783,132 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * @param void_fraction_cell The active cell associated with the void fraction DoFHandler
+   *
+   * @param void_fraction_solution The void fraction calculated with one of the methods of
+   * the VoidFractionBase class.
+   */
+
+  template <typename VectorType>
+  void
+  calculate_void_fraction_at_particle_location(
+    const Quadrature<dim>                                &q_particles,
```
blaisb: Same as above, good time to have a more verbose adequate name for the quadrature.

include/solvers/navier_stokes_scratch_data.h
blaisb: Or otherwise navier_stokes_solution since you use void_fraction_solution below.

include/solvers/navier_stokes_scratch_data.h
blaisb: Gather would be the adequate here I think

include/solvers/navier_stokes_scratch_data.h
blaisb: Agreed. Good point.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * The latter values are used in calculating the density and viscosity of the
+   * fluid at the particles' locations when VOF is used.
+   *
+   * @param q_particles Quadrature type object that contains the location of the particles
+   * relative to the cell's frame of reference.
+   *
+   * @param phase_cell The active cell associated with the vof DoFHandler
+   *
+   * @param current_filtered_solution The present value of the solution for the filtered phase
+   * fraction
```
blaisb: Well that for the VOF right? I would be more verbose in the wording here because the phase fraction to me could be related to the void fraction. I would say the filtered VOF solution.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+      {
+        calculate_void_fraction_at_particle_location(q_particles,
+                                                     void_fraction_cell,
+                                                     void_fraction_solution);
+      }
+    calculate_fluid_properties_at_particle_location();
+    calculate_force_parameters_at_particle_location();
+  }
+
+  // Version of the function when VOF is used with CFD-DEM
```
blaisb:
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+    if (number_of_particles == 0)
+      return;
+
+    Quadrature<dim> q_particles = gather_particles_reference_location();
+    calculate_fluid_fields_at_particle_location(q_particles,
+                                                velocity_cell,
+                                                previous_solution);
+
+    // Create a quadrature for the void fraction that is based on the particle
+    // reference location
```
blaisb: This comment is not relevant anymore no?
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * cell
+   */
+  template <typename VectorType>
+  void
+  reinit_particle_fluid_interactions(
+    const typename DoFHandler<dim>::active_cell_iterator &velocity_cell,
+    const typename DoFHandler<dim>::active_cell_iterator &void_fraction_cell,
+    const typename DoFHandler<dim>::active_cell_iterator &phase_cell,
+    const VectorType & /*current_solution*/,
+    const VectorType                      &previous_solution,
```
blaisb: Maybe a good idea to call the solution fluid_dynamics_solution and previous solutions here to be more clear.

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -839,14 +921,167 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+    const VectorType                      &void_fraction_solution,
+    const Particles::ParticleHandler<dim> &particle_handler,
+    const VectorType                      &current_filtered_solution)
+  {
+    pic = particle_handler.particles_in_cell(this->fe_values.get_cell());
+
+    // Calculate the particles properties
```
blaisb:
```suggestion
```
This comment is not adequate anymore I think

source/fem-dem/vans_assemblers.cc
```diff
@@ -1345,7 +1329,7 @@ VANSAssemblerGidaspow<dim>::calculate_particle_fluid_interactions(
             (18 * pow(cell_void_fraction, -3.65) *
              (1 + 0.15 * pow(Re_p[particle_number], 0.687))) *
             (particle_properties[DEM::CFDDEMProperties::PropertiesIndex::mass] *
-             kinematic_viscosity /
+             kinematic_viscosity[particle_number] /
```
blaisb: I'm not sure i'm a fan of this particle_number, I would rather it be i_particle or something similar, since it is not the number of particles and I find particle_number to be slightly confusing

blaisb (changes requested):

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: I put it in a separate function called ``reinit_particle_fluid_forces``

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: I changed it to ``velocity_pressure_solution``

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: It is now ``q_particles_location``

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: Changed it to "Gather"

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: Since we want to be more verbose anyway, and fluid_dynamics_solution or navier_stokes_solution may or may not include the void fraction, depending on how we see it, i will use ``velocity_pressure_solution`` or ``previous_velocity_pressure_solution``everywhere instead of just ``solution`` or ``navier_stokes_solution``.

source/fem-dem/vans_assemblers.cc
antonellasuccar: it is now ``i_particle`` in ``vans_assemblers.cc`` and ``navier_stokes_scratch_data.h``

source/fem-dem/vans_assemblers.cc
antonellasuccar: Replaced all pow with integer exponent with ``Utilities::fixed_power``.

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: This comment is removed

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: There is an `` if (number_of_particles == 0)
      return;`` in ``reinit_particle_fluid_interactions`` before calling calculate_force_parameters_at_particle_location();

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: We now use filter VOF solution

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -597,8 +592,7 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
                                                    this->curvature_values);
   }
 
-  /**
-   * @brief enable_void_fraction Enables the collection of the void fraction
+  /** @brief enable_void_fraction Enables the collection of the void fraction
```
voferreira: I think we usually have \@brief in the second line after /**

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -663,79 +656,81 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+          particle_velocity[i_particle][2] =
+            particle_properties[DEM::CFDDEMProperties::PropertiesIndex::v_z];
+
+        if (!interpolated_void_fraction)
+          total_particle_volume +=
+            M_PI *
+            pow(particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp],
                 dim) /
-              (2 * dim);
+            (2 * dim);
```
voferreira, line 714:
```suggestion
            (2.0 * dim);
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -926,38 +918,52 @@ VANSAssemblerDiFelice<dim>::calculate_particle_fluid_interactions(
...
       // Di Felice Drag Model CD Calculation
-      C_d = pow((0.63 + 4.8 / sqrt(Re_p[particle_number])), 2) *
+      C_d = Utilities::fixed_power<2>((0.63 + 4.8 / sqrt(Re_p[i_particle]))) *
             pow(cell_void_fraction,
-                2 - (3.7 -
-                     0.65 *
-                       exp(-pow((1.5 - log10(Re_p[particle_number])), 2) / 2)));
+                2 - (3.7 - 0.65 * exp(-Utilities::fixed_power<2>(
+                                        1.5 - log10(Re_p[i_particle])) /
+                                      2)));
```
voferreira:
```suggestion
                                      2.0)));
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -997,39 +994,51 @@ VANSAssemblerRong<dim>::calculate_particle_fluid_interactions(
...
-                       exp(-pow(1.5 - log10(Re_p[particle_number]), 2) / 2)));
+      C_d =
+        Utilities::fixed_power<2>((0.63 + 4.8 / sqrt(Re_p[i_particle]))) *
+        pow(cell_void_fraction,
+            2 -
+              (2.65 * (cell_void_fraction + 1) -
+               (5.3 - (3.5 * cell_void_fraction)) *
+                 Utilities::fixed_power<2>(cell_void_fraction) *
+                 exp(-Utilities::fixed_power<2>(1.5 - log10(Re_p[i_particle])) /
+                     2)));
```
voferreira:
```suggestion
                     2.0)));
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -997,39 +994,51 @@ VANSAssemblerRong<dim>::calculate_particle_fluid_interactions(
...
 
       double momentum_transfer_coefficient =
         (0.5 * C_d * M_PI *
-         pow(particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp],
-             2) /
+         Utilities::fixed_power<2>(
+           particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp]) /
          4) *
```
voferreira:
```suggestion
         4.0) *
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -1070,31 +1071,39 @@ VANSAssemblerDallavalle<dim>::calculate_particle_fluid_interactions(
...
+  unsigned int i_particle = 0;
 
   // Loop over particles in cell
   for (auto &particle : pic)
     {
       auto particle_properties = particle.get_properties();
 
       // Dallavalle Drag Model CD Calculation
-      C_d = pow((0.63 + 4.8 / sqrt(Re_p[particle_number])), 2);
+      C_d = Utilities::fixed_power<2>((0.63 + 4.8 / sqrt(Re_p[i_particle])));
```
voferreira:
```suggestion
      C_d = Utilities::fixed_power<2>(0.63 + 4.8 / sqrt(Re_p[i_particle]));
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -1070,31 +1071,39 @@ VANSAssemblerDallavalle<dim>::calculate_particle_fluid_interactions(
...
 
       double momentum_transfer_coefficient =
         (0.5 * C_d * M_PI *
-         pow(particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp],
-             2) /
+         Utilities::fixed_power<2>(
+           particle_properties[DEM::CFDDEMProperties::PropertiesIndex::dp]) /
          4) *
```
voferreira:
```suggestion
         4.0) *
```

include/solvers/navier_stokes_scratch_data.h
blaisb: To be honest this is not important. It use to be important in the 90 or 2000, nowadays any compiler will optimize this away.

source/fem-dem/vans_assemblers.cc
blaisb: Same here this is not important.

source/fem-dem/vans_assemblers.cc
blaisb: Here a better thing would be to replace the /2 by *.5, but even that I don't think makes a real difference anymore.

source/fem-dem/vans_assemblers.cc
blaisb: Same as above, here would be better to do *0.25 than /4 .

source/fem-dem/vans_assemblers.cc
blaisb: *0.25 instead

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -787,84 +812,295 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
 
-    // Create a quadrature for the void fraction that is based on the particle
-    // reference location
-    if (this->interpolated_void_fraction)
-      {
-        FEValues<dim> fe_values_particles_void_fraction(
-          this->fe_values_void_fraction->get_fe(), q_local, update_values);
+  /** @brief Interpolates the void fraction at the locations of the particles.
+   *
+   * @param q_particles_location Quadrature type object that contains the location of the particles
```
blaisb:
```suggestion
   * @param[in] q_particles_location Quadrature type object that contains the location of the particles
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -787,84 +812,295 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   * relative to the cell's frame of reference.
+   *
+   * @param void_fraction_cell The active cell associated with the void fraction DoFHandler
```
blaisb:
```suggestion
   * @param[in] void_fraction_cell The active cell associated with the void fraction DoFHandler
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -787,84 +812,295 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
+   *
+   * @param void_fraction_solution The void fraction calculated with one of the methods of
```
blaisb:
```suggestion
   * @param[in] void_fraction_solution The void fraction calculated with one of the methods of
```

source/fem-dem/vans_assemblers.cc
```diff
@@ -297,10 +299,13 @@ VANSAssemblerCoreModelB<dim>::assemble_rhs(
...
-                         9 * std::pow(4 * kinematic_viscosity / (h * h), 2)) :
-          1. / std::sqrt(std::pow(sdt, 2) + std::pow(2. * u_mag / h, 2) +
-                         9 * std::pow(4 * kinematic_viscosity / (h * h), 2));
+          1. / std::sqrt(Utilities::fixed_power<2>(2. * u_mag / h) +
+                         9 * Utilities::fixed_power<2>(4 * kinematic_viscosity /
+                                                       (h * h))) :
+          1. / std::sqrt(Utilities::fixed_power<2>(sdt) +
+                         Utilities::fixed_power<2>(2. * u_mag / h) +
+                         9 * Utilities::fixed_power<2>(4 * kinematic_viscosity /
+                                                       (h * h)));
```
blaisb: Same as before

source/fem-dem/vans_assemblers.cc
```diff
@@ -89,10 +89,13 @@ VANSAssemblerCoreModelB<dim>::assemble_matrix(
...
```
blaisb: We have a function now to calculate the stabilization. [You can see it here](https://github.com/chaos-polymtl/lethe/blob/master/include/solvers/stabilization.h#L30)

source/fem-dem/vans_assemblers.cc
```diff
@@ -454,10 +457,13 @@ VANSAssemblerCoreModelA<dim>::assemble_matrix(
...
```
blaisb: Same as before

source/fem-dem/vans_assemblers.cc
```diff
@@ -661,10 +667,13 @@ VANSAssemblerCoreModelA<dim>::assemble_rhs(
...
```
blaisb: Same as before

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: I multiplied by 0.5 instead

source/fem-dem/vans_assemblers.cc
antonellasuccar: I used *0.5

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

source/fem-dem/vans_assemblers.cc
antonellasuccar: This is done

include/solvers/navier_stokes_scratch_data.h
antonellasuccar: All the arguments in the functions i added are param[in], so i added it to all of them. I can relook at all the functions in navier_stokes_scratch_data.h and see their arguments to specify if param[in] or [out] if you think that is necessary, maybe in a separate PR?

PR #1562, opened 2025-06-25
Add surface area computation vof

hepap:
### Description

The surface area computation of the interface was added for vof simulations when mass conservation is enable. For that, the NonMatching FEValues features of deal.II are used. Additionally, the geometric volume computation was changed to use the same features.

### Testing

One unit test is added to check the surface and volume computations and their convergence rate in space. Both metrics converge to the second order.

Application tests outputting the mass conservation were updated because the additional surface area column was added to the outputted table.

### Documentation

No documentation is required.

Code related list:

Pull request related list:
- [ ] Labels are applied


tests/core/interface_tools_sphere_volume.cc
```diff
@@ -21,11 +21,12 @@
...
-  a level-set field using the InterfaceTools::compute_volume function. The
-  level-set field of interest is the one describing a sphere. The volume is
-  computed for 3 mesh refinements and the test checks the error on the volume
-  and the convergence rate of the method (formally 2).
+  /* This test checks the computation of the surface area of the level 0.1 of
+  a level-set field and the volume enclosed by it using the
+  InterfaceTools::compute_surface_and_volume function. The level-set field of
+  interest is the one describing a sphere. The surface and volume are computed
+  for 3 mesh refinements and the test checks the error on these metriics and the
+  convergence rate of the method (formally 2).
```
AmishgaAlphonius:
```suggestion
  /* This test checks the computation of the surface area of the level 0.1 of
  a level-set field and the volume enclosed by it using the
  InterfaceTools::compute_surface_and_volume function. The level-set field of
  interest is the one describing a sphere. The surface and volume are computed
  for 3 mesh refinements and the test checks the error on these metrics and the
  convergence rate of the method (formally 2).
```

AmishgaAlphonius (approved):
Nice work! Just one typo :)

source/core/interface_tools.cc
```diff
@@ -54,56 +54,97 @@ InterfaceTools::compute_cell_wise_volume(
...
+    dof_handler, level_set_vector_relevant_copy);
+  mesh_classifier.reclassify();
+
+  const hp::FECollection<dim> fe_collection(fe);
+
+  const QGauss<1> quadrature_1D(fe.degree + 1);
+
+  NonMatching::RegionUpdateFlags region_update_flags;
+  region_update_flags.inside  = update_JxW_values | update_quadrature_points;
+  region_update_flags.surface = update_JxW_values | update_quadrature_points;
```
blaisb: Do you really need to update the quadrature point location? Just the JxW values no?

tests/core/interface_tools_sphere_volume.cc
```diff
@@ -43,16 +44,18 @@ test()
...
+  Vector<double> error_surface(3);
 
   // Loop for the mesh convergence study
   for (unsigned int n = 0; n < 3; n++)
     {
       dof_handler.reinit(triangulation);
       dof_handler.distribute_dofs(fe);
 
-      Vector<double> signed_distance;
+      dealii::TrilinosWrappers::MPI::Vector signed_distance;
```
blaisb, line 57: I am curious, why this change? In theory all of the postprocess tools like this should kinda work with all kind of vectors no? For example deal.II parallel vectors

tests/core/interface_tools_sphere_volume.cc
```diff
@@ -61,29 +64,45 @@ test()
...
+    log(error_surface[2] / error_surface[1]) / log(0.5);
+  const double convergence_order_volume =
     log(error_volume[2] / error_volume[1]) / log(0.5);
 
-  deallog << "The convergence is: " << convergence_order << std::endl;
+  deallog << "The convergence rate for the surface is: "
+          << convergence_order_surface << std::endl;
+
+  deallog << "The convergence rate for the volume is: "
+          << convergence_order_volume << std::endl;
```
blaisb: Why not report the convergence rate between 0 and 1 and 1 and 2? It would be a great test.
You could consider using a deal.II error table, like we do in regular comparisons to analytical solutions, it would make the output clean and easy to understand.

tests/core/interface_tools_sphere_volume.output
blaisb, line 1: Since this test is compatible with Trilinos vectors, consider making a parallel version of it also by adding .mpirun=2. before the .output

tests/core/interface_tools_sphere_volume_cell_wise_sum.output
```diff
@@ -0,0 +1,5 @@
+
+DEAL::The volume error for ref. lev. 3 is: 0.0114409
+DEAL::The volume error for ref. lev. 4 is: 0.00285945
+DEAL::The volume error for ref. lev. 5 is: 0.000715903
+DEAL::The convergence rate is: 1.99790
```
blaisb: Same as above for the convergence rate and the potential parallel test :)

source/solvers/vof.cc
```diff
@@ -1269,40 +1280,55 @@ VolumeOfFluid<dim>::postprocess(bool first_iteration)
                   mass_column_name   = "mass_per_length_" + fluid_id;
                   geometric_volume_column_name =
                     "geometric_surface_" + fluid_id;
+                  area_column_name        = "length_" + fluid_id;
+                  volume_test_column_name = "surface_test_" + fluid_id;
```
AmishgaAlphonius: This seems to have been forgotten
```suggestion
```

source/solvers/vof.cc
```diff
@@ -1269,40 +1280,55 @@ VolumeOfFluid<dim>::postprocess(bool first_iteration)
...
                 }
               else if constexpr (dim == 3)
                 {
                   volume_column_name           = "volume_" + fluid_id;
                   mass_column_name             = "mass_" + fluid_id;
                   geometric_volume_column_name = "geometric_volume_" + fluid_id;
+                  area_column_name             = "surface_" + fluid_id;
+                  volume_test_column_name      = "volume_test_" + fluid_id;
```
AmishgaAlphonius: Same here
```suggestion
```

source/solvers/vof.cc
```diff
@@ -1191,6 +1198,8 @@ VolumeOfFluid<dim>::postprocess(bool first_iteration)
       std::string mass_column_name;
 
       std::string geometric_volume_column_name;
+      std::string volume_test_column_name;
```
AmishgaAlphonius:
```suggestion
```

AmishgaAlphonius (approved):

source/core/interface_tools.cc
hepap: yes you are right :)

tests/core/interface_tools_sphere_volume.cc
hepap: It's because of the copy vector I use to avoid the ghost cell problem. Since I'm constructing it with `VectorType level_set_vector_owned_copy(dof_handler.locally_owned_dofs(),                                         mpi_communicator);`, it doesn't work for `Vector<double>`.

I can fix it so it works for any vector type, but now it works for Trilinos and LinearAlgebra vector.

PR #1563, opened 2025-06-25
Mortar: Navier Stokes coupling

brunaccampos:
### Description

This PR couples the mortar method feature with the Navier Stokes matrix-based solver.

### Testing

An application test for a Taylor-Couette flow has been added. It uses the mortar feature to "bind" two `hyper_shell` meshes; the results show a smooth transition of velocity and pressure fields at the interface (i.e., mortar coupling terms ensure continuity at the interface).

### Documentation

The simulation parameters used in the mortar section have already been added to the documentation in previous PRs.

The NS mortar coupling was tested for Q2-Q1 and Q2-Q2 cases. The Q1-Q1 case, even with stabilization, does not lead to correct results. This will be further investigated.

Code related list:

Pull request related list:
- [x] Labels are applied


blaisb:
> I had to add a dealii version check when declaring `std::shared_ptr<MortarManagerCircle<dim>> mortar_manager` `std::shared_ptr<CouplingOperator<dim, double>> mortar_coupling_operator` so that the code compiles with dealii 9.6. Then, the version also had to be checked every time I used `this->mortar_coupling_operator`. It doesn't look too clean, so if there's a better way to do that I'd appreciate your ideas :)

What about the mortar coupling manager does not compile in deal.II 9.6? Is it the constructor, or it is related to a member of the class? That will help me regarding the deal.II version issue. Overall though, it won't be a big deal since 9.7 is coming soon and then we will stop support for 9.6

applications_tests/lethe-fluid/mortar_taylorcouette.output
```diff
@@ -0,0 +1,34 @@
+Running on 1 MPI rank(s)...
```
blaisb, line 1: If this work in parallel, could you also make sure this test is run in parallel ?

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -757,6 +776,19 @@ FluidDynamicsMatrixBased<dim>::assemble_system_rhs()
...
+#if DEAL_II_VERSION_GTE(9, 7, 0)
+  if (this->simulation_parameters.mortar.enable)
+    {
+      // Change sign of RHS to be compatible with mortar coupling terms
+      this->system_rhs *= -1.0;
+      this->mortar_coupling_operator->vmult_add(this->system_rhs,
+                                                this->evaluation_point);
+      // Return RHS to original sign
+      this->system_rhs *= -1.0;
+    }
```
peterrum: I think the better way would be to collect all residual contribution and only after that multiply the vector with `-1`. This is what we do for the matrix-free code.

applications_tests/lethe-fluid/mortar_taylorcouette.output
```diff
@@ -0,0 +1,34 @@
...
+   Number of active cells:       2048
+   Number of degrees of freedom: 19072
+   Volume of triangulation:      2.94524
+
+*****************************
+Steady iteration:        4/4
+*****************************
+   Number of active cells:       8192
+   Number of degrees of freedom: 75008
+   Volume of triangulation:      2.94524
```
blaisb: I would remove this iteration, on the debug CI a 10k cell Q2-Q2 is going to require a lot of time, anyway you already have the converged order before so I don't think it is necessary

applications_tests/lethe-fluid/mortar_taylorcouette.prm
```diff
@@ -0,0 +1,147 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 The Lethe Authors
```
blaisb:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
```

source/solvers/fluid_dynamics_matrix_based.cc
brunaccampos: I tried this way because `system_rhs` is already assembled with the opposite sign of the Jacobian. Also, this is because we are using the same `local_integrate` to add entries in both `system_matrix` and `system_rhs`; one alternative would be to have a new `local_integrate_residual`? But then `vmult_add` would also have to be modified

blaisb:
> > > I had to add a dealii version check when declaring `std::shared_ptr<MortarManagerCircle<dim>> mortar_manager` `std::shared_ptr<CouplingOperator<dim, double>> mortar_coupling_operator` so that the code compiles with dealii 9.6. Then, the version also had to be checked every time I used `this->mortar_coupling_operator`. It doesn't look too clean, so if there's a better way to do that I'd appreciate your ideas :)
> > 
> > 
> > What about the mortar coupling manager does not compile in deal.II 9.6? Is it the constructor, or it is related to a member of the class? That will help me regarding the deal.II version issue. Overall though, it won't be a big deal since 9.7 is coming soon and then we will stop support for 9.6
> 
> It doesn't compile the declaration of mortar_manager and mortar_coupling_operator in the NavierStokesBase class (`navier_stokes_base.h`); then, because I need to check the dealii version there, I also need to do so every time I use this shared pointers (i.e. in `fluid_dynamics_matrix_based.cc`)

I understand, but why do you "not compile them". I mean, could you isolate what in the class requires deal.II 9.7 and put that between ifdef instead? It's not a great idea to put a declaration of a member of a class between include guard if you can instead make the class compatible with 9.6 but using include guards within the class itself.
What i would try is to make the class compatible with 9.6 from a compilation perspective, then this way you would only need the #ifdef within the class itself.

PR #1568, opened 2025-07-02
SDIRK stage data

etiennemarin:
### Description

This PR introduces core functionalities required to support SDIRK (Singly Diagonally Implicit Runge-Kutta) time integration schemes within Lethe. It adds:

- `sdirk_table`: generates Butcher table `(A, b, c)` coefficients for specific SDIRK methods like `"SDIRK2"` and `"SDIRK3"`.
- `sdirk_stage_data`: retrieves the stage-specific coefficients `a_ij`, `c_i`, and `b_i` for a given stage `i` index.
- `SDIRKStageData`: a structure encapsulating the coefficients for a single stage. 

### Testing

- Added a new test: `tests/core/sdirk_stage.cc`.
- It validates stage coefficient extraction for both SDIRK2 and SDIRK3 methods.
- Ensures that `a_ij`, `c_i`, and `b_i` match known analytical values.

The test is run using the standard test framework and passes successfully.

This function will be used in a loop over all the stages of the SDIRK method. 
The choice of the coefficients in the Butcher's table is arbitrary. Other tables could be implemented, especially with more stages.

Code related list:

Pull request related list:
- [ ] Labels are applied


include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * @brief Extract the coefficients of a given stage from a SDIRK Butcher tableau.
+ *
+ * This function returns the data needed to compute the solution increment
+ * at a particular stage of a SDIRK (Singly Diagonally Implicit Runge-Kutta)
+ * method. The returned structure contains:
+ * - a_ij: the vector of coefficients from stage j to stage i (for j = 0 to i),
+ * - c_i: the time coefficient associated with stage i,
+ * - b_i: the weight for combining the result of stage i in the final solution.
```
blaisb: I would define the struct where the struct is defined instead of seperately.

include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,46 @@
+// get_sdirk_stage_data.h
+#ifndef SDIRK_STAGE_DATA_H
+#define SDIRK_STAGE_DATA_H
```
blaisb: I think both sdirk_stage_data and sdirk_tabel could be defined in the sdirk.h file. I don't think it's worth it to have two files for that :).

include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <deal.II/lac/full_matrix.h>
+
+#include <stdexcept>
```
blaisb: do you need this include?

include/core/sdirk_table.h
```diff
@@ -0,0 +1,93 @@
...
+#include <vector>
+
+using namespace dealii;
+
+struct SDIRKTable
+{
+  FullMatrix<double>  A;
+  std::vector<double> b;
+  std::vector<double> c;
+};
```
blaisb: the sdirk stage data takes as an argument a full_matrix, b and c, why does it not take into argument a SDIRKTable then?

include/core/sdirk_table.h
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param[in] method_name A string identifying the SDIRK method (e.g. "SDIRK2",
+ * "SDIRK3").
+ *
+ * @return SDIRKTable A structure containing the Butcher tableau (A, b, c) for the requested method.
+ *
+ * @throws std::invalid_argument if the method name is not recognized.
+ */
+
+inline SDIRKTable
```
blaisb: No need to inline it I think. It will be easier if you implement it in the .cc

include/core/sdirk_table.h
```diff
@@ -0,0 +1,93 @@
...
+sdirk_table(const std::string &method_name)
+{
+  SDIRKTable table;
+
+  if (method_name == "SDIRK2")
```
blaisb: There is already an ENUM for the time_stepping method (see in the parameters or simulation control for the definition). I would use the same enum and add sdirk2, sdirk 3... etc. To it.
I would also keep a nomenclature where you have sdirkOrderStage or sdirkStageOrder depending on what the literature uses. I think it's sdirkOrderStage so that SDIRK22 or SDIRK33 would mean order 2 with 2 stage and order 3 with 3 stages respectively.

blaisb: Using the enum will prevent you from having to use string for equality

include/core/sdirk_table.h
```diff
@@ -0,0 +1,93 @@
...
+      table.b.resize(2);
+      table.b[0] = 1.0 - alpha;
+      table.b[1] = alpha;
+
+      return table;
+    }
+
+  else if (method_name == "SDIRK3")
+    {
+      // SDIRK3 - 3-stage, 3rd-order L-stable method (Kennedy-Carpenter)
```
blaisb: Add the full reference with the DOI if you can.

include/core/sdirk_table.h
```diff
@@ -0,0 +1,93 @@
...
+      table.c = {gamma, 0.5, 1.0};
+      table.b = {1.0 / 6.0, 2.0 / 3.0, 1.0 / 6.0};
+
+      return table;
+    }
+
+  else
+    {
+      throw std::invalid_argument("Unknown SDIRK method: " + method_name);
+    }
```
blaisb: don't just throw. You can use an AssertThrow and define an appropriate exception. You can ctrl+f to find examples of AssertThrow

source/core/sdirk_stage_data.cc
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace dealii;
+
+SDIRKStageData
+sdirk_stage_data(const FullMatrix<double>  &butcher_table,
+                 const std::vector<double> &c,
+                 const std::vector<double> &b,
+                 const unsigned int         stage_i)
+{
+  const unsigned int n_stages = b.size();
```
blaisb: I would assert that c.size is equal to b.size no?

source/core/sdirk_stage_data.cc
```diff
@@ -0,0 +1,35 @@
...
+
+  if (stage_i >= n_stages)
+    throw std::invalid_argument("Invalid stage index.");
```
blaisb: Use AssertThrow here instead.

source/core/sdirk_stage_data.cc
blaisb: I would also AssertThrow for the size of the butcher table, making sure that everyone is consistent.

source/core/sdirk_stage_data.cc
```diff
@@ -0,0 +1,35 @@
...
+
+  SDIRKStageData data;
+  data.a_ij.resize(stage_i + 1);
```
blaisb: I think since you are resizing it from the outside, it could be a better idea to have SDIRKStageData be a class and the constructor could take asa n argument the size to be resized.
Here  it's not clear why it's stage_i+1. Be explicit and add a comment here.

tests/core/sdirk_stage.cc
```diff
@@ -0,0 +1,67 @@
...
+#include <core/sdirk_stage_data.h>
+#include <core/sdirk_table.h>
+
+// Tests (with common definitions)
+#include <../tests/tests.h>
+
+void
+test()
+{
+  deallog << "Testing SDIRK2 coefficients" << std::endl;
```
blaisb: I would also add a test for SDIRK33 coefficients. Maybe also print with more significant digits than here? It might be worth it to print with like 12 digits of accuracy to make sure that you really are calculaating what you think :)

blaisb (changes requested):
Some initial comments, please don't hesitate if you have questions.

include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: Copyright (c) 2019-2020, 2025 The Lethe Authors
```
blaisb:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
```

include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,81 @@
...
+#include <vector>
+
+using namespace dealii;
+
+struct SDIRKTable
+{
+  FullMatrix<double>  A;
+  std::vector<double> b;
+  std::vector<double> c;
+};
```
blaisb, line 42: I would add a brief here that explains how an SDIRK table works. What is the meaning of the A, b and c.
(also why is only A capital and b and c are lower case?) Try to explain why this is this way and/or point to an adequate reference

include/core/sdirk_stage_data.h
```diff
@@ -0,0 +1,81 @@
...
+   * @brief Extract the coefficients of a given stage from a SDIRK Butcher tableau.
+   *
+   * This function returns the data needed to compute the solution increment
+   * at a particular stage of a SDIRK (Singly Diagonally Implicit Runge-Kutta)
+   * method. The returned structure contains:
+   * - a_ij: the vector of coefficients from stage j to stage i (for j = 0 to
+   * i),
+   * - c_i: the time coefficient associated with stage i,
+   * - b_i: the weight for combining the result of stage i in the final
+   * solution.
```
blaisb, line 86: Could you give an example with a formula maybe? I find that text confusing to read. Not that it is not well written, I just don't fully understand what is the usage of a_iij, c_i and b_i.
Also now a is a lowercase letter while it was a upper case letter in the SDIRKTable, is that normal?

source/core/sdirk_stage_data.cc
```diff
@@ -0,0 +1,108 @@
+// SPDX-FileCopyrightText: Copyright (c) 2019-2020, 2025 The Lethe Authors
```
blaisb:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
```

source/core/sdirk_stage_data.cc
```diff
@@ -0,0 +1,108 @@
...
+
+/// Function to construct the Butcher's table for a given SDIRK method
+SDIRKTable
+sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
+{
+  SDIRKTable table;
+
+  // Important note : the nomenclature used for the name of the SDIRK methods
+  // are sdirkOrderStage sdirk22 means SDIRK with order 2 and 2 stages, sdirk33
+  // means SDIRK with order 3 and 3 stages
```
blaisb, line 28: This is better explained in the .h, but I would also keep that comment here.

blaisb (commented):
Not much work left, this version is much better :)

include/core/sdirk_stage_data.h
etiennemarin: I think a capital letter for a matrix and lower case for vectors are the norms in the papers that I read

include/core/sdirk_stage_data.h
etiennemarin: Now we are using the coefficients a_ij of the matrix A. I have seen this nomenclature many times in papers and in my studies. But if you think it is less confusing to name de coefficients A_ij, I'll make the change. And yes, I will add an example in order to get easier to understand

blaisb (approved):

PR #1580, opened 2025-07-13
Cfd dem void fraction linear solver

blaisb:
### Description

The calculation of the void fraction within the VANS and the CFD-DEM solvers can require the solution of a linear system of equation if the void fraction is calculated from particles.
The tolerance for the calculation of this void fraction was hard coded to 1e-15 and the parameters for the linear solver could not be controlled independently from those of the navier-stokes equations. This PR fixes that

### Testing

Since the default value now does not match the previous one, all of the examples and tests were modified so that the tolerance of 1e-15 appears explicitely instead of implicitely.

### Documentation

This new element is documented in the linear solver section.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
+++ b/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
@@ -209,6 +209,7 @@ subsection linear solver
   subsection void fraction
     set relative residual = 1e-15
     set minimum residual  = 1e-15
+    set verbopsity        = quiet
   end
   subsection fluid dynamics
     set method             = gmres
--- a/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
+++ b/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
@@ -206,6 +206,10 @@ end
 #---------------------------------------------------
 
 subsection linear solver
+  subsection void fraction
+    set relative residual = 1e-15
+    set minimum residual  = 1e-15
+  end
   subsection fluid dynamics
     set method             = gmres
     set max iters          = 200
--- a/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
+++ b/applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
@@ -209,7 +209,7 @@ subsection linear solver
   subsection void fraction
     set relative residual = 1e-15
     set minimum residual  = 1e-15
-    set verbopsity        = quiet
+    set verbosity        = quiet
   end
   subsection fluid dynamics
     set method             = gmres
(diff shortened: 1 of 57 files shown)


applications_tests/lethe-fluid-particles/dynamic_particle_insertion.prm
```diff
@@ -206,6 +206,11 @@ end
 #---------------------------------------------------
 
 subsection linear solver
+  subsection void fraction
+    set relative residual = 1e-15
+    set minimum residual  = 1e-15
+    set verbopsity        = quiet
```
flo-op:
```suggestion
    set verbosity        = quiet
```

include/core/multiphysics.h
```diff
@@ -51,8 +52,12 @@ get_physics_id(std::string physics_name)
     return PhysicsID::tracer;
   else if (physics_name == "VOF")
     return PhysicsID::VOF;
-  else
+  else if (physics_name == "cahn hilliard")
     return PhysicsID::cahn_hilliard;
+  else if (physics_name == "void fraction")
+    return PhysicsID::void_fraction;
+  else
+    AssertThrow(false, ExcMessage("An unknown Physics name was requested"));
```
mivaia:
```suggestion
    AssertThrow(false, dealii::StandardExceptions::ExcMessage("An unknown Physics name was requested"));
```

A test is crashing because of this. possible to import the namespace also.

source/fem-dem/fluid_dynamics_vans.cc
flo-op, line 1: Update copyright date to 2025

include/core/multiphysics.h
mivaia: Looks good on my side also! A single comment to address

mivaia (approved):

include/core/multiphysics.h
mivaia: There is a few tests that are not supposed to be changed. I suppose this is only the formatting script that removed some spaces?

applications_tests/lethe-fluid-vans/case0_vans_B.output
```diff
@@ -12,7 +12,7 @@ Void Fraction
 -------------------------------
 Volume-Averaged Fluid Dynamics
 -------------------------------
-Global continuity equation error: -2.41235e-18 s^-1
+Global continuity equation error: 7.58942e-19 s^-1
```
voferreira, line 15: Why did this change sign?

voferreira (commented):
I think it is all clear. I just don't understand why the global continuity equation has changed sign in some tests. The absolute values are similar, though.

applications_tests/lethe-fluid-vans/case0_vans_B.output
blaisb: The value is 1e-18, so this is below machine precision, meaning it is not portable accross machines. Changing compiler version of cpu architecture would do that. The reason why this passes is that numdiff sees <1e-12 as no changes

blaisb:
The warning CI breaking is an issue that is unrelated to this PR that I will fix this week in another PR :)

voferreira (approved):

PR #1590, opened 2025-07-21
Fix workflow status badges

lpsaavedra:
### Description

The workflow status badge in the `README.md` file was not pointing to the correct workflow in two cases: the clang-tidy and the example parameter files. In addition, two workflows were recently disabled, so in this PR the corresponding workflow status badge is disabled as well. The lines were simply commented and in the future they can be easily enabled once again.


blaisb:
Awesome catch. Thank you @lpsaavedra

PR #1597, opened 2025-07-28
Mortar ALE Assembler

brunaccampos:
### Description

This PR adds a mortar Arbitrary Lagrangian Eulerian (ALE) asssembler to allow the simulation of rotor-stator transient problems.

### Testing

An application test of a transient Taylor-Couette with mortar has been added.

### Documentation

The documentation has been updated to include both `rotor rotation angle` and `rotor angular velocity` in the mortar section parameters.

The `MortarManagerBase` and `MortarManagerCircle` functions have been updated regarding the order of the parameters in the constructor, as well as allowing the parameters `radius` and `n_subdivisions` to be public. This was needed so that the values obtained by calling `compute_n_subdivisions_and_radius` can be stored in the class parameters, and so that such parameters can be accessed by the `reinit_mortar` function incluided in scratch data.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/solvers/navier_stokes_base.h
+++ b/include/solvers/navier_stokes_base.h
@@ -478,7 +478,7 @@ protected:
    * twice
    */
   void
-  rotate_rotor_mapping(bool is_first);
+  rotate_rotor_mapping(const bool is_first);
 
   /**
    * @brief Update non-zero constraints if the boundary is time dependent.
--- a/source/solvers/navier_stokes_base.cc
+++ b/source/solvers/navier_stokes_base.cc
@@ -2031,7 +2031,8 @@ NavierStokesBase<dim, VectorType, DofsType>::reinit_mortar()
 
 template <int dim, typename VectorType, typename DofsType>
 void
-NavierStokesBase<dim, VectorType, DofsType>::rotate_rotor_mapping(bool is_first)
+NavierStokesBase<dim, VectorType, DofsType>::rotate_rotor_mapping(
+  const bool is_first)
 {
   if (!this->simulation_parameters.mortar_parameters.enable)
     return;


include/solvers/navier_stokes_base.h
```diff
@@ -450,17 +453,23 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
   }
 
   /**
    * @brief Rotate rotor mapping in mortar method
+   *
+   * @param[in] is_first Whether this is the first mapping rotation. This
+   * boolean is used only for printing output purposes; since the mapping needs
+   * to be rotated before setting up dofs, but also after the constraints are
+   * first defined, this function is called twice before the iterate() loop. The
+   * parameter is_first just prevents the mortar verbosity to be printed twice
```
blaisb:
```suggestion
   * parameter is_first just prevents the mortar verbosity from being printed twice
```

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -2658,3 +2658,102 @@ NavierStokesAssemblerALE<dim>::assemble_rhs(
...
+void
+NavierStokesAssemblerMortarALE<dim>::assemble_matrix(
+  const NavierStokesScratchData<dim>   &scratch_data,
+  StabilizedMethodsTensorCopyData<dim> &copy_data)
+{
+  /// Loop and quadrature informations
+  const auto                   &JxW_vec    = scratch_data.JxW;
+  const unsigned int            n_q_points = scratch_data.n_q_points;
+  const std::vector<Point<dim>> quadrature_points =
+    scratch_data.quadrature_points;
```
blaisb: you don't use these right?

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -2658,3 +2658,102 @@ NavierStokesAssemblerALE<dim>::assemble_rhs(
...
+            -scratch_data.grad_phi_u[q][j] * velocity_ale[q];
+        }
+
+      for (unsigned int i = 0; i < n_dofs; ++i)
+        {
+          const auto phi_u_i = scratch_data.phi_u[q][i];
+
+          for (unsigned int j = 0; j < n_dofs; ++j)
+            {
+              const Tensor<2, dim> grad_phi_u_j = scratch_data.grad_phi_u[q][j];
```
blaisb:
```suggestion
              const Tensor<2, dim>& grad_phi_u_j = scratch_data.grad_phi_u[q][j];
```

Otherwise you are copying the whole tensor, that will be slow. You need to copy the reference to it.

source/solvers/navier_stokes_assemblers.cc
```diff
@@ -2658,3 +2658,102 @@ NavierStokesAssemblerALE<dim>::assemble_rhs(
...
+void
+NavierStokesAssemblerMortarALE<dim>::assemble_rhs(
+  const NavierStokesScratchData<dim>   &scratch_data,
+  StabilizedMethodsTensorCopyData<dim> &copy_data)
+{
+  // Loop and quadrature informations
+  const auto                   &JxW_vec    = scratch_data.JxW;
+  const unsigned int            n_q_points = scratch_data.n_q_points;
+  const std::vector<Point<dim>> quadrature_points =
+    scratch_data.quadrature_points;
```
blaisb: Same here I don't think you use this.

source/solvers/navier_stokes_base.cc
```diff
@@ -1981,24 +1981,70 @@ NavierStokesBase<dim, VectorType, DofsType>::define_zero_constraints()
...
+          VectorType tmp_previous_solution = init_temporary_vector();
+          if constexpr (std::is_same_v<
+                          VectorType,
+                          LinearAlgebra::distributed::Vector<double>>)
+            tmp_previous_solution.update_ghost_values();
+
+          nonzero_constraints.distribute(tmp_previous_solution);
+          previous_solutions[i] = tmp_previous_solution;
+        }
+    }
```
blaisb: Why don't you move this to a function (like transfer solution). This way you could reuse that function at the end of setup dofs for example and remove some code duplication.

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -348,6 +348,50 @@ NavierStokesScratchData<dim>::enable_particle_fluid_interactions(
...
+{
+  gather_mortar = true;
+}
+
+template <int dim>
+void
+NavierStokesScratchData<dim>::reinit_mortar(
+  const typename DoFHandler<dim>::active_cell_iterator &cell,
+  const Parameters::Mortar<dim>                        &mortar_parameters,
+  const double                                         &radius)
```
blaisb, line 363: I think this could be implemented in the .cc. We should move the reinit to the .cc when possible :)

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -348,6 +348,50 @@ NavierStokesScratchData<dim>::enable_particle_fluid_interactions(
...
+  // Compute radius between center of rotation and current cell center
+  double radius_current =
+    cell_center.distance(mortar_parameters.center_of_rotation);
+
+  // Use prescribed rotor angular velocity only if cell is part of the rotor
+  double cell_rotor_angular_velocity;
+  if (radius_current > radius)
+    cell_rotor_angular_velocity = 0.0;
+  else
+    cell_rotor_angular_velocity = rotor_angular_velocity;
```
blaisb, line 382: I know it's a bit picky, but here we are "hard" assuming that the mortar is always at lower than the radius. Have we asserted this before? I don't think so. I would try to keep this in mind because one could imagine cases where the outside is rotating and not the inside :)

blaisb (changes requested):
Some few comments, otherwise nice

source/solvers/navier_stokes_base.cc
brunaccampos: I thought about it; there's a difference that, for mesh refinement, this transfer is done through `SolutionTransfer`, which I cannot use in this mortar update case because the mesh is not being refined. I could create a function like you said and add both cases? For the refinement or no refinement options

source/solvers/navier_stokes_scratch_data.cc
brunaccampos: You mean this `reinit_mortar()` implementation? It is being done in the .cc already I think

source/solvers/navier_stokes_scratch_data.cc
brunaccampos: You're right, this have not been asserted before. Even when we compute the radius of the inner domain we kind of assume that it is the rotor, but it doesn't have to be. I could add a parameter to identify if the inner/outer domain are rotor or stators and account for that at this place

source/solvers/navier_stokes_base.cc
blaisb: Yeah that could be an idea. I would at least put this block of code in a function. Like transfer_without_refinement and transfer_after_refinement could be the two functions for example?

source/solvers/navier_stokes_scratch_data.cc
blaisb: oupsie, my bad ):

source/solvers/navier_stokes_scratch_data.cc
blaisb: I don't think it's necessary in this PR,  but let's keep this in mind for now that we are "hard assuming" that the inside is the mortar.

include/core/mortar_coupling_manager.h
```diff
@@ -120,6 +120,11 @@ class MortarManagerBase
   std::vector<Tensor<1, dim, double>>
   get_normals(const Point<dim> &face_center) const;
 
+  /// Number of cells at the interface between inner and outer domains
+  unsigned int n_subdivisions;
+  /// Radius at the interface between inner and outer domains
+  double radius;
```
peterrum, line 126: I guess this is temporal?

source/solvers/navier_stokes_base.cc
```diff
@@ -1266,6 +1247,37 @@ NavierStokesBase<dim, VectorType, DofsType>::refine_mesh_uniform()
     this->update_multiphysics_time_average_solution();
 }
 
+template <int dim, typename VectorType, typename DofsType>
+void
+NavierStokesBase<dim, VectorType, DofsType>::transfer_without_refinement()
```
peterrum: Why do we need this? If we rotate the mesh, the vectors do not changes. So there is no need for transfer.

blaisb: There's no need to transfer that's true. Only the boundary conditions have to be updated.

brunaccampos: indeed! I was testing it at some point and forgot to remove it

include/core/mortar_coupling_manager.h
brunaccampos: if we're removing them late, you mean? these were already there, I just made them public

source/solvers/navier_stokes_base.cc
```diff
@@ -1987,24 +1968,41 @@ NavierStokesBase<dim, VectorType, DofsType>::define_zero_constraints()
...
+                        this->simulation_parameters.mesh_adaptation.frequency ==
+                      0;
+  else
+    refinement_step = this->simulation_control->get_step_number() == 0;
 
-  if (!this->simulation_parameters.mortar.enable)
+  if (this->simulation_control->is_at_start() || !refinement_step ||
+      this->simulation_parameters.mesh_adaptation.type ==
+        Parameters::MeshAdaptation::Type::none)
+    setup_dofs();
```
blaisb: I would add a comment here to explain what you are doing. There's a bunch of if, they make sense, but still it's not trivial to understand.

blaisb (approved):
Just one last comment and I think this is good to merge

source/solvers/navier_stokes_base.cc
```diff
@@ -2021,44 +2024,57 @@ NavierStokesBase<dim, VectorType, DofsType>::update_mortar_coupling()
...
-      this->simulation_parameters.mortar.sip_factor);
+      this->simulation_parameters.mortar_parameters.rotor_boundary_id,
+      this->simulation_parameters.mortar_parameters.stator_boundary_id,
+      this->simulation_parameters.mortar_parameters.sip_factor);
 }
 
 template <int dim, typename VectorType, typename DofsType>
 void
-NavierStokesBase<dim, VectorType, DofsType>::rotate_mortar_mapping()
+NavierStokesBase<dim, VectorType, DofsType>::rotate_rotor_mapping(bool is_first)
```
peterrum:
```suggestion
NavierStokesBase<dim, VectorType, DofsType>::rotate_rotor_mapping(const bool is_first)
```

PR #1598, opened 2025-07-29
SDIRK scheme implementation

etiennemarin:
### Description

This pull request implements the SDIRK (Singly Diagonally Implicit Runge-Kutta) time integration method for the unsteady Navier–Stokes equations.

The current implementation is quite brute-force and would surely benefit from a proper refactoring to improve efficiency and readability. The idea is to also use the SDIRK scheme for other applications beyond just `lethe-fluid`.

The code is structured to match as closely as possible the existing BDF structure. Here's how it works:

- At each time iteration, we loop over the stages. For "traditional" schemes (mainly BDF), the loop is straightforward: there is only one stage.
- At each stage, for the SDIRK scheme, we compute the vectors **kᵢ**.
- However, the structure of Lethe is designed to solve the nonlinear system on the velocity. So instead of directly solving `kᵢ = F(tᵢ, uᵢ)`, we solve the equivalent system:

 `(uᵢ* - uₙ) / (h aᵢᵢ) - ∑_{j=0}^{i-1} (aᵢⱼ / aᵢᵢ) kⱼ = F(tᵢ, uᵢ)`

- Once we have computed the $\mathbf{u}_i^*$ vector, we store the associated $\mathbf{k}_i$ vector in order to compute the next stages and the final sum.
- At the end of the loop, the solution is computed as:

`uₙ₊₁ = uₙ + h ∑ bᵢ kᵢ`

The main changes are implemented in the `solvers/navier_stokes_base.cc` file, within the `iterate()` function.  
The Butcher table is stored in `scratch-data` because it's required during matrix assembly.

---

### Testing

The code has been tested on the `lethe-fluid/application-test` Taylor–Green vortex (GLS formulation) and yields results close to the BDF2 scheme.  
The test outputs, at each time step, the enstrophy, the kinetic energy, and the L2 error.

The goal is now to test this new scheme on multiple examples and with different Butcher tables.

---

### Documentation

- The Butcher table is set in the `scratch-data`.
- The variable `sum_over_previous_stages` is now required for matrix assembly.

Code related list:

Pull request related list:
- [x] Labels are applied


applications_tests/lethe-fluid/taylor-green-vortex_gls_bdf2.prm
blaisb, line 1: So you added a new bdf2 test case? Is it to have some sort of stored comparison between bdf2,sdirk22 and sdirk33?
In that case, I would suggest you do a case with a larger time-step so that the accuracy in the case be limited by the time-step and not the mesh resolution. Do. like a time-step of 0.2 maybe

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.mpirun=2.output
```diff
@@ -0,0 +1,807 @@
...
+0.9000   4.246433e-03 
+0.9100   4.208380e-03 
+0.9200   4.170166e-03 
+0.9300   4.131813e-03 
+0.9400   4.093341e-03 
+0.9500   4.054771e-03 
+0.9600   4.016120e-03 
+0.9700   3.977408e-03 
+0.9800   3.938652e-03 
+0.9900   3.899868e-03
```
blaisb, line 806: You see here your error is only limited by your mesh. This test case will be relevant if. you really increase the time-step.
That's why both your sdirk 22 and sdirk 33 have the same last error

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019, 2021, 2023 The Lethe Authors
```
blaisb: Change year to 2025

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk22.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019, 2021, 2023 The Lethe Authors
```
blaisb: change year to 2025

applications_tests/lethe-fluid/taylor-green-vortex_gls_bdf2.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019, 2021, 2023 The Lethe Authors
```
blaisb: change year in the copyright

include/core/simulation_control.h
```diff
@@ -133,7 +134,7 @@ class SimulationControl
     bdf_coefs = calculate_bdf_coefficients(assembly_method, time_step_vector);
   }
 
-
+  unsigned int stage_i;
```
blaisb: Please add a comment that describe what is that variable.
I would give it a better and more verbose name.
Maybe current_sdirk_stage?
Be as explicit as possible.

include/core/simulation_control.h
```diff
@@ -230,6 +231,18 @@ class SimulationControl
            method == Parameters::SimulationControl::TimeSteppingMethod::bdf3;
   }
 
+  /**
+   * @brief Establish if the method is a sdirk method
+   *
+   */
```
blaisb: can you add the return and the description of the return

include/core/simulation_control.h
```diff
@@ -480,6 +493,18 @@ class SimulationControl
     return bdf_coefs;
   }
 
+  unsigned int
+  get_stage_i_number()
+  {
+    return stage_i;
+  }
```
blaisb: needs a brief and a description of the return. PLease be also a bit more verbose. What does i mean here?

include/core/simulation_control.h
```diff
@@ -480,6 +493,18 @@ class SimulationControl
...
+
+  void
+  set_stage_i_number(unsigned int s)
+  {
+    stage_i = s;
+  }
+
```
blaisb: Same as above comment.

include/solvers/navier_stokes_base.h
```diff
@@ -922,10 +923,25 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+  VectorType              previous_solutions_0;
+
+  std::vector<VectorType> previous_k_j_solutions;
+  VectorType              previous_k_j_solutions_p;
+
+  VectorType sum_bi_ki;
+  VectorType temp_sum_bi_ki;
+
+  VectorType sum_over_previous_stages;
+  VectorType temp_sum_over_previous_stages;
```
blaisb: Ok that's a lot of vectors... You really don't need that many! Please try to clean this to have as little as possible.

include/solvers/navier_stokes_base.h
```diff
@@ -990,6 +1006,8 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   /// Dynamic homogeneous constraints used for temperature-dependent solid
   /// domain constraints
   AffineConstraints<double> dynamic_zero_constraints;
+
+  std::unique_ptr<NavierStokesScratchData<dim>> scratch_data;
```
blaisb: ??? You don't need this

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1086,6 +1101,7 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
   std::vector<Tensor<1, dim>>              pressure_gradients;
   std::vector<std::vector<double>>         previous_pressure_values;
   std::vector<std::vector<Tensor<1, dim>>> previous_velocity_values;
+  std::vector<Tensor<1, dim>>              u_sum_over_stages;
```
blaisb: I would maybe call it something else than u_sum because it's not the velocity. that's a bit confusing.

source/core/utilities.cc
```diff
@@ -547,8 +547,7 @@ std::string
 get_last_value_of_parameter(const std::string &file_name,
                             const std::string &parameter_name)
 {
-  std::string return_value;
-
+  std::string   return_value;
```
blaisb, line 550: Why is this changed?

source/solvers/fluid_dynamics_block.cc
```diff
@@ -605,23 +607,58 @@ FluidDynamicsBlock<dim>::setup_dofs_fd()
...
+  this->temp_sum_bi_ki.reinit(this->locally_owned_dofs, this->mpi_communicator);
+
+
+  this->sum_over_previous_stages.reinit(this->locally_owned_dofs,
+                                        this->locally_relevant_dofs,
+                                        this->mpi_communicator);
+  this->temp_sum_over_previous_stages.reinit(this->locally_owned_dofs,
+                                             this->mpi_communicator);
+
+
```
blaisb: Can you please add comments here?
I would group all of the "SDIRK-related " vectors inside of a struct or a similar thing

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -87,26 +87,67 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
   // If enabled, create mortar coupling
   this->update_mortar_coupling();
 
+  // Important note : calculus on the following vectors (addition,
+  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
+  // locally_relevant_dofs. This is why most of them are reinitialized both with
+  // and without locally_relevant_dofs.
```
blaisb:
```suggestion
  // Important note : calculation on the following vectors (addition,
  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
  // locally_relevant_dofs. This is why most of them are reinitialized both with
  // and without locally_relevant_dofs.
```
Yes, but you don't need to have a copy of each,  you can reuse a locally_owned vectors for many things.

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
-      // is necessary because of the checkpointing mechanism. You spent an
-      // evening debugging this, trust me.
-      multiphysics->percolate_time_vectors(true);
-    }
-  else
-    {
-      // Fluid dynamics is not to be solved, but rather specified. Update
-      // condition and move on.
+  auto &previous_solutions   = this->previous_solutions;
+  auto &previous_solutions_0 = this->previous_solutions_0;
```
blaisb: ? why do. you need these reference copies

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+
+  // Intermediate variable to make the calculus easier to understand
+  // This will be stored in the previous_k_j_solutions vector
+  auto &present_k_i_solution      = this->present_k_i_solution;
+  auto &temp_present_k_i_solution = this->temp_present_k_i_solution;
+
+  // Very important variable because \sum_{j=1}^{i-1} a_{ij} k_j has to be
+  // calculated at each stage
+  auto &previous_k_j_solutions   = this->previous_k_j_solutions;
+  auto &previous_k_j_solutions_p = this->previous_k_j_solutions_p;
```
blaisb: Here you are just making copies of the reference of the member vectors. Why do this? You can use them directly.

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+  // Temporary variable to compute the sum over the previous stages
+  auto &tmp = this->tmp;
 
-      if (this->simulation_parameters.initial_condition->type ==
-          Parameters::InitialConditionType::average_velocity_profile)
+  const unsigned int n_stages = SimulationControl::get_number_of_stages(method);
+
+  // At each time iteration, we reset the value of sum_bi_ki
+  temp_sum_bi_ki = sum_bi_ki;
+  temp_sum_bi_ki = 0;
```
blaisb: why do you equalize it then reset it to zero? I would just reset it to zero

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+
+          if (method !=
+                Parameters::SimulationControl::TimeSteppingMethod::steady &&
+              method !=
+                Parameters::SimulationControl::TimeSteppingMethod::steady_bdf &&
+              method !=
+                Parameters::SimulationControl::TimeSteppingMethod::bdf1 &&
+              method !=
+                Parameters::SimulationControl::TimeSteppingMethod::bdf2 &&
+              method != Parameters::SimulationControl::TimeSteppingMethod::bdf3)
```
blaisb: You had a simulation_control method that said "is_sdirk", use it here.

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+
+              // At each stage, the value of the (a_ij), (b_i) and (c_i)
+              // coefficients are differents
+              SDIRKStageData stage_data(scratch_data->sdirk_table, stage + 1);
+              const double   a_ii = stage_data.a_ij[stage];
+
+              // At each stage, we need to recompute \sum_{j=1}^{i-1} a_{ij} k_j
+              // = sum_over_previous_stages
+              temp_sum_over_previous_stages = sum_over_previous_stages;
+              temp_sum_over_previous_stages = 0;
```
blaisb: ? why do you = and then = 0 ?

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+
+              if (stage > 0)
```
blaisb: You don't need the if, because the loop below loops over stage, so for stage=0 it will just not do anything.

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+                {
+                  // At the first stage, the sum_over_previous_stages is set to
+                  // 0. But for the next stages, we need to update this
+                  // sum_over_previous_stages
+                  for (unsigned int p = 0; p < stage; ++p)
+                    {
+                      previous_k_j_solutions_p = previous_k_j_solutions[p];
+                      tmp                      = previous_k_j_solutions[0];
```
blaisb:
```suggestion
```
You don't need this line since you are equalizing below.

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+                      tmp.equ(stage_data.a_ij[p] / a_ii,
+                              previous_k_j_solutions_p);
+                      temp_sum_over_previous_stages.add(1.0, tmp);
```
blaisb: Why don't you just do instead
```suggestion
                      temp_sum_over_previous_stages.add(stage_data.a_ij[p] / a_ii, previous_k_j_solutions[p]);
```

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+              temp_sum_over_previous_stages = 0;
+
+              if (stage > 0)
+                {
+                  // At the first stage, the sum_over_previous_stages is set to
+                  // 0. But for the next stages, we need to update this
+                  // sum_over_previous_stages
+                  for (unsigned int p = 0; p < stage; ++p)
+                    {
+                      previous_k_j_solutions_p = previous_k_j_solutions[p];
```
blaisb:
```suggestion
```

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+                      tmp                      = previous_k_j_solutions[0];
+                      tmp.equ(stage_data.a_ij[p] / a_ii,
+                              previous_k_j_solutions_p);
```
blaisb:
```suggestion
                              previous_k_j_solutions[p]);
```

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
+
+          if (method ==
+                Parameters::SimulationControl::TimeSteppingMethod::steady ||
+              method ==
+                Parameters::SimulationControl::TimeSteppingMethod::steady_bdf ||
+              method ==
+                Parameters::SimulationControl::TimeSteppingMethod::bdf1 ||
+              method ==
+                Parameters::SimulationControl::TimeSteppingMethod::bdf2 ||
+              method == Parameters::SimulationControl::TimeSteppingMethod::bdf3)
```
blaisb: This is only when it's not sdirk, use the function you wrote..

source/solvers/navier_stokes_base.cc
```diff
@@ -553,74 +565,227 @@ template <int dim, typename VectorType, typename DofsType>
...
-      // Solve and percolate the auxiliary physics that should be treated
-      // AFTER the fluid dynamics
-      multiphysics->solve(true,
-                          simulation_parameters.simulation_control.method);
-      multiphysics->percolate_time_vectors(true);
+  if (method != Parameters::SimulationControl::TimeSteppingMethod::steady &&
+      method != Parameters::SimulationControl::TimeSteppingMethod::steady_bdf &&
+      method != Parameters::SimulationControl::TimeSteppingMethod::bdf1 &&
+      method != Parameters::SimulationControl::TimeSteppingMethod::bdf2 &&
+      method != Parameters::SimulationControl::TimeSteppingMethod::bdf3)
```
blaisb: Same as above use the is_sdirk

blaisb (changes requested):

applications_tests/lethe-fluid/taylor-green-vortex_gls_bdf2.prm
etiennemarin: Yes that was my idea since the BDF2 is broadly used in Lethe. I found it more relevant than BDF1. I will do the simulation again with the larger time step

include/core/simulation_control.h
etiennemarin: I will also add a return to the is_bdf() function. It was probably an omission when it was implemented in the past

include/core/simulation_control.h
etiennemarin: Actually, since we are calculating the sum of a_ij * k_j in the loop in the iterate() function, I think we don't need these two functions. In the assembler, we only need to have the diagonal term, so we don't need to know at which stage we are. In the loop, we already know the stage. These two functions can maybe be usefull for more general implicit schemes, so I don't know if you want me to keep them. But for SDIRK, it is not necessary

include/core/simulation_control.h
etiennemarin: See the comment just above about the need (or not) to keep these two functions in the code

include/solvers/navier_stokes_base.h
etiennemarin: I need the Butcher's table in the iterate() function to build the sums. The table is in the scratch data. But I also need this table for the matrix assembly. Should I store the table in an other location ?

source/core/utilities.cc
etiennemarin: Mistake from me I think

include/solvers/navier_stokes_scratch_data.h
etiennemarin: I called it like this because the object FEValues in the scratch_data is associated to the velocity field :
this->fe_values[velocities].get_function_values(sum_over_previous_stages,
                                                    this->u_sum_over_stages);
Is it better to just call it sum_a_ij_k_j and explain what are a_ij coefficients and k_j ?

blaisb: yeah or sdirk_stage_sum, something that thint that this is SDIRK related

include/solvers/navier_stokes_base.h
blaisb: You can recalculate the Butcher table when. you need it there. Just remake it and not store it as part of the class. It will be simpler. There,s no point in having a scratch data live just for a table. Just reget the table. Otherwise if you want to store it, store it in simulation control where the time related stuff is stored.

include/core/simulation_control.h
blaisb: If you don't use them, then you can remove them for now i presume.

source/solvers/navier_stokes_base.cc
etiennemarin: Yes, I don't need to have all these lines, your solution works. But we still need to have a non-locally relevant vector in the .add operation, and previous_k_j_solutions[p] is locally-relevant

source/solvers/navier_stokes_base.cc
etiennemarin: In the operations (addition, multiplication, etc.) we need a non-locally relevant dofs vector, I think

source/solvers/navier_stokes_base.cc
blaisb: I understand, but why create a new locally_owned vector? You can use local evaluation point to store your temporary solution, this will remove one more vector.
Try to minimize the amount of local vectors you generate.

doc/source/parameters/cfd/simulation_control.rst
```diff
@@ -107,6 +107,9 @@ This subsection is the most important in a simulation and therefore, the most co
 	* ``bdf1``: 1st order backward differentiation
 	* ``bdf2``: 2nd order backward differentiation
 	* ``bdf3``: 3rd order backward differentiation
+	* ``sdirk22``: 2nd order 2 stages singly diagonally implicit Runge-Kutta
```
brunaccampos, line 110: it looks like the user does not need to enter additional parameters for the sdirk method, right? you could add a `note` about this after the BDF scheme parameters sections, just to make it clear

include/solvers/navier_stokes_assemblers.h
```diff
@@ -308,6 +308,54 @@ class GLSNavierStokesAssemblerBDF : public NavierStokesAssemblerBase<dim>
...
+ * For a given stage \( i \), the following expression is assembled and treated
+ * implicitly: \f[ \frac{\mathbf{u}_i^* - \mathbf{u}_n}{h a_{ii}} -
+ * \sum_{j=0}^{i-1} \frac{a_{ij}}{a_{ii}} \mathbf{k}_j \f] where:
+ * - \( \mathbf{u}_i^* \) is the current stage solution,
+ * - \( \mathbf{u}_n \) is the solution at the previous time step,
+ * - \( \mathbf{k}_j \)coefficients are the solution increments at the previous
+ * stages,
+ * - \( h \) is the time step size,
+ * - \( a_{ij} \), \( a_{ii} \) are coefficients from the SDIRK Butcher tableau.
+ *
```
brunaccampos, line 325: Reading these expressions in latex form is a bit confusing; maybe you could remove some parts like \frac \mathbf etc. to make the description cleaner

include/solvers/navier_stokes_assemblers.h
```diff
@@ -308,6 +308,54 @@ class GLSNavierStokesAssemblerBDF : public NavierStokesAssemblerBase<dim>
...
+ * integration for the Navier Stokes equations.
+ * For a given stage \( i \), the following expression is assembled and treated
+ * implicitly: \f[ \frac{\mathbf{u}_i^* - \mathbf{u}_n}{h a_{ii}} -
+ * \sum_{j=0}^{i-1} \frac{a_{ij}}{a_{ii}} \mathbf{k}_j \f] where:
+ * - \( \mathbf{u}_i^* \) is the current stage solution,
+ * - \( \mathbf{u}_n \) is the solution at the previous time step,
+ * - \( \mathbf{k}_j \)coefficients are the solution increments at the previous
+ * stages,
+ * - \( h \) is the time step size,
+ * - \( a_{ij} \), \( a_{ii} \) are coefficients from the SDIRK Butcher tableau.
```
brunaccampos: Are these a_ij, a_ii coefficients fixed? how are they chosen?

include/solvers/navier_stokes_base.h
```diff
@@ -357,6 +358,30 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+  virtual void
+  multi_stage_precond(unsigned int                                      stage,
+                      Parameters::SimulationControl::TimeSteppingMethod method);
+
+  /**
+   * @brief multi_stage_postcond
+   * Update the \sum_{i=1}^{stage} b_i \times k_i
+   */
+  virtual void
+  multi_stage_postcond(unsigned int                                      stage,
```
brunaccampos: what `precond` and `postcond` are referring to?

source/core/sdirk_stage_data.cc
```diff
@@ -66,12 +66,50 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
       table.A(0, 0) = gamma;
```
brunaccampos, line 69: gamma and alpha are fixed coefficients for the butcher's table? maybe add this description here

source/core/sdirk_stage_data.cc
```diff
@@ -66,12 +66,50 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
...
+      return table;
+    }
+
+  else if (method == Parameters::SimulationControl::TimeSteppingMethod::sdirk43)
+    {
+      // SDIRK43 - 3-stage, 4th-order method
+      // Reference: L. Ferracina, M.N. Spijker, "Strong stability of
+      // singly-diagonally-implicit Runge–Kutta methods" Applied Numerical
+      // Mathematics 58 (2008) 1675–1686 DOI: doi:10.1016/j.apnum.2007.10.004
+      // const double gamma = 0.128886400515;
```
brunaccampos:
```suggestion
```

source/core/simulation_control.cc
```diff
@@ -158,6 +158,7 @@ SimulationControl::get_number_of_stages(
         return 2;
 
       case Method::sdirk33:
+      case Method::sdirk43:
```
brunaccampos, line 161: I would reformulate this:      
// Important note : the nomenclature used for the name of the SDIRK
// methods are sdirkOrderStage sdirk22 means SDIRK with order 2 and 2
// stages, sdirk33 means SDIRK with order 3 and 3 stages
Like this:
When referencing to SDIRK methods the pattern used is sdirkOrderStage. For instance, sdirk22 indicates SDIRK method with order 2 and 2 stages

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
```diff
@@ -0,0 +1,151 @@
...
+
+set dimension = 2
+
+#---------------------------------------------------
+# Simulation Control
+#---------------------------------------------------
+
+subsection simulation control
+  set method            = sdirk33
+  set time step         = 0.2 # Time step
```
brunaccampos: why is the time step for this one 0.2 and for the other tests is 0.1?

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
brunaccampos, line 1: Do you have a test for the sdirk43 too?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1225,6 +1243,8 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
   std::vector<std::vector<std::vector<Tensor<2, dim>>>> face_grad_phi_u;
   std::vector<std::vector<std::vector<double>>>         face_phi_p;
   std::vector<std::vector<std::vector<Tensor<1, dim>>>> face_grad_phi_p;
+
```
brunaccampos, line 1275:
```suggestion
// Butcher table with coefficients for the SDIRK time-stepping method
```

source/solvers/fluid_dynamics_block.cc
```diff
@@ -602,10 +604,15 @@ FluidDynamicsBlock<dim>::setup_dofs_fd()
   // Zero constraints
   this->define_zero_constraints();
 
+  // Important note : calculation on the following vectors (addition,
+  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
+  // locally_relevant_dofs. This is why most of them are reinitialized both with
```
brunaccampos: // Important note : operations with the following vectors can only be done ...

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -87,10 +87,16 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
   // If enabled, create mortar coupling
   this->update_mortar_coupling();
 
+
+  // Important note : calculation on the following vectors (addition,
+  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
+  // locally_relevant_dofs. This is why most of them are reinitialized both with
```
brunaccampos: // Important note : operations with the following vectors can only be done ...

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -100,6 +101,15 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
                      update_values | update_quadrature_points |
                        update_JxW_values | update_gradients | update_hessians |
                        update_normal_vectors)
+    , sdirk_table(
```
brunaccampos: So the SDIRK method is initialized even when it is not used? could we have it separate and be called only if stepping method = sdirk?

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -103,10 +109,37 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
...
+  this->sdirk_vectors.sum_over_previous_stages.reinit(
+    this->locally_owned_dofs,
+    this->locally_relevant_dofs,
+    this->mpi_communicator);
+  this->sdirk_vectors.temp_sum_over_previous_stages.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
+
+  this->sdirk_vectors.locally_owned_for_calculus.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
+
```
brunaccampos: do these reinit need to always be called or only whe sdirk method is chosen?

source/solvers/navier_stokes_base.cc
```diff
@@ -567,81 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
-      multiphysics->percolate_time_vectors(true);
-    }
-  else
+  // The following variables are used for the SDIRK method
+  // u_{n+1} = u_n + \sum_{i=1}^{s} b_i k_i
+  auto &temp_sum_bi_ki = this->sdirk_vectors.temp_sum_bi_ki;
+  // At each time iteration, we reset the value of sum_bi_ki
+  temp_sum_bi_ki = 0;
+
+  for (unsigned int stage = 0; stage < n_stages; ++stage)
```
brunaccampos, line 708: is n_stages used for all time-stepping methods? initialized as zero for bdf?

brunaccampos (changes requested):
Some comments and some questions :) nice work

source/solvers/navier_stokes_base.cc
etiennemarin: Yes, for bdf, n_stages is initialized as one, so the loop has only one iteratation.

source/solvers/fluid_dynamics_matrix_based.cc
etiennemarin: They need to be called only when the sdirk method is chosen. Since most of the computational time is based on the assembling of matrices, I didn't add a condition. So, even when a BDF scheme is chosen, these are reinit. But maybe it's clearer to add an if ?

source/core/sdirk_stage_data.cc
etiennemarin: Yes they are fixed, I'll add that on the description

include/solvers/navier_stokes_base.h
etiennemarin: The terms post and precond are probably not really appropriate here. For sdirk, before solving the non linear system, we need to recalculate the sum(a_ij * k_j). After solving the non linear system, we need to calculate and store the k_i variable. We also update the sum(b_i*k_i). I'll change the name of the functions by post_resolution() and pre_resolution()

include/solvers/navier_stokes_assemblers.h
etiennemarin: Yes, it is given data by the choosen method (stored in the sdirk_stage_data.cc file). They are chosen to ensure that the method is consistent. Since the operations to find them is a bit long, I just take them from litterature.

include/solvers/navier_stokes_assemblers.h
etiennemarin: Right, I'll changed that :)

doc/source/parameters/cfd/simulation_control.rst
etiennemarin: Exactly, the only thing to change in the .prm file is the 'method = ' section. I'll add a note

include/solvers/navier_stokes_assemblers.h
etiennemarin: I wonder now, is Dioxygen converting this format in to nice formula since it is in the @brief section ?

brunaccampos: Oh indeed, I forgot about it!

source/solvers/fluid_dynamics_matrix_based.cc
brunaccampos: Makes sense. I would add an if just to make it clearer and only call them when needed, but it shouldn't be a big deal either way :)

include/solvers/navier_stokes_scratch_data.h
etiennemarin: Yes, it is initialized as empty if a BDF or steady method is chosen. I would say that it doesn't take much computional cost, so I Implemented like this. But I don't know if there is an easier way to implement this

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
etiennemarin: Actually the SDIRK43 doesn't reach order 4 as expected (approximately 3.5) so I don't know if, in the end, this scheme will be kept in the code...

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
etiennemarin: You're right, I will harmonize this :)

source/solvers/fluid_dynamics_matrix_based.cc
etiennemarin: If I don't reinit them even if I don't use them, I have this assertion when the tests are running :
The violated condition was: 
    values.size() == this->get_dof_handler().n_dofs()
For now, I just add a comment and remove the if condition

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
brunaccampos: Sure, I would remove the mentions of the scheme order 4 from the code if this is not being used/tested yet, just to keep it clear for users :)

doc/source/parameters/cfd/simulation_control.rst
```diff
@@ -107,6 +107,12 @@ This subsection is the most important in a simulation and therefore, the most co
...
+	* ``sdirk33``: 3rd order 3 stages singly diagonally implicit Runge-Kutta
+	* ``sdirk43``: 4th order 3 stages singly diagonally implicit Runge-Kutta
+
+.. note::
+	For now, the SDIRK schemes are only available for a matrix based resolution.
```
blaisb: That's cool, Maybe put a warning that the sdirk 43 is still experimental, but that's very cool!

include/core/simulation_control.h
```diff
@@ -230,6 +230,21 @@ class SimulationControl
            method == Parameters::SimulationControl::TimeSteppingMethod::bdf3;
   }
 
+  /**
+   * @brief Establish if the method is a sdirk method
+   *
+   * @return true if the method is sdirk22 or sdirk33, false otherwise.
```
blaisb:
```suggestion
   * @return true if the method is sdirk, false otherwise
```
we can understand that the sdirkXX methods are all sdirk, that's cool :)

include/solvers/navier_stokes_assemblers.h
```diff
@@ -308,6 +308,56 @@ class GLSNavierStokesAssemblerBDF : public NavierStokesAssemblerBase<dim>
...
+ * These coefficients are given data by the method and constant for the all
```
blaisb:
```suggestion
 * These coefficients are given data by the method and constant for the entire
```

include/solvers/navier_stokes_base.h
```diff
@@ -357,6 +358,33 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   virtual void
   dynamic_flow_control();
 
+  /**
+   * @brief multi_stage_preresolution
+   * It updates the sum_over_previous_stages variable to use it in the solver.
+   */
```
blaisb: Can you add the @param :)

include/solvers/navier_stokes_base.h
```diff
@@ -357,6 +358,33 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+  /**
+   * @brief multi_stage_postresolution
+   * Calculate k_i from u*_i : k_i = (u*_i - u_n)/time_step*a_ii - sum(a_ij *
+   * k_j) Update the sum(b_i * k_i)
+   */
+  virtual void
+  multi_stage_postresolution(
+    unsigned int                                      stage,
+    Parameters::SimulationControl::TimeSteppingMethod method,
+    double                                            time_step);
```
blaisb, line 367: Same as above

include/solvers/navier_stokes_base.h
```diff
@@ -357,6 +358,33 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+
+  /**
+   * @brief update_multi_stage_solution
+   * u_{n+1} = u_n + time_step * sum(b_i * k_i)
+   */
```
blaisb: same as above

include/solvers/navier_stokes_base.h
```diff
@@ -941,11 +978,38 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
   std::vector<VectorType> previous_solutions;
 
+  /**
+   * @brief Structure that stores all SDIRK-related vectors used during the time integration process.
+   */
+  struct SDIRKVectors
+  {
+    /// Vector to hold the locally non-relevant part of the solution (for
+    /// calculus purposes)
+    VectorType locally_owned_for_calculus;
```
blaisb: calculus means something else in English, maybe calculation?

etiennemarin: You're right, I just checked, it is used when we say for instance 'differential calculus'. I'll change that :)

include/core/simulation_control.h
hepap: And it includes also sdirk43 :)

include/solvers/navier_stokes_assemblers.h
```diff
@@ -308,6 +308,56 @@ class GLSNavierStokesAssemblerBDF : public NavierStokesAssemblerBase<dim>
   const std::shared_ptr<SimulationControl> simulation_control;
 };
 
+/**
+ * @brief Class that assembles the transient time arising from SDIRK time
```
hepap:
```suggestion
 * @brief Class that assembles the transient term arising from SDIRK time
```

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +360,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   virtual void
   dynamic_flow_control();
 
+  /**
+   * @brief multi_stage_preresolution
+   * It updates the sum_over_previous_stages variable to use it in the solver.
```
hepap:
```suggestion
   * @brief multi_stage_preresolution Updates the sum_over_previous_stages variable to use it in the solver.
```

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +360,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+   *
+   * @param stage An unsigned integer which gives the index of the current stage
+   * @param method SDIRK methods for now (BDF methods are single stage)
```
hepap:
```suggestion
   * @param method SDIRK method for now (BDF methods are single stage)
```

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +360,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+   */
+  virtual void
+  multi_stage_preresolution(
+    unsigned int                                      stage,
+    Parameters::SimulationControl::TimeSteppingMethod method);
+
+  /**
+   * @brief multi_stage_postresolution
+   * Calculate k_i from u*_i : k_i = (u*_i - u_n)/time_step*a_ii - sum(a_ij *
+   * k_j) Update the sum(b_i * k_i)
```
hepap: Indent

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -268,6 +280,11 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
       compute_cell_measure_with_JxW(this->fe_values.get_JxW_values());
     this->cell_size = compute_cell_diameter<dim>(cell_measure, fe.degree);
 
+    // For the SDIRK methods, \sum_{j=1}^{i-1} a_{ij} k_j is needed for the
+    // assembler
+    this->fe_values[velocities].get_function_values(sum_over_previous_stages,
+                                                    this->sdirk_stage_sum);
```
blaisb: You have access to the simulation control right? I wonder if  we could get_function_values of this field only and only if the method is sdirk? This would save additional calculations.

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -103,10 +109,36 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
...
+  this->sdirk_vectors.sum_over_previous_stages.reinit(
+    this->locally_owned_dofs,
+    this->locally_relevant_dofs,
+    this->mpi_communicator);
+  this->sdirk_vectors.temp_sum_over_previous_stages.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
+
+  this->sdirk_vectors.locally_owned_for_calculus.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
```
blaisb: I would reinitialize these vector only if the method is sdirk. you have this nice function in your simulation control to check it :)
Will save calculations for the rest of simulations.

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
+  auto &previous_k_j_solutions = this->sdirk_vectors.previous_k_j_solutions;
+
+  // sum(a_{ij} *k_j) when we are at least at stage 2
+  // If we are at stage 1, sum_over_previous_stages is set to 0.
+  auto &sum_over_previous_stages = this->sdirk_vectors.sum_over_previous_stages;
+  auto &temp_sum_over_previous_stages =
+    this->sdirk_vectors.temp_sum_over_previous_stages;
+
+  auto &locally_owned_for_calculus =
+    this->sdirk_vectors.locally_owned_for_calculus;
```
blaisb: calculus -> calculation

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
+  // calculated at each stage
+  auto &previous_k_j_solutions = this->sdirk_vectors.previous_k_j_solutions;
+
+  // sum(a_{ij} *k_j) when we are at least at stage 2
+  // If we are at stage 1, sum_over_previous_stages is set to 0.
+  auto &sum_over_previous_stages = this->sdirk_vectors.sum_over_previous_stages;
+  auto &temp_sum_over_previous_stages =
+    this->sdirk_vectors.temp_sum_over_previous_stages;
+
+  auto &locally_owned_for_calculus =
```
blaisb: Here I would add a comment that first you copy the reference to some of the vectors to enhance readability below. I have nothing agasint that, it is more readable, but I would explain that more clearly :)

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
 // Do an iteration with the NavierStokes Solver
 // Handles the fact that we may or may not be at a first
 // iteration with the solver and sets the initial condition
 template <int dim, typename VectorType, typename DofsType>
 void
 NavierStokesBase<dim, VectorType, DofsType>::iterate()
 {
+  // The locally_relevant_dofs are used for the assembly,
+  // while the non-locally_relevant_dofs are used for the calculus on the
+  // vectors (addition, multiplication, etc.).
```
blaisb:
```suggestion
  // while the non-locally_relevant_dofs are used for the calculations on the
  // vectors (addition, multiplication, etc.).
```

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
-      // is necessary because of the checkpointing mechanism. You spent an
-      // evening debugging this, trust me.
-      multiphysics->percolate_time_vectors(true);
-    }
-  else
+  // The following variables are used for the SDIRK method
+  // u_{n+1} = u_n + sum(b_i * k_i)
+  auto &temp_sum_bi_ki = this->sdirk_vectors.temp_sum_bi_ki;
+  // At each time iteration, we reset the value of sum_bi_ki
+  temp_sum_bi_ki = 0;
```
blaisb: I would do this only if this is an sdirk method. You don,t need these tables and operation if it's not sdirk :).

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
-                          simulation_parameters.simulation_control.method);
-      // Dear future Bruno, percolating auxiliary physics before fluid dynamics
-      // is necessary because of the checkpointing mechanism. You spent an
-      // evening debugging this, trust me.
-      multiphysics->percolate_time_vectors(true);
-    }
-  else
+  // The following variables are used for the SDIRK method
+  // u_{n+1} = u_n + sum(b_i * k_i)
+  auto &temp_sum_bi_ki = this->sdirk_vectors.temp_sum_bi_ki;
```
blaisb: Here I would avoid creating this temporary variable and just use the sdirk_vectors.temp_sum_bi_ki variable directly down there.

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -39,6 +39,8 @@ NavierStokesScratchData<dim>::allocate()
   this->velocity_hessians    = std::vector<Tensor<3, dim>>(n_q_points);
   this->shear_rate           = std::vector<double>(n_q_points);
 
+  this->sdirk_stage_sum = std::vector<Tensor<1, dim>>(n_q_points);
```
blaisb: I would consider only allocating if the method is sdirk

blaisb (changes requested):
A few last comments. Very good PR and very good work this was not easy.
Your implementation is 10X better than the one I had done in the past. It's a lot more reusable.

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +359,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+   *
+   * @param stage An unsigned integer which gives the index of the current stage
+   * @param method SDIRK methods for now (BDF methods are single stage)
```
hepap:
```suggestion
   * @param method SDIRK method for now (BDF methods are single stage)
```

include/solvers/navier_stokes_base.h
hepap: And I think the equations need to be between \f to appear like math in the doxygen

source/core/sdirk_stage_data.cc
```diff
@@ -50,28 +54,75 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
       return table;
     }
 
-  // Important note : the nomenclature used for the name of the SDIRK methods
-  // are sdirkOrderStage sdirk22 means SDIRK with order 2 and 2 stages, sdirk33
-  // means SDIRK with order 3 and 3 stages
+  // Important note : When referencing to SDIRK methods the pattern used is
+  // sdirkOrderStage. For instance, sdirk22 indicates SDIRK method with order 2
+  // and 2 stages
```
hepap:
```suggestion
```

source/core/sdirk_stage_data.cc
```diff
@@ -50,28 +54,75 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
...
   else if (method == Parameters::SimulationControl::TimeSteppingMethod::sdirk33)
     {
       // SDIRK33 - 3-stage, 3rd-order L-stable method (Kennedy-Carpenter)
       // Reference: J. F. G. Kennedy and A. D. Carpenter, "Diagonally Implicit
       // Runge-KuttaMethods for Ordinary Differential Equation"
       // NASA/TM–2016–219173 pp. 77, March 2016. No DOI available URL :
       // https://ntrs.nasa.gov/api/citations/20160005923/downloads/20160005923.pdf
+
+      // gamma is a constant coefficient chosen to ensure the method is stable
+      // and consistent
```
hepap, line 65: Is this based on the ref?

source/core/sdirk_stage_data.cc
```diff
@@ -50,28 +54,75 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
...
+      table.b[1] = -0.6443631707;
+      table.b[2] = gamma;
+
+      return table;
+    }
+
+
+  // Important note : When referencing to SDIRK methods the pattern used is
+  // sdirkOrderStage. For instance, sdirk22 indicates SDIRK method with order 2
+  // and 2 stages
```
hepap:
```suggestion
```

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk33.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
+
```
OGaboriault, line 8: Can you add a comment in the prm to know what this is testing ?

applications_tests/lethe-fluid/taylor-green-vortex_gls_sdirk22.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
```
OGaboriault, line 3: Can you add a comment in the prm to know what this is testing ?

applications_tests/lethe-fluid/taylor-green-vortex_gls_bdf2.prm
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
```
OGaboriault, line 3: Can you add a comment in the prm to know what this is testing ?

include/core/parameters.h
```diff
@@ -81,7 +81,8 @@ namespace Parameters
       bdf2,
       bdf3,
       sdirk22,
-      sdirk33
+      sdirk33,
+      sdirk43
```
OGaboriault, line 85: do we need the second number? Can't this just be sdirk2, sdirk3 and sdirk4 ?

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +359,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
   virtual void
   dynamic_flow_control();
 
+  /**
+   * @brief multi_stage_preresolution
+   * It updates the sum_over_previous_stages variable to use it in the solver.
```
OGaboriault:
```suggestion
  /**
   * @brief Update the sum_over_previous_stages variable to used in the solver.
```

include/solvers/navier_stokes_base.h
```diff
@@ -358,6 +359,42 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+   * @param stage An unsigned integer which gives the index of the current stage
+   * @param method SDIRK methods for now (BDF methods are single stage)
+   */
+  virtual void
+  multi_stage_preresolution(
+    unsigned int                                      stage,
+    Parameters::SimulationControl::TimeSteppingMethod method);
+
+  /**
+   * @brief multi_stage_postresolution
```
OGaboriault: Don't write the name of the function in the brief, just the description of what it is doing.

OGaboriault: I think the file was using this convention before your PR, but other files are not. Could you fix this problem for every function in this file?

source/solvers/fluid_dynamics_block.cc
```diff
@@ -602,10 +604,15 @@ FluidDynamicsBlock<dim>::setup_dofs_fd()
   // Zero constraints
   this->define_zero_constraints();
 
+  // Important note : operations on the following vectors (addition,
+  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
+  // locally_relevant_dofs. This is why most of them are reinitialized both with
```
hepap:
```suggestion
  // multiplication, etc.) can only be done if these are reinitialized WITHOUT
  // locally_relevant_dofs (i.e. without ghost DoFs). This is why most of them are reinitialized both with
```

source/solvers/fluid_dynamics_block.cc
```diff
@@ -618,10 +625,35 @@ FluidDynamicsBlock<dim>::setup_dofs_fd()
...
+  this->sdirk_vectors.temp_sum_bi_ki.reinit(this->locally_owned_dofs,
+                                            this->mpi_communicator);
+
+
+  this->sdirk_vectors.sum_over_previous_stages.reinit(
+    this->locally_owned_dofs,
+    this->locally_relevant_dofs,
+    this->mpi_communicator);
+  this->sdirk_vectors.temp_sum_over_previous_stages.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
```
hepap:
```suggestion
  this->sdirk_vectors.local_sum_over_previous_stages.reinit(
    this->locally_owned_dofs, this->mpi_communicator);
```

source/core/parameters.cc
```diff
@@ -78,11 +78,13 @@ namespace Parameters
...
-                        "steady",
-                        Patterns::Selection("steady|steady_bdf|bdf1|bdf2|bdf3"),
-                        "The kind of solver for the linear system. "
-                        "Choices are <steady|steady_bdf|bdf1|bdf2|bdf3>.");
+      prm.declare_entry(
+        "method",
+        "steady",
+        Patterns::Selection(
+          "steady|steady_bdf|bdf1|bdf2|bdf3|sdirk22|sdirk33|sdirk43"),
+        "The kind of solver for the linear system. "
```
OGaboriault: Is this really a kind of solver?

source/core/sdirk_stage_data.cc
```diff
@@ -50,28 +54,75 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
       return table;
     }
 
-  // Important note : the nomenclature used for the name of the SDIRK methods
-  // are sdirkOrderStage sdirk22 means SDIRK with order 2 and 2 stages, sdirk33
-  // means SDIRK with order 3 and 3 stages
+  // Important note : When referencing to SDIRK methods the pattern used is
```
OGaboriault: I think you can remove this comment since it is written at line 27

source/core/sdirk_stage_data.cc
```diff
@@ -50,28 +54,75 @@ sdirk_table(const Parameters::SimulationControl::TimeSteppingMethod method)
...
+      table.b.resize(3);
+      table.b[0] = 1.208496649;
+      table.b[1] = -0.6443631707;
+      table.b[2] = gamma;
+
+      return table;
+    }
+
+
+  // Important note : When referencing to SDIRK methods the pattern used is
```
OGaboriault: Same here

source/solvers/fluid_dynamics_block.cc
```diff
@@ -618,10 +625,35 @@ FluidDynamicsBlock<dim>::setup_dofs_fd()
...
+
+  this->sdirk_vectors.locally_owned_for_calculus.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
+
+  // Initialize vector of previous hk_j solutions
```
hepap:
```suggestion
  // Initialize vector of previous k_j solutions
```

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -103,10 +109,36 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
                       this->mpi_communicator);
     }
 
+  // Reinitialize vectors used for the SDIRK methods
+  this->sdirk_vectors.sum_bi_ki.reinit(this->locally_owned_dofs,
+                                       this->locally_relevant_dofs,
+                                       this->mpi_communicator);
+  this->sdirk_vectors.temp_sum_bi_ki.reinit(this->locally_owned_dofs,
```
hepap:
```suggestion
  this->sdirk_vectors.local_sum_bi_ki.reinit(this->locally_owned_dofs,
```

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -103,10 +109,36 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
...
+                                            this->mpi_communicator);
+
+
+  this->sdirk_vectors.sum_over_previous_stages.reinit(
+    this->locally_owned_dofs,
+    this->locally_relevant_dofs,
+    this->mpi_communicator);
+  this->sdirk_vectors.temp_sum_over_previous_stages.reinit(
```
hepap:
```suggestion
  this->sdirk_vectors.local_sum_over_previous_stages.reinit(
```

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -103,10 +109,36 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
...
+    this->locally_owned_dofs, this->mpi_communicator);
+
+  this->sdirk_vectors.locally_owned_for_calculus.reinit(
+    this->locally_owned_dofs, this->mpi_communicator);
+
+  // Initialize vector of previous hk_j solutions
```
hepap:
```suggestion
  // Initialize vector of previous k_j solutions
```

source/solvers/fluid_dynamics_matrix_based.cc
```diff
@@ -87,10 +87,16 @@ FluidDynamicsMatrixBased<dim>::setup_dofs_fd()
   // If enabled, create mortar coupling
   this->reinit_mortar();
 
+
+  // Important note : operations on the following vectors (addition,
```
OGaboriault: You can remove the "Important note :" part from this comment. 

Every comment is an important note ;)

source/solvers/navier_stokes_base.cc
```diff
@@ -567,82 +572,242 @@ NavierStokesBase<dim, VectorType, DofsType>::finish_time_step()
...
+
+template <int dim, typename VectorType, typename DofsType>
+void
+NavierStokesBase<dim, VectorType, DofsType>::multi_stage_preresolution(
+  unsigned int                                      stage,
+  Parameters::SimulationControl::TimeSteppingMethod method)
+{
+  // Very important variable because sum(a_{ij} * k_j) has to be
+  // calculated at each stage
+  auto &previous_k_j_solutions = this->sdirk_vectors.previous_k_j_solutions;
```
OGaboriault: I would reformulate this comment. 

something like : "sum(a_{ij} * k_j) has to be calculated at each stage. "

source/solvers/navier_stokes_base.cc
```diff
@@ -172,6 +172,10 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
   // of the BDF schemes
   previous_solutions.resize(maximum_number_of_previous_solutions());
 
+  sdirk_vectors.previous_k_j_solutions.resize(
+    SimulationControl::get_number_of_stages(
+      this->simulation_control->get_assembly_method()));
```
hepap: I guess this also could only be done if sdirk

source/solvers/navier_stokes_base.cc
```diff
@@ -172,6 +172,10 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
   // of the BDF schemes
   previous_solutions.resize(maximum_number_of_previous_solutions());
```
hepap, line 182: Do we need this if it is sdirk?

include/core/parameters.h
etiennemarin: There are methods of order 2 with 3 stages, or order 4 with 4 stages for example that are not yet implemented here. For now, I don't really know what are the most relevant Butcher's tables to use for the simulations, but maybe we will need all of these in the code. With this nomenclature we can have all of them :)

source/core/parameters.cc
etiennemarin: I don't know. It was already there, but actually I don't really know what does it mean. Maybe "method" is more accurate @blaisb ?

source/solvers/navier_stokes_base.cc
etiennemarin: Yes, I just removed these comments, I think it is better just to put them where the calculations are made

etiennemarin:
> Here are my comments!
> 
> I have just one question, if we are also setting VOF = true or any other physics, which scheme is taken? Or do we have a flag already since SDIRK is only implemented for NS?

source/solvers/navier_stokes_base.cc
etiennemarin: We need the previous_solutions[0] vector only, not the others !

etiennemarin: Maybe I can resize it with only one element ?

etiennemarin:
I corrected with all your advices. Thank you all for your time :)). For the comment of Gabo to harmonize the "@brief" in the navier_stokes_base.h file, I removed all the name of the functions from the brief, because I think it is already written by Dioxygen

source/core/sdirk_stage_data.cc
etiennemarin: Yes it is

include/solvers/navier_stokes_scratch_data.h
blaisb: No this is fine. But instead of manually checking for every SDIRK method, you should call your is_sdirk() method

source/solvers/fluid_dynamics_block.cc
blaisb: @etiennemarin I would add this suggestion

source/solvers/navier_stokes_base.cc
```diff
@@ -168,9 +168,20 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
...
 
-  // Pre-allocate memory for the previous solutions using the information
-  // of the BDF schemes
-  previous_solutions.resize(maximum_number_of_previous_solutions());
+
+  if (this->simulation_control->is_sdirk())
+    {
+      // Pre-allocate memory for the previous solutions using the information
+      // of the SDIRK schemes
+      previous_solutions.resize(1);
```
blaisb:
```suggestion
```
Let's wait a little for this optimization :) This can come afterwards ok? :)

blaisb (approved):
3 last things to address, then this will be good for merge in my opinion.
Please wait for the ok from @hepap and @brunaccampos  before marking reviewed and ready to merge please :)

source/solvers/fluid_dynamics_matrix_based.cc
blaisb, line 1: Instead of putting the assertions in navier_stokes_base, could you please put them in their respective physics (like in the constructor)
Just put these assertion in the constructor of:
- The tracer
- The heat transfer
- VOF
- CHN

:)

blaisb (approved):
Great work @etiennemarin . I know this is a lot of work, but it's normal :). Keep it up and one more day and it will be merged I think :)

applications_tests/lethe-fluid/taylor-green-vortex_gls_bdf2.prm
```diff
@@ -0,0 +1,153 @@
...
+# Testing the results of the BDF (2nd order BDF) scheme on the TGV example
```
brunaccampos:
```suggestion
# Testing the results of the 2nd order BDF scheme on the TGV example
```

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 - MINOR Prototype of slice output feature using deal.II's [DataOutputResample](https://dealii.org/current/doxygen/deal.II/classDataOutResample.html). The prototype works in parallel for the cases tested. This is the first step towards having a slice output feature, which will allow users to output only a slice or a region of the domain. [#1612](https://github.com/chaos-polymtl/lethe/pull/1612)
+### [Master] - 2025-08-03
```
```suggestion

### [Master] - 2025-08-03
```

CHANGELOG.md
```diff
@@ -59,7 +64,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 - MINOR This PR adds a mortar Arbitrary Lagrangian Eulerian (ALE) asssembler to allow the simulation of rotor-stator transient problems. [#1597](https://github.com/chaos-polymtl/lethe/pull/1597)
-
 ### [Master] - 2025-07-31
```
brunaccampos, line 69:
```suggestion

### [Master] - 2025-07-31
```

include/solvers/navier_stokes_base.h
```diff
@@ -334,33 +322,60 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
    */
   void
   postprocessing_torques(const VectorType &evaluation_point);
 
   /**
-   * @brief dynamic_flow_control
-   * If set to enable, dynamic_flow_control allows to control the flow by
-   * executing space-average velocity and beta coefficient force calculation at
-   * each time step.
+   * @brief If set to enable, dynamic_flow_control allows to control the flow by executing space-average velocity and beta coefficient force calculation at each time step.
```
brunaccampos, line 339: Nice cleanup on the briefs!

include/solvers/navier_stokes_assemblers.h
```diff
@@ -308,6 +308,56 @@ class GLSNavierStokesAssemblerBDF : public NavierStokesAssemblerBase<dim>
   const std::shared_ptr<SimulationControl> simulation_control;
 };
 
+/**
+ * @brief Class that assembles the transient term arising from SDIRK time
+ * integration for the Navier Stokes equations.
+ * For a given stage \( i \), the following expression is assembled and treated
+ * implicitly: \f[ \frac{\mathbf{u}_i^* - \mathbf{u}_n}{h a_{ii}} -
+ * \sum_{j=0}^{i-1} \frac{a_{ij}}{a_{ii}} \mathbf{k}_j \f] where:
+ * - \( \mathbf{u}_i^* \) is the current stage solution,
```
brunaccampos: Does each of these lines (317 - 322) need a \f too?

source/solvers/navier_stokes_base.cc
```diff
@@ -168,6 +168,14 @@ NavierStokesBase<dim, VectorType, DofsType>::NavierStokesBase(
   multiphysics = std::make_shared<MultiphysicsInterface<dim>>(
     simulation_parameters, triangulation, simulation_control, this->pcout);
 
+
+  if (this->simulation_control->is_sdirk())
+    {
+      //Pre-allocate memory for the previous k_j solutions using the number of stages of the method
```
brunaccampos:
```suggestion
      // Pre-allocate memory for the previous k_j solutions using the number of stages of the method
```

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 - MINOR Prototype of slice output feature using deal.II's [DataOutputResample](https://dealii.org/current/doxygen/deal.II/classDataOutResample.html). The prototype works in parallel for the cases tested. This is the first step towards having a slice output feature, which will allow users to output only a slice or a region of the domain. [#1612](https://github.com/chaos-polymtl/lethe/pull/1612)
+### [Master] - 2025-08-03
+
+### Added
+
+- MINOR A new time stepping scheme to solve the unsteady Navier-Stokes equations with a matrix based solver. Before, only BDF schemes were used in Lethe. This PR adds the Singly Diagonnaly Implicit Runge-Kutta (SDIRK) schemes with order 2, 3 and 4. [#1598](https://github.com/chaos-polymtl/lethe/pull/1598/)
```
hepap: This should be moved up with the date of when the PR will be merged.

include/solvers/navier_stokes_base.h
```diff
@@ -334,33 +322,60 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+  multi_stage_preresolution(
+    unsigned int                                      stage,
+    Parameters::SimulationControl::TimeSteppingMethod method);
+
+  /**
+   * @brief Calculate \f$ k_i \f$ from \f$ u*_i : k_i = \frac{u*_i - u_n}{time_step*a_ii} - \sum{j=1}^{i-1} a_ij * k_j \f$
+   * Update \f$ \sum_{i=1}^{N_{stages}} b_i * k_i \f$
+   *
+   * @param stage An unsigned integer which gives the index of the current stage
+   * @param method SDIRK methods for now (BDF methods are single stage)
```
hepap:
```suggestion
   * @param method SDIRK method for now (BDF methods are single stage)
```

include/solvers/navier_stokes_scratch_data.h
hepap: I would add this description

source/core/parameters.cc
hepap:
```suggestion
        "The time integration scheme. "
```

tests/core/sdirk_stage.cc
hepap, line 1: Maybe a description of the test can be added

blaisb:
@etiennemarin One last thing to check with one of  your test, then this is good for a mergggge :)

PR #1606, opened 2025-08-05
Mortar: remove second partitioner

peterrum:
... and `MortarMangerBase::get_indices()` and treat indices as cell quantities. This should make the transition to 3D easier.


blaisb:
@peterrum can you just add a small description of the change (like the why)

brunaccampos (approved):

PR #1612, opened 2025-08-08
Output slice prototype

voferreira:
### Description

This prototype tests the DataOutResample class of deal.II, which can be used to output slices of your dataset instead of its entirety. The steps include generating a new triangulation and resampling the results in the intersection points between the new and the original ones.

### Testing

The solution seems to work in the canonical case I have tested. In the future, when the final version will be implemented, the prototype will become a test.

### Documentation

No need for the time being.

Deploying this feature requires some refactoring that will be done in other PRs. This includes creating structs that will carry the output fields and the necessary information to generate the vtu files, which will be used to fill the data_out container.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+### [Master] - 2025-08-11
+
+### Added
+
+- MINOR Prototype of slice output feature using deal.II's [DataOutputResample](https://dealii.org/current/doxygen/deal.II/classDataOutResample.html). The prototype works in parallel for the cases tested. This is the first step towards having a slice output feature, which will allow users to output only a slice or a region of the domain. [#1612](https://github.com/chaos-polymtl/lethe/pull/1612)
+
 ### [Master] - 2025-08-08
 
 ### Fixed 


prototypes/output_slice/output_slice_parallel.cc
```diff
@@ -0,0 +1,1584 @@
...
+                                      MPI_COMM_WORLD,
+                                      3);
+}
+
+template <int dim, int fe_degree>
+void
+MatrixBasedPoissonProblem<dim, fe_degree>::output_slice_results(
+  const unsigned int cycle)
+{
+  if (triangulation.n_global_active_cells() > 1e6)
```
brunaccampos: Why do you need this check?

prototypes/output_slice/parameters.prm
```diff
@@ -0,0 +1,52 @@
...
+
+set dim = 3
+
+# Order of FE element <1|2|3>
+
+set element order = 1
+
+# Geometry <hyperball|hypercube|hyperrectangle>
+
+set geometry = hyperball
```
brunaccampos: Did you get the chance to try with different geometries?

brunaccampos (approved):
Nice! Just a couple questions :)

prototypes/output_slice/output_slice_parallel.cc
voferreira: I am not sure, but I think it is to make sure the test does not explode if we put a number of refinements that would lead to a super-heavy mesh.

prototypes/output_slice/parameters.prm
voferreira: Yes! It worked for the three cases. That's why I kept them there when I copied the prototype.

prototypes/output_slice/output_slice_parallel.cc
blaisb: Yeah I guess but it should still work nonetheless right, just take up more ressources? I'd remove that check and just print out a warning anyway. It will be important to prototype this application over large number of cores (>1000) and large meshes or degrees of freedom (>100M) to see if the approach scales for very large simulations since this is where we need it :) . So on the contrary, I'd check with this prototype if you can go to big things.

voferreira: Good point! I can launch this one on the cluster to see what happens. Thanks!

blaisb: Yeah parallel I/O is really the thing i would troubleshoot. This feature is for very big simulations so let'S try to use the prototype to see if it works for very big simulations (like I dunno, >10M or more DOfs). Otherwise, we never really need to do slices :).
If it slows down to a crawl for big parallel stuff, then maybe we need to rethink the way we do this. It does not prevent doing the refactoring in the output though :)

prototypes/output_slice/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_executable(output_slice_parallel output_slice_parallel.cc)
+deal_ii_setup_target(output_slice_parallel)
```
blaisb: If the prototype is called output_slice_parallel I would call the folder in the same way so that things are easier to understand :)

prototypes/output_slice/output_slice_parallel.cc
blaisb, line 1: That's a very big prototype for just prototyping output slices. Is there any reason why this uses a solver and not just like, pitching a solution on a DofHandler and then slicing it?
The goal of prototypes is often to test a feature right. Here you don't need a full GMG  matrix-based non-linear poisson solver to test slicing.
I would try to simplify this prototype by just interpolating a function then slicing it. Then also it would be a bit easier to understand and it will also be a LOT easier for you to test it over like 1000+ cores if you don't need to solve a problem but just output it.

Simple is good for these things.

blaisb (changes requested):
I would suggest  trimming down the prototype to just write a solution inside of the dof handler using vector tools instead of solving a problem. it will accomplish the same thing as far as I know but will be easier to understand for everyone and will be more durable.

prototypes/output_slice/output_slice_parallel.cc
```diff
@@ -0,0 +1,1584 @@
...
+                         DataOut<dim - 1, dim>::curved_inner_cells);
+  pcout << "slice output" << std::endl;
+  DataOutBase::VtkFlags flags;
+  flags.compression_level = DataOutBase::CompressionLevel::best_speed;
+  data_out.set_flags(flags);
+  data_out.write_vtu_with_pvtu_record(parameters.output_path,
+                                      parameters.output_name +
+                                        std::to_string(dim) + "d" + "_slice",
+                                      cycle,
+                                      MPI_COMM_WORLD,
```
blaisb: In the end this is the only part you want to test :). Where the solution comes from does not matter :)

voferreira:
@blaisb I did a clean-up of the prototype, and I think now we have a much simpler case to test it. I will test it on the cluster and let you know. So far, creating a slice and interpolating is not very costly, but I will check with a super fine mesh.

prototypes/CMakeLists.txt
```diff
@@ -14,3 +14,4 @@ add_subdirectory(matrix_free_mortar_rotate)
 add_subdirectory(modified_zonal_approach)
 add_subdirectory(template)
 add_subdirectory(vof_advection)
+add_subdirectory(output_slice_parallel)
```
oguevremont: Here I would move `output_slice_parallel` to its place in the alphabetical order.

prototypes/output_slice_parallel/parameters.prm
```diff
@@ -0,0 +1,40 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 The Lethe Authors
```
oguevremont:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
```

prototypes/output_slice_parallel/output_slice_parallel.cc
```diff
@@ -0,0 +1,516 @@
...
+                    Patterns::Integer(),
+                    "Repetitions in z direction for the hyper rectangle");
+  prm.declare_entry("output",
+                    "true",
+                    Patterns::Bool(),
+                    "Output vtu files <true|false>");
+  prm.declare_entry("output slice",
+                    "true",
+                    Patterns::Bool(),
+                    "Output vtu files <true|false>");
```
oguevremont:
```suggestion
                    "Output slice vtu files <true|false>");
```

prototypes/output_slice_parallel/output_slice_parallel.cc
```diff
@@ -0,0 +1,516 @@
...
+  , pcout(std::cout, Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0)
+  , computing_timer(MPI_COMM_WORLD,
+                    pcout,
+                    TimerOutput::never,
+                    TimerOutput::wall_times)
+  , mpi_communicator(MPI_COMM_WORLD)
+  , parameters(parameters)
+{}
+
+
```
oguevremont:
```suggestion
```

blaisb (approved):
Much better. Also much more compact. I think we can merge it no matter what. If performance is an issue, this is still the prototype we would use.

PR #1624, opened 2025-08-18
Refactor output results into struct

voferreira:
### Description

The goal of this refactor was to separate the write_output_results function into two: one that gathers the results in a single container (vector of structs) and another that takes the struct data and writes the vtu outputs. This is a step before adding a new feature that will allow results to be output in a mesh different than the one of the simulation. This can be used, for example, to output only a slice of the domain instead of the entirety of the results. With the gathering of the results and the instantiation of the feeding of the data_out object being done in two different functions, other instances of DataOut can be generated.

This PR also deprecates the smoothers of post-processing fields.

### Testing

Test cases included the example case flow around cylinder, a simple case for the tracer in a micromixer, and the validation cases launched on Lucille.

### Documentation

No documentation needed.

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/output_struct.h
```diff
@@ -0,0 +1,92 @@
...
+ * attributes such as name and data component interpretation. This version of
+ * the struct uses Postprocessors.
+ */
+template <int dim, typename VectorType>
+struct OutputStructPostprocessor
+{
+  /**
+   * @brief Constructor for when data is in data_postprocessor.
+   *
+   * @param[in] solution Solution field.
```
AmishgaAlphonius, line 25:
```suggestion
   * @param[in] dof_handler DoFHandler containing solution.
   *
   * @param[in] solution Solution field.
```

include/solvers/navier_stokes_base.h
```diff
@@ -853,7 +856,23 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+   *
+   * @param[out] solution_output_structs Vector of OutputStructs that will be
+   * used to write the output results as VTU files
+   *
+   * @param[in] solution Vector of present solution
+   */
+  void
+  gather_output_results(
+    std::vector<OutputStruct<dim, VectorType>> &solution_output_structs,
+    const VectorType                           &solution);
```
OGaboriault: I think we should change the order of the input parameter. Since the first one is the output, makes more sense to be the last one.

include/solvers/output_struct.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef lethe_output_struct_h
+#define lethe_output_struct_h
+
+#include <solvers/postprocessors.h>
+
+using namespace dealii;
+
+/**
+ * @brief Struct containing information about the solution output.
+ * It is used to pass all the information required upon calling
```
OGaboriault:
```suggestion
 * It is used to pass the information required upon calling
```

include/solvers/output_struct.h
```diff
@@ -0,0 +1,92 @@
...
+   * @brief Constructor for when data is not related to a data_postprocessor.
+   *
+   * @param[in] dof_handler DoFHandler containing solution.
+   *
+   * @param[in] solution Solution field.
+   *
+   * @param[in] solution_names Vector with strings containing solution name.
+   *
+   * @param[in] data_component_interpretation Vector with data component
+   * interpretation of the solution.
```
OGaboriault: Are we putting spaces between @param lines?  I think it is better to write wiothout the extra lines.

include/solvers/output_struct.h
```diff
@@ -0,0 +1,92 @@
...
+ * add_data_vector() for the DataOut instance without losing track of its
+ * attributes such as name and data component interpretation. This version of
+ * the struct uses Postprocessors.
+ */
+template <int dim, typename VectorType>
+struct OutputStructPostprocessor
+{
+  /**
+   * @brief Constructor for when data is in data_postprocessor.
+   *
```
OGaboriault, line 23:
```suggestion
   * @param[in] dof_handler ...
```

OGaboriault (approved):
Not sure I'm the best person to review this PR, but seems ok for me. 

I would really change the extra line between @param .

AmishgaAlphonius (approved):
Very clean! :)
Only one comment on my side.

include/solvers/navier_stokes_base.h
blaisb: Yeah agreed. Ideally we should always put the const argument first and the modified at the last. It's not a "hard" convention, but it is a good habit we should develop

include/solvers/fluid_dynamics_matrix_free.h
```diff
@@ -11,6 +11,8 @@
 
 #include <deal.II/lac/precondition.h>
 #include <deal.II/lac/solver_gmres.h>
+#include <deal.II/lac/trilinos_precondition.h>
+#include <deal.II/lac/trilinos_solver.h>
```
blaisb, line 15: This should not be necessary since the fluid_dynamics_matrix_free solver is independent of trilinos
What you might need to import is the global_vector_type.h file

include/solvers/navier_stokes_base.h
```diff
@@ -252,7 +253,9 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
         this->postprocess_fd(true);
         multiphysics->postprocess(true);
         if (this->simulation_control->is_output_iteration())
-          this->write_output_results(this->present_solution);
+          {
+            this->write_output_results(this->present_solution);
+          }
```
blaisb:
```suggestion
            this->write_output_results(this->present_solution);
```

One line if conditions are fine, they take less space and they say the same things as regular ifs. I understand it might seem like the {} make the code "more structured" but it is really just more stuff for a person to parse.

source/solvers/output_struct.cc
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+#include <solvers/output_struct.h>
+
+#include <deal.II/base/vector_tools>
```
blaisb: Do you need this include? I would presume what you need is the global vector type include and the dealii parallel vector one

source/solvers/postprocessors_smoothing.cc
blaisb, line 1: Glad to see this option go. Maybe another day we will remake it but cleaner.

source/solvers/navier_stokes_base.cc
```diff
@@ -2801,33 +2788,39 @@ NavierStokesBase<dim, VectorType, DofsType>::write_output_results(
...
   // Create the post-processors to have derived information about the velocity
   // They are generated outside the if condition for smoothing to ensure
   // that the objects still exist when the write output of DataOut is called
   // Regular discontinuous postprocessors
-  QCriterionPostprocessor<dim> qcriterion;
-  DivergencePostprocessor<dim> divergence;
-  VorticityPostprocessor<dim>  vorticity;
-  data_out.add_data_vector(solution, vorticity);
+  // They are created as shared pointers to outlive the function
+  auto qcriterion = std::make_shared<QCriterionPostprocessor<dim>>();
```
blaisb: I would rather have the types be explicit than auto. It's a lot easier for people IDE to parse them this way.

source/solvers/navier_stokes_base.cc
```diff
@@ -2866,98 +2862,115 @@ NavierStokesBase<dim, VectorType, DofsType>::write_output_results(
...
+  const unsigned int iter          = simulation_control->get_step_number();
+  const double       time          = simulation_control->get_current_time();
+  const unsigned int subdivision = simulation_control->get_number_subdivision();
+  const unsigned int group_files = simulation_control->get_group_files();
+
+  // Vector containing structs with vtu output information
+  std::vector<OutputStruct<dim, VectorType>> solution_output_structs;
+
+  // Gather all results in vector container
+  gather_output_results(solution_output_structs, this->present_solution);
```
blaisb: Since everytime you call gather_output_results you use it to fill a vector,  you could also just have the solution_output_structs return a vector instead and combine the line above with this one.
I mean it should not have a real impact on performance.

source/solvers/navier_stokes_base.cc
```diff
@@ -2866,98 +2862,115 @@ NavierStokesBase<dim, VectorType, DofsType>::write_output_results(
...
+                                   postprocessor_struct->solution,
+                                   *postprocessor_struct->data_postprocessor);
+        }
+    }
+  Vector<float> subdomain(this->triangulation->n_active_cells());
+  for (unsigned int i = 0; i < subdomain.size(); ++i)
+    subdomain(i) = this->triangulation->locally_owned_subdomain();
+  data_out.add_data_vector(subdomain,
+                           "subdomain",
+                           DataOut<dim>::type_cell_data);
```
blaisb, line 2980: This one is treated differently because it's a cell data?

include/solvers/navier_stokes_base.h
voferreira: I was calling two different functions while testing, that's why :)

include/solvers/navier_stokes_base.h
voferreira: Perfect!

include/solvers/fluid_dynamics_matrix_free.h
voferreira: Not importing the first one gives me a compilation error " ‘PreconditionILU’ is not a member of ‘dealii::TrilinosWrappers’" whereas not importing the second one gives me " ‘dealii::TrilinosWrappers::SolverDirect’ has not been declared".
Those were being imported with the deprecated postprocess smoothing.

I am not sure, though, if these are the correct files to use.

voferreira: Let me double-check

source/solvers/navier_stokes_base.cc
voferreira: Exactly!

include/solvers/fluid_dynamics_matrix_free.h
blaisb: Ohh these are for the preconditioner solver in the GMG, That could make sense yeah. It's a wonder where they were being imported from then.

source/solvers/navier_stokes_base.cc
blaisb: I would just add a comment here that clarifies this point for a reader. Something like "Since the subdomain is a cell data, it is processed seperately"

blaisb (approved):
Other than the comment i left above, good to go from my side. Great pre-cleanup

PR #1626, opened 2025-08-19
Fix bugs with constraints when using local smoothing

lpsaavedra:
### Description

The Taylor-Couette transient case was not converging when using local smoothing. 

### Solution

Several bugs were identified and fixed:
- Pressure constraint was not correctly set for local smoothing.
- A bug in the periodic boundary conditions was introduced since an additional boundary type `periodic_neighbor` was introduced in [](https://github.com/chaos-polymtl/lethe/pull/1330).
- A timer that was throwing an error in debug mode.

### Testing

A minimal Taylor-Couette test with periodic boundary conditions was added to ensure that this does not happen again when using `lsmg` and `gcmg` as preconditioners for the matrix-free application. 

### Documentation

It does not change anything regarding the documentation. 

Code related list:

Pull request related list:
- [ ] Labels are applied


source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -1133,6 +1143,11 @@ MFNavierStokesPreconditionGMGBase<dim>::reinit(
                       .periodic_direction.at(id),
                     level_constraint);
                 }
+              else if (type ==
+                       BoundaryConditions::BoundaryType::periodic_neighbor)
+                {
+                  /*already taken into account in the previous else if*/
+                }
```
blaisb, line 1158: might be better to migrate to if statements here instead of if else if else if else since I feel like these problems are bout to arise.
Maybe at least you could have the "else statement" throw an error instead something like this boundary condition is not defined. Otherwise i am afraid these errors can happen again in the future

lpsaavedra: I agree, I have updated it accordingly. Now we check for each of the cases and throw in case a non-supported BC is specified.

blaisb:
Will merge the other PR and rebase this one after when I wake up tmrow
 :)

PR #1632, opened 2025-08-24
Update instructions for launching simulations on clusters

voferreira:
### Description

Small comment on using Trillium. According to the support, the new Trillium cluster needs an additional line in the launching script. Additionally, it uses mpirun instead of srun.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/doc/source/installation/digital_alliance.rst
+++ b/doc/source/installation/digital_alliance.rst
@@ -206,10 +206,3 @@ To upload this local key to your Compute Canada Database account (CCDB) use:
   :class: copy-button
 
     ssh-copy-id username@clustername.computecanada.ca
-
-.. code-block:: text
-  :class: copy-button
-
-  cat ~/.ssh/$KEY_ID.pub
-
-where ``$KEY_ID.pub`` is the public key file located in ``~/.ssh/``. For more information, see `SSH documentation <https://docs.scinet.utoronto.ca/index.php/SSH#SSH_Keys>`_.
--- a/doc/source/installation/digital_alliance.rst
+++ b/doc/source/installation/digital_alliance.rst
@@ -181,7 +181,7 @@ The job is sent using:
   .. code-block:: text
     :class: copy-button
 
-    mpirun -np $(( $SLURM_NNODES * $SLURM_NTASKS_PER_NODE )) $HOME/lethe/inst/bin/$lethe_application_name_wanted $lethe_application_name_wanted $parameter_file_name.prm
+    mpirun $HOME/lethe/inst/bin/$lethe_application_name_wanted $lethe_application_name_wanted $parameter_file_name.prm
 
 Status can be followed with the ``sq`` command: under ``ST``, ``PD`` indicates a pending job, and ``R`` a running job.
 
--- a/doc/source/installation/digital_alliance.rst
+++ b/doc/source/installation/digital_alliance.rst
@@ -122,7 +122,7 @@ We use `Globus <https://alliancecan.ca/en/services/advanced-research-computing/n
 Launching Simulations
 ---------------------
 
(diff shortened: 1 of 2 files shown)


doc/source/installation/digital_alliance.rst
```diff
@@ -168,6 +168,21 @@ The job is sent using:
 
   sbatch job.sh
 
+.. important::
```
blaisb: I would move this warning before line 164 (aka put the warning before the instructions on how to submit jobs. Otherwise it makes the reading of it kinda confusing. (see the image below, it's a bit weird I think)
Screenshot 2025-08-25 at 09 44 15

blaisb (approved):

blaisb:
Just one comment to address then this will be ready for merge

doc/source/installation/digital_alliance.rst
OGaboriault, line 1: "For your convenience, an example of job.sh used on Beluga is given below:" 

Can you change beluga for any other cluster name since it will be shutdown soon?

OGaboriault: Also, you could add :

"##SBATCH --dependency=$previous-slurm-job-id" and explain what is a use case scenario of using this command.

doc/source/installation/digital_alliance.rst
```diff
@@ -168,6 +168,21 @@ The job is sent using:
...
+  While using `Trillium <https://docs.alliancecan.ca/wiki/Trillium>`_, add the following line after ``source $HOME/.dealii`` in your ``.sh`` launching script:
```
OGaboriault: It is weird, because we are saying previously in the file :

```
"This file needs to be sourced every time you launch a job or you compile deal.II and/or Lethe. Once the file has been created, you can then source it on the terminal with:

source $HOME/.dealii
and use it in your .sh script when launching a job (see [:ref:`Launching Simulations<Launching Simulations>`](https://github.com/chaos-polymtl/lethe/blob/a7947db3a67eac8585d72e95781053e5f68a0bde/doc/source/installation/digital_alliance.rst#id1) below)."
```

Should we say to always source the .dealii regardless of the cluster you are using?

doc/source/installation/digital_alliance.rst
```diff
@@ -168,6 +168,21 @@ The job is sent using:
...
+
+  .. code-block:: text
+    :class: copy-button
+
+    source /scinet/vast/etc/vastpreload-openmpi.bash
+
+  Additionally, instead of using ``srun``, use ``mpirun`` to launch your application:
```
OGaboriault: does it work if you use mpirun on other cluster?

doc/source/installation/digital_alliance.rst
OGaboriault: Could you add trillium in the cluster specification table at the bottom of the page? :) (You can remove niagara I think)

OGaboriault (approved):
Seems good to me. 

I added comment regarding the page in general. If you have time to apply those comment, that would be great. For example, we sill talk about beluga et niaga... Otherwise this is good to go.

doc/source/installation/digital_alliance.rst
blaisb: Yeah. I would also suggest putting the .sh include for VAST in the .dealii instead of in the slurm file. The goal is to ensure that the .dealii file is enough to configure the whole system. Stuff that are required for each jobs should be in the .dealii

doc/source/installation/digital_alliance.rst
blaisb: I think it does, but I am not 100% sure

OGaboriault: It would be smart/better to have one way that works regardless of the cluster being used. I can test when I'll be at poly.

voferreira: I agree. I will test it as well!

doc/source/installation/digital_alliance.rst
voferreira: Sure! Good one

doc/source/installation/digital_alliance.rst
blaisb: I would suggest testing it on Rorqual. On Nibi I know mpirun works. I have no clue what was srun except an alias for mpirun with the number of cores specified

PR #1634, opened 2025-08-28
Cleanup scratch data for particle-fluid interaction and add defensive assertions

blaisb:
### Description

There were some issues with the scratch data for particle-fluid interaction:
1. Many functions were in the .h when they could have easily been in the .cc -> This has been fixed by moving them to the .cc
2. There are many locations were assumptions that a reinit is only called if the physics has been enabled were never asserted for. This is not a good defensive programming practice -> I have added a bunch of assertions whenever an enable is required or assumed
3. For a weird reason, gather the fluid information at the particle location required that the fluid information would have already been reinit and gathered. This meant that we could not gather the particle information without gathering the fluid information first, which is a fallacy since both are independent (they use different quadrature etc.) -> This has been fixed by making sure that the velocity_cell is used instead of the fe_values.get_cell() function.
4. The fluid properties array at the particle location was being resized at every iteration. This is not required for this array. The only reason why the other array are being resized is because of a weird quirk in fe_values that requires that the array used for the calculation be of the same size as the number of point in the quadrature.

### Testing

No tests should be changed by this.

### Documentation

No documentation change

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -916,19 +840,21 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
-            kinematic_viscosity_at_particle_location[i_particle] =
-              kinematic_viscosity_scale;
-            density_at_particle_location[i_particle] = density_scale;
-          }
+        // Gather the kinematic viscosity and density at the particle
+        // location assuming a constant kinematic viscosity and density in a
+        // single fluid
+        kinematic_viscosity_at_particle_location[i_particle] =
+          kinematic_viscosity_scale;
+        density_at_particle_location[i_particle] = density_scale;
```
antonellasuccar, line 857: Does returning inside the if condition instead of having ``else`` save time?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -956,7 +884,7 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
           fluid_particle_relative_velocity_at_particle_location[i_particle];
 
         Re_particle[i_particle] =
-          1e-3 +
+          Re_particle_tolerance +
```
antonellasuccar, line 887: Is this added in case where the particle velocity relative to the fluid is near zero?

blaisb: Is zero, yeah. Which is always the case at the first newton step of the first iteration for example.

PR #1640, opened 2025-08-31
Dem rolling resistance naming convention

### LINKED ISSUE #1420: Rolling resistance model names
Rolling resistance model are declared as `constant_resistance`,  `viscous_resistance`, `no_resistance` and  `epsd_resistance`. They should be declared without the "resistance" in their names, or to "none" in the case of `no_resistance`.

OGaboriault:
### Description

The naming convention related to rolling resistance models is redundant. When setting the type of model : 

"set rolling resistance torque model = constant_resistance "

The "_resistance" is not necessary. This PR is resolve #1420 . 

### Testing
Every test are passing and I tested that the warning message informing the user that he is using a soon to be deprecated parameter naming convention appears when using the old versions. 

### Documentation

The documentation was updated using the new convention. 

N/A

Code related list:
- [X] If parameters are modified, the tests and the documentation of examples are up to date
- [X] Changelog (CHANGELOG.md) is up to date if the refactor affects the user experience or the codebase

Pull request related list:
- [X] No other PR is open related to this refactoring
- [X] Labels are applied
- [X] If any future works is planned, an issue is opened

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,17 +3,24 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+<<<<<<< HEAD
 ### [Master] - 2025-09-01
 
 ### Added
 
 - MINOR Adds the capability to project the particle-fluid forces within the ParticleProjector class. This is still an experimental feature that is not usable from the parameter file. [#1642](https://github.com/chaos-polymtl/lethe/pull/1642)
 
+### [Master] - 2025-08-31
+
+### Changed
+
+- MAJOR The rolling resistance models naming convention was redundant. Every option had an extra \"_resistance\" at the end, which was not necessary. Options are now "none", "constant", "viscous" and "epsd". The old naming convention is still compatible for now, but will be deprecated down the line. Currently, a warning message is shown when using the previous naming convention. [#1640](https://github.com/chaos-polymtl/lethe/pull/1640)
+
 ### [Master] - 2025-08-30
 
 ### Changed
 
-- MINOR The `create_random_number_container`, previously only used during the volume insertion in the DEM, has been moved to the `utilities.h` file. This way, this function will be accessible outside of the volume insertion, which will limit duplicated code. [#1639](https://github.com/chaos-polymtl/lethe/pull/1639)
+- MINOR The `create_random_number_container`, previously only used during the volume insertion in the DEM, has been moved to the `utilities.h` file. This way, this function will be accessible outside the volume insertion, which will limit duplicated code. [#1639](https://github.com/chaos-polymtl/lethe/pull/1639)
 
 ### [Master] - 2025-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,18 +3,17 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
-<<<<<<< HEAD
+### [Master] - 2025-09-02
+
+### Changed
+
+- MAJOR The rolling resistance models naming convention was redundant. Every option had an extra \"_resistance\" at the end, which was not necessary. Options are now "none", "constant", "viscous" and "epsd". The old naming convention is still compatible for now, but will be deprecated down the line. Currently, a warning message is shown when using the previous naming convention. [#1640](https://github.com/chaos-polymtl/lethe/pull/1640)
+
 ### [Master] - 2025-09-01
 
 ### Added
 
 - MINOR Adds the capability to project the particle-fluid forces within the ParticleProjector class. This is still an experimental feature that is not usable from the parameter file. [#1642](https://github.com/chaos-polymtl/lethe/pull/1642)
-
-### [Master] - 2025-08-31
-
-### Changed
-
-- MAJOR The rolling resistance models naming convention was redundant. Every option had an extra \"_resistance\" at the end, which was not necessary. Options are now "none", "constant", "viscous" and "epsd". The old naming convention is still compatible for now, but will be deprecated down the line. Currently, a warning message is shown when using the previous naming convention. [#1640](https://github.com/chaos-polymtl/lethe/pull/1640)
 
 ### [Master] - 2025-08-30
 
--- a/doc/source/parameters/dem/model_parameters.rst
+++ b/doc/source/parameters/dem/model_parameters.rst
@@ -49,9 +49,9 @@ In this subsection, contact detection, force models, time integration, load bala
     set integration method                     = velocity_verlet
 
     # Rolling resistance method
-    # Choices are no_resistance|constant_resistance|viscous_resistance|epsd_resistance
-    set rolling resistance torque method       = constant_resistance
-    set f coefficient                          = 0.0 # if rolling resistance torque method = epsd_resistance
+    # Choices are none|constant|viscous|epsd
+    set rolling resistance torque method       = constant
+    set f coefficient                          = 0.0 # if rolling resistance torque method = epsd
 
     subsection adaptive sparse contacts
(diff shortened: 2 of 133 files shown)


blaisb:
Yessssssssssssssss
Nice

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+### [Master] - 2025-08-31
+
+### Changed
+
+- MINOR The rolling resistance models naming convention was redundant. Every option had an extra \"_resistance\" at the end, which was not necessary. Options are now "none", "constant", "viscous" and "epsd". [#1640](https://github.com/chaos-polymtl/lethe/pull/1640)
```
```suggestion
- MAJOR The rolling resistance models naming convention was redundant. Every option had an extra \"_resistance\" at the end, which was not necessary. Options are now "none", "constant", "viscous" and "epsd". [#1640](https://github.com/chaos-polymtl/lethe/pull/1640)
```

blaisb: I think we also put a bigger warning when there is a syntax change. Can ctrl+f syntax in the changelog and change? I think any syntax change is a major change

doc/source/parameters/dem/model_parameters.rst
```diff
@@ -49,9 +49,9 @@ In this subsection, contact detection, force models, time integration, load bala
     set integration method                     = velocity_verlet
 
     # Rolling resistance method
-    # Choices are no_resistance|constant_resistance|viscous_resistance|epsd_resistance
-    set rolling resistance torque method       = constant_resistance
-    set f coefficient                          = 0.0 # if rolling resistance torque method = epsd_resistance
+    # Choices are none|constant|viscous|epsd
+    set rolling resistance torque method       = constant
```
blaisb, line 53:
```suggestion
    set rolling resistance torque method       = constant
```

Not in this PR, but do you think we should change the default to EPSD? I think so...

include/dem/particle_wall_contact_force.h
```diff
@@ -496,30 +496,28 @@ class ParticleWallContactForce
...
+    if constexpr (rolling_friction_model == viscous)
       {
         return viscous_rolling_torque<PropertiesIndex>(particle_radius,
                                                        particle_properties,
                                                        rolling_friction_coeff,
                                                        normal_force_norm,
                                                        normal_unit_vector);
       }
-    if constexpr (rolling_friction_model == epsd_resistance)
+    if constexpr (rolling_friction_model == epsd)
```
blaisb, line 520: These ones why don't they have the RollingResistanceMethod:: 
prefix?

Let's try to make it uniform (I know it was already like this before)

source/core/parameters_lagrangian.cc
```diff
@@ -882,13 +882,14 @@ namespace Parameters
...
+          "Cut-off threshold above which the Van der Waal forces are "
+          "ignored for the DMT model relative to the pull-off force");
 
         prm.declare_entry(
           "rolling resistance torque method",
-          "constant_resistance",
+          "constant",
           Patterns::Selection(
-            "no_resistance|constant_resistance|viscous_resistance|epsd_resistance"),
+            "none|no_resistance|constant|constant_resistance|viscous|viscous_resistance|epsd|epsd_resistance"),
```
blaisb, line 892:
```suggestion
            "none|no_resistance|constant|constant_resistance|viscous|viscous_resistance|epsd|epsd_resistance"),
```

Ahh wow ok, you marked everything for deprecation. That's pretty dope.
We will deprecate these parameter after the next numbered release.

blaisb: Maybe we should keep a track of what we have currently deprecated, but I don't think it's that big of a deal.

source/core/parameters_lagrangian.cc
```diff
@@ -1107,30 +1108,64 @@ namespace Parameters
...
-        if (rolling_resistance_torque == "no_resistance")
+
+        if (Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) == 0)
+          {
+            if (rolling_resistance_torque == "no_resistance")
+              std::cout
+                << "Warning, the \"no_resistance\" entry to the \"rolling "
+                   "resistance torque method\" parameter will be deprecated. "
+                   "Please use \"none\" instead."
+                << std::endl;
```
blaisb, line 1119: I think Audrey had a special option to mark for deprecation but I don't remember what is was. the parameter handler supports that in-situ I think

source/core/parameters_lagrangian.cc
```diff
@@ -1107,30 +1108,64 @@ namespace Parameters
...
+                 rolling_resistance_torque == "epsd")
           {
-            rolling_resistance_method =
-              RollingResistanceMethod::epsd_resistance;
+            rolling_resistance_method = RollingResistanceMethod::epsd;
           }
         else
           {
             throw(
               std::runtime_error("Invalid rolling resistance torque method "));
```
blaisb: Can you AssertThrow instead, I am trying to keep it cleaner with AssertTrhwo instead of regular throws everywhere
like AssertThrow(false, ExcMessage()).

blaisb (changes requested):
Very few changes, thx for this

source/core/parameters_lagrangian.cc
OGaboriault: I did remember this while coding for the change! You are talking about the declare_alias.The gravity parameter is a good example of this.

Back in the days, we had three parameters for gravity (gx, gy, and gz) now we have one. 

The difference, and why I did it like I did, is that Audrey changed the parameter name while I changed the parameter entry. I don't think that dealii supports an "alias" for parameters entries.

OGaboriault: Although, wecould think about adding this to dealii

blaisb: Yeah but I understand your point, it's fine then
One thing i don't like about aliases is we will never know when to deprecate them. Here we will deprecate after next numbered release (which will be after 1.0 paper is published I think).

source/core/parameters_lagrangian.cc
OGaboriault: I don't know how we should do that... 

We could change the declare_entry function in dealii? We could add an input parameter, to the declare-entry function... something like const std::string &deprecated_entries = "" ? If the patterns is a selection, it checks if the entry is deprecated and writes a warning message if true. 

When we launch a new release, we would just have to set those strings to back to ""

source/core/parameters_lagrangian.cc
OGaboriault: I don't remember how the release notation is gonna work for us, but I think we should remove deprecated input paramters and entries once a year.

doc/source/parameters/dem/model_parameters.rst
OGaboriault: Mehh, I don't know. we didn't benchmark a simulation with both models. Maybe the EPSD is really expensive...  The constant can still be useful in a dynamic problem. Like a rotating drum...

source/core/parameters_lagrangian.cc
OGaboriault: Wait a second. I don't think we need the throw at all. Since the pattern used when declaring the parameter is a Selection, it already throws an exception if it doesn't match one of the written options.

source/core/parameters_lagrangian.cc
blaisb: We will do 2-3 releases a year I think. One when the deal.II 9.7 version release is finalized and another in the spring.
I will do 1.0.2 release soon I think since deal.II 9.7 is released.

source/core/parameters_lagrangian.cc
blaisb: Yeah that is true

blaisb (approved):

PR #1655, opened 2025-09-03
Inlet polynomial function expression script

voferreira:
### Description

Create a script to generate an inlet flow rate function expression according to the desired polynomial degree.

### Testing

No test needed. It is just a script

### Documentation

Documentation written in tools.

Code related list:

Pull request related list:
- [x] Labels are applied


contrib/utilities/inlet_profile_generator/inlet_profile_generator.py
```diff
@@ -0,0 +1,110 @@
...
+parser = argparse.ArgumentParser(description="Generate inlet polynomial velocity profile with given flow rate")
+parser.add_argument("--polynomial-degree", type=int, default=2, help="Degree k of the polynomial shape (k>=0)")
+parser.add_argument("--shape", type=str, default="rectangle", choices=["rectangle", "circle"],
+                    help="Inlet cross-section shape")
+parser.add_argument("--flow-rate", type=float, default=1.0, help="Desired volumetric flow rate Q")
+parser.add_argument("--center", type=float, nargs="+", default=[0.0, 0.0],
+                    help="Center coordinates [x0, y0]")
+parser.add_argument("--length", type=float, nargs="+", default=[1.0, 1.0],
+                    help="Rectangle: lengths [ax, ay]. Circle: diameter [D]")
+parser.add_argument("--show", type=bool, default=False, help="Show a 1D centerline plot (rectangle only)")
```
blaisb: You don't need the show to request the user to give a value like True.
What I would suggest is instead something like:
https://github.com/chaos-polymtl/lethe/blob/master/examples/incompressible-flow/2d-lid-driven-cavity/post_process_Reynolds_400.py#L25

Where the fact that the argument is present creates an action that stores true. This way you could just write --show

contrib/utilities/inlet_profile_generator/inlet_profile_generator.py
```diff
@@ -0,0 +1,110 @@
...
+# Parse arguments
+parser = argparse.ArgumentParser(description="Generate inlet polynomial velocity profile with given flow rate")
+parser.add_argument("--polynomial-degree", type=int, default=2, help="Degree k of the polynomial shape (k>=0)")
+parser.add_argument("--shape", type=str, default="rectangle", choices=["rectangle", "circle"],
+                    help="Inlet cross-section shape")
+parser.add_argument("--flow-rate", type=float, default=1.0, help="Desired volumetric flow rate Q")
+parser.add_argument("--center", type=float, nargs="+", default=[0.0, 0.0],
+                    help="Center coordinates [x0, y0]")
+parser.add_argument("--length", type=float, nargs="+", default=[1.0, 1.0],
+                    help="Rectangle: lengths [ax, ay]. Circle: diameter [D]")
```
blaisb: I tend to like having short handed version of the arguments also like -s, -fr, etc, but in the hand that's your call.

doc/source/tools/inlet_profile_generator/inlet_profile_generator.rst
```diff
@@ -0,0 +1,49 @@
...
+
+It can be found in the ``/contrib/utilities/inlet_profile_generator`` folder of the Lethe repository.
+
+To run the script you just need to have `SymPy <https://www.sympy.org/en/index.html>`_ installed in your Python environment. You can install it via pip:
+
+.. code-block:: bash
+    :class: copy-button
+
+    pip install sympy
+
```
blaisb: I am not sure we need to provide the sympy install instruction. Especially since on ubuntu machines this does not really work like this anymore and you need to have your virtual environment not to conflict with system packages. I would remove the installation instruction and just say you just need to have Sympy and Matplotlib.

doc/source/tools/inlet_profile_generator/inlet_profile_generator.rst
```diff
@@ -0,0 +1,49 @@
...
+* ``--center``: Coordinates of the center of the inlet. Default is ``0. 0.``.
+* ``--length``: Length associated to the inlet shape. If the shape is a rectangle, use ``a,b``, where ``a`` is the length in the x direction and ``b`` is the length in the y direction. If the shape is a circle, give the diameter of. Default is ``1. 1.`` for rectangle and ``1.`` for the circle.
+* ``show``: Used for plotting the generated profile obtained by the script. If ``True`` plots of velocity as a function of ``x`` at center ``y``` and velocity as a function of ``y``` at center ``x`` are shown. Default is ``False``.
+
+Running the script is as simple as:
+
+.. code-block:: bash
+    :class: copy-button
+
+    python3 inlet_profile_generator.py --polynomial-degree 4 --shape circle --flow-rate 2 --center 0. 0. --length 1. --show True
```
blaisb:
```suggestion
    python3 inlet_profile_generator.py --polynomial-degree 4 --shape circle --flow-rate 2 --center 0. 0. --length 1. --show 
```

After my comment this would be the new command

contrib/utilities/inlet_profile_generator/inlet_profile_generator.py
blaisb: When I think about it, don't you think naming the command plot would also be better?

blaisb (changes requested):
Just a very very minor comments. Great work! Tested it it works super well

blaisb:
For the x-y plot I would also consider making a plot along the diagonal. But really it's nice. Tested it with polynomial order 20 and the results is noice.

doc/source/tools/inlet_profile_generator/inlet_profile_generator.rst
```diff
@@ -0,0 +1,49 @@
...
+
+The script has the following arguments:
+
+* ``--polynomial-degree``: Degree of the polynomial to be generated. Default is ``2```. To prevent negative flow velocities, we always use even degrees.
+* ``--shape``: Shape of the inlet. Supported shapes are ``rectangle`` and ``circle``. Default is rectangle.
+* ``--flow-rate``: Volumetric flow rate to be imposed at the inlet. Default is ``1```.
+* ``--center``: Coordinates of the center of the inlet. Default is ``0. 0.``.
+* ``--length``: Length associated to the inlet shape. If the shape is a rectangle, use ``a,b``, where ``a`` is the length in the x direction and ``b`` is the length in the y direction. If the shape is a circle, give the diameter of. Default is ``1. 1.`` for rectangle and ``1.`` for the circle.
+* ``show``: Used for plotting the generated profile obtained by the script. If ``True`` plots of velocity as a function of ``x`` at center ``y``` and velocity as a function of ``y``` at center ``x`` are shown. Default is ``False``.
+
```
brunaccampos, line 23: I would add a comment about the use in 2D/3D cases, Could we use the rectangle case with one of the dimensions = 0? Or add the dimension as an input parameter?

brunaccampos (commented):
Nice and useful script!

doc/source/tools/inlet_profile_generator/inlet_profile_generator.rst
voferreira: I added a parameter for dimensions. The script now works for both cases :)

blaisb:
I'll wait for the review of @brunaccampos , but on my side this is good. thanks :)!. I'm already using it this is super useful.

contrib/utilities/inlet_profile_generator/inlet_profile_generator.py
```diff
@@ -0,0 +1,154 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+import sympy as sp
+import argparse
+from math import ceil
+
+# Parse arguments
+parser = argparse.ArgumentParser(description="Generate inlet polynomial velocity profile with given flow rate")
+parser.add_argument("--dimensions", "-d", type=int, default=2, help="Number of spatial dimensions (1 or 2)")
+parser.add_argument("--polynomial-degree", '-pd', type=int, default=2, help="Degree k of the polynomial shape (k>=0)")
```
brunaccampos, line 11: I would just add that 1 and 2D refer to the inlet dimension and not the whole problem dimension (as we discussed)

brunaccampos (approved):
Good to go on my side :)

PR #1609, opened 2025-08-07
MortarManagerBase/CouplingOperator: enable 3D

peterrum:
Example:

Screenshot from 2025-08-07 16-09-16

PR #1669, opened 2025-09-08
Mortar: Compute n_subdivisions and radius in 3D case

brunaccampos:
### Description

After PR #1609 , the function `compute_n_subdivisions_and_radius` in the `mortar_coupling_manager` had to be updated to be compatible with 3D cases. This PR fixes this.

### Solution

The arguments returned by the function `compute_n_subdivisions_and_radius` have been updated. Instead of `int n_subdivisions`, it now returns the number of subdivisions in the radial and axial directions. Also, instead of returning `double radius`, it now returns the radius and the length in the axial direction.

### Testing

Previous tests `rotate_mapping` and `rotate_mapping_3d` have been updated.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/core/mortar_coupling_manager.cc
+++ b/source/core/mortar_coupling_manager.cc
@@ -571,7 +571,8 @@ compute_n_subdivisions_and_radius(
 
                               const auto aux = cross_product_3d(
                                 (v - mortar_parameters.center_of_rotation),
-                                mortar_parameters.rotation_axis);
+                                mortar_parameters.rotation_axis /
+                                  mortar_parameters.rotation_axis.norm());
                               const double num = aux.norm();
                               const double den =
                                 mortar_parameters.rotation_axis.norm();


source/core/mortar_coupling_manager.cc
```diff
@@ -440,24 +440,84 @@ MortarManagerBase<dim>::get_config(const Point<dim> &face_center,
...
+      // First check if the vector has a unit norm
+      bool is_unit_axis =
+        mortar_parameters.rotation_axis.norm() == 1 ? true : false;
+
+      // Now check if the vector represents the x, y, or z directions
+      // specifically (we assume those are the only options for now)
+      for (unsigned int i = 0; i < dim; i++)
+        if (mortar_parameters.rotation_axis[i] != 0. &&
+            mortar_parameters.rotation_axis[i] != 1.)
+          is_unit_axis = false;
```
antonellasuccar, line 483: Why do we do the first verification if only x, y and z directions are allowed? Is it for later?

source/core/mortar_coupling_manager.cc
```diff
@@ -493,6 +566,9 @@ compute_n_subdivisions_and_radius(
                           // is the rotation axis
                           if constexpr (dim == 3)
                             {
+                              vertex_min = std::min(vertex_min, v[direction]);
+                              vertex_max = std::max(vertex_max, v[direction]);
+
                               const auto aux = cross_product_3d(
                                 (v - mortar_parameters.center_of_rotation),
                                 mortar_parameters.rotation_axis);
```
antonellasuccar: I would maybe add /1 or the norm of the rotation direction, just to remember it's there, in case someone will change something. Although the rotation direction should mostly be a unit direction.

source/core/mortar_coupling_manager.cc
brunaccampos: This is because the verification for subdivisions in the radial direction assumes that the rotation axis is either x, y, or z, so I added the assert just to make sure. In the future the goal is to generalize it!

PR #1661, opened 2025-09-04
Fix issue by attaching dof_handler to DataOut instance + refactor structs to include cell data

voferreira:
### Description

The output_field_hook functions were not working due to the last refactoring of the write output results, PR #1624. The reason is that we were not attaching a default DoF handler to the DataOut instance. Additionally, the inclusion of cell fields, mainly the subdomain field, was done without the struct.

### Solution

Call attach_dof_handler upon instantiating DataOut.
Refactor output_field_hook function to use the new struct.
Add a variant of the struct for the case of cell data.

### Testing

This was first observed when @blaisb was running a sharp example. I tested the "Sedimentation of One Particle" sharp example, and it worked just fine without the hotfix of PR #1658.

### Documentation

No need. Just a fix.

### Miscellaneous (will be removed when merged)

This solves the issue related to it, but we want to also add a sharp validation case to both validate the code and test the output results.

### Checklist (will be removed when merged)
See [this page](https://chaos-polymtl.github.io/lethe/documentation/contributing.html#pull-requests) for more information about the pull request process.

Code related list:
- [x] All in-code documentation related to this PR is up to date (Doxygen format)
- [x] Copyright headers are present and up to date
- [x] Lethe documentation is up to date
- [x] Fix has unit test(s) (preferred) or application test(s), and restart files are in the generator folder
- [x] The branch is rebased onto master
- [x] Changelog (CHANGELOG.md) is up to date
- [x] Code is indented with indent-all and .prm files (examples and tests) with prm-indent

Pull request related list:
- [x] Labels are applied
- [ ] There are at least 2 reviewers (or 1 if small feature) excluding the responsible for the merge
- [ ] If this PR closes an issue or is related to a project, it is linked in the "Projects" or "Development" section
- [ ] If the fix is temporary, an issue is opened
- [x] The PR description is cleaned and ready for merge

PR #1678, opened 2025-09-13
Refactor output multiphysics to involve output struct

voferreira:
### Description

This is the last PR of the output refactoring started in #1624 and fixed in #1661. It includes using the new output struct to replace the previous "attach output" to DataOut pattern.

### Testing

Launched Damn break example, cooling fin example, and herringbone micromixer tracer and results match with master.

### Documentation

No documentation required.

The next step is to implement the output on patch mesh. I was doing it in this PR, but decided to break the refactoring and the new feature into two different PRs.

Code related list:

Pull request related list:
- [ ] Labels are applied


include/core/interface_tools.h
```diff
@@ -9,6 +9,8 @@
 #include <core/utilities.h>
 #include <core/vector.h>
 
+#include <solvers/output_struct.h>
```
voferreira, line 12: @blaisb Do you think it would make more sense for the output struct to be in the core folder instead?

blaisb: It would be moved to core the moment you start using it for DEM I think

include/core/interface_tools.h
```diff
@@ -968,6 +969,9 @@ namespace InterfaceTools
     /// Solution vector of the signed distance (write only)
     LinearAlgebra::distributed::Vector<double> signed_distance;
 
+    /// Solution vector of the signed distance as used for output only
+    GlobalVectorType signed_distance_output;
```
blaisb, line 973: I'm not sure I understand, why do you need a GlobalVectorType for the output? Why not use directly the LinearAlgebra::distributed::Vector solutions below?
Is the limitation that the output must all be of the same vector type? If so, what happens when you do multiphysics with the matrix-free solver where all the physics are Trillinos vectors and the Navier-Stokes are regular vectors.

Or is the limitation that all of the output vector of a single physics must be of the same type?

include/core/parameters.h
```diff
@@ -1864,6 +1864,13 @@ namespace Parameters
       {
         return vector_of_string;
       }
+    else
+      {
+        AssertThrow(false,
+                    ExcMessage("Error: convert_string_to_vector only works for "
+                               "int, double and string types."));
+      }
```
blaisb:
```suggestion

        AssertThrow(false,
                    ExcMessage("Error: convert_string_to_vector only works for "
                               "int, double and string types."));
      
```

If you have reached this point there is nothing to be done, so there's no need for an if statement.

include/solvers/auxiliary_physics.h
```diff
@@ -7,11 +7,13 @@
 #include <core/parameters.h>
 #include <core/physics_solver.h>
 
+#include <solvers/output_struct.h>
 #include <solvers/simulation_parameters.h>
 
 #include <deal.II/dofs/dof_handler.h>
 
 #include <deal.II/numerics/data_out.h>
+#include <deal.II/numerics/data_out_resample.h>
```
blaisb: Is this include needed in this PR?

source/core/interface_tools.cc
```diff
@@ -1389,13 +1394,28 @@ InterfaceTools::SignedDistanceSolver<dim, VectorType>::output_signed_distance(
...
+  std::vector<std::string> solution_names(1, "signed_distance");
+  std::vector<DataComponentInterpretation::DataComponentInterpretation>
+    solution_data_component_interpretation(
+      1, DataComponentInterpretation::component_is_scalar);
+  OutputStructSolution<dim, GlobalVectorType> output_struct(
+    this->dof_handler,
+    this->signed_distance_output,
+    solution_names,
+    solution_data_component_interpretation);
+  solution_output_structs.push_back(output_struct);
```
blaisb: You could emplace it back directly instead of making it and pushing it back.

source/solvers/cahn_hilliard.cc
```diff
@@ -284,34 +284,40 @@ CahnHilliard<dim>::copy_local_rhs_to_global_rhs(
...
-
   std::vector<DataComponentInterpretation::DataComponentInterpretation>
     data_component_interpretation(
       2, DataComponentInterpretation::component_is_scalar);
 
-  data_out.add_data_vector(dof_handler,
-                           present_solution,
-                           solution_names,
-                           data_component_interpretation);
+  solution_output_structs.emplace_back(
```
blaisb, line 301: This is a better approach I think

include/solvers/auxiliary_physics.h
voferreira: Nope. Thanks!

include/solvers/multiphysics_interface.h
```diff
@@ -180,23 +181,45 @@ class MultiphysicsInterface
...
       {
-        iphys.second->attach_solution_to_output(data_out);
+        std::vector<OutputStruct<dim, GlobalBlockVectorType>> output_structs =
+          iphys.second->gather_output_hook();
+        for (auto &output_struct : output_structs)
+          solution_output_structs.push_back(output_struct);
       }
+
+    return solution_output_structs;
   }
```
antonellasuccar, line 223: Could this be replaced by one templated function?

include/core/interface_tools.h
voferreira: I think the limitation is more at the single physics level. As we talked, I need to return a vector of output structs with the same vector type currently. Though, since we don't have a templated VectorType and almost all physics are GlobalVectorType, I thought it would be cleaner to just convert it here.

include/solvers/multiphysics_interface.h
```diff
@@ -19,6 +19,7 @@
 #include <core/vector.h>
 
 #include <solvers/auxiliary_physics.h>
+#include <solvers/output_struct.h>
```
AmishgaAlphonius: Already in `solvers/auxiliary_physics.h>`
```suggestion
```

include/solvers/multiphysics_interface.h
voferreira: As we discussed, yes, but we would need to template the entire class. I was trying to avoid doing it here.

include/solvers/output_struct.h
```diff
@@ -4,7 +4,7 @@
 #ifndef lethe_output_struct_h
 #define lethe_output_struct_h
 
-#include <solvers/postprocessors.h>
+#include <deal.II/numerics/data_postprocessor.h>
```
AmishgaAlphonius, line 7: maybe change the comment starting at `line 104` for:
```
/**
 * @brief Variant handler of the three output structs (OutputStructPostprocessor,
 * OutputStructSolution and OutputStructCellVector).
 * This is used to allow the output of both postprocessors with their respective
 * DoF handlers, and the straightforward cell data struct in a single vector of
 * OutputStruct.
 */
```

AmishgaAlphonius (approved):
Great work! :)
I've also tested it with a few examples.

brunaccampos (approved):
Nice work!

PR #1687, opened 2025-09-18
Added AssertThrow in the read_checkpoint function in the DEM

OGaboriault:
### Description
If the checkpoint controller does not exist while reading a restart at the beginning of a simulation, no error message was being written and the simulation was crashing. This was annoying when you are copy-pasting a .prm file that already exist, change some parametesr manually and then start your new simulation while forgetting that the `restart` parameter is still at `true`. (Aka, me losing 10 minutes this morning wondering why my prm was not working) 

Now, with the new `AssertThrow`, the user is informed that the checkpoint_controller is not found, thus he know that we are trying to restart a simulation. If he doesn't want to restart, he knows right away what to do. On the other hand, if he wants to restart, well he knows there is a problem. 

Code related list:

Pull request related list:
- [X] Labels are applied

### DIFF
--- a/include/dem/read_checkpoint.h
+++ b/include/dem/read_checkpoint.h
@@ -43,4 +43,20 @@ read_checkpoint(
   std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solid_surfaces,
   CheckpointControl &checkpoint_controller);
 
+/**
+ * @brief Asserts that a file exists. If the file does not exist, an exception is
+ * @param[in] f Object associated with the file to check. The file
+ * should be opened before passing it to this function.
+ * @param[in] filename Name of the file that is not found.
+ *
+ */
+inline void
+assert_restart_file_exists(std::ifstream &f, const std::string &filename)
+{
+  AssertThrow(f.is_open(),
+              ExcMessage(
+                std::string("You are trying to restart a previous computation, "
+                            "but the restart file <") +
+                filename + "> does not appear to exist!"));
+}
 #endif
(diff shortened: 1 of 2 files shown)


include/dem/read_checkpoint.h
```diff
@@ -43,4 +43,20 @@ read_checkpoint(
   std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solid_surfaces,
   CheckpointControl &checkpoint_controller);
 
+/**
+ * @brief Asserts that a file exists. If the file does not exist, an exception is
+ * @param[in] f Object associated with the file to check. The file
+ * should be opened before passing it to this function.
+ * @param[in] filename Name of the file that is not found.
```
blaisb:
```suggestion
 * @param[in] f FIle stream of the file. The file
 * should already be open when it is passed to this function.
 * @param[in] filename Name of the file.
```

include/dem/read_checkpoint.h
```diff
@@ -43,4 +43,20 @@ read_checkpoint(
...
+ */
+inline void
+assert_restart_file_exists(std::ifstream &f, const std::string &filename)
+{
+  AssertThrow(f.is_open(),
+              ExcMessage(
+                std::string("You are trying to restart a previous computation, "
+                            "but the restart file <") +
+                filename + "> does not appear to exist!"));
+}
```
blaisb: Instead you could use directly the deal.II exception no?
https://dealii.org/current/doxygen/deal.II/group__Exceptions.html#ga2f31207e832b937688f75c04e939cc21

Then I am not sure you need the function.

blaisb (commented):
I'm not sure if it would not be better to juste use the deal.II default exception for files?

include/dem/read_checkpoint.h
OGaboriault: Good point. I notice that we weren't consistent regarding this in the code.

blaisb: I'd rather use the deal.II exceptions when we can, but I (and many others) forget about them sometimes
It's life, we get better as we go go

blaisb (approved):

PR #1692, opened 2025-09-24
Mortar: use constraints_extended in all functions

peterrum:
This should fix some ghosting issues.

@brunaccampos Could you try it out in your multigrid code?


brunaccampos (approved):
I've tested with a couple examples and different refinement levels. This fixes the constraints issue!

blaisb (approved):

PR #1696, opened 2025-09-28
Fix bug with sparsity pattern memory overflowing due to the mortar coupling interface

blaisb:
### Description

The mortar coupling interface uses an additional sparsity pattern which is stored as a member of the solver. This sparsity pattern is copied from the dynamic sparsity pattern and, I think, is reused to update the coupling interface at every time-step. The issue is that the call "copy_form(dsp)" would make the solver run out of memory on clusters when using more than 100M degrees of freedom or so.

### Solution

Right now I have encapsulated so that it runs only when the mortar coupling interface is used. However, this is a temporary solution and there is no reason why the mortar interface should induce such a deadset memory limitation. @brunaccampos We should fix this in a coming PR. Do we still really need this sparsity pattern?

### Testing

I have tested it in the cluster and now my simulations works just fine with 100M+ cells.

### Documentation

Nothing special to document here.

Code related list:

Pull request related list:
- [ ] Labels are applied


brunaccampos:
Indeed using the `sparsity_pattern` was not necessary! I remove it from `matrix_based` and also the Poisson mortar tests.

PR #1704, opened 2025-10-03
Clean up includes

OGaboriault:
Removed unnecessary includes from the code.

### DIFF
--- a/include/fem-dem/vans_assemblers.h
+++ b/include/fem-dem/vans_assemblers.h
@@ -28,9 +28,9 @@ class ParticleFluidAssemblerBase
 {
 public:
   /**
-   * @brief calculate_particle_fluid_interactions calculates the particle-fluid interaction for all of the particles.
+   * @brief Calculates the particle-fluid interaction for all of the particles.
    * The forces calculated on the particles is stored within the fem_force. The
-   * overall beta coefficient for the cell, whcih is defined as the drag
+   * overall beta coefficient for the cell, which is defined as the drag
    * coefficient divided by the cell volume, is also calculated within this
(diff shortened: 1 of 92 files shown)


blaisb:
I think you missed a few includes there :)

blaisb:
I think the DEBUG CI crashing is a deal.II update and not related to your PR.

include/fem-dem/vans_assemblers.h
```diff
@@ -698,11 +691,10 @@ class VANSAssemblerFPI : public NavierStokesAssemblerBase<dim>
 };
 
 /**
- * @brief Calculate gamma grad-div stabilization constant for the VANS equations
+ * @brief Calculate gamma grad-div stabilization constant for the VANS equations.
  *
  * @param[in] velocity Magnitude of the velocity at the quadrature point
  * @param[in] kinematic_viscosity
- * @param[in] h The element size
```
blaisb, line 705: Are you sure that this renders well in the doxygen? (does doxygen really ignores the fact that the third parameter is not used?
I think it might be better to leave h and add [[maybe_unused]] for the time being or just refactor the function in a follow-up PR.

OGaboriault: I don't know if it renders well. CLion just warned me that the parameter wasn't used.

OGaboriault: I will put it back

blaisb (approved):

blaisb:
@OGaboriault Could you please add a CHANGELOG and a PR description that is a bit more thorough? I will then merge :)

PR #1712, opened 2025-10-10
dpg electromagnetic solver in parallel

OresteMarquis:
### Description

This is the current version of the EM-DPG solver in parallel that I used to produce the figures of convergence with Matthias meeting on October 9 2025. Do you think I should change the format to follow the same nomenclature as the other prototype (e.g., ScratchData,  CopyData) before merging and starting the implementation in Lethe?

### Testing

I ran convergence analysis for DG0, DG1 and DG2.

### Documentation

I don't think I need doc for this prototype, but we can start too think of examples! Do I need to add the changelog and copyright headers?

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/prototypes/emdpg_waveguide/EM_waveguide.cc
+++ b/prototypes/emdpg_waveguide/EM_waveguide.cc
@@ -1,3 +1,6 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2025 The Lethe Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
 /**
  * @brief Necessary includes for the implementation of the DPG method for the
  * time-harmonic maxwell equations.


prototypes/emdpg_waveguide/parameters.prm
```diff
@@ -0,0 +1,54 @@
+# Listing of Parameters
+# ---------------------
+subsection Parameters
+
+
+  # relative permittivity of medium
+  set epsilon r = {1.,0.}
```
blaisb: The {} are for complex numbers? nice

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+    return result;
+  }
+
+  // We create a class to handle the input parameters.
+  template <int dim>
+  class Parameters : public ParameterAcceptor
+  {
+  public:
+    // CONSTRUCTOR
```
blaisb:
```suggestion
    // Constructor
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    const Parameters<dim>     &parameters;
+    const double               k;
+    const double               k_x;
+    const double               k_y;
+    const double               k_z;
+    const double               k_c;
+    const std::complex<double> Z_r;
+  };
+
+  template <int dim>
```
blaisb, line 316: I would add a comment here to the effect that the solution for E is
Real(E_x), Real(E_y), Real(E_z), Im(E_x), Im(E_y), Im(E_z).

This makes sense, but it is worth to repeat it.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                                        dofs_per_cell_trial_interior);
+
+    LAPACKFullMatrix<double> tmp_matrix2(dofs_per_cell_trial_skeleton,
+                                         dofs_per_cell_trial_skeleton);
+
+    LAPACKFullMatrix<double> tmp_matrix3(dofs_per_cell_trial_skeleton,
+                                         dofs_per_cell_test);
+
+    Vector<double> tmp_vector(dofs_per_cell_trial_interior);
+    Vector<double> tmp_vector2(dofs_per_cell_test);
```
blaisb: I know these are temporary vectors, but is there a way you can give them a more descriptive name since they don't have the same size. Just something to distinguish tmp_vector and tmp_vector2

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+            G_matrix     = 0;
+            B_matrix     = 0;
+            B_hat_matrix = 0;
+            l_vector     = 0;
+
+            // We also need to reinitialize the $M_1$ condensation matrix
+            // between each iteration on cell to get rid of its inverse status.
+            M1_matrix = 0;
+
+            // Reset the dofs relationships containers
```
blaisb: Can you define what these containers aim to do. Like just a short paragraph here will enhance the readabilityu.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+            // We fill the dofs relationship containers at the cell level. To do
+            // so, we first loop on the test space dofs.
+            for (unsigned int i : fe_values_test.dof_indices())
+              {
+                // Get the information on which element the dof is
+                const unsigned int current_element_test_i =
+                  fe_test.system_to_base_index(i).first.first;
+
+                // Fill the l vector relationship
```
blaisb: What is the 1 vector? Please be more descriptive here.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                if ((current_element_test_i == 0) ||
+                    (current_element_test_i == 1))
+                  {
+                    l_F.emplace_back(i);
+                  }
+
+                // Loop over the dofs test a second time to fill the dofs
```
blaisb:
```suggestion
                // Loop over the test dofs a second time to fill the dofs
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                         (current_element_test_i == 3)) &&
+                        ((current_element_test_j == 2) ||
+                         (current_element_test_j == 3)))
+                      {
+                        G_II.emplace_back(i, j);
+                      }
+                  }
+
+                // Then we loop over the dofs trial space to fill the dofs
+                // relationship for the B matrix (bilinear form)
```
blaisb:
```suggestion
                // Then we loop over the trial space dofs to fill the dofs
                // relationship for the B matrix (bilinear form)
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    // B_hat matrix for the bilinear form and relationships between skeleton
+    // trial and test
+    std::vector<std::pair<unsigned int, unsigned int>> B_hat_FE;
+    std::vector<std::pair<unsigned int, unsigned int>> B_hat_IE;
+    std::vector<std::pair<unsigned int, unsigned int>> B_hat_FH;
+
+    // l vector for the linear form and relationships between test functions
+    std::vector<unsigned int> l_F;
+
+    // Reserve memory to avoid reallocations of each matrix
```
blaisb: These are not matrices right, these are vectors of pairs. I would refer to them as vectors here.

```suggestion
    // Reserve memory to avoid reallocations of each relationship vectors
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                      imag * fe_values_trial_interior[extractor_E_imag].value(
+                               i, q_point);
+                    H[i] =
+                      fe_values_trial_interior[extractor_H_real].value(
+                        i, q_point) +
+                      imag * fe_values_trial_interior[extractor_H_imag].value(
+                               i, q_point);
+                  }
+
+                // Now we loop on each relationship container to assemble the
```
blaisb, line 1401: That's pretty nice.... I actually think it's quite an efficient implementation...

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+        locally_relevant_solution_interior  = locally_owned_solution_interior;
+        locally_relevant_residual_estimator = locally_owned_residual_estimator;
+      }
+    else
+      {
+        system_matrix.compress(VectorOperation::add);
+        system_rhs.compress(VectorOperation::add);
+      }
+  }
+
```
blaisb, line 1886: I have to say that's impressive work @OresteMarquis ... Good job seriously

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  Time_Harmonic_Maxwell_DPG<dim>::solve_skeleton()
+  {
+    pcout << "*--- Solving system ---*" << std::endl;
+
+    TimerOutput::Scope t(computing_timer, "solve");
+
+    LA::MPI::Vector completely_distributed_solution(
+      locally_owned_dofs_trial_skeleton, mpi_communicator);
+    SolverControl solver_control(dof_handler_trial_skeleton.n_dofs(),
+                                 1e-10 * system_rhs.l2_norm());
```
blaisb:
```suggestion
    SolverControl solver_control(dof_handler_trial_skeleton.n_dofs(),
                                 std::max(1e-10 * system_rhs.l2_norm(),1e-12));
```
Always try to keep a buffer so you don't specify an overly small tolerance either

blaisb (approved):
Just a few comments. I think this implementation will be ready for the full Lethe implementation.
Great work Oreste, this is really good.

prototypes/emdpg_waveguide/parameters.prm
OresteMarquis: I think its an error from my side!

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+namespace LA
+{
+  using namespace dealii::LinearAlgebraTrilinos;
+} // namespace LA
+namespace EM_DPG
+{
+  using namespace dealii;
+
+  // This is a helper function that maps a 3D tensor to a 2D face in
+  // H^-1/2(curl).
```
hepap: I don't really mind for a prototype, but if you want you, yon can add the doxygen briefs

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    double frequency;
+    int    mode_x;
+    int    mode_y;
+
+    // Mesh parameters
+    unsigned int              n_refinements;
+    std::vector<unsigned int> n_repetitions;
+    unsigned int              initial_refinements;
+    bool                      adaptive_refinement;
+    int                       max_refinement_level;
```
hepap:
```suggestion
    unsigned int                       max_refinement_level;
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  }
+
+  /* Create analytical solution class for the electric field (E).
+   * The corresponding solution for the electric field follows the physicist's
+   * time convention with the imaginary unit in the exponent being negative
+   * (i.e. exp(-i wt) ). The function manage both the real and complex
+   * components of the solution. Since the frequency, the material properties,
+   * and the geometry are arguments for the parameter file, they are used in the
+   * constructor to initialize the different wavenumber. Note that everything is
+   * addimensionalize using the following : \tilde{\mathbf{E}} =
```
hepap:
```suggestion
   * adimensionalize using the following : \tilde{\mathbf{E}} =
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    degree = 0;
+    add_parameter("degree", degree, "order of the DGQ finite element space");
+
+    delta_degree = 1;
+    add_parameter("delta degree",
+                  delta_degree,
+                  "order difference between the trial and test space");
+  }
+
+  /* Create analytical solution class for the electric field (E).
```
hepap:
```suggestion
  /* Create analytical solution class for the electric field (\f$\mathbf{E}\f$).
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+   * \frac{\mathbf{E}}{E_0}, \tilde{\mathbf{H}} = \frac{Z_0}{E_0} \mathbf{H},
+   *       \varepsilon_r = \frac{\varepsilon}{\varepsilon_0},
+   *       \mu_r = \frac{\mu}{\mu_0},
+   *       Z_r = \frac{Z}{Z_0},
+   *       \sigma_r = \frac{\sigma}{\omega \varepsilon_0},
+   *       \tilde{\omega} =  \frac{\omega}{k_0 c_0},
+   *       \tilde{\mathbf{J}} = \frac{1}{\omega \varepsilon E_0 } \mathbf{J}
+   *       \tilde{k} = L \mathbf{k},
+   *       \tilde{\nabla} = L \nabla.
+   */
```
hepap: if you want the math expressions to appear correctly in the doxygen, you need to add them between \f$ \f$.

hepap: But I don't think it's necessary for the prototype  :)

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    // Constructor
+    AnalyticalSolution_H(const Parameters<dim> &parameters)
+      : Function<dim>(6)
+      , parameters(parameters)
+      , k(2. * M_PI * parameters.frequency / parameters.speed_of_light)
+      , k_x(parameters.mode_x * M_PI / parameters.waveguide_a)
+      , k_y(parameters.mode_y * M_PI / parameters.waveguide_b)
+      , k_z(std::sqrt(k * k - k_x * k_x - k_y * k_y))
+      , k_c(std::sqrt(k_x * k_x + k_y * k_y))
+    {}
```
hepap, line 399: Is it the same constructor as the solution for E? If so, you could have a base class and both solutions inherit from it with only the value function changing.

Maybe not for the prototype, but when it will be in lethe!

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+        .imag();
+    else if (component == 5)
+      return (factor * std::cos(k_x * p[0]) * std::cos(k_y * p[1])).imag();
+    else
+      throw std::runtime_error(
+        "Too many components for the analytical solution");
+  }
+
+  // Main class for the DPG solver. The method rely on multiple DoFHandler and
+  // FESystem to manage the different functional spaces. The DoFHandlers that we
```
hepap:
```suggestion
  // Main class for the DPG solver. The method relies on multiple DoFHandlers and
  // FESystems to manage the different functional spaces. The DoFHandlers that we
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  // rely on are the following:
+  // - The <code>dof_handler_trial_interior</code> is for the unknowns in the
+  // interior of the cells;
+  // - The <code>dof_handler_trial_skeleton</code> is for the unknowns in the
+  // skeleton;
+  // - The <code>dof_handler_test</code> is for the test functions. Although we
+  // do not use the unknowns associated with this DoFHandler, it enables us to
+  // evaluate the test function we will use in DPG.
```
hepap:
```suggestion
  // - The <code>dof_handler_test</code> is for the test functions. Although we
  // do not use the unknowns associated with this DoFHandler, it enables us to
  // evaluate the test functions we will use in DPG.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+  // The same applies for the three FESystem:
```
hepap:
```suggestion
  // The same applies for the three FESystems:
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  template <int dim>
+  class Time_Harmonic_Maxwell_DPG : public ParameterAcceptor
+  {
+  public:
+    Time_Harmonic_Maxwell_DPG(const Parameters<dim> &parameters);
+    void
+    run();
+
+  private:
+    // <code>make_grid</code> creates the initial mesh.
```
hepap: If you do the brief, you don't need to repeat the name of the function

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    // skeleton unknowns. This is achieved by local condensation.
+    // - When <code>solve_interior = true</code> the system is assembled and the
+    // skeleton degrees of freedom solution is used to
+    // reconstruct the interior solution and the built-in error estimator right
+    // after.
+    void
+    assemble_system(bool solve_interior = false);
+
+    // <code>solve_linear_system_skeleton</code> solves the linear system of
+    // equations for the skeleton degree of freedom.
```
hepap:
```suggestion
    // equations for the skeleton degrees of freedom.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    void
+    solve_skeleton();
+
+    // <code>output_results</code> write the skeleton and the interior unknowns
```
hepap:
```suggestion
    // <code>output_results</code> writes the skeleton and the interior unknowns
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+    // The vector container for the energy norm on each cell used for adaptive
+    // refinement.
+    Vector<double> estimated_error_per_cell;
+
+    // A container for the $L^2$ error and other related quantities.
+    ConvergenceTable error_table;
+
+    // Extractors that will be used at multiple places in the class to
+    // select the relevant components for the calculation. Those are created
```
hepap:
```suggestion
    // select the relevant components for the calculations. Those are created
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  // In the constructor, we assign the relevant finite element to each FESystem
+  // following the nomenclature described above in the class description:
+  // - <code>fe_system_trial_interior</code> contains $\Re(\mathbf{E})$,
+  // $\Im(\mathbf{E})$, $\Re(\mathbf{H})$, $\Im(\mathbf{H})$ ;
+  // - <code>fe_system_trial_skeleton</code> contains $\Re(\hat{E}_\parallel)$,
+  // $\Im(\hat{E}_\parallel)$, $\Re(\hat{H}_\parallel)$,
+  // $\Im(\hat{H}_\parallel)$;
+  // - <code>fe_system_test</code> contains $\Re(\mathbf{F})$,
+  // $\Im(\mathbf{F})$, $\Re(\mathbf{I})$, $\Im(\mathbf{I})$.
+  // We also initialize probleme dependent parameters and parallel related
```
hepap:
```suggestion
  // We also initialize problem dependent parameters and parallel related
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    , extractor_H_imag(3 * dim)
+  {
+    AssertThrow(dim == 3,
+                ExcMessage("The implementation only works for dim==3."));
+
+    AssertThrow(parameters.delta_degree >= 1,
+                ExcMessage("The delta_degree needs to be at least 1."));
+
+    AssertThrow(k > 0,
+                ExcMessage("The wavenumber must be positive, its the "
```
hepap:
```suggestion
                ExcMessage("The wavenumber must be positive, it is the "
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                           "magnitude of the wave vector."));
+  }
+
+  // Here the <code>make_grid</code> function creates the initial rectangular
+  // mesh using a structured grid using
+  // dealii::GridGenerator::subdivided_hyper_rectangle.
```
hepap:
```suggestion
  // mesh using a structured grid
  // dealii::GridGenerator::subdivided_hyper_rectangle.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+        // We added the possibility to limit the maximum refinement level.
+        for (auto &cell : triangulation.active_cell_iterators())
+          {
+            if (cell->is_locally_owned())
+              {
+                if (cell->level() >= parameters.max_refinement_level &&
+                    cell->refine_flag_set())
+                  cell->clear_refine_flag();
+              }
+          }
```
hepap, line 736: I don't know if you need it, but you don't limit the coarsening level?

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+    // We initialize vectors to store the locally owned solution
+    LA::MPI::Vector locally_owned_solution_interior(
+      locally_owned_dofs_trial_interior, mpi_communicator);
+    LA::MPI::Vector locally_owned_residual_estimator(locally_owned_dofs_test,
+                                                     mpi_communicator);
+
+    // We first define quadrature rules and related variables. Since the
+    // quadrature formula should be the same for both trial and test FE and that
+    // the test space have higher polynomial degree than the others by
```
hepap:
```suggestion
    // the test space have an higher polynomial degree than the others by
```

prototypes/emdpg_waveguide/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+#     DEAL_II_WITH_UMFPACK        = ${DEAL_II_WITH_UMFPACK}
+#     DEAL_II_WITH_COMPLEX_VALUES = ${DEAL_II_WITH_COMPLEX_VALUES}
+# This conflicts with the requirements."
+#     )
+# endif()
+
+deal_ii_initialize_cached_variables()
+set(CLEAN_UP_FILES *.log *.vtu *.pvtu)
+project(${TARGET})
+deal_ii_invoke_autopilot()
```
Julamipolymtl: Now that this is a prototype and not a standalone project, shouldn't the `CMakeLists.txt` be changed so hat it is properly built with other prototypes during a Lethe install ? I think other prototypes' CMakesLists are simply something of the sort:
```
add_executable(project_name project_name.cc)
deal_ii_setup_target(project_name)
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    std::vector<Tensor<1, dim, std::complex<double>>> H(
+      dofs_per_cell_trial_interior);
+    std::vector<Tensor<1, dim, std::complex<double>>> E_hat(
+      dofs_per_cell_trial_skeleton);
+    std::vector<Tensor<1, dim, std::complex<double>>> n_cross_E_hat(
+      dofs_per_cell_trial_skeleton);
+    std::vector<Tensor<1, dim, std::complex<double>>> n_cross_H_hat(
+      dofs_per_cell_trial_skeleton);
+
+    // Also, to avoid multiple if calls to understand where assemble each term
```
hepap:
```suggestion
    // Also, to avoid multiple if calls to understand where we assemble each term
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    locally_relevant_residual_estimator.reinit(locally_owned_dofs_test,
+                                               locally_relevant_dofs_test,
+                                               mpi_communicator);
+  };
+
+  // Here we assemble the system matrix and right-hand side and apply the Robin
+  // boundary conditions during the assembly.
+  template <int dim>
+  void
+  Time_Harmonic_Maxwell_DPG<dim>::assemble_system(const bool solve_interior)
```
hepap, line 963: I think all the initialization here makes it a bit heavy to read. Do you plan to use a structure for that in Lethe? Maybe like ScratchData if it is suitable?

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    const std::complex<double> ikZr      = imag * k * Z_r;
+    const std::complex<double> conj_ikZr = conj(ikZr);
+    const std::complex<double> ikepsilon_Zr =
+      imag * k / Z_r * (1. - imag * parameters.sigma_r / parameters.epsilon_r);
+    const std::complex<double> conj_ikepsilon_Zr = conj(ikepsilon_Zr);
+    const std::complex<double> kz_kZr            = k_z / (k * Z_r);
+    const std::complex<double> conj_kz_kZr       = conj(kz_kZr);
+
+    // We also create objects to store the incident fields that will be used to
+    // compute the excitation for the robin boundary condition.
```
hepap:
```suggestion
    // compute the excitation for the Robin boundary condition.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    // conditions that rely on AffineConstraints
+    void
+    setup_system();
+
+    // The <code>assemble_system</code> assembles both the right-hand side and
+    // the system matrix. This function is used twice per problem and it has
+    // two functions.
+    // - When <code>solve_interior = false</code> the system is assembled and is
+    // locally condensed such that the resulting system only contains the
+    // skeleton unknowns. This is achieved by local condensation.
```
Julamipolymtl: Repeated information
```suggestion
    // - When <code>solve_interior = false</code> the system is assembled and is
    // locally condensed such that the resulting system only contains the
    // skeleton unknowns.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+  };
+
+  // DPG constructor
+  // In the constructor, we assign the relevant finite element to each FESystem
+  // following the nomenclature described above in the class description:
+  // - <code>fe_system_trial_interior</code> contains $\Re(\mathbf{E})$,
+  // $\Im(\mathbf{E})$, $\Re(\mathbf{H})$, $\Im(\mathbf{H})$ ;
+  // - <code>fe_system_trial_skeleton</code> contains $\Re(\hat{E}_\parallel)$,
+  // $\Im(\hat{E}_\parallel)$, $\Re(\hat{H}_\parallel)$,
+  // $\Im(\hat{H}_\parallel)$;
```
Julamipolymtl: Here does `\parallel` have something to do with the fact that we are in mpi or is it simply used as a symbol to represent the skeleton ?

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+    // Waveguide geometry
+    double waveguide_a; // Width of the waveguide
+    double waveguide_b; // Height of the waveguide
+    double waveguide_length;
+
+    // Frequency and excitation modes
+    double frequency;
+    int    mode_x;
+    int    mode_y;
```
Julamipolymtl: Very picky but couldn't this also be unsigned int? is there such a thing as a mode -N?
```suggestion
    unsigned int    mode_x;
    unsigned int    mode_y;
```

prototypes/emdpg_waveguide/EM_waveguide.cc
Julamipolymtl: Small typo
```suggestion
    // the test space has a higher polynomial degree than the others by
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+      {
+        error_table.add_value("n_iter", n_CG_iter);
+      }
+  }
+
+  // This function is standard and is in charge of outputting the
+  // solution to a file that can be visualized with Paraview or Visit (VTU and
+  // PVTU format). However, because the skeleton solution lives only on the mesh
+  // faces, we made use of the DataOutFaces class to output it properly. Note
+  // that the subdomains are only outputed in the interior solution file.
```
Julamipolymtl:
```suggestion
  // that the subdomains are only outputted in the interior solution file.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                // Then we compute the residual estimator and also map it to the
+                // global residual estimator vector.
+                B_matrix.vmult(tmp_vector2, cell_interior_solution);
+                B_hat_matrix.vmult_add(tmp_vector2, cell_skeleton_solution);
+                l_vector -= tmp_vector2;
+                G_matrix.vmult(cell_residual, l_vector);
+                cell_test->distribute_local_to_global(
+                  cell_residual, locally_owned_residual_estimator);
+              }
+            // If the flag solves interior is set to false, we have to compute
```
Julamipolymtl:
```suggestion
            // If the flag solve_interior is set to false, we have to compute
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+    // Output the skeleton solution
+    data_out_faces.add_data_vector(locally_relevant_solution_skeleton,
+                                   solution_skeleton_names,
+                                   DataOutFaces<dim>::type_automatic,
+                                   data_component_interpretation_skeleton);
+
+    data_out_faces.build_patches(fe_trial_skeleton.degree);
+    data_out_faces.write_vtu_with_pvtu_record(
+      "./output/", "solution-face_waveguide", cycle, mpi_communicator, 3);
```
Julamipolymtl, line 2062: Ran as is, if `/output` does not yet exist, it is not created and writing to it returns an error.

prototypes/emdpg_waveguide/parameters.prm
OresteMarquis: Actually its a mistake hahaha, we don't need the brackets

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    B_FE.reserve(dofs_per_cell_trial_interior * dofs_per_cell_test);
+    B_IE.reserve(dofs_per_cell_trial_interior * dofs_per_cell_test);
+    B_FH.reserve(dofs_per_cell_trial_interior * dofs_per_cell_test);
+    B_IH.reserve(dofs_per_cell_trial_interior * dofs_per_cell_test);
+
+    B_hat_FE.reserve(dofs_per_cell_trial_skeleton * dofs_per_cell_test);
+    B_hat_IE.reserve(dofs_per_cell_trial_skeleton * dofs_per_cell_test);
+    B_hat_FH.reserve(dofs_per_cell_trial_skeleton * dofs_per_cell_test);
+
+    l_F.reserve(dofs_per_cell_test);
```
hepap, line 1102: To improve readability, you could put the declaration and allocation together, ex:

`std::vector<std::pair<unsigned int, unsigned int>> G_FF(dofs_per_cell_test * dofs_per_cell_test)`

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+    LAPACKFullMatrix<double> M1_matrix(dofs_per_cell_trial_interior,
+                                       dofs_per_cell_trial_interior);
+    LAPACKFullMatrix<double> M2_matrix(dofs_per_cell_trial_interior,
+                                       dofs_per_cell_trial_skeleton);
+    LAPACKFullMatrix<double> M3_matrix(dofs_per_cell_trial_skeleton,
+                                       dofs_per_cell_trial_skeleton);
+    LAPACKFullMatrix<double> M4_matrix(dofs_per_cell_trial_interior,
+                                       dofs_per_cell_test);
+    LAPACKFullMatrix<double> M5_matrix(dofs_per_cell_trial_skeleton,
+                                       dofs_per_cell_test);
```
hepap: Can you add a comment for the definition of those 5 matrices, because the name doesn't tell it?

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+            B_FE.clear();
+            B_IE.clear();
+            B_FH.clear();
+            B_IH.clear();
+
+            l_F.clear();
+
+            // We fill the dofs relationship containers at the cell level. To do
+            // so, we first loop on the test space dofs.
```
hepap, line 1259: In which case the relationship would change between cells?

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                      }
+                  }
+              }
+
+            // Now we loop over all quadrature points of the cell
+            for (unsigned int q_point = 0; q_point < n_q_points; ++q_point)
+              {
+                // To avoid unnecessary computation, we fill the shape values
+                // containers for the real and imaginary parts of the electric
+                // and magnetic fields and the Dof relationship at the current
```
hepap:
```suggestion
                // and magnetic fields and the dof relationship at the current
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                // To do so, we first loop on the test space dofs.
+                for (unsigned int i : fe_face_values_test.dof_indices())
+                  {
+                    // Get the information on which element the dof is
+                    const unsigned int current_element_test_i =
+                      fe_test.system_to_base_index(i).first.first;
+
+                    // If we are at either boundary 4 or 5, we are on the Robin
+                    // B.C. and the load vector and the B_hat matrix will have a
+                    // contribution in addition to a modification of the riesz
```
hepap:
```suggestion
                    // contribution in addition to a modification of the Riesz
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                                 (current_element_trial_j == 1)))
+                              {
+                                B_hat_IE.emplace_back(i, j);
+                              }
+                          }
+                      }
+                    else
+                      {
+                        // If not on Robin B.C., assemble all the other
+                        // relevantskeleton terms
```
hepap:
```suggestion
                        // relevant skeleton terms
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                        n_cross_I_face_conj[i] = cross_product_3d(
+                          normal,
+                          fe_face_values_test[extractor_H_real].value(i,
+                                                                      q_point) -
+                            imag * fe_face_values_test[extractor_H_imag].value(
+                                     i, q_point));
+                      }
+
+                    // Then, similarly we loop over the trial dofs to fill the
+                    // face values containers. Note that the to be in
```
hepap:
```suggestion
                    // face values containers. Note that to be in
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                                .value(i, q_point) +
+                              imag *
+                                fe_face_values_trial_skeleton[extractor_H_imag]
+                                  .value(i, q_point),
+                            normal));
+                      }
+
+                    // Here we get the excitation at the inlet port. To do so we
+                    // force the electromagnetic fields with our analytical
+                    // solution at the boundary.
```
hepap:
```suggestion
                    // force the electromagnetic fields of our analytical
                    // solution at the boundary.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+            // can perform at the same time the residual estimator (\Psi =
+            // G^{-1}(l-B u_h
+            // - \hat{B}\hat{u}_h)).
+            if (solve_interior)
+              {
+                // We first get the solution vector for this cell.
+                cell_skeleton->get_dof_values(
+                  locally_relevant_solution_skeleton, cell_skeleton_solution);
+
+                // Then we do the matrix-vector product to obtain the interior
```
hepap:
```suggestion
                // Then we do the matrix-vector products to obtain the interior
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                l_vector -= tmp_vector2;
+                G_matrix.vmult(cell_residual, l_vector);
+                cell_test->distribute_local_to_global(
+                  cell_residual, locally_owned_residual_estimator);
+              }
+            // If the flag solves interior is set to false, we have to compute
+            // the local matrix and the local RHS for the condensed system.
+            else
+              {
+                // So the cell matrix is obtained with the formula $(M_3 -
```
hepap:
```suggestion
                // The cell matrix is obtained with the formula $(M_3 -
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+        locally_relevant_residual_estimator = locally_owned_residual_estimator;
+      }
+    else
+      {
+        system_matrix.compress(VectorOperation::add);
+        system_rhs.compress(VectorOperation::add);
+      }
+  }
+
+  // The solve function is in charge of solving the linear system assembled and
```
hepap:
```suggestion
  // The solve function is in charge of solving the assembled linear system and
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+      data_component_interpretation(
+        4 * dim, DataComponentInterpretation::component_is_part_of_vector);
+
+    // Output the solution
+    data_out.add_data_vector(locally_relevant_solution_interior,
+                             solution_interior_names,
+                             DataOut<dim>::type_automatic,
+                             data_component_interpretation);
+
+    // Out the subdomains
```
hepap:
```suggestion
    // Output the subdomains
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+            for (const unsigned int q_index :
+                 fe_values_trial_interior.quadrature_point_indices())
+              {
+                const double JxW      = fe_values_trial_interior.JxW(q_index);
+                const auto  &position = quadrature_points[q_index];
+
+                // Loop over dimensions to compute each component of the error
+                // because for a vector field is simply the sum of the squared
+                // scalar L² errors of each component.
```
hepap:
```suggestion
                // scalar L2 errors of each component.
```

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+                for (unsigned int i = 0; i < dim; ++i)
+                  {
+                    // Calculate the L2 error for E
+                    L2_error_E_real +=
+                      pow((local_E_values_real[q_index][i] -
```
hepap, line 2242: you can maybe use fixed_power

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2451 @@
...
+
+                fe_face_values_trial_skeleton[extractor_H_real]
+                  .get_function_values(locally_relevant_solution_skeleton,
+                                       local_H_hat_values_real);
+                fe_face_values_trial_skeleton[extractor_H_imag]
+                  .get_function_values(locally_relevant_solution_skeleton,
+                                       local_H_hat_values_imag);
+
+                // The energy norm only has a face contribution the face where
+                // we applied a Robin Boundary condition, i.e., the inlet port
```
hepap:
```suggestion
                // The energy norm only has a face contribution at the face where
                // we applied a Robin Boundary condition, i.e., the inlet port
```

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: True, but I am not sure I would need those class in Lethe, they are only used for convergence. Maybe for a validation test?

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: Its the parallel component of the electric field at a given face.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: At the moment, the adaptive refinement doesn't work so I will not add this functionality.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: @blaisb Do you think it is worth to do this in the prototype or should i do it only in Lethe implementation?

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: The dimension reserved is only the maximum theoretical size that can be filling the relationship between the test function. This would only occur if the test space is based on a unique FE, which is not our case. Therefore, we don't know how many dofs relationship will be in stored in those vector in advance and consequently, the reserve is more appropriate I think since it keep their size to zero until we store them with the exact number of relation needed.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: This is a L, for load hahah, I will change it.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: @blaisb DO you know if there is a major difference between pow and fixed_pow, both are dealii function.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2643 @@
...
+                for (const auto &[i, j] : B_IH)
+                  {
+                    B_matrix(i, j) -= (iwmu_r * H[j] * I_conj[i] * JxW).real();
+                  }
+
+                for (const auto &i : l_F)
+                  {
+                    l_vector[i] += 0.0;
+                  }
+              }
```
Julamipolymtl, line 1461: Looking at step 81, I believe the .real() part of the matrix is taken a bit too early. To stick to step 81's approach, I think it should be M1.real(), but since `M1 = B_CT*G_inv*B`, there are a few more terms coming from the complex product of the submatrices. Plus your conjugate transpose on B would be well defined here.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2643 @@
...
+            // Finally, after having assembled all the matrices and vectors, we
+            // build the condensed version of the system.
+
+            // We only need the inverse of the Gram matrix $G$, so we
+            // invert it.
+            G_matrix.invert();
+
+            // We construct $M_4 = B^\dagger G^{-1}$ and $M_5 = \hat{B}^\dagger
+            // G^{-1}$ with it:
+            B_matrix.Tmmult(M4_matrix, G_matrix);
```
Julamipolymtl, line 1786: In your documentation, this conjugate transpose of B (\dagger) is used, but since here it was already turned into a real valued matrix the conjugate transpose was the same definition as the transpose. treating B as complex valued would stick more closely to your documentation.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2643 @@
...
+              {
+                // We first get the solution vector for this cell.
+                cell_skeleton->get_dof_values(
+                  locally_relevant_solution_skeleton, cell_skeleton_solution);
+
+                // Then we do the matrix-vector products to obtain the interior
+                // unknowns.
+                M2_matrix.vmult(tmp_vector_interior, cell_skeleton_solution);
+                M4_matrix.vmult(cell_interior_rhs, l_vector);
+                cell_interior_rhs -= tmp_vector_interior;
```
Julamipolymtl, line 1817: Based on step 81, I think you should take the real part of the right hand side. If both `M4` and `l` are technically complex matrices, their complex product will create a few more terms than simply the product of their real parts.

prototypes/emdpg_waveguide/EM_waveguide.cc
```diff
@@ -0,0 +1,2643 @@
...
+                // matrix so we can perform the distribution to the global
+                // system below.
+                cell_matrix = tmp_matrix_M2M1M2;
+
+                // Then we compute the cell RHS using $(M_5 -
+                // M_2^\dagger M_1^{-1} M_4)l -
+                // G$.
+                tmp_matrix_M2M1.mmult(tmp_matrix_M2M1M4, M4_matrix);
+                M5_matrix.add(-1.0, tmp_matrix_M2M1M4);
+                M5_matrix.vmult(cell_skeleton_rhs, l_vector);
```
Julamipolymtl, line 1856: same as above

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: Actually you are right, I can precompute the relationship only once as long has we don't have p-refinement because the dofs relationship don't change. Maybe this would be useful if there is a lot of cell (I haven't tested if for the face but my guess it's that it is the same). I will try to implement it when doing the official version in Lethe.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: I cannot use B as complex without having dealii build with complex on, which we want to avoid.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: When everything is transform into real component, the complex component product are taken into account directly within the structure of the matrix.

prototypes/emdpg_waveguide/EM_waveguide.cc
OresteMarquis: We need to take the .real() to build the matrix in real form. The relationships between complex and real value when performing multiplication is obtain by using the complex form of test and trial shape functions which encodes everything in the matrix.

prototypes/emdpg_waveguide/EM_waveguide.cc
Julamipolymtl: True plus after our discussions, once the system is decomplexified, the transpose of the decomplexified B matrix seems to have the same definition as the conjugate transpose of the complex B matrix.

Julamipolymtl (approved):
After a few (lots) of dicussions and math proofs with @OresteMarquis, I believe that the current implementation is the proper way to do it! Good job it's an amazing job!

prototypes/emdpg_waveguide/EM_waveguide.cc
blaisb: Fixed pow is fixed at compile time, so it is always more efficient than the pow

PR #1721, opened 2025-10-14
Explicit add the velocity divergence within the Tracer DG formulation to account for non-divergence free velocity field.

voferreira:
### Description

During tests with the tracer in the herringbone micromixer, Bruno noticed that the DG tracer would have big jumps at hanging nodes located at sharp corners. To solve the problem, he added a velocity divergence term to the weak form, which compensantes for the jump.
One thing we observed is that the new formulation is more conservative when we use a lower CFL (1 or less). This was observed, for example, when running the dg_tracer_pipe application test, which had a CFL of approximately 6 and presented a concentration undershooting, reaching a min tracer concentration of -11. 

### Testing

This has been tested with the herringbone micromixer and the application tests existent.
dg_tracer_pipe application was sensitive to the change and was adapted accordingly. The time-step of the test used to be 0.1 and now it is 0.025, changing the CFL from 6.4 to 1.6.

### Documentation

A comment on the need for lower CFL was added accordingly.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/solvers/tracer_scratch_data.cc
+++ b/source/solvers/tracer_scratch_data.cc
@@ -103,11 +103,11 @@ TracerScratchData<dim>::calculate_physical_properties()
           throw std::runtime_error(
             "Unsupported number of fluids (2) with the tracer physics.");
           // In this case, we need both density and viscosity
-          // const auto diffusivity_models =
-          //   properties_manager.get_tracer_diffusivity_vector();
-          //
-          // diffusivity_models[0]->vector_value(fields, tracer_diffusivity_0);
-          // diffusivity_models[1]->vector_value(fields, tracer_diffusivity_1);
+          const auto diffusivity_models =
+            properties_manager.get_tracer_diffusivity_vector();
+
+          diffusivity_models[0]->vector_value(fields, tracer_diffusivity_0);
+          diffusivity_models[1]->vector_value(fields, tracer_diffusivity_1);
 
           // TODO Incomplete at the present time because the tracer VOF
           // complete is not finished Blend the physical properties


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### [Master] - 2025-10-14
 
+### Changed
+
+- MAJOR Changed DG Tracer formulation to be - D * laplacian T +  u * gradT + T div(u) - f=0, adding the term T div(u). It has proven to be more conservative at edges. The dg_tracer_pipe application test was sensitive to this change, and will be used to track the change. The use of the new formulation seem more sensitive to the CFL to keep the solution closer to bounded.
```
```suggestion
- MAJOR Changed DG Tracer formulation to be - D * laplacian T +  u * gradT + T div(u) - f=0, adding the term T div(u). It has proven to be more conservative at edges. The dg_tracer_pipe application test was sensitive to this change, and will be used to track it. The use of the new formulation seems more sensitive to the CFL to keep the solution closer to bounded.
```

source/solvers/tracer_assemblers.cc
```diff
@@ -267,24 +267,28 @@ TracerAssemblerDGCore<dim>::assemble_matrix(
...
+          const auto phi_T_i      = scratch_data.phi[q][i];
+
 
           for (unsigned int j = 0; j < n_dofs; ++j)
             {
               const Tensor<1, dim> grad_phi_T_j = scratch_data.grad_phi[q][j];
               const auto           phi_T_j      = scratch_data.phi[q][j];
 
-              // Weak form : - D * laplacian T +  u * gradT - f=0
+              // Weak form : - D * laplacian T +  u * gradT + T div(u) - f=0
```
brunaccampos: maybe add a comment here about why T div(u) is added to the formulation

applications_tests/lethe-fluid/dg_tracer_pipe.output
brunaccampos, line 1: is this the only test for the tracer or the only one affected by the new term added? the CFL in other tests were already close to 1?

source/solvers/tracer_assemblers.cc
```diff
@@ -267,24 +267,28 @@ TracerAssemblerDGCore<dim>::assemble_matrix(
...
               // Note that the advection term has been weakened for it to appear
               // explicitly in the weak form as a boundary term.
               local_matrix(i, j) += (diffusivity * grad_phi_T_i * grad_phi_T_j -
-                                     grad_phi_T_i * velocity * phi_T_j) *
+                                     grad_phi_T_i * velocity * phi_T_j -
```
OresteMarquis: I am not sure, but the sign don't match here with (- D * laplacian T +  u * gradT + T div(u)), if I take the weak form I get ( D\*gradT_i\*gradT_j + u\*gradT_i\*T_j + div(u)\*T_i\* T_j). Also, if we multiply by a test function, the term T div(u) should have a phi_T_j no? Maybe it is because we are solving the equation in a non-linear way, I haven't check with the jacobian and everything.

source/solvers/tracer_assemblers.cc
```diff
@@ -327,9 +333,10 @@ TracerAssemblerDGCore<dim>::assemble_rhs(
           const auto phi_T_i      = scratch_data.phi[q][i];
           const auto grad_phi_T_i = scratch_data.grad_phi[q][i];
 
-          // rhs for : - D * laplacian T +  u * grad T - f=0
+          // rhs for : - D * laplacian T +  u * grad T - T div(u) - f=0
```
OresteMarquis: In the above you have + T div(u), which one is right?

OresteMarquis (changes requested):

applications_tests/lethe-fluid/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ file(COPY lpbf_benchmark_mesh.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY lpbf_benchmark_mesh_3D.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY cylinder_close_outlet.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 file(COPY jurins-law-2d-mesh-dimensioned.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY dg_tracer_sharp_edge_90_degrees_curve.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
blaisb, line 17: Can you add the .geo file to generate that mesh in the repository? It's always useful to keep the mesh generator.
If you made it with pointwise then it's ok, no need to add the pointwise file.

applications_tests/lethe-fluid/dg_tracer_pipe.output
```diff
@@ -17,65 +17,257 @@ Tracer flow rate at the boundaries:
...
+------------------
+Tracer flow rates
+------------------
+Tracer flow rate at the boundaries: 
+	 boundary 0: 0
+	 boundary 1: 0.0286279
+	 boundary 2: -0.0266201
+
+*******************************************************************************
+Transient iteration: 16       Time: 0.4      Time step: 0.025    CFL: 1.60881
```
blaisb, line 260: Is there a reason for simulating until 0.4s? Does that lead to steady-state or something in particular?

applications_tests/lethe-fluid/dg_tracer_pipe.prm
```diff
@@ -1,13 +1,12 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023 The Lethe Authors
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025 The Lethe Authors
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
 set dimension = 3
 
 subsection simulation control
   set method           = bdf1
-  set time step        = 0.1
+  set time step        = 0.025
```
blaisb, line 12: Could you add an paragraph of comment at the top that explains exactly what the test is doing, what the expected result is and what is the main feature the test is testing. We are trying to move to a point where the prm files of the tests are clearer in a way.

applications_tests/lethe-fluid/dg_tracer_sharp_edge_90_degrees_curve.prm
```diff
@@ -0,0 +1,101 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+set dimension = 2
+
+subsection simulation control
```
blaisb, line 11: Same as above, can you add a paragraph comment that describes what the test does, what is the expected results, what the test is testing. Something that describes what's the goal here.

doc/source/parameters/cfd/fem.rst
```diff
@@ -44,6 +44,10 @@ This subsection specifies the characteristics of the finite element method used
 
 * ``tracer uses dg`` specifies if the Discontinuous Galerkin (DG) formulation is used instead of the Continuous Galerkin (CG) that is used by default, for the tracer physics. This formulation allows discontinuities between elements and at boundaries, and only penalizes them; this prevents oscillations and enforces local conservation of mass, at the cost of more numerous degrees of freedom.
 
+.. important::
+
+    The DG formulation is sensitive to CFL. Use a small time-step to avoid over/undershoots. From our experience, a CFL of 1 or lower is recommended.
```
blaisb:
```suggestion
    The DG formulation is sensitive to the CFL value. Use a small time-step to  keep the tracer value bounded. From our experience, a CFL of 1 or lower is recommended.
```

source/solvers/tracer_assemblers.cc
```diff
@@ -267,25 +267,33 @@ TracerAssemblerDGCore<dim>::assemble_matrix(
...
           const auto grad_phi_T_i = scratch_data.grad_phi[q][i];
+          const auto phi_T_i      = scratch_data.phi[q][i];
 
           for (unsigned int j = 0; j < n_dofs; ++j)
             {
               const Tensor<1, dim> grad_phi_T_j = scratch_data.grad_phi[q][j];
               const auto           phi_T_j      = scratch_data.phi[q][j];
 
-              // Weak form : - D * laplacian T +  u * gradT - f=0
+              // Weak form: - D * laplacian T + u * gradT - f=0
```
blaisb:
```suggestion
              // Weak form of: - D * laplacian T + u * gradT - f=0
```

source/solvers/tracer_assemblers.cc
```diff
@@ -267,25 +267,33 @@ TracerAssemblerDGCore<dim>::assemble_matrix(
...
               // explicitly in the weak form as a boundary term.
               local_matrix(i, j) += (diffusivity * grad_phi_T_i * grad_phi_T_j -
                                      grad_phi_T_i * velocity * phi_T_j) *
                                     JxW;
+
+              // This term is added to correct for the influence of the
+              // non-divergence-free velocity field on the concentration. The
+              // term is introduced in the strong form as - div u * T.
+              local_matrix(i, j) +=
+                -phi_T_i * velocity_divergence * phi_T_j * JxW;
```
blaisb, line 295:
```suggestion
              // This term is added to correct for the influence of the
              // non-divergence-free velocity field on the concentration.
              local_matrix(i, j) +=
                -phi_T_i * velocity_divergence * phi_T_j * JxW;
```

It's not introduced in the strong form per say, it's because we have weakened the advection term.

source/solvers/tracer_assemblers.cc
```diff
@@ -327,11 +336,16 @@ TracerAssemblerDGCore<dim>::assemble_rhs(
...
+          // rhs for : - D * laplacian T + u * grad T - f=0
           local_rhs(i) -= (diffusivity * grad_phi_T_i * tracer_gradient -
                            grad_phi_T_i * velocity * tracer_value -
                            scratch_data.source[q] * phi_T_i) *
                           JxW;
+
+          // This term is added to correct for the influence of the
+          // non-divergence-free velocity field on the concentration. The term
+          // is introduced in the strong form as - div u * T.
+          local_rhs(i) -= -phi_T_i * velocity_divergence * tracer_value * JxW;
```
blaisb, line 346:
```suggestion
          // This term is added to correct for the influence of the
          // non-divergence-free velocity field on the concentration.
          local_rhs(i) -= -phi_T_i * velocity_divergence * tracer_value * JxW;
```

source/solvers/tracer_scratch_data.cc
```diff
@@ -101,11 +103,11 @@ TracerScratchData<dim>::calculate_physical_properties()
...
-          const auto diffusivity_models =
-            properties_manager.get_tracer_diffusivity_vector();
-
-          diffusivity_models[0]->vector_value(fields, tracer_diffusivity_0);
-          diffusivity_models[1]->vector_value(fields, tracer_diffusivity_1);
+          // const auto diffusivity_models =
+          //   properties_manager.get_tracer_diffusivity_vector();
+          //
+          // diffusivity_models[0]->vector_value(fields, tracer_diffusivity_0);
+          // diffusivity_models[1]->vector_value(fields, tracer_diffusivity_1);
```
blaisb: Why is this part of the code commented out? it should not be I think.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### [Master] - 2025-10-14
 
+### Changed
+
+- MAJOR Changed DG Tracer formulation to be - D * laplacian T +  u * gradT + T div(u) - f=0, adding the term T div(u). It has proven to be more conservative at edges. The dg_tracer_pipe application test was sensitive to this change, and will be used to track it. The use of the new formulation seems more sensitive to the CFL to keep the solution closer to bounded.
```
```suggestion
- MAJOR The DG Tracer formulation for the equation - D * Laplacian T +  u * gradT - f=0 is formulated in the weak sense. Consequently, we solve it by solving -D * Laplacian T + div(u*T) - f = 0 and taking the weak form of that expression. However, this assumes that the divergence of the velocity field is exactly zero, yet the divergence of the velocity field is not exactly zero if the velocity field arises from a CG formulation. This is especially true near corners or hanging nodes. To fix this, we explicitly account for the divergence of the velocity field and solve for -D * Laplacian T + div(u*T) -T*div(u) - f = 0. This essentially explicitly removes the term T div(u). It has proven to be more conservative at edges and corners. The dg_tracer_pipe application test was sensitive to this change, and will be used to track it. The use of the new formulation seems more sensitive to the CFL to keep the solution closer to its bounds.
```

I fixed the explanation. The way you were explaining it was not correct. We are not changing the formulation to be what you had mentioned because we weaken the full u*T term in DG.

applications_tests/lethe-fluid/CMakeLists.txt
voferreira: It was pointwise

applications_tests/lethe-fluid/dg_tracer_pipe.output
voferreira: Yeah, it was to get to steady-state. I just decreased the time step, that's why it is longer. But still it is a cheap simulation.

source/solvers/tracer_assemblers.cc
```diff
@@ -267,25 +267,32 @@ TracerAssemblerDGCore<dim>::assemble_matrix(
...
-              // Weak form : - D * laplacian T +  u * gradT - f=0
+              // Weak form of: - D * laplacian T + u * gradT - f=0
               // Note that the advection term has been weakened for it to appear
               // explicitly in the weak form as a boundary term.
               local_matrix(i, j) += (diffusivity * grad_phi_T_i * grad_phi_T_j -
                                      grad_phi_T_i * velocity * phi_T_j) *
                                     JxW;
+
+              // This term is added to correct for the influence of the
+              // non-divergence-free velocity field on the concentration.
```
OresteMarquis, line 293: I like that! Super clear :)

OresteMarquis (approved):
For me it is good to go !

blaisb (approved):

PR #1722, opened 2025-10-15
Remove a substantial amount of CLang Tidy V20 warnings. Mostly in the core library.

OGaboriault:
### Description

In our efforts to migrate from Clang Tidy v12 to Clang Tidy v20, this PR does the first batch of refactoring of the core library of lethe. The main things that is changed is that many loops over the dimensions used to use unsigned integer as indices and, when compared to the dimension (dim), this would raise a warning regarding comparison between signed and unsigned integer. This PR fixes that by using integers for these loops.

Other small things are to use enum with a smaller data footprint and some usage of range-based interators. Some spelling mistakes there and there were also fixed. Overall this PR removes half of the CLang V20 warnings.

### Testing

None of the tests results change.

Code related list:

Pull request related list:
- [ ] Labels are applied

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 All notable changes to the Lethe project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+### [Master] - 2025-10-14
+
+### Fixed
+
+- MINOR In our efforts to migrate from Clang Tidy v12 to Clang Tidy v20, this PR does the first batch of refactoring of the core library of lethe. The main things that is changed is that many loops over the dimensions used to use unsigned integer as indices and, when compared to the dimension (dim), this would raise a warning regarding comparison between signed and unsigned integer. This PR fixes that by using integers for these loops. Other small things are to use enum with a smaller data footprint and some usage of range-based interators. Some spelling mistakes there and there were also fixed. A flag was also added to the CMakeList (-std=c++20) to ensure that compilation still works on Apple Clang. Overall this removes half of the CLang V20 warnings. [#1722](https://github.com/chaos-polymtl/lethe/pull/1722)
+
+
 ## Release of Lethe v1.0.2 - 2025-10-15
 
 The Lethe v1.0.2 requires the latest major release of the deal.II library (deal.ii v9.7).


include/core/parameters_lagrangian.h
```diff
@@ -189,49 +189,89 @@ namespace Parameters
...
       initialize_containers(
-        std::unordered_map<unsigned int, double> &particle_average_diameter,
-        std::unordered_map<unsigned int, double> &particle_size_std,
-        std::vector<SizeDistributionType>        &distribution_type,
+        std::unordered_map<unsigned int, double> &p_average_diameter,
+        std::unordered_map<unsigned int, double> &p_size_std,
+        std::vector<SizeDistributionType>        &dist_types,
         std::unordered_map<unsigned int, std::vector<double>>
-          &particle_custom_diameter,
+          &p_custom_diameter,
```
blaisb, line 252: why is here p_ a prefix and for the other parameters it is _p a suffix?
I would make it uniform to be honest. Maybe _p for everyone and that's it.

include/core/parameters_lagrangian.h
```diff
@@ -545,7 +583,7 @@ namespace Parameters
       double grid_rotational_speed;
 
       // Rotational axis of rotating grid. Similar to deal.II, we use 0=x axis,
-      // 1=y axis, 2=z axis.
+      // 1=y-axis, 2=z axis.
```
blaisb: why is 1 now - axis?

include/core/dimensionality.h
```diff
@@ -78,7 +78,7 @@ namespace Parameters
      * Scaling laws. These are constants that are used to rescale quantities
      * arising frequently. Some of these quantities are not trivial to calculate
      * (e.g. thermal conductivity). Thus, it is better to pre-calculate these
-     * scalings and to reuse them instead of calculating them on the fly
+     * scaling and to reuse them instead of calculating them on the fly
      * everywhere.
```
blaisb, line 82: these scaling? We say these before, so there are multiple scalings not just one.

```suggestion
    * (e.g. thermal conductivity). Thus, it is better to pre-calculate these
     * scalings and to reuse them instead of calculating them on the fly
    * everywhere.
```

include/core/mobility_cahn_hilliard_model.h
```diff
@@ -271,7 +271,7 @@ class MobilityCahnHilliardModelQuartic : public MobilityCahnHilliardModel
 
   void
   vector_jacobian(const std::map<field, std::vector<double>> &field_vectors,
```
blaisb: Here should be fields_vectors if we wish to be consistent (field should take an s)

include/core/rheological_model.h
```diff
@@ -170,7 +170,7 @@ class RheologicalModel : public PhysicalPropertyModel
   /**
    * @brief is_non_newtonian_rheological_model Returns a boolean indicating if
    * the model is a non-Newtonian rheological model.
-   * @return Boolean value of if the model corresponds to a non newtonian
+   * @return Boolean value of if the model corresponds to a non-newtonian
```
blaisb:
```suggestion
   * @return Boolean value of if the model corresponds to a non-Newtonian
```

include/core/rheological_model.h
```diff
@@ -207,34 +207,34 @@ class Newtonian : public RheologicalModel
    *
    * @param p_kinematic_viscosity The constant newtonian viscosity
    */
-  Newtonian(const double &p_kinematic_viscosity)
+  explicit Newtonian(const double &p_kinematic_viscosity)
```
blaisb, line 210: Nice, I had never used that keyword

include/core/rheological_model.h
```diff
@@ -846,9 +854,9 @@ class PhaseChangeRheology : public RheologicalModel
    */
   virtual void
   get_dynamic_viscosity_for_stabilization_vector(
-    const double &p_density_ref,
-    const std::map<field, std::vector<double>> & /*field_vectors*/,
-    std::vector<double> &property_vector) override;
+    const double                               &p_density_ref,
+    const std::map<field, std::vector<double>> &field_vectors,
```
blaisb: I'm surprised this one was commented but does not need the [[maybe_unused]] here

include/core/serial_solid.h
```diff
@@ -112,7 +112,7 @@ class SerialSolid
     if constexpr (spacedim == 3)
       return this->current_translational_velocity;
 
-    if constexpr (spacedim == 2)
+    else
```
blaisb: If that's what clang tidy suggest why not. I guess it evaluates the full expression at compile time.

include/core/shape.h
```diff
@@ -80,7 +78,7 @@ class Shape : public AutoDerivativeFunction<dim>
    * @param position The position to set the shape at
    * @param orientation The orientation to set the shape at
    */
-  Shape(double              radius,
+  Shape(const double        radius,
```
blaisb: Are you sure that const is necessary? This is a pass by copy so by definition in the prototype of the function the fact that it's const does not signify anything.

include/core/rheological_model.h
OGaboriault: maybe I should add it actually...

include/core/parameters_lagrangian.h
OGaboriault: The warning I got here is that the class already has an attribute called ``particle_custom_diameter``, thus the input parameter of the method cannot be called the same thing as an attribute. The same thing goes for the other attributes and input parameters.

I will change all the input parameters so that they all start with the same prefix.

include/core/shape.h
OGaboriault: you are right, I think warning is made by the IDE and not Clang directly.

include/core/serial_solid.h
OGaboriault: Because he doesn't know that dim can only be equal to 2 or 3 ...

I can put is back, but I get this warning : 

``Not all control paths return a value"

I can put it back like it was. An other solution would be to add an ``assert`` or something for when dim is not equal to 2 or 3.  It is a yellow warning btw, not a red one.

source/core/pvd_handler.cc
```diff
@@ -37,10 +37,10 @@ PVDHandler::read(const std::string &prefix)
   for (unsigned int i = 0; i < size; ++i)
     {
       double      time;
-      std::string filename;
+      std::string filename_i;
       input >> time;
-      input >> filename;
-      append(time, filename);
+      input >> filename_i;
+      append(time, filename_i);
```
blaisb, line 43: Ahh ok just so that you don't double declare kinda. I get what you mean. Interesting,. It's not really a fault per se but i get the point,.

source/core/serial_solid.cc
```diff
@@ -255,55 +255,58 @@ SerialSolid<dim, spacedim>::get_triangulation()
...
+                                 &)
 {
   GridTools::rotate(angle, *solid_tria);
 }
 
 template <>
 void
-SerialSolid<2, 2>::rotate_grid(double angle, const Tensor<1, 3> /*axis*/)
+SerialSolid<2, 2>::rotate_grid(const double angle,
+                               const Tensor<1, 3> & /*axis*/)
```
blaisb:
```suggestion
                               [[maybe unused]] const Tensor<1, 3> & axis)
```
no?

source/core/serial_solid.cc
```diff
@@ -255,55 +255,58 @@ SerialSolid<dim, spacedim>::get_triangulation()
 
 template <>
 void
-SerialSolid<1, 2>::rotate_grid(const double angle, const Tensor<1, 3> /*axis*/)
+SerialSolid<1, 2>::rotate_grid(const double angle,
+                               const Tensor<1, 3> /*axis*/
```
blaisb: maybe unused instead here?

source/core/lethe_grid_tools.cc
```diff
@@ -594,7 +594,7 @@ LetheGridTools::project_to_d_linear_object(
   //                   \partial_m phi_j
   // we start at xi=(0.5, 0.5).
   Point<structdim> xi;
-  for (unsigned int d = 0; d < structdim; ++d)
+  for (unsigned int d = 0; std::cmp_less(d, structdim); ++d)
```
blaisb, line 597: Why not make d an int instead here?
I don't understand why we need the cmp_less (which is tailored made for two different type here....

source/core/lethe_grid_tools.cc
```diff
@@ -662,7 +662,7 @@ LetheGridTools::project_to_d_linear_object(
       normal[1]                = temp[0];
     }
   bool inside = true;
-  for (unsigned int i = 0; i < structdim; ++i)
+  for (int i = 0; i < structdim; ++i)
```
blaisb, line 665: You see here you just used int, that's much better no?

source/core/lethe_grid_tools.cc
```diff
@@ -1607,14 +1605,6 @@ LetheGridTools::find_point_triangle_distance(
...
               s *= inv_det;
               t *= inv_det;
-
-              // In region 0, normal vector is the face normal vector
-              // Cast unit_normal to a tensor<1, 3>
-              if constexpr (dim == 3)
-                normal_unit_3d = normal_unit;
-
-              if constexpr (dim == 2)
-                normal_unit_3d = tensor_nd_to_3d(normal_unit);
```
blaisb, line 1617: What happened to this equality? where is normal_unit_3d assigned  now?

blaisb:
Can you update the PR description @OGaboriault ?

source/core/pvd_handler.cc
OGaboriault: yes, It is because filename is already declared that I did that.

blaisb (approved):

PR #1726, opened 2025-10-16
Change all of the scratch data to use deal.II Tables instead of std::vector<std::vector....>>

blaisb:
### Description

The scratch data for all of the FEM-based physics used std::vector<type> for both the evaluation of the solution at the gauss point, but also to store the shape function (or their gradient, divergence, etc.) at the gauss points for the degrees of freedom. The issue with this is that it creates std::vector<std::vector<type>> data structures. These vectors of vectors are not contiguous in memory and may lead to a large number of cache misses as the degree of the polynomial of the shape function is increased. This PR fixes this by changing the data structure used to store the shape function (and their gradient, divergence, etc.) to use the deal.II Table<n_dim,type> data structure. These data structures are identical to vectors of vectors in the way they are used, but they store the information in a contiguous memory block (see https://dealii.org/current/doxygen/deal.II/classTableBase.html for more information). This change has minimal impact on the code, but should lead to much better scalability of the matrix and RHS assembly when the degree of the polynomial is increased and the matrix-based solvers are used. The improvement will be problem dependent, but it should never worsen performance.

### Testing

None of the test output should change, this is just a change in the data structure.

### Documentation

Nothing to document here.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/solvers/cahn_hilliard_scratch_data.h
+++ b/include/solvers/cahn_hilliard_scratch_data.h
@@ -382,7 +382,7 @@ public:
   std::vector<bool>         is_boundary_face;
   std::vector<unsigned int> boundary_face_id;
 
-  // First vector is face number, second quadrature point, third DOF
+  // First is face number, second quadrature point, third DOF
   Table<3, Tensor<1, dim>> grad_phi_face_phase;
   // First vector is face number, second quadrature point
   std::vector<std::vector<Tensor<1, dim>>> face_phase_grad_values;
--- a/include/solvers/heat_transfer_scratch_data.h
+++ b/include/solvers/heat_transfer_scratch_data.h
@@ -505,7 +505,7 @@ public:
 
   // First is face number, second quadrature point, third DOF
   Table<3, double> phi_face_T;
-  // First  is face number, second quadrature point
+  // First vector is face number, second quadrature point
   std::vector<std::vector<double>> temperature_face_value;
 };
 
--- a/include/solvers/navier_stokes_scratch_data.h
+++ b/include/solvers/navier_stokes_scratch_data.h
@@ -364,8 +364,7 @@ public:
           std::vector<std::vector<Tensor<1, dim>>>(
             n_faces, std::vector<Tensor<1, dim>>(n_faces_q_points));
 
-        this->face_normal = std::vector<std::vector<Tensor<1, dim>>>(
-          n_faces, std::vector<Tensor<1, dim>>(n_faces_q_points));
+        this->face_normal.reinit(n_faces, n_faces_q_points);
 
         this->face_quadrature_points = std::vector<std::vector<Point<dim>>>(
           n_faces, std::vector<Point<dim>>(n_faces_q_points));
@@ -1315,8 +1314,7 @@ public:
   std::vector<Tensor<1, dim>>              sdirk_stage_sum;
 
   // Shape functions
-  Table<2, double> div_phi_u;
-
+  Table<2, double>         div_phi_u;
   Table<2, Tensor<1, dim>> phi_u;
   Table<2, Tensor<3, dim>> hess_phi_u;
   Table<2, Tensor<1, dim>> laplacian_phi_u;
@@ -1440,9 +1438,9 @@ public:
   std::vector<unsigned int> boundary_face_id;
 
   // Quadrature
-  Table<2, double>                         face_JxW;
-  std::vector<std::vector<Point<dim>>>     face_quadrature_points;
-  std::vector<std::vector<Tensor<1, dim>>> face_normal;
+  Table<2, double>                     face_JxW;
+  std::vector<std::vector<Point<dim>>> face_quadrature_points;
+  Table<2, Tensor<1, dim>>             face_normal;
 
   // Velocity and pressure values
   // First vector is face number, second quadrature point
--- a/source/solvers/heat_transfer_scratch_data.cc
+++ b/source/solvers/heat_transfer_scratch_data.cc
@@ -16,7 +16,7 @@ HeatTransferScratchData<dim>::allocate()
   this->n_dofs     = fe_values_T.get_fe().n_dofs_per_cell();
 
   // Initialize arrays related to quadrature
-  this->JxW.reinit(TableIndices<1>(n_q_points));
+  this->JxW.reinit(n_q_points);
 
   // Forcing term array
   this->source = std::vector<double>(n_q_points);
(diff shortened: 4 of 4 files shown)


include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1344,13 +1315,14 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
...
   // Shape functions
-  std::vector<std::vector<double>>         div_phi_u;
-  std::vector<std::vector<Tensor<1, dim>>> phi_u;
-  std::vector<std::vector<Tensor<3, dim>>> hess_phi_u;
-  std::vector<std::vector<Tensor<1, dim>>> laplacian_phi_u;
-  std::vector<std::vector<Tensor<2, dim>>> grad_phi_u;
-  std::vector<std::vector<double>>         phi_p;
-  std::vector<std::vector<Tensor<1, dim>>> grad_phi_p;
+  Table<2, double> div_phi_u;
+
```
brunaccampos:
```suggestion
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1483,14 +1455,15 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
 
   // Shape functions
   // First vector is face number, second quadrature point, third DOF
```
brunaccampos, line 1455:
```suggestion
  // First is face number, second quadrature point, third DOF
```

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -1468,7 +1440,7 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
   std::vector<unsigned int> boundary_face_id;
 
   // Quadrature
-  std::vector<std::vector<double>>         face_JxW;
+  Table<2, double>                         face_JxW;
   std::vector<std::vector<Point<dim>>>     face_quadrature_points;
   std::vector<std::vector<Tensor<1, dim>>> face_normal;
```
brunaccampos: Could face_normal also be stored in a table?

include/solvers/cahn_hilliard_scratch_data.h
```diff
@@ -389,7 +383,7 @@ class CahnHilliardScratchData : public PhysicsScratchDataBase
   std::vector<unsigned int> boundary_face_id;
 
   // First vector is face number, second quadrature point, third DOF
```
brunaccampos:
```suggestion
  // First is face number, second quadrature point, third DOF
```

include/solvers/heat_transfer_scratch_data.h
```diff
@@ -507,9 +503,9 @@ class HeatTransferScratchData : public PhysicsScratchDataBase
   std::vector<unsigned int> boundary_face_id;
 
 
-  // First vector is face number, second quadrature point, third DOF
-  std::vector<std::vector<std::vector<double>>> phi_face_T;
-  // First vector is face number, second quadrature point
+  // First is face number, second quadrature point, third DOF
+  Table<3, double> phi_face_T;
+  // First  is face number, second quadrature point
```
brunaccampos:
```suggestion
  // First vector is face number, second quadrature point
```

include/solvers/heat_transfer_scratch_data.h
```diff
@@ -507,9 +503,9 @@ class HeatTransferScratchData : public PhysicsScratchDataBase
...
   std::vector<std::vector<double>> temperature_face_value;
```
brunaccampos, line 509: If temperature_face_value is still a vector<vector<double>>, no need to remove "vector" from the description. Or can it be stored in a table?

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -16,7 +16,7 @@ HeatTransferScratchData<dim>::allocate()
   this->n_dofs     = fe_values_T.get_fe().n_dofs_per_cell();
 
   // Initialize arrays related to quadrature
-  this->JxW = std::vector<double>(n_q_points);
+  this->JxW.reinit(TableIndices<1>(n_q_points));
```
brunaccampos: In the cahn_hillard_scratch_data this is just `this->JxW.reinit(n_q_points);`, why do we need to specify TableIndices<1> here? Is this because of the field type of heat transfer?

source/solvers/heat_transfer_scratch_data.cc
```diff
@@ -50,16 +50,13 @@ HeatTransferScratchData<dim>::allocate()
       maximum_number_of_previous_solutions(),
       std::vector<Tensor<1, dim>>(n_q_points));
 
+
```
brunaccampos:
```suggestion
```

include/solvers/tracer_scratch_data.h
```diff
@@ -537,7 +537,7 @@ class TracerScratchData : public PhysicsScratchDataBase
   double       cell_size;
 
   // Quadrature
-  std::vector<double>     JxW;
+  Table<1, double>        JxW;
```
brunaccampos, line 543: Since this was a vector<double>, and not a higher-dimensional array, is this change still improving the performance? or is it just to keep the same table pattern as in the other shape function related structures?

blaisb: That's a very good question. I have stored what could be stored in a Table in a Table. For single dimension table, the performance will be the same.
The only variables that cannot be stored in a table are the one that we update based on a solution (like when we call get_function_values) because these get_function_values and etc. require an argument which is an std::vector and not a Table, so this does not work. However, everything that is parsed on a one-by-one basis like the shapes is actually better as a Table :).

include/solvers/navier_stokes_scratch_data.h
blaisb: Yes, great point!

include/solvers/heat_transfer_scratch_data.h
blaisb: Since it is a value it cannot be stored in a table due to the fe_values interface :(

source/solvers/heat_transfer_scratch_data.cc
blaisb: No I think it's a mistake from my part. we don't need to specify the TableIndices here. It's just that TableIndices<1> and an integer must likely give he same results so the code are equivalent. I'll still fix it.

PR #1729, opened 2025-10-21
Kokkos prototypes: use MGTwoLevelTransferCopyToHost

peterrum:
Feature has been introduced in https://github.com/dealii/dealii/pull/18922.


blaisb:
I will review it soon too :)

blaisb (approved):
Wow that is indeed much shorter :)!
I'll test it on my machine :)

PR #1730, opened 2025-10-23
Add void fraction output at iteration 0

antonellasuccar:
### Description

In the CFD-DEM solver, the void fraction was not output at iteration 0. This is a very minor bug and this PR fixes it.

### Solution

The output was done in ``set_initital_condition()`` before initializing the void fraction. The ``set_initial_conditions`` was defined in ``NavierStokesBase``. A new version has been defined in FluidDynamicsVANS that skips the output. The solution is now explicitly written after the void fraction has been initialized.

### Testing

This has been tested by launching a simulation and visualizing the void fraction in the vtu files.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/fem-dem/cfd_dem_coupling.h
+++ b/include/fem-dem/cfd_dem_coupling.h
@@ -129,9 +129,10 @@ private:
    * @brief Set the initial condition. The version in this class does not output
    * the solution, since the void fraction is not initialized yet.
    *
-   * @param initial_condition_type Type of method  use to impose initial condition.
+   * @param[in] initial_condition_type Type of method  use to impose initial
+   *condition.
    *
-   * @param restart Indicator if the simulation is being restarted or not.
+   * @param[in] restart Indicator if the simulation is being restarted or not.
    *
    **/
 


include/fem-dem/fluid_dynamics_vans.h
```diff
@@ -109,6 +109,40 @@ class FluidDynamicsVANS : public FluidDynamicsMatrixBased<dim>
...
+   *
+   * @param initial_condition_type Type of method  use to impose initial condition.
+   *
+   * @param restart Indicator if the simulation is being restarted or not.
+   *
+   **/
+
+  void
+  set_initial_condition(
+    Parameters::FluidDynamicsInitialConditionType initial_condition_type,
```
blaisb:
```suggestion
    const Parameters::FluidDynamicsInitialConditionType initial_condition_type,
```

include/fem-dem/fluid_dynamics_vans.h
```diff
@@ -109,6 +109,40 @@ class FluidDynamicsVANS : public FluidDynamicsMatrixBased<dim>
...
+    bool                                          restart = false) override
```
blaisb:
```suggestion
    const bool                                          restart = false) override
```

include/fem-dem/fluid_dynamics_vans.h
blaisb: I would modify the base class to maintain consistency. But these two parameters should be const.

applications_tests/lethe-fluid-particles/restart_gas_solid_fluidized_bed.mpirun=2.output
```diff
@@ -9,6 +9,7 @@ DEM time-step is 3.65413% of Rayleigh time step
    Number of active cells:       2560
    Number of degrees of freedom: 16100
    Volume of triangulation:      0.00108
+---------------------------------------------------------------
```
blaisb, line 12: I am not sure I understand why this only prints a single line? Is it the void fraction calculation that print this line? Should it not print "Void Fraction"?

include/fem-dem/cfd_dem_coupling.h
```diff
@@ -125,6 +125,44 @@ class CFDDEMSolver : public FluidDynamicsVANS<dim>
...
+        ref_iter++;
+      }
+    while (
+      ref_iter <
+        (this->simulation_parameters.mesh_adaptation.initial_refinement + 1) &&
+      restart == false);
+
+    this->pcout
+      << "---------------------------------------------------------------"
+      << std::endl;
```
blaisb, line 164: I would use the announce_string here instead and have a more descriptive string than a --------
Maybe something like: Calculating initial condition or Calculating void fraction

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -1242,17 +1243,19 @@ template <int dim>
...
   if (this->simulation_control->is_output_iteration())
     {
       write_dem_output_results();
     }
+  if (!first_iteration)
+    {
+      this->pcout
+        << "---------------------------------------------------------------"
+        << std::endl;
+    }
```
blaisb, line 1258: What is this ------?
I am not sure this serves a purpose. Generally we announce the string when there's something that is calculated or that is done, not to separate iteration. I would remove this.

blaisb (changes requested):
A few things to fix. I don't think this "------------------" manual line is a good idea.
You can announce a string for the calculation of the void fraction when it is calculated if you want, but I would refrain from having these ----------------------- lines appear if they are not providing explicit information in between.

Could you also identify a second reviewer please? :)

include/fem-dem/cfd_dem_coupling.h
```diff
@@ -125,6 +125,44 @@ class CFDDEMSolver : public FluidDynamicsVANS<dim>
   void
   read_dem() override;
 
+  /**
+   * @brief Set the initial condition. The version in this class does not output
+   * the solution, since the void fraction is not initialized yet.
+   *
+   * @param initial_condition_type Type of method  use to impose initial condition.
+   *
+   * @param restart Indicator if the simulation is being restarted or not.
```
blaisb:
```suggestion
   * @param[in] initial_condition_type Type of method  use to impose initial condition.
   *
   * @param[in] restart Indicator if the simulation is being restarted or not.
```

PR #1743, opened 2025-10-29
Update slip BC for mortar

brunaccampos:
### Description

This PR changes the parameter `use_manifold_for_normal`, used in slip BCs, to `false` when mortar is enabled. This way only the mapping is used to compute normal vectors.

### Solution

The parameter `use_manifold_for_normal` has been explicitly declared when setting slip BCs, and it is set to `false` only when mortar is enabled.

### Testing

No new tests have been added. All previous tests should still pass.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/solvers/navier_stokes_base.cc
+++ b/source/solvers/navier_stokes_base.cc
@@ -1977,7 +1977,6 @@ NavierStokesBase<dim, VectorType, DofsType>::define_non_zero_constraints()
         {
           std::set<types::boundary_id> no_normal_flux_boundaries;
           no_normal_flux_boundaries.insert(id);
-
           VectorTools::compute_no_normal_flux_constraints(
             this->dof_handler,
             0,
@@ -2091,7 +2090,6 @@ NavierStokesBase<dim, VectorType, DofsType>::define_zero_constraints()
         {
           std::set<types::boundary_id> no_normal_flux_boundaries;
           no_normal_flux_boundaries.insert(id);
-
           VectorTools::compute_no_normal_flux_constraints(
             this->dof_handler,
             0,


source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -1468,12 +1468,17 @@ MFNavierStokesPreconditionGMGBase<dim>::reinit(
                   std::set<types::boundary_id> no_normal_flux_boundaries;
                   no_normal_flux_boundaries.insert(id);
 
+                  bool use_manifold_for_normal = true;
+                  if (this->simulation_parameters.mortar_parameters.enable)
+                    use_manifold_for_normal = false;
```
blaisb: Can you add a comment before that describes why you are doing this.
you and I both understand it, but it is important to explain it

source/solvers/navier_stokes_base.cc
```diff
@@ -1970,12 +1970,18 @@ NavierStokesBase<dim, VectorType, DofsType>::define_non_zero_constraints()
         {
           std::set<types::boundary_id> no_normal_flux_boundaries;
           no_normal_flux_boundaries.insert(id);
+
+          bool use_manifold_for_normal = true;
+          if (this->simulation_parameters.mortar_parameters.enable)
+            use_manifold_for_normal = false;
+
```
blaisb: Same as above

source/solvers/navier_stokes_base.cc
```diff
@@ -2082,12 +2088,18 @@ NavierStokesBase<dim, VectorType, DofsType>::define_zero_constraints()
         {
           std::set<types::boundary_id> no_normal_flux_boundaries;
           no_normal_flux_boundaries.insert(id);
+
+          bool use_manifold_for_normal = true;
+          if (this->simulation_parameters.mortar_parameters.enable)
+            use_manifold_for_normal = false;
```
blaisb: Same as above.
Makes me thing, maybe this should be a member of the class used everywhere to prevent confusiom?

brunaccampos: you mean a member of boundary_conditions maybe?

blaisb: No but a member of the Navier-Stokes base class.
I don't think it's a big deal though. What you did is just fine

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -1468,12 +1468,22 @@ MFNavierStokesPreconditionGMGBase<dim>::reinit(
                   std::set<types::boundary_id> no_normal_flux_boundaries;
                   no_normal_flux_boundaries.insert(id);
 
+                  // If mortar is enabled, we need to change one default
+                  // parameter of the function that computes normal vectors. In
+                  // this case, we cannot use the manifold to compute them,
+                  // since it leads to issues at the mortar interface. Hence we
+                  // use the mapping instead
```
blaisb:
```suggestion
                  // use the mapping instead.
```

source/solvers/navier_stokes_base.cc
```diff
@@ -1970,12 +1970,22 @@ NavierStokesBase<dim, VectorType, DofsType>::define_non_zero_constraints()
         {
           std::set<types::boundary_id> no_normal_flux_boundaries;
           no_normal_flux_boundaries.insert(id);
+
+          // If mortar is enabled, we need to change one default parameter of
+          // the function that computes normal vectors. In this case, we cannot
+          // use the manifold to compute them, since it leads to issues at the
+          // mortar interface. Hence we use the mapping instead
```
blaisb:
```suggestion
          // mortar interface. Hence we use the mapping instead.
```

source/solvers/navier_stokes_base.cc
```diff
@@ -2082,12 +2092,22 @@ NavierStokesBase<dim, VectorType, DofsType>::define_zero_constraints()
...
```
blaisb:
```suggestion
          // mortar interface. Hence we use the mapping instead.
```

blaisb (approved):
Just a few . missing

source/solvers/navier_stokes_base.cc
brunaccampos: Oh okay. It might be better indeed, though. Because the constraints are redefined at every transient iteration for mortar, so it could save us some time if this bool is only defined once in the class. I would need to define it both in the NS base and also once for the matrix-free operator, but still

PR #1745, opened 2025-10-29
Migrate Clang TIDY V20 to a custom docker image and self-hosted runners

blaisb:
This PR migrates the CLang V20 Tidy CI to use self-hosted runners. An additional feature is that the runner now uses a custom Docker image that I have made. This is significantly faster than before (10-15 minutes vs 2 hours). It can still be improved, but for now this is much better than what it was before.

### DIFF
--- a/.github/workflows/clang-tidy-v20.yml
+++ b/.github/workflows/clang-tidy-v20.yml
@@ -5,6 +5,7 @@ name: CI-Clang-tidy-v20
 
 
 on:
+  #Runs on pushes to master and on PRs, ignoring documentation and contrib changes
   push:
     branches: [ master ]
   pull_request:


.github/workflows/clang-tidy-v20.yml
```diff
@@ -3,63 +3,37 @@
 
 name: CI-Clang-tidy-v20
 
-on: 
+
+on:
   push:
-    # Runs on every push on master branch
```
OGaboriault, line 8: Can you put back the comments that you removed? If still applicable

blaisb: Done, I added a more descriptive comment

.github/workflows/clang-tidy-v20.yml
```diff
@@ -3,63 +3,37 @@
...
   pull_request:
     paths-ignore:
       - 'doc/**'
       - 'contrib/**'
 
-# Cancels running instances when a pull request is updated by a commit
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.run_id }}
-  cancel-in-progress: ${{github.event_name == 'pull_request'}}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
OGaboriault, line 18:
```suggestion
  cancel-in-progress: ${{github.event_name == 'pull_request'}}
```

OGaboriault: (If it doesn't make the CI crash)

blaisb: I think the spaces here are syntactic actually. We have them everywhere.

PR #1746, opened 2025-10-29
Dem - Triangle double contact

OGaboriault:
### Description
Contacts between particles and solid surfaces had a problem when the contact was occurring with many triangles at the same time. Contacts we being duplicated even if the solid surface represent a single surface. The logic used to delete invalid contacts is taken from : 

https://doi.org/10.1002/nme.4487 

### Testing

Three new application tests we made. Each takes 3 seconds in release. 

### Documentation

N/A

The tangential displacement could be passes from one triangle to an other is some cases. The same logic could be used on the domain boundary when using a cylindrical domain.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/applications_tests/lethe-particles/particle_solid_surface_CPES_double_edge_contact.prm
+++ b/applications_tests/lethe-particles/particle_solid_surface_CPES_double_edge_contact.prm
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
 # Test description:
 # In this test, a single particle is bouncing on a solid-surface made of two triangles that are coplanar (CP) and edge sharing (ES).
--- a/applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
+++ b/applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
 # Test description:
 # In this test, a single particle is bouncing on a solid-surface made of two triangles that are non-coplanar (NP) and edge sharing (ES).
--- a/applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
+++ b/applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
 # Test description:
 # In this test, a single particle is bouncing on a solid-surface made of two triangles that are non-coplanar (NP) and edge sharing (ES).


include/core/lethe_grid_tools.h
```diff
@@ -431,7 +438,56 @@ namespace LetheGridTools
...
+   */
+  template <int dim, int spacedim>
+  void
+  vertices_cell_mapping(
+    const Triangulation<dim, spacedim> &tria,
+    std::map<
+      unsigned int,
+      std::set<typename Triangulation<dim, spacedim>::active_cell_iterator>>
+      &vertices_cell_map);
+
```
OGaboriault, line 462: Maybe I could add a template to the previous vertices_cell_mapping function.

include/core/lethe_grid_tools.h
```diff
@@ -280,18 +291,19 @@ namespace LetheGridTools
    * @param n_particles_in_base_cell Number of particles in the base cell
    *
    * @return A tuple in which 0. a vector of bools to determine if the particle is
-   * close to the triangle plane, 1. a vector of projected location of particles
-   * on the triangle, 2. a vector of normal vectors of the triangles
+   * less than a radius away from the triangle plane, 1. a vector of projected
+   * location of particles on the triangle, 2. a vector of normal vectors of the
+   * triangles, 3. the type of contact between the particle and the triangle.
```
Julamipolymtl:
```suggestion
   * triangle, 3. a vector of the type of contact between the particle and the triangle.
```

include/core/lethe_grid_tools.h
```diff
@@ -431,7 +443,55 @@ namespace LetheGridTools
     }
   };
 
-} // namespace LetheGridTools
+  /**
+   * @brief Function that checks if two cells are coplanar within a given tolerance.
```
Julamipolymtl:
```suggestion
   * @brief Checks whether two cells are coplanar within a given tolerance.
```

include/core/lethe_grid_tools.h
```diff
@@ -431,7 +443,55 @@ namespace LetheGridTools
...
+  template <int dim, int spacedim>
+  bool
+  triangle_cells_are_coplanar(
+    const typename Triangulation<dim, spacedim>::active_cell_iterator &cell_A,
+    const typename Triangulation<dim, spacedim>::active_cell_iterator &cell_B,
+    double tol = 1e-6);
 
 
+  /**
+   * @brief Map the vertex index to the cells that includes that vertex.
```
Julamipolymtl:
```suggestion
   * @brief Map the vertex index to the cells which include that vertex.
```

applications_tests/lethe-particles/solid_surfaces_mesh/90deg_triangles.geo
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: Copyright (c) 2023-2025 The Lethe Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
```
blaisb, line 3: Could you add a one liner comment that decribes what this gometry is?

applications_tests/lethe-particles/solid_surfaces_mesh/90deg_triangles.geo
blaisb, line 1: We put the geo files in: applications_tests/geo_files

applications_tests/lethe-particles/solid_surfaces_mesh/flat_triangles.geo
blaisb, line 1: Same comment regarding a comment on the file and moving it to /applications_tests/geo_files

applications_tests/lethe-particles/CMakeLists.txt
```diff
@@ -40,6 +40,9 @@ file(COPY moving_solid_surface_files/square.msh DESTINATION "${CMAKE_CURRENT_BIN
 file(COPY moving_solid_surface_files/square.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/moving_solid_surface_jkr.${_build_type}")
 file(COPY insert_and_remove_with_files_files/particles_00.input DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/insert_and_remove_with_files.${_build_type}")
 file(COPY insert_and_remove_with_files_files/particles_01.input DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/insert_and_remove_with_files.${_build_type}")
+file(COPY solid_surfaces_mesh/90deg_triangles.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/particle_solid_surface_NPES_double_edge_contact.${_build_type}")
+file(COPY solid_surfaces_mesh/90deg_triangles.msh DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/particle_solid_surface_NPES_double_face_contact.${_build_type}")
```
blaisb, line 44: Unless I am mistaken, you care copying twice?

applications_tests/lethe-particles/particle_solid_surface_CPES_double_edge_contact.prm
```diff
@@ -0,0 +1,133 @@
+# Listing of Parameters
+#----------------------
+
+set dimension = 3
+
```
blaisb, line 13: Can you add a comment describing the test and what it does here? What does the CPES acronym mean also?

applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
```diff
@@ -0,0 +1,132 @@
+# Listing of Parameters
+#----------------------
+
+set dimension = 3
```
blaisb, line 13: Same comment regarding description of the test
Can you also define in that comment the meaning of NPES

applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
```diff
@@ -0,0 +1,133 @@
+# Listing of Parameters
+#----------------------
+
+set dimension = 3
+
```
blaisb, line 15: Same comment regarding description of the test
Can you also define in that comment the meaning of NPES

include/core/lethe_grid_tools.h
```diff
@@ -260,6 +259,18 @@ namespace LetheGridTools
                                                   &vertices_cell_map,
     std::vector<SerialSolid<structdim, spacedim>> &list_of_objects);
 
+
+  /** brief
+   * @enum ContactIndicator
+   * Enum to indicate the type of contact between a particle and a triangle.
+   */
+  enum class ContactIndicator : std::uint8_t
```
blaisb: I would rename that to ParticleTriangleContactIndicator

include/core/lethe_grid_tools.h
```diff
@@ -431,7 +443,55 @@ namespace LetheGridTools
     }
   };
 
-} // namespace LetheGridTools
+  /**
+   * @brief Function that checks if two cells are coplanar within a given tolerance.
+   *
+   * @param cell_A First cell to compare.
+   * @param cell_B Second cell to compare.
+   * @param tol Tolerance for coplanarity check.
```
blaisb: the coplanarity checks uses the angle between the normals? can you specify how coplanarity is checked?

include/core/serial_solid.h
```diff
@@ -177,6 +184,34 @@ class SerialSolid
...
+    DEM::cut_cell_comparison<spacedim>>
+    triangulation_cell_neighbors_map;
+
+  /**
+   * @brief Returns the neighboring cells maps of the solid object triangulation.
+   * These maps store for each cell of the solid object which neighboring cells
+   * are sharing a vertex or an edge (two vertex). Also stores which of these
+   * neighboring cells are coplanar or non-coplanar. 1. coplanar edge-sharing,
+   * 2. coplanar vertex-sharing, 3. non-coplanar edge-sharing, 4. non-coplanar
+   * vertex-sharing.
```
blaisb, line 197:
```suggestion
   * @brief Returns the neighboring cells maps of the solid object triangulation.
   * These maps store for each cell of the solid object which neighboring cells
   * are sharing a vertex or an edge (two vertex). Also stores which of these
   * neighboring cells are coplanar (cp) or non-coplanar (ncp). 1. coplanar edge-sharing,
   * 2. coplanar vertex-sharing, 3. non-coplanar edge-sharing, 4. non-coplanar
   * vertex-sharing.
```

I would define here the acronym you use before
I would rather use cp and ncp than cp and np because it's coplanaer and non-coplanar.

include/core/serial_solid.h
```diff
@@ -309,6 +342,14 @@ class SerialSolid
   std::shared_ptr<Function<spacedim>> solid_temperature;
   Parameters::ThermalBoundaryType     thermal_boundary_type;
   double                              current_solid_temperature;
+
+  // Neighboring cells maps
+  // CP : coplanar
+  // NP : non-coplanar
```
blaisb: I would use NCP as an abbreviation instead of CP

source/core/lethe_grid_tools.cc
```diff
@@ -1852,3 +1886,176 @@ LetheGridTools::rotate_mapping(const DoFHandler<3> &dof_handler,
...
+  v_A = points_A[2] - points_A[0];
+  u_B = points_B[1] - points_B[0];
+  v_B = points_B[2] - points_B[0];
+
+  // Normal of each triangle
+  normal_A = cross_product_3d(u_A, v_A);
+  normal_B = cross_product_3d(u_B, v_B);
+
+  if (std::abs(scalar_product(normal_A, normal_B)) >=
+      normal_A.norm() * normal_B.norm() * (1.0 - tol))
```
blaisb: You could do the product of the norm squared and the sqrt afterwards to avoid doing 2X sqrt (one in each of the norm)

source/core/lethe_grid_tools.cc
```diff
@@ -1852,3 +1886,176 @@ LetheGridTools::rotate_mapping(const DoFHandler<3> &dof_handler,
...
+      // Since this is a serial triangulation, no need to check if the cells
+      // are locally own.
+      const unsigned int vertices_per_cell = cell->n_vertices();
+      for (unsigned int i = 0; i < vertices_per_cell; i++)
+        {
+          // First obtain vertex index
+          unsigned int v_index = cell->vertex_index(i);
+
+          // Insert the cell into the set of cell around that vertex.
+          vertices_cell_map[v_index].insert(cell);
```
blaisb, line 1928: If you want to read it at v_index (only if v_index exists) then I would use .at(v_index) instead of []
Otherwise, this is the right choice

source/core/serial_solid.cc
```diff
@@ -607,7 +614,92 @@ SerialSolid<dim, spacedim>::setup_dof_handler()
...
+        cp_es_main_cell, np_es_main_cell, cp_vs_main_cell, np_vs_main_cell;
+
+      // Number of vertices of the main cell
+      const unsigned int n_vertices_main_cell = main_cell->n_vertices();
+
+      // Store the vertices index of the main cell in a vector
+      std::vector<unsigned int> main_cell_vertex_indices;
+      main_cell_vertex_indices.reserve(n_vertices_main_cell);
+      for (unsigned int v = 0; v < n_vertices_main_cell; ++v)
+        main_cell_vertex_indices.push_back(main_cell->vertex_index(v));
```
blaisb, line 642: I wonder if we could pre-allocate this before the loop over the main_cell

source/core/serial_solid.cc
```diff
@@ -607,7 +614,92 @@ SerialSolid<dim, spacedim>::setup_dof_handler()
...
+      // Loop on the neighboring cells of the main cell
+      for (const auto &neigh_cell : neighboring_cells)
+        {
+          // Skip the first iterator since it is the main cell itself
+          if (neigh_cell == main_cell)
+            continue;
+
+          // Check if the neighboring and main cells are coplanar. This bool
+          // is used later in the loop.
+          bool coplanar =
```
blaisb:
```suggestion
          const bool coplanar =
```

blaisb (changes requested):
One file left Yo !

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
+
+      using particle_triangle_contact_record =
+        ankerl::unordered_dense::map<types::particle_index,
+                                     particle_triangle_contact_description>;
+
+      particle_triangle_contact_record contact_record;
+
+      // In general, a particle should not be in contact with a huge number of
+      // triangle. 10 contact is a reasonable upper limit for most cases.
+      contact_record.reserve(10);
```
blaisb: Might be better to not reserve no? I mean by default most particle will not contact any triangles so might as well start from scratch?
Here you will allocate 10 contact per particle in the simulation, which is kinda a lot I feel

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
-                                      PropertiesIndex,
-                                      DEM::DEMMPProperties::PropertiesIndex>)
+                      // If C1 is a face contact and C2 is a vertex
+                      // contact.
+                      // C2 is always invalid.
+                      clear_contact_info(contact_info_C2);
+                      contact_2 = this_contact_record.erase(contact_2);
+                      continue;
+                    }
+
```
blaisb, line 367:
```suggestion
```

include/core/lethe_grid_tools.h
OGaboriault: It uses the scalar product of the normal unit vector of each triangle. If it is equal to 1 (or -1) they are coplanar

source/core/lethe_grid_tools.cc
OGaboriault: Sure, but currently this is only being called a few times at the start of the simulation, so it is not a big deal.

source/dem/particle_wall_contact_force.cc
OGaboriault: 4 is enough?

include/core/lethe_grid_tools.h
Julamipolymtl: Does the function take in 1 or multiple triangles ?

include/dem/particle_wall_fine_search.h
```diff
@@ -56,11 +56,11 @@ particle_floating_wall_fine_search(
...
  * @brief Iterate over the contact candidates from particle-floating mesh broad
  * search (particle_floating_mesh_contact_candidates) to add new contact pairs
- * to the particle_floating_mesh_in_contact container
+ * to the particle_floating_mesh_potentially_in_contact container
  *
  * @param particle_floating_mesh_contact_candidates The output of particle-floating mesh
  * broad search which shows contact pair candidates
- * @param particle_floating_mesh_in_contact A map of maps which stores
+ * @param particle_floating_mesh_potentially_in_contact A map of maps which stores
  * all the particle-floating mesh pairs which are in contact
```
Julamipolymtl:
```suggestion
 * @param particle_floating_mesh_potentially_in_contact A map of maps which stores
 * all the particle-floating mesh pairs which are potentially in contact
```

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
+                      if (contact_indicator_C2 ==
+                          LetheGridTools::ContactIndicator::edge_contact)
                         {
-                          contact_info.tangential_displacement.clear();
-                          contact_info.rolling_resistance_spring_torque.clear();
+                          // If T1 and T2 are non-coplanar and vertex
+                          // sharing, C1 and C2 are valid. This could happen
+                          // with two concave triangles.
+                          if (std::ranges::find(T1_np_vs_neighbors, T2_cell) !=
+                              T1_np_vs_neighbors.end())
```
Julamipolymtl: From the paper's discussion (3.2.2), it seems a face-edge contact when co-planar and vertex sharing is impossible. Therefore the check on coplanarity is not necessary here as we'd only need to check wether or not they are vertex sharing.

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
+                          LetheGridTools::ContactIndicator::edge_contact)
+                        {
+                          // C1 is always valid.
+                          // C2 is invalid if T1 and T2 are edge sharing.
+                          // This case means that both contacts occur on the
+                          // same edge which is connecting both triangles.
+                          if (std::ranges::find(T1_cp_es_neighbors, T2_cell) !=
+                                T1_cp_es_neighbors.end() ||
+                              std::ranges::find(T1_np_es_neighbors, T2_cell) !=
+                                T1_np_es_neighbors.end())
```
Julamipolymtl: This basically only depends wether the triangles are edge sharing.

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
-                              (normal_overlap > 0))
+                          // C2 is valid
+                          // C1 is invalid if T1 and T2 are vertex sharing.
+                          // This case means that both contacts occur on the
+                          // same vertex which is connecting both triangles,
+                          // thus C1 and C2 are double contacts.
+                          if (std::ranges::find(T1_cp_vs_neighbors, T2_cell) !=
+                                T1_cp_vs_neighbors.end() ||
+                              std::ranges::find(T1_np_vs_neighbors, T2_cell) !=
+                                T1_np_vs_neighbors.end())
```
Julamipolymtl: Same as above, you'd only need to know wether the triangles are vertex sharing.

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,455 @@ ParticleWallContactForce<dim,
...
-
-      auto &particle_floating_mesh_contact_pair =
-        particle_floating_mesh_in_contact[solid_counter];
-
-      for (auto &[cut_cell, map_info] : particle_floating_mesh_contact_pair)
+      // Extract the neighboring cell maps of this solid objects
+      const auto &[this_solid_cp_es_neighbors,
+                   this_solid_cp_vs_neighbors,
+                   this_solid_np_es_neighbors,
+                   this_solid_np_vs_neighbors] =
```
Julamipolymtl: After checking the paper and speaking with @OGaboriault, I'm not sure we actually need to know wether the triangles neighbours are coplanar or not. See my comments below for cases where this seems to be the case.

include/dem/particle_wall_fine_search.h
```diff
@@ -56,11 +56,11 @@ particle_floating_wall_fine_search(
 /**
  * @brief Iterate over the contact candidates from particle-floating mesh broad
  * search (particle_floating_mesh_contact_candidates) to add new contact pairs
- * to the particle_floating_mesh_in_contact container
+ * to the particle_floating_mesh_potentially_in_contact container
  *
  * @param particle_floating_mesh_contact_candidates The output of particle-floating mesh
  * broad search which shows contact pair candidates
- * @param particle_floating_mesh_in_contact A map of maps which stores
+ * @param particle_floating_mesh_potentially_in_contact A map of maps which stores
```
Julamipolymtl, line 63: With a floating mesh, when do the contact go from "potential" contacts to be actual contacts. Is the contact invalidity check that is being currently added all that i needed to get the actual contacts ?

Julamipolymtl:
Unsure what is the best way to discuss about this on github, but here goes!

The paper's implementation also seems to consider a particle contact at a much smaller lengthscale than that of the shape of the triangle mesh. In their cases, they always consider that contacts happen with neighbours, never neigbours of neigbours. Plus, in an edge-edge or vertex-vertex contact validity check, it is always assumed that the edge or vertex in question is the shared one. 

I can see this being valid for cases where the particle is either small compared to the mesh, or behaves very rigidely. It is unclear to me wether that is the case irregardless of the wall mesh size (i.e. if the triangles are very small compared to the sphere contact lengthscale)

image

For example in this figure below, the current implementation **FC-EC1** contact check would invalidate **EC1** properly as they are edge sharing, but **FC-EC2** would never be checked as **T1** and **T3** are not neighbours. Furthermore, **EC1-EC2** check would validate both as **T2** and **T3** are vertex sharing. Therefore, both **FC** and **EC2** would be considered valid, which bothers me if they are coplanar.

particle-mesh-contact


Finally, in a **EC1-EC3** check, since they are edge sharing, the current implementation considers the contact is on the shared edge and drops one of them, even though the contact is not on the shared edge. Plus, the paper considers that a face-edge contact is impossible for coplanar vertex sharing triangles, therefore there is no prescribed validity to the **FC-EC3** check.

All in all, if the mesh is coarse enough compared to the contact, there should be no issues, but otherwise there might be more multiplicity in the contacts.

include/core/lethe_grid_tools.h
OGaboriault: 1 triangle, multiple particles

include/dem/particle_wall_fine_search.h
OGaboriault: There's no actual contact list... 

Particle that are close enough to the triangle are potential contact. aka the overlap need to be check at each time step. 

The invalidity check is for when the particle is touching multiple triangles at the same time.

source/core/lethe_grid_tools.cc
OGaboriault: actually, I don't think I can.. 

either way, this is no longer being used in the code... we could remove it.

source/core/lethe_grid_tools.cc
OGaboriault: you cannot wrtie in the container when using .at(). 

I'm inserting a cell, thus I can't use it.

source/core/serial_solid.cc
OGaboriault: long term, will this be used for cell that are not simplexes? 

If this is not the cas, we know from the get go that  n_vertices_main_cell = 3, thus we could pre-allocate the vector, but idk what is preferable...

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,445 @@ ParticleWallContactForce<dim,
...
+                         contact_indicator_C2,
+                         contact_info_C2] = *contact_2;
+
+                  // First, we check if both triangle are neighbors. If they
+                  // are not neighbors, C1 and C2 are automatically valid.
+                  // (Disconnected triangles)
+                  if (std::ranges::find(T1_es_neighbors, T2_cell) ==
+                        T1_es_neighbors.end() &&
+                      std::ranges::find(T1_es_neighbors, T2_cell) ==
+                        T1_es_neighbors.end())
```
Julamipolymtl: Isn't this twice the same check twice?
```suggestion
                  if (std::ranges::find(T1_es_neighbors, T2_cell) ==
                        T1_es_neighbors.end())
```

applications_tests/lethe-particles/CMakeLists.txt
blaisb: oh it's edge and face. Wow ok. I did not catch this clearly :)

applications_tests/lethe-particles/particle_solid_surface_CPES_double_edge_contact.prm
```diff
@@ -0,0 +1,138 @@
+# Test description:
+# In this test, a single particle is bouncing on a solid-surface made of two triangles that are coplanar (CP) and edge sharing (ES).
+# The particle is bouncing exactly on the sharing edge.
+# This test checks that the code doesn't apply the contact force twice on the particle.
```
blaisb:
```suggestion
# The particle is bouncing exactly on the edge shared by the two triangles.
# This test checks that the code doesn't apply the contact force twice on the particle.
```

applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
```diff
@@ -0,0 +1,138 @@
+# Test description:
+# In this test, a single particle is bouncing on a solid-surface made of two triangles that are non-coplanar (NP) and edge sharing (ES).
+# The two triangles are making a 90 deg angle on the sharing edge, making an UPSIDE-DOWN "v".
```
blaisb:
```suggestion
# The two triangles are making a 90 deg angle on the edge they share, forming an UPSIDE-DOWN "v".
```

applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
```diff
@@ -0,0 +1,140 @@
+# Test description:
+# In this test, a single particle is bouncing on a solid-surface made of two triangles that are non-coplanar (NP) and edge sharing (ES).
+# The two triangles are making a 90 deg angle on the sharing edge, making an UPRIGHT "v".
+# The particle is LOCATED exactly on the sharing edge, but doesn't contact that edge due to the angle of the two triangles.
```
blaisb:
```suggestion
# The particle is LOCATED exactly on the edge shared by the two triangles, but does not contact that edge due to the angle between the two triangles.
```

applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
```diff
@@ -0,0 +1,140 @@
...
+# The particle contact both triangle faces, thus two contact forces should be applied since it is not the same contact point.
```
blaisb:
```suggestion
# The particle is in contact with both triangle faces, thus two contact forces should be applied since it is not the same contact point.
```

include/core/lethe_grid_tools.h
```diff
@@ -280,18 +291,20 @@ namespace LetheGridTools
    * @param n_particles_in_base_cell Number of particles in the base cell
    *
    * @return A tuple in which 0. a vector of bools to determine if the particle is
```
blaisb:
```suggestion
   * @return A tuple in which: 0. a vector of bools to determine if the particle is
```

include/core/lethe_grid_tools.h
```diff
@@ -431,7 +444,56 @@ namespace LetheGridTools
     }
   };
 
-} // namespace LetheGridTools
+  /**
+   * @brief Checks whether two cells are coplanar within a given tolerance by
+   * using the scalar product of each normal unit vector. If the cells are
+   * coplanar, the scalar product should be equal to 1.
```
blaisb:
```suggestion
   * @brief Checks whether two cells are coplanar within a given tolerance by
   * using the scalar product of each normal unit vector. If the cells are
   * coplanar, the absolute value of the scalar product should be above (1-tolerance).
```

include/core/serial_solid.h
```diff
@@ -177,6 +184,30 @@ class SerialSolid
...
+   * are sharing a vertex or an edge (two vertex). Also stores which of these
+   * neighboring cells are coplanar (cp) or non-coplanar (ncp). 1. coplanar
+   * edge-sharing,
+   * 2. coplanar vertex-sharing, 3. non-coplanar edge-sharing, 4. non-coplanar
+   * vertex-sharing.
+   *
+   * @return A tuple containing the four neighboring cells maps.
+   */
+  inline std::tuple<triangulation_cell_neighbors_map,
+                    triangulation_cell_neighbors_map>
```
blaisb, line 202: But there are two maps here? Why so you say 4 maps then?

include/core/serial_solid.h
```diff
@@ -177,6 +184,30 @@ class SerialSolid
...
+    DEM::cut_cell_comparison<spacedim>>
+    triangulation_cell_neighbors_map;
+
+  /**
+   * @brief Returns the neighboring cells maps of the solid object triangulation.
+   * These maps store for each cell of the solid object which neighboring cells
+   * are sharing a vertex or an edge (two vertex). Also stores which of these
+   * neighboring cells are coplanar (cp) or non-coplanar (ncp). 1. coplanar
+   * edge-sharing,
+   * 2. coplanar vertex-sharing, 3. non-coplanar edge-sharing, 4. non-coplanar
```
blaisb: You only return two maps right now.

source/core/serial_solid.cc
blaisb: I think it is fine the current way you implemented it for now :)

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The changelog for the previous releases of Lethe are located in the release_note
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 
+### [Master] - 2025-11-22
+
+### Fixed
+
+- MINOR Contacts between particles and solid surfaces had a problem when the contact was occurring with many triangles at the same time. Contacts we being duplicated even if the solid surface represent a single surface. In this PR, duplicated contacts are identified and removed following the logic described in https://doi.org/10.1002/nme.4487. [#1745](https://github.com/chaos-polymtl/lethe/pull/1745)
```
```suggestion
- MINOR Contacts between particles and solid surfaces had a problem when the contact was occurring with more than one triangle from the same solid object at the same time. Contacts were duplicated even if the solid surface represent a single surface. This resulting in a particle contacting the same solid object twice, meaning that the particle would see a surface that has twice the stiffness at this location. In this PR, duplicated contacts are identified and removed following the logic described in https://doi.org/10.1002/nme.4487. This ensures that there is only one contact between the particle and a solid object and keeps the contact regular as a particle moves over a solid object. [#1745](https://github.com/chaos-polymtl/lethe/pull/1745)
```

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,445 @@ ParticleWallContactForce<dim,
...
+      // or vertex) 4. The contact info associated.
+      using particle_triangle_contact_description = std::vector<
+        std::tuple<typename Triangulation<dim - 1, dim>::active_cell_iterator,
+                   double,
+                   LetheGridTools::ParticleTriangleContactIndicator,
+                   particle_wall_contact_info<dim>>>;
+
+      using particle_triangle_contact_record =
+        ankerl::unordered_dense::map<types::particle_index,
+                                     particle_triangle_contact_description>;
```
blaisb: Don't you want instead to use typedefs in the .h of that file? It will be a lot clearer to use a typedef instead of a using locally I find. I'm a bit skeptical of these using statements in only a single routine. These are quite big structures, In that case it's better to have a typedef generally.

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,445 @@ ParticleWallContactForce<dim,
...
+                &T1_es_neighbors = this_solid_es_neighbors.at(T1_cell);
+              const std::vector<
+                typename Triangulation<dim - 1, dim>::active_cell_iterator>
+                &T1_vs_neighbors = this_solid_vs_neighbors.at(T1_cell);
+
+              // Initialize variables. We need to compare C1 to every other
+              // contact that was compared yet.
+              bool erase_contact_1 = false;
+              auto contact_2       = std::next(contact_1);
+              while (contact_2 != this_contact_record.end())
```
blaisb: Instead of calling them _1 and _2, I would try to use a more descriptive name like contact_with_triangle_2 or contact_triangle_2. Something that is a bit more descriptive than just 1 and 2. Especially since you have C1 and C2 after that it makes it a bit confusing.

source/core/lethe_grid_tools.cc
```diff
@@ -1219,7 +1219,10 @@ LetheGridTools::find_cells_cut_by_object(
 
 
 template <int dim, typename PropertiesIndex>
-std::tuple<std::vector<bool>, std::vector<Point<3>>, std::vector<Tensor<1, 3>>>
+std::tuple<std::vector<bool>,
+           std::vector<Point<3>>,
+           std::vector<Tensor<1, 3>>,
+           std::vector<LetheGridTools::ParticleTriangleContactIndicator>>
```
blaisb, line 1225: Is there a reason to make such a huge tuple instead of making a struct instead? I find that structs are easier to understand and follow than tuples since they group set of parameters more intuitively than tuples.

source/core/lethe_grid_tools.cc
```diff
@@ -1453,57 +1477,80 @@ LetheGridTools::find_particle_triangle_projection(
...
       pass_distance_check[k] = true;
       normal_vectors[k]      = unit_normal_3d;
+      contact_indicators[k]  = this_contact_indicator;
       ++k;
     }
   return std::make_tuple(pass_distance_check,
                          projection_points,
-                         normal_vectors);
+                         normal_vectors,
+                         contact_indicators);
```
blaisb, line 1486: Maybe we could consider making this a struct. But that might be btter for a follow-up PR. In that case, I would consider opening an issue.

source/core/lethe_grid_tools.cc
```diff
@@ -1852,3 +1899,176 @@ LetheGridTools::rotate_mapping(const DoFHandler<3> &dof_handler,
                                const double        &rotation_angle,
                                const Point<3>      &center_of_rotation,
                                const Tensor<1, 3>  &rotation_axis);
+
+
+template <int dim, int spacedim>
+bool
+LetheGridTools::triangle_cells_are_coplanar(
```
blaisb:
```suggestion
LetheGridTools::triangular_cells_are_coplanar(
```

source/core/lethe_grid_tools.cc
```diff
@@ -1852,3 +1899,176 @@ LetheGridTools::rotate_mapping(const DoFHandler<3> &dof_handler,
...
+LetheGridTools::triangle_cells_are_coplanar<1, 2>(
+  const typename Triangulation<1, 2>::active_cell_iterator &cell_A,
+  const typename Triangulation<1, 2>::active_cell_iterator &cell_B,
+  const double                                              tol);
+
+template bool
+LetheGridTools::triangle_cells_are_coplanar<2, 2>(
+  const typename Triangulation<2, 2>::active_cell_iterator &cell_A,
+  const typename Triangulation<2, 2>::active_cell_iterator &cell_B,
+  const double                                              tol);
```
blaisb: These functions are not ok. They will not work because you "hardcode" the loop to 3!

source/core/lethe_grid_tools.cc
```diff
@@ -1852,3 +1899,176 @@ LetheGridTools::rotate_mapping(const DoFHandler<3> &dof_handler,
...
+  const typename Triangulation<dim, spacedim>::active_cell_iterator &cell_A,
+  const typename Triangulation<dim, spacedim>::active_cell_iterator &cell_B,
+  const double                                                       tol)
+{
+  // Point of each triangle cell
+  std::vector<Point<spacedim>> points_A, points_B;
+  points_A.reserve(3);
+  points_B.reserve(3);
+
+  for (unsigned int i = 0; i < 3; ++i)
```
blaisb: This functoin will not work if you cell has spacedim=2 and dim=2. Yet you are instantiating the template for these variants below. If you want to create these variants, I would create empty functions that just AssertThrow(false) with a message that says they are not implemented instead of having something with a false premise of working.

source/core/lethe_grid_tools.cc
OGaboriault: We use to have a tuple, this is why I added an extra vector. I don't see why we would not be able to use a struct.

source/dem/particle_wall_contact_force.cc
OGaboriault: I can rename "contact_1" and "contact_2" to C1 and C2 respectively. Maybe that would help? But at the end of the day, we are comparing two contacts... _1 and _2 seems a good way to differentiate them 

Otherwise, i and j ?

source/core/lethe_grid_tools.cc
blaisb: Yeah we can move to a struct. Maybe for a quick follow-up PR afterwards if you want.

include/core/lethe_grid_tools.h
```diff
@@ -280,18 +291,20 @@ namespace LetheGridTools
    * @param n_particles_in_base_cell Number of particles in the base cell
    *
    * @return A tuple in which 0. a vector of bools to determine if the particle is
-   * close to the triangle plane, 1. a vector of projected location of particles
-   * on the triangle, 2. a vector of normal vectors of the triangles
+   * less than a radius away from the triangle plane, 1. a vector of projected
+   * location of particles on the triangle, 2. a vector of normal vectors of the
+   * triangles, 3. a vector of the contact type between the particle and the
+   * triangle.
```
Julamipolymtl:
```suggestion
   * less than a radius away from the triangle plane, 1. a vector of projected
   * location of particles on the triangle, 2. a vector of normal vectors of the
   * triangle, 3. a vector of the contact type between the particle and the
   * triangle.
```

I know its not the in the PR per say, but I'm not sure I understand exactly what the normal vectors are in this function. Is it simply the normal of the triangle (a plane) so its basically always the same for a given triangle ? If so it looks like its being evaluated differently based on what region we are in.

applications_tests/lethe-particles/particle_solid_surface_CPES_double_edge_contact.prm
```diff
@@ -0,0 +1,138 @@
+# Test description:
+# In this test, a single particle is bouncing on a solid-surface made of two triangles that are coplanar (CP) and edge sharing (ES).
```
blaisb, line 5: License text indication for the prm file is missing. I know this is annoying but every file must have the license file ideally.

applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
```diff
@@ -0,0 +1,138 @@
+# Test description:
```
blaisb, line 4: License text missing

applications_tests/lethe-particles/particle_solid_surface_NPES_double_edge_contact.prm
```diff
@@ -0,0 +1,138 @@
...
+# In this test, a single particle is bouncing on a solid-surface made of two triangles that are non-coplanar (NP) and edge sharing (ES).
+# The two triangles are making a 90 deg angle on the edge they share, forming an UPSIDE-DOWN "v".
+# The particle is BOUNCING exactly on the sharing edge.
```
blaisb:
```suggestion
# The particle is BOUNCING exactly on the edge that is shared.
```

applications_tests/lethe-particles/particle_solid_surface_NPES_double_face_contact.prm
```diff
@@ -0,0 +1,140 @@
+# Test description:
```
blaisb, line 4: License information missing here.

include/core/lethe_grid_tools.h
```diff
@@ -260,6 +259,18 @@ namespace LetheGridTools
                                                   &vertices_cell_map,
     std::vector<SerialSolid<structdim, spacedim>> &list_of_objects);
 
+
+  /** brief
+   * @enum ParticleTriangleContactIndicator
+   * Enum to indicate the type of contact between a particle and a triangle.
+   */
```
blaisb:
```suggestion
  /** @brief
   * @enum ParticleTriangleContactIndicator
   * Enum to indicate the type of contact between a particle and a triangle.
   */
```

Brief is a keyword no? Could you double check how this renders in the Doxygen?

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,434 @@ ParticleWallContactForce<dim,
...
+              // Assigning the triangle neighboring list of T1;
+              const std::vector<
+                typename Triangulation<dim - 1, dim>::active_cell_iterator>
+                &T1_es_neighbors = this_solid_es_neighbors.at(T1_cell);
+              const std::vector<
+                typename Triangulation<dim - 1, dim>::active_cell_iterator>
+                &T1_vs_neighbors = this_solid_vs_neighbors.at(T1_cell);
+
+              // Initialize variables. We need to compare C1 to every other
+              // contact (C2) that was compared yet.
```
blaisb: What does was compared yet mean?
We need to compare C1 to every contact that has been compared thus far or that has not been compared thus far?
Just be a bit more explicit in the comment.

source/dem/particle_wall_contact_force.cc
```diff
@@ -158,209 +165,434 @@ ParticleWallContactForce<dim,
...
+                          continue;
+                        }
+                      // If C1 is a face contact and C2 is an edge contact.
+                      if (contact_indicator_C2 ==
+                          LetheGridTools::ParticleTriangleContactIndicator::
+                            edge_contact)
+                        {
+                          // If T1 and T2 are vertex
+                          // sharing, C1 and C2 are valid. This could happen
+                          // with two concave triangles.
```
blaisb:
```suggestion
                          // If T1 and T2 are vertex-sharing
                          // , C1 and C2 are valid. This could happen
                          // with two concave triangles.
```

If you mean in the sense that they share a vertex, then this requires an hyphen.

include/core/lethe_grid_tools.h
OGaboriault: I have looked elsewhere in the code. For example here is what we do for the drag_coupling : 

```/**
   * @enum DragCoupling
   * @brief Defines the numerical coupling strategy used for drag force computation between fluid and particles.
   *
   * This enumeration specifies how the drag interaction term is treated in time
   * integration schemes for coupled CFD-DEM or multiphase simulations. The
   * choice of coupling affects both stability and computational cost. The ...
  ```
I will use the same syntax here.

source/dem/particle_wall_contact_force.cc
OGaboriault: missing a "not" after the was

PR #1757, opened 2025-11-06
Matrix free implicit coupling

blaisb:
### Description

This introduces the particle-fluid coupling the VANS matrix-free operator. It was a lot more work than anticipated, but the results are nice.

Essentially, this PR introduces the capability to couple implicitly, explicitly or semi-implicitly the particle-fluid forces onto the fluid. The coupling leverages the QCM projection to project the drag on the particle, the particle velocity and the momentum transfer coefficient onto the CFD mesh. Then, these projected fields are used in the VANS formulation within the coupling.

There is one thing special here. When the semi-implicit and implicit couplings are used, the drag force applied on the fluid is the momentum_transfer_term * (u_particle-u_fluid). This is not numerically fully identical to the filtered drag force. However, this is the only way I have managed to make the coupling robust and stable, so I think this is the way to go forward. Consequently, this PR meant I had to add a new property on the particles (their momentum exchange coefficient), but there is absolutely no way around that right now. The results are very good though for all of the coupling modes and the linear solver is behaving much better than I expected.

### Testing

Three new tests have been added that are the particle-fluid coupling with drag explicit, semi-implicit and implicit. These tests perform a simulation through a packed bed and measure the pressure drop through the bed.

### Documentation

There is nothing to document for now since this is a very experimental solver. 

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/fem-dem/fluid_dynamics_vans_matrix_free_operators.h
+++ b/include/fem-dem/fluid_dynamics_vans_matrix_free_operators.h
@@ -148,7 +148,8 @@ protected:
   Table<2, Tensor<1, dim, VectorizedArray<number>>> particle_fluid_force;
 
   /// Table with correct alignment for vectorization to store the values of the
-  /// particle-fluid drag. The drag force in this table is volumetric, resulting from the projection used.
+  /// particle-fluid drag. The drag force in this table is volumetric, resulting
+  /// from the projection used.
   Table<2, Tensor<1, dim, VectorizedArray<number>>> particle_fluid_drag;
 
   /// Table with correct alignment for vectorization to store the values of the
--- a/include/fem-dem/particle_projector.h
+++ b/include/fem-dem/particle_projector.h
@@ -702,11 +702,16 @@ private:
 
 public:
   /// Projector used to save the particle velocity field onto the mesh
+  /// This field is not volumetric, meaning that the integral over the
+  /// entire domain of this field will not yield the sum of the particle
+  /// velocity over all the particles.
   ParticleFieldQCM<dim, 3, DEM::CFDDEMProperties::PropertiesIndex::v_x>
     particle_velocity;
 
   /// Projector used to save the forces used in the two-way coupling scheme,
-  /// without drag.
+  /// without drag. This field is volumetric. The integral of this field over
+  /// the entire domain will result in the sum of the force (without drag) over
+  /// all of the particles.
   ParticleFieldQCM<
     dim,
     3,
@@ -714,11 +719,16 @@ public:
     fluid_force_on_particles_two_way_coupling;
 
   /// Projector used to store the drag force. The drag is stored as a separate
-  /// field since we may want to make implicit calculations of it.
+  /// field since we may want to make implicit calculations of it. This field
+  /// is volumetric. The integral of this field over the entire domain will
+  /// result in the sum of the drag force over all the particles.
   ParticleFieldQCM<dim, 3, DEM::CFDDEMProperties::PropertiesIndex::fem_drag_x>
     fluid_drag_on_particles;
 
-  /// Projector used to save the momentum transfer coefficient (beta)
+  /// Projector used to save the momentum transfer coefficient (beta). This
+  /// field is volumetric. The integral of this field over the entire domain
+  /// will result in the sum of the momentum transfer coefficient over all
+  /// particles.
   ParticleFieldQCM<
     dim,
     1,


CHANGELOG.md
```diff
@@ -5,12 +5,15 @@ The changelog for the previous releases of Lethe are located in the release_note
...
...s a fully implicit and a fully explicit coupling mode. In the fully implicit mode, the particle-fluid forces are calculated using the full velocity solution at time t+delta t. Surprisingly, this is a lot more stable than I thought and it leads to very good results. In the fully explicit formulation, the particle-fluid forces are calculated at time t and are applied to both the particle and the fluid. In this case, there is no Jacobian matrix and the solver is fully explicit. This means that these is a stability criterion. To make this work well, I had to move the CFD-DEM parameters to the core solver so that the library may be used within the scratch data (since I need to know the drag mode in the calculation of the velocity at the particle location). I also found some slight issues with the two-way coupling for the lift forces and there were some signs that were conmfusing (for example the lift force was applied to both the fluid and the particle with the same sign and the - sign came later), this was kinda confusing and I took the opportunity to clarify this here. As a consequence, some tests results have slightly changed. [#1752](https://github.com/chaos-polymtl/lethe/pull/1752)
```
antonellasuccar, line 24:
```suggestion
- MAJOR The particle-fluid coupling in the VANS and CFD-DEM solvers currently relies on a semi-implicit coupling where the drag force is calculated explicitly and partially treated implicitly on the fluid side. This is a bit problematic for some cases and I want to introduce the capability to control what we are essentially doing. This change introduces 3 coupling modes. The semi-implicit (which is the one we actually use right now) as well as a fully implicit and a fully explicit coupling mode. In the fully implicit mode, the particle-fluid forces are calculated using the full velocity solution at time t+delta t. Surprisingly, this is a lot more stable than I thought and it leads to very good results. In the fully explicit formulation, the particle-fluid forces are calculated at time t and are applied to both the particle and the fluid. In this case, there is no Jacobian matrix and the solver is fully explicit. This means that there is a stability criterion. To make this work well, I had to move the CFD-DEM parameters to the core solver so that the library may be used within the scratch data (since I need to know the drag mode in the calculation of the velocity at the particle location). I also found some slight issues with the two-way coupling for the lift forces and there were some signs that were confusing (for example the lift force was applied to both the fluid and the particle with the same sign and the - sign came later). This was kinda confusing and I took the opportunity to clarify this here. As a consequence, some tests results have slightly changed. [#1752](https://github.com/chaos-polymtl/lethe/pull/1752)
```

include/fem-dem/fluid_dynamics_vans_matrix_free.h
```diff
@@ -103,6 +103,30 @@ class MFNavierStokesVANSPreconditionGMG
...
+  /// Particle velocity for each of the levels of the global
+  /// coarsening algorithm
+  MGLevelObject<MGTwoLevelTransfer<dim, MGVectorType>>
+    transfers_particle_velocity;
+
+  /// Transfer operator for global coarsening for the particle velocity
+  std::shared_ptr<GCTransferType> mg_transfer_gc_particle_velocity;
+
+  /// Particle velocity DoF handlers for each of the levels of the global
+  /// coarsening algorithm
```
antonellasuccar:
```suggestion
  /// Particle momentum transfer coefficient DoF handlers for each of the levels of the global
  /// coarsening algorithm
```
or if it is clearer fluid-drag like line 123?

include/fem-dem/fluid_dynamics_vans_matrix_free_operators.h
```diff
@@ -125,11 +142,23 @@ class VANSOperator : public NavierStokesOperatorBase<dim, number>
...
-  /// particle-fluid forces. These are the resulting forces applied on the
-  /// particles due to hydrodynamic interaction, normalized by the volume
-  /// of the cells.
+  /// particle-fluid forces (without drag). These are the resulting forces
+  /// applied on the particles due to hydrodynamic interaction, normalized by
+  /// the volume of the cells.
   Table<2, Tensor<1, dim, VectorizedArray<number>>> particle_fluid_force;
 
+  /// Table with correct alignment for vectorization to store the values of the
+  /// particle-fluid drag.
```
antonellasuccar: I guess the value here is normalized by the volume of the cell too (through the projection itself)?

include/fem-dem/particle_projector.h
```diff
@@ -136,23 +136,13 @@ class ParticleFieldQCM
...
       {
-        if constexpr (n_components == 1)
-          fe = FE_SimplexP<dim>(fe_degree);
-        else
-          {
-            const FE_SimplexP<dim> fe_temp(fe_degree);
-            fe = std::make_shared<FESystem<dim>>(fe_temp, n_components);
-          }
+        const FE_SimplexP<dim> fe_temp(fe_degree);
+        fe = std::make_shared<FESystem<dim>>(fe_temp, n_components);
```
antonellasuccar, line 140: Does the VANS solver work with a simplex mesh?

include/solvers/navier_stokes_scratch_data.h
```diff
@@ -787,6 +786,10 @@ class NavierStokesScratchData : public PhysicsScratchDataBase
       q_particles_location,
       update_values);
 
+    Assert(
+      cell_void_fraction.size() == q_particles_location.size(),
+      ExcMessage(
+        "The vector for the void fraction at the particle location does not have the same size as the quadrature used to evaluate it."));
```
antonellasuccar, line 792: Is there a reason we initialize ``cell_void_fraction`` with a size of ``n_global_max_particles_per_cell`` if it is going to be resized in each cell to a size equal to the number of particles in the cell? Is it for performance?

source/fem-dem/fluid_dynamics_vans_matrix_free.cc
```diff
@@ -155,37 +223,70 @@ MFNavierStokesVANSPreconditionGMG<dim>::initialize(
       MGLevelObject<MFNavierStokesVANSPreconditionGMG::MGVectorType>
         mg_pf_drag_solution(this->minlevel, this->maxlevel);
 
+      MGLevelObject<MFNavierStokesVANSPreconditionGMG::MGVectorType>
+        mg_particle_velocity_solution(this->minlevel, this->maxlevel);
+
+      MGLevelObject<MFNavierStokesVANSPreconditionGMG::MGVectorType>
+        mg_momentum_tranfer_coefficient_solution(this->minlevel,
```
antonellasuccar:
```suggestion
        mg_momentum_transfer_coefficient_solution(this->minlevel,
```

source/fem-dem/fluid_dynamics_vans_matrix_free.cc
```diff
@@ -155,37 +223,70 @@ MFNavierStokesVANSPreconditionGMG<dim>::initialize(
...
+        mg_particle_velocity_solution,
+        particle_projector.particle_velocity.particle_field_solution);
+
+      // Momentum transfer coefficient
+      particle_projector.momentum_transfer_coefficient.particle_field_solution
+        .update_ghost_values();
+
+      this->mg_transfer_gc_momentum_transfer_coefficient->interpolate_to_mg(
+        particle_projector.momentum_transfer_coefficient.dof_handler,
+        mg_momentum_tranfer_coefficient_solution,
```
antonellasuccar:
```suggestion
        mg_momentum_transfer_coefficient_solution,
```

source/fem-dem/fluid_dynamics_vans_matrix_free.cc
```diff
@@ -155,37 +223,70 @@ MFNavierStokesVANSPreconditionGMG<dim>::initialize(
...
+          .particle_field_solution);
+
 
       for (unsigned int l = min_level; l <= max_level; l++)
         {
           mg_void_fraction_solution[l].update_ghost_values();
           mg_pf_forces_solution[l].update_ghost_values();
           mg_pf_drag_solution[l].update_ghost_values();
+          mg_particle_velocity_solution[l].update_ghost_values();
+          mg_momentum_tranfer_coefficient_solution[l].update_ghost_values();
```
antonellasuccar:
```suggestion
          mg_momentum_transfer_coefficient_solution[l].update_ghost_values();
```

source/fem-dem/fluid_dynamics_vans_matrix_free.cc
```diff
@@ -195,11 +296,15 @@ MFNavierStokesVANSPreconditionGMG<dim>::initialize(
...
+              mf_operator->compute_particle_fluid_interaction(
                 this->pf_force_dof_handlers[l],
                 mg_pf_forces_solution[l],
                 this->pf_drag_dof_handlers[l],
-                mg_pf_drag_solution[l]);
+                mg_pf_drag_solution[l],
+                this->particle_velocity_dof_handlers[l],
+                mg_particle_velocity_solution[l],
+                this->momentum_transfer_coefficient_dof_handlers[l],
+                mg_momentum_tranfer_coefficient_solution[l]);
```
antonellasuccar:
```suggestion
                mg_momentum_transfer_coefficient_solution[l]);
```

source/fem-dem/fluid_dynamics_vans_matrix_free.cc
```diff
@@ -353,6 +458,83 @@ FluidDynamicsVANSMatrixFree<dim>::read_dem()
     }
 }
 
+template <int dim>
+void
+FluidDynamicsVANSMatrixFree<dim>::assemble_system_rhs()
+{
+  // First we update the particle-fluid coupling terms.
+  // The base matrix-free operator is not aware of the various VANS
+  // coupling term. We must do a cast here to ensure that the operator is
```
antonellasuccar:
```suggestion
  // coupling terms. We must do a cast here to ensure that the operator is
```

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -106,62 +106,159 @@ VANSOperator<dim, number>::compute_void_fraction(
...
+                               fp_force_dof_handler.get_fe(),
                                this->matrix_free.get_quadrature(),
                                update_values);
 
-  std::vector<Tensor<1, dim>> cell_pf_force(
+  FEValues<dim> fe_values_particle_velocity(
+    *(this->matrix_free.get_mapping_info().mapping),
+    particle_velocity_dof_handler.get_fe(),
+    this->matrix_free.get_quadrature(),
+    update_values);
```
antonellasuccar, line 174: Do we need different FEValues object for the different vector projected particle fields? The dofs, quadrature points, mapping and polynomial degrees are the same.

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -106,62 +106,159 @@ VANSOperator<dim, number>::compute_void_fraction(
...
+                    -cell_fp_drag[q][c] * inv_density;
+
+                  // The particle velocity is not divided by the fluid density.
+                  particle_velocity[cell][q][c][lane] =
+                    cell_particle_velocity[q][c];
                 }
+              // The momentum transfer coefficient does not need to be divided
+              // by the density since it is by construction divided by density.
+              // This is something to keep in mind later on if we allow for
+              // variable density.
```
antonellasuccar:
```suggestion
              // a variable density.
```

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -170,17 +267,26 @@ VANSOperator<dim, number>::compute_particle_fluid_force(
...
- * \- (∇·v, ε δp) - (v,p∇ε)
- * \+ ε*ν(∇v,∇δu)  +  ν(v,∇ε·∇δu)   (VANS equations),
- * plus three additional terms in the case of SUPG-PSPG stabilization:
- * \+ (ε ∂t δu + ε(u·∇)δu +  ε(δu·∇)u +  ε∇δp -  εν∆δu)τ·∇q (PSPG
+ * (Continuity equation)
+ * (q,ε∇·δu) + (q,δu·∇ε) +
+ * (VANS equations)
+ * \+ (v, ε ∂t δu)  -> Time derivative
+ * \+ (v, ε (u·∇)δu) + (v, ε (δu·∇)u) -> Advection
+ * \- (∇·v, ε δp) - (v,p∇ε) -> Pressure
```
antonellasuccar:
```suggestion
 * \- (∇·v, ε δp) - (v,p∇ε) -> Pressure
```
Should it be delta p in the second term?

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -211,21 +317,32 @@ VANSOperator<dim, number>::do_cell_integral_local(
...
       // Gather particle-fluid force, will be zero if they have not been
       // gathered.
-      auto pf_value = this->particle_fluid_force(cell, q);
+      auto pf_force_value = this->particle_fluid_force(cell, q);
+      auto pf_drag_value  = this->particle_fluid_drag(cell, q);
+      auto p_velocity     = this->particle_velocity(cell, q);
+      auto beta_momentum_transfer =
+        this->momentum_transfer_coefficient(cell, q);
+
+      // Add to source term the particle-fluid force and the  explicitdrag force
```
antonellasuccar:
```suggestion
      // Add to source term the particle-fluid force and the  explicit drag force
```

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -211,21 +317,32 @@ VANSOperator<dim, number>::do_cell_integral_local(
 
   for (const auto q : integrator.quadrature_point_indices())
     {
+      // Gather void fraction value and gradient
+      // We gather the void fraction force since we will use it in the forcing
+      // terms (beta and source terms).
```
antonellasuccar: What is the void fraction term? Aren't the forces stored in ``particle_fluid_force``, ``particle_fluid_drag``, etc.?

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -281,6 +393,8 @@ VANSOperator<dim, number>::do_cell_integral_local(
           value_result[dim] += vf_value * gradient[i][i];
           // +(q,∇ɛ·δu)
           value_result[dim] += vf_gradient[i] * value[i];
+          // +(v, βu)
```
antonellasuccar: Shouldn't be delta u instead of u?

source/fem-dem/particle_projector.cc
```diff
@@ -1158,13 +1155,18 @@ ParticleProjector<dim>::calculate_field_projection(
                                   volumetric_contribution]) :
                           particle_volume_in_sphere;
 
-
-
-                      for (int d = 0; d < n_components; ++d)
+                      if constexpr (n_components == 1)
+                        particle_field_in_sphere +=
+                          volumetric_contribution *
+                          particle_properties[property_start_index];
```
antonellasuccar, line 1162: What is the advantage of having the conditional type (double or vector) as compared to doing one iteration in the loop if there is one component?

source/fem-dem/vans_assemblers.cc
```diff
@@ -968,16 +968,24 @@ VANSAssemblerDiFelice<dim>::calculate_particle_fluid_interactions(
 
       if (cfd_dem.drag_coupling == Parameters::DragCoupling::fully_implicit ||
           cfd_dem.drag_coupling == Parameters::DragCoupling::semi_implicit)
-        beta_drag += momentum_transfer_coefficient;
+        {
+          particle_properties[DEM::CFDDEMProperties::PropertiesIndex::
+                                momentum_transfer_coefficient] =
+            momentum_transfer_coefficient;
+          beta_drag += momentum_transfer_coefficient;
```
antonellasuccar, line 975: I feel that now the code has two filters: the cell filter and the QCM one. Maybe after this progresses further, we can choose the filter and then choose either the ``beta_drag``(line 999) or the momentum transfer coefficient in the particle properties (this is more for the matrix-based case).

antonellasuccar (commented):
Very nice work!

CHANGELOG.md
blaisb: It says this is an outdated suggestion. Im not sure waht is the change proposed here I don't see it?

include/fem-dem/fluid_dynamics_vans_matrix_free_operators.h
blaisb: The projection itself, when in conservative mode, always gives fields for which the integral over the domain gives the sum of the field over all the particles, so yeah.

include/fem-dem/particle_projector.h
blaisb: It does, but we have never tested it really. It is maybe something for the future. I would rather implement it correctly now and when we want to do simplices we can check :). The only thing that does not work well for simplices now I think is the DEM.

include/solvers/navier_stokes_scratch_data.h
blaisb: No. I'll remove that initialization that's a good point.

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
blaisb: Yup we do. Every dof handler needs a different FeValues

include/solvers/navier_stokes_scratch_data.h
blaisb: I have fixed this.

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
blaisb: It should be delta p, great catch!

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
blaisb: Indeed!

source/fem-dem/particle_projector.cc
blaisb: because particle_field_in_sphere is a double if we have only one component and it is an array if we have 3, so the call is different (in one case we have a [d] and in the other one we don't. Hence this constexpr to havethe two variants :)

source/fem-dem/vans_assemblers.cc
blaisb: Yup, for the matrix-based case we indeed have two filters now. In fact, we could also have an arbitrary number of filters if we use a Wendland or a Gaussian function instead of the top-hat filter we have right now. that is something I will explore for sure once the matrix-free coupling is finalized!

include/fem-dem/fluid_dynamics_vans_matrix_free_operators.h
antonellasuccar: Maybe I would add it in the comment like the previous one?
```suggestion
  /// Table with correct alignment for vectorization to store the values of the
  /// particle-fluid drag. The drag force in this table is volumetric, resulting from the projection used.
```

source/fem-dem/vans_assemblers.cc
antonellasuccar: Same here!

source/fem-dem/particle_projector.cc
antonellasuccar: Ah true! I thought it was an array with one element if ``n_component==1`` before.

PR #1781, opened 2025-11-06
Clang tidy v20 fluid_dynamics_sharp.cc

voferreira:
Fix comparison between unsigned int and int, pre-allocate container size instead of pushing back, use sqrt2 instead of sqrt(2)


blaisb (approved):
Awesome. Thanks, super clear. Let's just wait for CI

voferreira:
@blaisb As we decided, we will keep this warning and try to fix it in another PR. Issue #1787

PR #1799, opened 2025-11-13
Refactor solve linear system

OresteMarquis:
### Description

Remove the deprecated parameter "renewed_matrix" in the definition of "solve_linear_sytem". I saw that it was deprecated from a TODO comment in the code. 

### Testing
 
This has been tested in debug and release, but nothing more.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/solvers/fluid_dynamics_block.cc
+++ b/source/solvers/fluid_dynamics_block.cc
@@ -385,12 +385,7 @@ FluidDynamicsBlock<dim>::setup_preconditioner()
     }
   else
     AssertThrow(
-      this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
-            .preconditioner ==
-          Parameters::LinearSolver::PreconditionerType::ilu ||
-        this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
-            .preconditioner ==
-          Parameters::LinearSolver::PreconditionerType::amg,
+      false,
       ExcMessage(
         "This linear solver does not support this preconditioner. Only <ilu> and <amg> preconditioners are supported."));
 }


source/solvers/fluid_dynamics_block.cc
```diff
@@ -370,7 +370,30 @@ FluidDynamicsBlock<dim>::assemble_system_rhs()
...
+  else
+    AssertThrow(
+      this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
+            .preconditioner ==
+          Parameters::LinearSolver::PreconditionerType::ilu ||
+        this->simulation_parameters.linear_solver.at(PhysicsID::fluid_dynamics)
+            .preconditioner ==
+          Parameters::LinearSolver::PreconditionerType::amg,
+      ExcMessage(
+        "This linear solver does not support this preconditioner. Only <ilu> and <amg> preconditioners are supported."));
```
blaisb, line 390:
```suggestion
    AssertThrow(
      false,
      ExcMessage(
        "This linear solver does not support this preconditioner. Only <ilu> and <amg> preconditioners are supported."));
```
You already have done the check before for that so :)

blaisb (approved):
Awesome, I just have one comment, nothing big.

Another thing that this PR made me realize. Once merge this, could you make a follow-up PR to remove the initial_step parameter? It seems like we never call the solver linear system with something else than initial_step=false. I think this is a remnant of the past.
Still better to do it in another PR since this one is already nice and close to merge :). Great work btw.

source/solvers/fluid_dynamics_block.cc
OresteMarquis: The initial_step is used to apply the dirichlet b.c. for the non linear system, which seems coherent to me for what i remember of GCH8108E. But I don't know how we apply them in Lethe, maybe it is done somewhere else and this becomes obsolete?

blaisb: Yeah it was done like that in the beggining, but now since we always have an initial condition, the initial condition applies the dirichlet boundary condition. I think this is nothing more than a relic from the past now :).
It will be worth it to delete it in a next PR :)

OresteMarquis: Perfect! I will open the issue and try to do it next week!

PR #1801, opened 2025-11-14
CFD-DEM matrix free solver

blaisb:
### Description

This PR introduces the matrix-free CFD-DEM solver. This solver is still in experimental form and will be refactored extensively in the near future. It should therefore not be used for production run.

The solver essentially uses the recent matrix free VANS solver in a coupled CFD-DEM fashion. Right now there is some (500ish lines) of code duplication with the regular CFD-DEM solver. However, I wanted to first introduce the solver and then refactor the two CFD-DEM solver at the same time in small pull requests. 

The solver works quite well actually. It is fast, performance is good. Now will just be a matter of revisiting the various test cases wth it.

- [x] Add the time derivative of the void fraction inside of the VANS solver

### Testing

Three tests are added

- [x] Implicit 1 particle sedimentation
- [x] Semi-Implicit 1 particle sedimentation
- [x] Explicit 1 particle sedimentation

### Documentation

TODO
- [x]  Document the existence of the solver in the application list

This is a really experimental solver. Step by step I hope it can become mature in a month or so.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/dem/insertion.cc
+++ b/source/dem/insertion.cc
@@ -1,7 +1,6 @@
 // SPDX-FileCopyrightText: Copyright (c) 2020-2025 The Lethe Authors
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
-#include "fem-dem/cfd_dem_coupling_matrix_free.h"
 #include <dem/insertion.h>
 #include <dem/insertion_file.h>
 #include <dem/insertion_list.h>
--- a/source/fem-dem/cfd_dem_coupling_matrix_free.cc
+++ b/source/fem-dem/cfd_dem_coupling_matrix_free.cc
@@ -20,7 +20,6 @@
 #include <dem/set_particle_wall_contact_force_model.h>
 #include <dem/velocity_verlet_integrator.h>
 #include <fem-dem/cfd_dem_coupling_matrix_free.h>
-#include <fem-dem/fluid_dynamics_vans_matrix_free.h>
 #include <fem-dem/fluid_dynamics_vans_matrix_free_operators.h>
 #include <fem-dem/postprocessing_cfd_dem.h>
 
@@ -1272,7 +1271,7 @@ template <int dim>
 void
 CFDDEMMatrixFree<dim>::dem_contact_build(unsigned int counter)
 {
-  // If this is not the lasts DEM iteration before the next CFD iteration, check
+  // If this is not the last DEM iteration before the next CFD iteration, check
   // if a contact detection step is necessary
   if (counter != (coupling_frequency - 1))
     check_contact_detection_method(counter);
@@ -1384,7 +1383,6 @@ CFDDEMMatrixFree<dim>::solve()
   initialize_dem_parameters();
 
   // Calculate first instance of void fraction once particles are set up
-  this->vertices_cell_mapping();
   if (!dem_action_manager->check_restart_simulation())
     this->particle_projector.initialize_void_fraction(
       this->simulation_control->get_current_time());
@@ -1416,7 +1414,6 @@ CFDDEMMatrixFree<dim>::solve()
       if (!this->simulation_control->is_at_start())
         {
           this->refine_mesh();
-          this->vertices_cell_mapping();
         }
 
       // We calculate the void fraction and the particle-fluid interaction using
--- a/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
+++ b/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
@@ -692,7 +692,7 @@ VANSOperator<dim, number>::local_evaluate_residual(
           gradient_result[dim] += tau * vf_value * gradient[dim];
 
           // Grad-div stabilization
-          // (∇·v,γ(ɛ∇·u+u·∇ɛ))
+          // (∇·v,γ(∂tɛ+ɛ∇·u+u·∇ɛ))
           if (cfd_dem_parameters.grad_div == true)
             {
               for (int i = 0; i < dim; ++i)
--- a/source/fem-dem/particle_projector.cc
+++ b/source/fem-dem/particle_projector.cc
@@ -57,9 +57,11 @@ ParticleFieldQCM<dim, n_components, component_start>::setup_dofs()
 
   particle_field_constraints.close();
 
-  // The particle field matrix sparsity pattern is always a Diagonal matrix
-  // meaning that the particle fields are never coupled with one another.
-  // We enforce this directly inside the sparsity pattern.
+  // The particle field matrix sparsity pattern is always a block diagonal
+  // matrix meaning that the components of the particle fields are never coupled
+  // with one another. We can enforce this directly inside the sparsity pattern.
+  // This makes it significantly cheaper to solve the linear systems associated
+  // with the projection.
   Table<2, DoFTools::Coupling> coupling_table(n_components, n_components);
   for (int i = 0; i < n_components; ++i)
     for (int j = 0; j < n_components; ++j)
--- a/include/fem-dem/fluid_dynamics_vans_matrix_free.h
+++ b/include/fem-dem/fluid_dynamics_vans_matrix_free.h
@@ -218,16 +218,6 @@ protected:
   initialize_GMG() override;
 
   /**
-   * @brief Create mapping between vertices and cells.
-   *
-   * Establishes the relationship between mesh vertices and the cells
-   * that contain them, which is needed for efficient void fraction
-   * and particle-fluid coupling calculations.
-   */
-  void
-  vertices_cell_mapping();
-
-  /**
    * @brief Evaluate the time derivative of the void fraction.
    * This evaluates the time-derivative as a solution vector which can be passed
    * to the matrix-free operator.
@@ -252,18 +242,6 @@ protected:
   bool           has_periodic_boundaries;
   Tensor<1, dim> periodic_offset;
   unsigned int   periodic_direction;
-
-
-  /// Mapping from vertex indices to sets of cells containing each vertex
-  std::map<unsigned int,
-           std::set<typename DoFHandler<dim>::active_cell_iterator>>
-    vertices_to_cell;
-
-  /// Mapping from vertex indices to sets of periodic cells containing each
-  /// vertex
-  std::map<unsigned int,
-           std::set<typename DoFHandler<dim>::active_cell_iterator>>
-    vertices_to_periodic_cell;
 
   /// Vector to store the time derivative of the void fraction
   VectorType time_derivative_void_fraction;
--- a/source/fem-dem/fluid_dynamics_vans_matrix_free.cc
+++ b/source/fem-dem/fluid_dynamics_vans_matrix_free.cc
@@ -624,21 +624,6 @@ FluidDynamicsVANSMatrixFree<dim>::gather_output_hook()
 
 template <int dim>
 void
-FluidDynamicsVANSMatrixFree<dim>::vertices_cell_mapping()
-{
-  // Find all the cells around each vertex
-  TimerOutput::Scope t(this->computing_timer, "Map vertices to cell");
-
-  LetheGridTools::vertices_cell_mapping(this->particle_projector.dof_handler,
-                                        vertices_to_cell);
-
-  if (has_periodic_boundaries)
-    LetheGridTools::vertices_cell_mapping_with_periodic_boundaries(
-      this->particle_projector.dof_handler, vertices_to_periodic_cell);
-}
-
-template <int dim>
-void
 FluidDynamicsVANSMatrixFree<dim>::evaluate_time_derivative_void_fraction()
 {
   this->time_derivative_void_fraction = 0;


include/fem-dem/cfd_dem_coupling_matrix_free.h
```diff
@@ -0,0 +1,499 @@
...
+
+  /// Container for particle interaction outcomes (forces and torques)
+  ParticleInteractionOutcomes<DEM::CFDDEMProperties::PropertiesIndex>
+    contact_outcome;
+
+  /// Reference to torque vector from contact outcomes
+  std::vector<Tensor<1, 3>> &torque = contact_outcome.torque;
+
+  /// Reference to force vector from contact outcomes
+  std::vector<Tensor<1, 3>> &force = contact_outcome.force;
```
antonellasuccar: Not sure why these are needed as members of the class if we have ``contact_outcome`` as a member

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+
+  if (this->cfd_dem_simulation_parameters.cfd_parameters.restart_parameters
+        .restart)
+    dem_action_manager->restart_simulation();
+
+  // Initialize the DEM parameters and generate the ghost particles
+  initialize_dem_parameters();
+
+  // Calculate first instance of void fraction once particles are set up
+  this->vertices_cell_mapping();
```
antonellasuccar: Is this till needed here? I think ``setup_dofs`` of ``ParticleProjector`` calls `` LetheGridTools::vertices_cell_mapping(this->dof_handler, vertices_to_cell);`` before calculating the void fraction in the next line. This function modifies the member variable ``vertices_to_cell``, but I am not sure where it is used next.

source/dem/insertion.cc
```diff
@@ -1,7 +1,12 @@
 // SPDX-FileCopyrightText: Copyright (c) 2020-2025 The Lethe Authors
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
+#include "fem-dem/cfd_dem_coupling_matrix_free.h"
```
antonellasuccar: Curious to know why we need this include and the subsequent includes here. Is it because of the dealii vector type in matrix-free?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+#include <dem/insertion_file.h>
+#include <dem/insertion_list.h>
+#include <dem/insertion_plane.h>
+#include <dem/insertion_volume.h>
+#include <dem/particle_handler_conversion.h>
+#include <dem/set_particle_particle_contact_force_model.h>
+#include <dem/set_particle_wall_contact_force_model.h>
+#include <dem/velocity_verlet_integrator.h>
+#include <fem-dem/cfd_dem_coupling_matrix_free.h>
+#include <fem-dem/fluid_dynamics_vans_matrix_free.h>
```
antonellasuccar: Isn't this file already included in ``cfd_dem_coupling_matrix_free.h``?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+{}
+
+template <int dim>
+void
+CFDDEMMatrixFree<dim>::setup_distribution_type()
+{
+  // Use namespace and alias to make the code more readable
+  using namespace Parameters::Lagrangian;
+  const unsigned int this_mpi_process =
+    Utilities::MPI::this_mpi_process(this->mpi_communicator);
```
antonellasuccar, line 40: Why isn't ``mpi_process`` a member like in ``CFDDEMSolver``?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+{
+  load_balancing.check_load_balance_iteration();
+
+  // If not a load balance iteration, exit the function
+  if (!dem_action_manager->check_load_balance())
+    return;
+
+  // Otherwise, we do not support load balancing at the present time, throw and
+  // exit gracefully.
+  AssertThrow(false, ExcMessage("Load balancing is currently not supportedi"));
```
antonellasuccar: Why not do this from the beginning without the previous if, since it is not supported anyway? Also, there is an extra "i" at the end.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+    contact_manager.get_local_ghost_periodic_adjacent_particles(),
+    contact_manager.get_ghost_local_periodic_adjacent_particles(),
+    dem_time_step,
+    contact_outcome);
+
+  // Particles-walls contact force:
+  particle_wall_contact_force();
+
+  // Add fluid-particle interaction force to the force container
+  add_fluid_particle_interaction();
```
antonellasuccar, line 1212: Should this condition be added inside ``add_fluid_particle_interaction``:
``// Add fluid-particle interaction torque to the torque container
  if (this->cfd_dem_simulation_parameters.cfd_dem.rotational_viscous_torque ||
      this->cfd_dem_simulation_parameters.cfd_dem.vortical_viscous_torque)``

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+    }
+
+  dem_action_manager->reset_triggers();
+}
+
+template <int dim>
+void
+CFDDEMMatrixFree<dim>::dem_contact_build(unsigned int counter)
+{
+  // If this is not the lasts DEM iteration before the next CFD iteration, check
```
antonellasuccar:
```suggestion
  // If this is not the last DEM iteration before the next CFD iteration, check
```

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+  // if a contact detection step is necessary
+  if (counter != (coupling_frequency - 1))
+    check_contact_detection_method(counter);
+
+  // Otherwise, force a contact search at the last DEM iteration before a CFD
+  // iteration to ensure that the particles are adequately located before
+  // calculating the coupling between particle and fluid.
+  else
+    dem_action_manager->last_dem_of_cfddem_iteration_step();
```
antonellasuccar, line 1283: Should this also be changed in the ``CFDDEMSolver`` class?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+         .restart)
+    read_dem();
+
+  this->computing_timer.leave_subsection("Read mesh, manifolds and particles");
+
+
+  this->setup_dofs();
+
+  this->particle_projector.initialize_void_fraction(
+    this->simulation_control->get_current_time());
```
antonellasuccar: Why is the void fraction initialized twice? If it is initialized here and the simulation is restarted, are the particles data read at this point?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+      if (!this->simulation_control->is_at_start())
+        {
+          this->refine_mesh();
+          this->vertices_cell_mapping();
+        }
+
+      // We calculate the void fraction and the particle-fluid interaction using
+      // the particle projector.
+      this->particle_projector.calculate_void_fraction(
+        this->simulation_control->get_current_time());
```
antonellasuccar, line 1422: Should we also make the same call in ``CFDDEMSolver`` directly instead of having a class function that makes this call?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+        NavierStokesScratchData<dim>(
+          this->simulation_control,
+          this->simulation_parameters.physical_properties_manager,
+          *this->fe,
+          *this->cell_quadrature,
+          *this->mapping,
+          *this->face_quadrature));
+
+      // The base matrix-free operator is not aware of the various VANS
+      // coupling term. We must do a cast here to ensure that the operator is
```
antonellasuccar:
```suggestion
      // coupling terms. We must do a cast here to ensure that the operator is
```

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -0,0 +1,1534 @@
...
+      // of the right type.
+      if (auto mf_operator = dynamic_cast<VANSOperator<dim, double> *>(
+            this->system_operator.get()))
```
antonellasuccar, line 1460: Why a dynamic cast?

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -684,7 +701,8 @@ VANSOperator<dim, number>::local_evaluate_residual(
                     {
                       gradient_result[i][i] +=
                         grad_div_gamma(cell, q) *
-                        (vf_value * gradient[k][k] + value[k] * vf_gradient[k]);
+                        (vf_value * gradient[k][k] + value[k] * vf_gradient[k] +
```
antonellasuccar, line 704: this term should be added in line 695:
∂tɛ+

source/fem-dem/particle_projector.cc
```diff
@@ -55,11 +57,22 @@ ParticleFieldQCM<dim, n_components, component_start>::setup_dofs()
 
   particle_field_constraints.close();
 
+  // The particle field matrix sparsity pattern is always a Diagonal matrix
+  // meaning that the particle fields are never coupled with one another.
```
antonellasuccar: Do you mean the particle field components?

source/fem-dem/particle_projector.cc
```diff
@@ -55,11 +57,22 @@ ParticleFieldQCM<dim, n_components, component_start>::setup_dofs()
 
   particle_field_constraints.close();
 
+  // The particle field matrix sparsity pattern is always a Diagonal matrix
```
antonellasuccar: You mean the matrix at a given node?

source/fem-dem/particle_projector.cc
```diff
@@ -847,6 +870,11 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
               sum_quadrature_weights += fe_values_void_fraction.JxW(q);
             }
 
+          Assert(
+            sum_quadrature_weights > 0,
+            ExcMessage(
+              "The sum of the quadrature weight should be strictly positive."));
```
antonellasuccar, line 878: In what case can this condition be broken?

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: The explanation is directly written one line above :)

source/fem-dem/particle_projector.cc
blaisb: I mean a diagonal block matrix when referring to the components. I will update that.

source/fem-dem/particle_projector.cc
blaisb: In no cases, but it was helpful during development and this costs nothing.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: I will check if it is used somewhere. I think it is used in the contact calculation

source/dem/insertion.cc
blaisb: That's a mistake. it should not have to be there.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: Let me double check if these includes are still necessary.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: I'm trying to move away from having members that are kinda trivial to get to make the classes as lightweight as possible. This is what we do in deal.II in general.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: Because for now we want to keep the logic of calling the function, but if someone enables load balancing it will abort gracefully. There's not an easy otherwise to check if load balancing is on or off other than if it's being called.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: It is absolutely  not useful. Removing it.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: We definitely should do that Indeed. In another PR :). You can open an issue if you want.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: Yes, it is not a dramatic impact though, just easier to read.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
blaisb: No. The torque is by default zero. We should add to the torque if we have something to add to it, but I'd rather we always sum it up instead of trying to avoid summing it sometiimes. These kind of premature optimization often lead to errors down the road and do not increase the performance in any way

PR #1949, opened 2026-03-19
DEM: adding multiple periodicity

karlwilson:
### Description

This PR enables multiple pairs of periodic boundaries for the DEM module, whereas before only one pair of periodic boundaries at a time was supported. This allows particles to traverse periodic corners of the domain at vertices where many periodic boundaries meet.

Most of this PR involves refactoring scalar quantities identifying attributes of a single periodic boundary pair to ``std::vector`` or ``std::unordered_map`` quantities for multiple periodic boundary pairs.

### Testing

Two new application tests for ``lethe-particles`` were added, simulating particle collisions through corners in a 2D periodic square and a 3D periodic cube.

### Documentation

The Boundary Conditions section for the DEM Parameters Guide has been updated to provide more details about the parameters for periodic boundaries, and now mentions support for multiple periodicity.

Code related list:

Pull request related list:
- [x] Labels are applied


CMakeFiles/cmake.check_cache
OGaboriault, line 1: Remove this file

include/core/parameters_lagrangian.h
```diff
@@ -808,14 +808,10 @@ namespace Parameters
...
-
-      /// Second periodic boundary ID.
-      types::boundary_id periodic_boundary_1;
-
-      /// Direction of periodicity.
-      types::boundary_id periodic_direction;
+      // Periodic boundary IDs
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_0;
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_1;
+      std::unordered_map<unsigned int, unsigned int>       periodic_direction;
```
OGaboriault: We talked about this, but I don't think this should be an unordered_map. Then again, due to the size of those containers, I don't think it will make a difference.

include/core/parameters_lagrangian.h
```diff
@@ -848,7 +844,8 @@ namespace Parameters
        * @param[in] prm The parameter handler.
        */
       void
-      parse_boundary_conditions(const ParameterHandler &prm);
+      parse_boundary_conditions(const ParameterHandler &prm,
+                                const unsigned int      i_bc);
```
OGaboriault, line 856: YOu need to add the @param[in] i_bc

include/core/parameters_lagrangian.h
OGaboriault, line 869: parameters description.

include/core/parameters_lagrangian.h
```diff
@@ -870,18 +867,13 @@ namespace Parameters
...
-        std::vector<unsigned int>                      &outlet_boundaries_id,
-        std::vector<BoundaryType>                      &boundaries_types) const;
+        std::unordered_map<unsigned int, types::boundary_id> &periodic_bndry_0,
+        std::unordered_map<unsigned int, types::boundary_id> &periodic_bndry_1,
+        std::unordered_map<unsigned int, unsigned int>       &periodic_dir,
+        std::vector<unsigned int> &outlet_boundaries_id,
+        std::vector<BoundaryType> &boundaries_types) const;
     };
 
-    /**
```
OGaboriault, line 877: I think you should keep this.

include/dem/data_containers.h
```diff
@@ -45,7 +45,7 @@ namespace DEM
    * @brief DEM data structure containers
    * To improve readability of containers, there's a simple description above
    * all declarations which follow these representation :
-   * <> : map or unordered map
+   * <> : map or unordered map or unordered multimap
```
OGaboriault, line 48: what do you mean by multimap? 

If you have a map inside a map, it should be <, <>>

include/dem/data_containers.h
```diff
@@ -191,9 +191,9 @@ namespace DEM
       cell_set;
 
     // <cell id, periodic cells info>
-    typedef ankerl::unordered_dense::
-      map<types::global_cell_index, periodic_boundaries_cells_info_struct<dim>>
-        periodic_boundaries_cells_info;
+    typedef std::unordered_multimap<types::global_cell_index,
```
OGaboriault, line 204: why is it no longer an ankerl map?

include/dem/particle_particle_contact_force.h
```diff
@@ -70,14 +72,23 @@ class ParticleParticleContactForceBase
...
+
+  inline void
+  set_combined_offsets(const std::vector<Tensor<1, dim>> &offsets)
+  {
+    this->combined_offsets = offsets;
   }
 
 protected:
-  Tensor<1, dim> periodic_offset;
+  std::unordered_map<types::boundary_id, Tensor<1, dim>> periodic_offsets;
```
OGaboriault, line 110: Again, I don't think this needs to be an unordered_map. 

Just a std::vector<Tensor<1,dim>> should work.

You just need to to periodic_offsets.at(b_id) to call the right Tensor<1,dim>

include/dem/particle_particle_contact_force.h
```diff
@@ -252,18 +264,21 @@ class ParticleParticleContactForce
...
+  get_periodic_location(const Particles::ParticleIterator<dim> &particle,
+                        const Tensor<1, 3> &periodic_offset) &
   {
     if constexpr (dim == 3)
-      return (particle->get_location() - this->periodic_offset);
+      return (particle->get_location() + periodic_offset);
 
     if constexpr (dim == 2)
-      return point_nd_to_3d(particle->get_location() - this->periodic_offset);
+      return point_nd_to_3d(particle->get_location()) + periodic_offset;
```
OGaboriault, line 310: You can remove the two  "if constexpr (dim == X)" and just keep the 
`` return point_nd_to_3d(particle->get_location()) + periodic_offset;`` 

since there is a constexpr inside the ``point_nd_to_3d`` function.

OGaboriault: FYI, I know is was like this before your PR.

include/dem/particle_particle_contact_force.h
```diff
@@ -1737,7 +1752,8 @@ class ParticleParticleContactForce
                       contact_type ==
                         ContactType::ghost_local_periodic_particle_particle)
           {
-            particle_two_location = get_periodic_location(particle_two);
+            particle_two_location =
+              get_periodic_location(particle_two, contact_info.periodic_offset);
```
OGaboriault, line 1784: I have just realised, but this means that we now track a new tensor insid the contact_info? This may affect performance.

@le-zouave, I can show you next week how to assess the code performance if you want.

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -89,22 +100,45 @@ class PeriodicBoundariesManipulator
...
+   * It is calculated from the first pair of cells on periodic boundaries,
+   * all pairs of cells on that boundary are assumed to have the same offset.
    *
+   * @param[in] boundary_id ID of the boundary to query
    * @return Offset distance between periodic boundaries.
    */
   inline Tensor<1, dim>
-  get_periodic_offset_distance()
+  get_periodic_offset_distance(const types::boundary_id boundary_id)
+  {
```
OGaboriault, line 116: Instead of checking if we reach the end of the vector, wouldn't be more efficient to fill the map for every boundary ids with ``Tensor<1, dim>()`` except or the key that need to return something different? 

By default, every boundary has a zero periodic_offsets. If they they are periodic, we change the value. (Tell me if this is clear or not)

source/dem/particle_particle_fine_search.cc
```diff
@@ -48,19 +50,44 @@ particle_particle_fine_search(
           auto &particle_two = adjacent_pair_information.particle_two;
 
           // Finding the properties of the particles in contact
-          Point<dim, double> particle_two_location =
-            particle_two->get_location() - periodic_offset;
+          Point<dim, double> particle_two_base_location =
```
OGaboriault: maybe real_location?

source/dem/particle_particle_fine_search.cc
```diff
@@ -48,19 +50,44 @@ particle_particle_fine_search(
...
+            particle_two->get_location();
 
-          // Finding distance
-          const double square_distance =
-            particle_one_location.distance_square(particle_two_location);
-          if (square_distance > neighborhood_threshold)
+          // Find minimum periodic distance squared
```
OGaboriault: I think you need to add comments here. I don't understand what you are doing.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -12,12 +12,17 @@
 #include <deal.II/fe/fe_q.h>
 #include <deal.II/fe/fe_values.h>
 
+#include <algorithm>
+#include <unordered_map>
+
+// TODO: data structure periodic_boundaries_cells_info must be defined
+// as std::unordered_multimap
```
OGaboriault: ahhh this is actually a thing.... I didn't know.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -133,6 +154,26 @@ PeriodicBoundariesManipulator<dim>::check_and_move_particles(
        &particles_in_cell,
   bool &particle_has_been_moved)
 {
+  // Retrieve correct offset for this specific boundary interaction.
```
OGaboriault, line 159: I don't understand what is going on here.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -178,6 +219,25 @@ PeriodicBoundariesManipulator<dim>::check_and_move_particles(
     }
 }
 
+template <int dim>
+void
+PeriodicBoundariesManipulator<dim>::compute_combined_offsets()
+{
+  this->combined_offsets.clear();
+  this->combined_offsets.push_back(Tensor<1, dim>()); // Initialized as zeros
```
OGaboriault: So the first Tensor is a zero tensor, why?

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -178,6 +219,25 @@ PeriodicBoundariesManipulator<dim>::check_and_move_particles(
...
+
+  for (auto const &[id, offset] : this->periodic_offsets)
+    {
+      size_t current_size = this->combined_offsets.size();
+
+      for (size_t i = 0; i < current_size; ++i)
+        {
+          this->combined_offsets.push_back(this->combined_offsets[i] + offset);
+          this->combined_offsets.push_back(this->combined_offsets[i] - offset);
```
OGaboriault: Why one is + offset and the other is - offset?

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -204,9 +266,7 @@ PeriodicBoundariesManipulator<dim>::execute_particles_displacement(
           auto cell          = boundaries_cells_content.cell;
           auto periodic_cell = boundaries_cells_content.periodic_cell;
 
-          // Check and execute displacement of particles crossing a periodic
-          // wall.
-          // Case when particle goes from periodic boundary 0 to 1
+          // Main cell (pb0 side)
```
OGaboriault: I would keep/adapt the lines of comment you remove.

OGaboriault: I feel that those comments made the code easier to follow. Just saying "Main cell (pbo side)" does not give the information about what is going on that main cell side.

OGaboriault (changes requested):
First round of comments. I will need to read it again because I got lost in the middle.

include/dem/data_containers.h
karlwilson: A multimap is a map that allows multiple entries for the same key, so one key -> one or more values. It is not a map inside a map, so I think the notation <> can still be used

include/dem/data_containers.h
karlwilson: ``ankerl::unordered_dense`` does not provide a stand-in for ``std::unordered_multimap``. A multimap is needed here because some cells can touch multiple periodic boundaries (corners of a periodic square, edges and corners of a periodic cube), and thus be associated to many ``periodic_boundaries_cells_info_struct``. 

If we want to keep using ``ankerl::unordered_dense::map``, maybe this could work:
```suggestion
    typedef ankerl::unordered_dense::map<types::global_cell_index,
      std::vector<periodic_boundaries_cells_info_struct<dim>>> periodic_boundaries_cells_info;
```

include/dem/particle_particle_contact_force.h
karlwilson: You want to index this vector with `i_bc`, the boundary condition subsection number? That would break, if for example you had 

  `subsection boundary condition 0`: periodic
  `subsection boundary condition 1`: outlet
  `subsection boundary condition 2`: periodic
  
Since there are 2 periodic BCs, the `periodic_offsets` vector should be of length 2. Then, you can't use non-sequential BC subsection numbers to index it. A map is better for using these numbers as non-sequential keys

include/core/parameters_lagrangian.h
```diff
@@ -795,6 +795,9 @@ namespace Parameters
       /// Boundary IDs designated as outlets.
       std::vector<unsigned int> outlet_boundaries;
 
+      /// Index of boundary conditions that are periodic
```
jatred:
```suggestion
      /// Index of boundary conditions that are periodic.
```

include/core/parameters_lagrangian.h
```diff
@@ -808,14 +811,12 @@ namespace Parameters
       /// Point on the rotational axis of each rotating boundary.
       std::unordered_map<unsigned int, Point<3>> point_on_rotation_axis;
 
-      /// First periodic boundary ID.
-      types::boundary_id periodic_boundary_0;
-
-      /// Second periodic boundary ID.
-      types::boundary_id periodic_boundary_1;
+      // Periodic boundary IDs
```
jatred:
```suggestion
      // Periodic boundary IDs.
```

include/core/parameters_lagrangian.h
```diff
@@ -808,14 +811,12 @@ namespace Parameters
...
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_0;
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_1;
 
-      /// Direction of periodicity.
-      types::boundary_id periodic_direction;
+      // Directions of periodicity
```
jatred:
```suggestion
      // Directions of periodicity.
```

include/core/parameters_lagrangian.h
```diff
@@ -846,9 +847,11 @@ namespace Parameters
        * @brief Parse boundary condition parameters for a single boundary.
        *
        * @param[in] prm The parameter handler.
+       * @param[in] i_bc Index of the boundary condition subsection
```
jatred:
```suggestion
       * @param[in] i_bc Index of the boundary condition subsection.
```

include/dem/contact_info.h
```diff
@@ -24,6 +24,10 @@ struct particle_particle_contact_info
   Particles::ParticleIterator<dim> particle_two;
   Tensor<1, 3>                     tangential_displacement;
   Tensor<1, 3>                     rolling_resistance_spring_torque;
+
+  // Store the translation vector that brings particle_two into contact
+  // with particle_one (default to zero for non-periodic)
+  Tensor<1, 3> periodic_offset;
 };
```
jatred, line 39: I would rewrite the `brief` so that you don't need to add a separate comment for the periodic_offset, to keep it visually similar to the other structures in this file

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
```diff
@@ -0,0 +1,116 @@
+# SPDX-FileCopyrightText: Copyright (c) 2020-2024 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
```
blaisb, line 22: Could you add a short description of the content of the test and what it does? How does one measure that the test is doing what it should be doing :)

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
```diff
@@ -0,0 +1,116 @@
...
+
+#---------------------------------------------------
+# Simulation Control
+#---------------------------------------------------
+
+subsection simulation control
+  set time step        = 1e-5
+  set time end         = 2
+  set log frequency    = 10000
+  set output frequency = 0
```
blaisb: Great :)

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
```diff
@@ -0,0 +1,116 @@
...
+  set list y              = 0.500, 0.975, 0.025, 0.975
+  set list z              = 0.000, 0.000, 0.000, 0.000
+  set list velocity x     = 1.000, 0.000, 0.000, 0.000
+  set list velocity y     = -1.000, 0.000, 0.000, 0.000
+  set list velocity z     = 0.000, 0.000, 0.000, 0.000
+  set list omega x        = 0.000, 0.000, 0.000, 0.000
+  set list omega y        = 0.000, 0.000, 0.000, 0.000
+  set list omega z        = 0.000, 0.000, 0.000, 0.000
+  set list diameters      = 0.050, 0.050, 0.050, 0.050
+  set list temperatures   = 0.000, 0.000, 0.000, 0.000
```
blaisb, line 104: Also great

applications_tests/lethe-particles/multiperiodic_collisions_3d.prm
```diff
@@ -0,0 +1,119 @@
+# SPDX-FileCopyrightText: Copyright (c) 2020-2024 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
+
```
blaisb, line 20: Same thing about a description of the test. What does it mean to test, how does one assess that it is working well, and so on and so forth :)

doc/source/parameters/dem/boundary_conditions.rst
```diff
@@ -44,22 +44,36 @@ In this subsection, the boundary conditions of the DEM simulation are defined. F
...
+      set type               = periodic
+
+      # ID of principal boundary
+      set periodic id 0      = 3
+
+      # ID of associated periodic boundary
+      set periodic id 1      = 2
+
+      # Direction along axis separating the periodic boundary pair
+      set periodic direction = 1
```
blaisb:
```suggestion
      # Direction normal to the periodic boundary faces (x=0, y=1, z=2)
      set periodic direction = 1
```

doc/source/parameters/dem/boundary_conditions.rst
```diff
@@ -44,22 +44,36 @@ In this subsection, the boundary conditions of the DEM simulation are defined. F
...
+* The ``type`` parameter specifies the type of the boundary condition. Acceptable types are: ``fixed_wall``, ``outlet``, ``rotational``, ``translational`` and ``periodic``. The default boundary condition type is ``fixed_wall``.
 
-.. note::
-        Only periodic boundaries which have co-linear normal vectors which align along one axis of the problem (e.g., x axis) are currently supported.
+* The ``periodic id 0`` and ``periodic id 1`` parameters specify the periodic boundaries ID for which the periodic boundary condition should be applied. By convention, ``periodic id 0`` should correspond to the boundary ID for which the boundary is further along the coordinate axis (this is the principal boundary).
 
-* The ``type`` parameter specifies the type of the boundary condition. Acceptable types are: ``fixed_wall``, ``outlet``, ``rotational``, ``translational`` and ``periodic``. The default boundary condition type is ``fixed_wall``.
+.. note::
+        Only periodic boundaries which have co-linear normal vectors which align along an axis of the problem (e.g., x axis) are currently supported. Multiple simultaneous periodic directions are supported.
```
blaisb, line 74: Nice that you added this.

include/dem/periodic_boundaries_manipulator.h
karlwilson: Changed this function for a `const` reference, now using `.at()`. This function is not called in non-periodic setups, and is assumed to be called for an existing `boundary_id` key

include/core/parameters_lagrangian.h
```diff
@@ -808,14 +811,12 @@ namespace Parameters
...
       std::unordered_map<unsigned int, Point<3>> point_on_rotation_axis;
 
-      /// First periodic boundary ID.
-      types::boundary_id periodic_boundary_0;
-
-      /// Second periodic boundary ID.
-      types::boundary_id periodic_boundary_1;
+      // Periodic boundary IDs
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_0;
+      std::unordered_map<unsigned int, types::boundary_id> periodic_boundary_1;
```
blaisb, line 818: Can you document both parameters individually
Like main periodic boundary and secondary periodic boundary for example.

include/core/parameters_lagrangian.h
blaisb:
```suggestion
      /// Directions of periodicity
```
For Doxygen purposes we generally keep ///

blaisb (changes requested):
I am still reviewing this. I will finish tomorrow :)

include/dem/data_containers.h
blaisb: Because he needs to store more than one value. Hence using a multimap. There's more than one info.

include/dem/data_containers.h
blaisb: A multimap is not a map within a map. A multimap is a map with one key, but multiple values. It's different from a map.

source/dem/periodic_boundaries_manipulator.cc
karlwilson: `boundaries_cells_content` contains a single cell on a periodic boundary and its associated periodic cell. The offset between these cells is obtained from `periodic_offsets` to displace particles across these cells. `relevant_offset` is a vector pointing from `pb0` to `pb1`, so we check on which side the current particles are and offset their position accordingly (+/- offset)

source/dem/periodic_boundaries_manipulator.cc
karlwilson: it's a default value. since `combined_offsets` is also used in non-periodic setups, the zero offset is the real position of the particle in contact

source/dem/periodic_boundaries_manipulator.cc
karlwilson: because a particle next to a periodic boundary can be next to `pb0` or `pb1`, we want check its images across all periodic directions

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
```diff
@@ -0,0 +1,122 @@
+# SPDX-FileCopyrightText: Copyright (c) 2020-2024 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# This DEM test checks the contact of 4 identical particles across the corners
+# of a fully periodic 2D square. A particle begins at the domain center and
+# moves diagonally towards the bottom-right corner, where it makes simultaneous
+# contact with the other particles in the other corners. The output particle
+# positions should be reproduced to validate this test.
```
blaisb: This is very nice, thank you!

include/dem/data_containers.h
blaisb: It's allright to use a multimap. This is not the biggest data structure. Worst case if we find that it affects performance, we can improve that down the road.

include/dem/particle_particle_contact_force.h
blaisb: I agree. If you want to use the boundary ID of the mesh as your index (which you should do), it's fine to use a map or an unordered map here. it's going to be a tiny one.
We avoid using vectors for boundary conditions now since this way we can directly use the boundary ID without any need for conversion.

include/dem/particle_particle_contact_force.h
```diff
@@ -239,31 +251,26 @@ class ParticleParticleContactForce
...
    */
   inline Point<3>
   get_location(const Particles::ParticleIterator<dim> &particle) &
   {
-    if constexpr (dim == 3)
-      return particle->get_location();
-
-    if constexpr (dim == 2)
-      return point_nd_to_3d(particle->get_location());
+    return point_nd_to_3d(particle->get_location());
```
blaisb: I don't think this change is a good idea. There's a non-negligible overhead with point_nd_to_3d since it generates a new Tensor<1,3> even if the starting one is a Tensor<1,3>. I would not change that part of the code.

Is there a reason for this change?

include/dem/particle_particle_contact_force.h
```diff
@@ -239,31 +251,26 @@ class ParticleParticleContactForce
...
-  get_periodic_location(const Particles::ParticleIterator<dim> &particle) &
+  get_periodic_location(const Particles::ParticleIterator<dim> &particle,
+                        const Tensor<1, 3> &periodic_offset) &
   {
-    if constexpr (dim == 3)
-      return (particle->get_location() - this->periodic_offset);
-
-    if constexpr (dim == 2)
-      return point_nd_to_3d(particle->get_location() - this->periodic_offset);
+    return point_nd_to_3d(particle->get_location()) + periodic_offset;
```
blaisb: Same here as above, why this change?

include/dem/particle_particle_contact_force.h
blaisb: Yeah, I don't think the consequences on performance will be very major however. I don't expect much here.

include/dem/particle_particle_fine_search.h
```diff
@@ -42,8 +44,8 @@ particle_particle_fine_search(
   typename DEM::dem_data_structures<dim>::adjacent_particle_pairs
     &adjacent_particles,
   const typename DEM::dem_data_structures<dim>::particle_particle_candidates
-                      &contact_pair_candidates,
-  const double         neighborhood_threshold,
-  const Tensor<1, dim> periodic_offset = Tensor<1, dim>());
+                                    &contact_pair_candidates,
+  const double                       neighborhood_threshold,
+  const std::vector<Tensor<1, dim>> &combined_offsets = {Tensor<1, dim>()});
```
blaisb: I would suggest a name that is slightly more explicit? combined_periodic_offsets maybe?

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -37,31 +39,42 @@ class PeriodicBoundariesManipulator
...
+        DEMActionManager::get_action_manager()
+          ->set_periodic_boundaries_enabled();
+
+        this->periodic_boundaries_ids = periodic_boundary_ids_0;
+        this->directions              = periodic_directions;
+        this->periodic_bc_index       = periodic_bc_ind;
+
+        // Initialize offset map
+        this->periodic_offsets.clear();
+      }
```
blaisb: What I would suggest is to use the reverse strategy here.

```suggestion
    if (periodic_boundary_ids_0.empty()) return;


        periodic_boundaries_enabled = true;

        // Communicate to the action manager that there are periodic boundaries
        DEMActionManager::get_action_manager()
          ->set_periodic_boundaries_enabled();

        this->periodic_boundaries_ids = periodic_boundary_ids_0;
        this->directions              = periodic_directions;
        this->periodic_bc_index       = periodic_bc_ind;

        // Initialize offset map
        this->periodic_offsets.clear();
      
```

It makes things a lot easier to understand in my opinion.

include/dem/read_mesh.h
```diff
@@ -34,6 +34,7 @@ read_mesh(const Parameters::Mesh   &mesh_parameters,
 template <int dim, int spacedim>
 void
 match_periodic_boundaries(Triangulation<dim, spacedim>        &triangulation,
-                          const Parameters::Lagrangian::BCDEM &bc_params);
+                          const Parameters::Lagrangian::BCDEM &bc_params,
+                          const std::vector<unsigned int> &periodic_bc_indices);
```
blaisb: Can you update the documentation of the function? 
Just to be sure, why do you need to pass an additional vector here? The periodic boundaries id are already contained within the bc_params no?

source/dem/dem.cc
```diff
@@ -130,15 +132,16 @@ DEMSolver<dim, PropertiesIndex>::setup_parameters()
         {
           periodic_boundaries_object.set_periodic_boundaries_information(
             parameters.boundary_conditions.periodic_boundary_0,
-            parameters.boundary_conditions.periodic_direction);
+            parameters.boundary_conditions.periodic_direction,
+            parameters.boundary_conditions.periodic_bc_index);
           break;
         }
     }
```
blaisb: I'm not sure I understand what you mean by your comment. So once you identify one periodic boundary condition, you set everything related to periodic and that's it? (hence the breaking?) I would be a bit more explicit in the comments.

source/dem/dem_contact_manager.cc
```diff
@@ -419,23 +419,23 @@ DEMContactManager<dim, PropertiesIndex>::execute_particle_particle_fine_search(
         local_local_periodic_adjacent_particles,
         local_contact_pair_periodic_candidates,
         neighborhood_threshold,
-        periodic_offset);
+        combined_offsets);
```
blaisb: definitely something like combined_periodic_offsets would be great here as a name.

source/dem/particle_particle_fine_search.cc
```diff
@@ -48,19 +50,47 @@ particle_particle_fine_search(
...
-          const double square_distance =
-            particle_one_location.distance_square(particle_two_location);
-          if (square_distance > neighborhood_threshold)
+          // Find minimum periodic distance squared between particle one and
+          // the set of periodic images of particle two. The nearest periodic
+          // image corresponds to a given periodic offset, indicating if the
+          // particles are in contact through a periodic edge, face, or corner.
+          double min_square_distance = std::numeric_limits<double>::max();
+          Tensor<1, dim> nearest_translation;
+
```
blaisb: I understand why you are doing this, but I wonder if this will have a big impact on performance, because essentially you are doing it for every particle.
I think the best approach would be to "flag" particles which may or may not be periodic. 99.9% of the particles in a simulation will not be periodic, so it makes no sense to do all of this substraction and work if the particle is potentially not periodic.

You could create a very small set that would say particle_is_potentially_periodic and if that is true, then you do the whole substraction and offset. If it is not true, then you don't.

The good thing is that this reaching into the set would only occur in the first loop, and not the second one. So that boolean variable would be constant for the second loop. So actually you could just do particle_two_real_location, calculate the current square distance, and then if it is a periodic particle, you could do the translation process to find if there's another distance.
We can talk about it if you want.

blaisb (changes requested):
Some more important comments. I understand a lot of work went into this, great work. There's one major issue however that will have a big impact on performance.

source/dem/periodic_boundaries_manipulator.cc
blaisb: I see, I think it's worth it to add that as a comment to be as explicit as possible.

source/dem/periodic_boundaries_manipulator.cc
blaisb: Yup, this is important especially with the diagonals. Still worth it to add a comment to explain exactly what you just wrote above :)

blaisb (changes requested):
Last few comments. I think this is very good work for a first PR, I'm very much impressed. Let's make sure performance is not affected by the new change to the fine contact search however.

include/dem/particle_particle_contact_force.h
OGaboriault: I think it is me who ask him to do it haha

include/dem/contact_info.h
```diff
@@ -24,6 +24,10 @@ struct particle_particle_contact_info
   Particles::ParticleIterator<dim> particle_two;
   Tensor<1, 3>                     tangential_displacement;
   Tensor<1, 3>                     rolling_resistance_spring_torque;
+
+  // Store the translation vector that brings particle_two into contact
+  // with particle_one (default to zero for non-periodic)
+  Tensor<1, 3> periodic_offset;
```
OGaboriault, line 38: So store the offset inside de particle-particle contact info...

include/dem/dem_contact_manager.h
```diff
@@ -186,16 +188,16 @@ class DEMContactManager
...
-   * periodic boundaries.
+   * @param[in] offsets Combined periodic offsets for periodic
+   * boundaries, used for determining periodic contacts
    */
   inline void
-  set_periodic_offset(const Tensor<1, dim> &offset)
+  set_combined_offsets(const std::vector<Tensor<1, dim>> &offsets)
   {
-    this->periodic_offset = offset;
+    this->combined_offsets = offsets;
```
OGaboriault: combined_periodic_offsets ?

include/dem/dem_contact_manager.h
```diff
@@ -383,7 +385,12 @@ class DEMContactManager
   typename DEM::dem_data_structures<dim>::cell_vector periodic_cells_container;
 
 private:
-  Tensor<1, dim> periodic_offset = Tensor<1, dim>();
+  /**
+   * @brief Storage for all precomputed periodic translation vectors.
```
OGaboriault: The precomputed periodic translation vectors are between what and what? Boundaries? particles in contact?

include/dem/particle_particle_contact_force.h
```diff
@@ -70,14 +72,23 @@ class ParticleParticleContactForceBase
...
-  void
-  set_periodic_offset(const Tensor<1, dim> &periodic_offset)
+  inline void
+  set_periodic_offset(const Tensor<1, dim>    &offset,
+                      const types::boundary_id boundary_id)
   {
-    this->periodic_offset = periodic_offset;
+    this->periodic_offsets[boundary_id] = offset;
+  }
+
```
OGaboriault, line 90: You need the description of the function here.

include/dem/particle_particle_contact_force.h
OGaboriault: oupsii

include/dem/particle_particle_contact_force.h
OGaboriault: We could have a new contact_info for the periodic contact?

source/dem/particle_particle_fine_search.cc
OGaboriault: I agree with Bruno. Maybe we could take inspiration from what we are doing in the ``update_fine_search_candidates.cc`` file.

OGaboriault (changes requested):

include/dem/particle_particle_contact_force.h
blaisb: No problem, that's why we have code reviews, It's life. Most likely a good part of the increase computational overhead comes from that changes. It means grenerating millions of Tensor<1,3>

include/dem/find_cell_neighbors.h
jatred: Should be 2020-2026, the space between 2020- and 2025 made it so the pre-commit.sh missed it

include/dem/particle_particle_contact_force.h
```diff
@@ -239,31 +251,26 @@ class ParticleParticleContactForce
   /**
    * @brief Get the location of the particle.
    *
-   * @param particle The particle to get the location from.
+   * @param[in] particle The particle to get the location from.
    */
```
jatred: Might be redundant, but I've seen other functions with similar structures to this one repeat the `brief` in the `return`, probably to tell the user that the output is not just a void.

```suggestion
   * @param[in] particle The particle to get the location from.
   * @return The location of the particle
  */
```

include/dem/particle_particle_contact_force.h
```diff
@@ -239,31 +251,26 @@ class ParticleParticleContactForce
...
 
   /**
-   * @brief Get the shifted location of the particle on the periodic boundary.
+   * @brief Get the shifted location of a particle on a periodic boundary.
    *
-   * @param particle The particle to get the location from.
+   * @param[in] particle The particle to get the location from.
+   * @param[in] periodic_offset The periodic offset with which to shift the
+   * location.
    */
```
jatred, line 301:
```suggestion
   * @param[in] particle The particle to get the location from.
   * @param[in] periodic_offset The periodic offset with which to shift the
   * location.
   * @return The shifted location of the particle
  */
```

include/dem/particle_particle_fine_search.h
jatred: I know it was already like this, but I would put `const` before `typename`. I'm pretty sure both ways are the same, but 'const typename' is used everywhere else.

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -89,22 +102,50 @@ class PeriodicBoundariesManipulator
...
+
+  /**
+   * @brief Return the combined periodic offsets
+   */
+  inline const std::vector<Tensor<1, dim>> &
+  get_combined_offsets() const
   {
-    return constant_periodic_offset;
+    return combined_offsets;
   }
```
jatred, line 158: I would add a 'return' next to the 'brief' for all new functions.

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -89,22 +102,50 @@ class PeriodicBoundariesManipulator
       &periodic_boundaries_cells_information);
 
   /**
-   * @brief Return the periodic offset distance, it is calculated from the first
-   * pair of cells on periodic boundaries, all pair of cells are assumed to
-   * have the same offset.
+   * @brief Return the periodic offset distance for a specific boundary ID.
+   * It is calculated from the first pair of cells on periodic boundaries,
```
jatred: What do you mean by 'It is calculated from the first pair of cells on periodic boundaries'?

Do you mean from the first pair of cells on the boundary specified in input? And if so, what is considered as the 'first pair of cells'?

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -89,22 +102,50 @@ class PeriodicBoundariesManipulator
...
+  inline const std::vector<Tensor<1, dim>> &
+  get_combined_offsets() const
   {
-    return constant_periodic_offset;
+    return combined_offsets;
   }
 
 private:
   /**
    * @brief Gets boundaries information related to the face at periodic
```
jatred:
```suggestion
   * @brief Get boundaries information related to the face at periodic
```

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -120,8 +161,7 @@ class PeriodicBoundariesManipulator
 
   /**
    * @brief Checks if particle is outside the cell, if so, modifies the
```
jatred:
```suggestion
   * @brief Check if particle is outside the cell, if so, modifies the
```

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -120,8 +161,7 @@ class PeriodicBoundariesManipulator
...
-   * location of the particle with the distance (offset) between the periodic
-   * faces.
+   * location of the particle using the specific offset for boundary.
```
jatred: Might need a new line to make it fit.

```suggestion
   * location of the particle using the specific offset for its associated boundary.
```

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -139,31 +179,54 @@ class PeriodicBoundariesManipulator
...
    */
-  types::boundary_id periodic_boundary_0;
+  std::unordered_map<unsigned int, types::boundary_id> periodic_boundaries_ids;
 
   /**
-   * @brief Direction of the periodic boundary, it is the perpendicular axis of
-   * the periodic boundaries.
+   * @brief Map storing offset distance between periodic boundaries, keyed by
+   * the boundary ID (pb0). It is calculated from the first pair of cells on
+   * periodic boundaries, all pair of cells are assumed to have the same offset.
```
jatred: The same offset on a unique periodic boundary, or do all periodic boundaries need to have the same offset and only differ in their direction?

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -139,31 +179,54 @@ class PeriodicBoundariesManipulator
          &particles_in_cell,
     bool &particle_has_been_moved);
 
+  /**
+   * @brief Computes the combined periodic offsets and stores them in
```
jatred:
```suggestion
   * @brief Compute the combined periodic offsets and store them in
```

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -192,6 +254,8 @@ PeriodicBoundariesManipulator<dim>::execute_particles_displacement(
 
   if (!periodic_boundaries_cells_information.empty())
     {
+      // Iterate over all entries. If using multimap, this correctly handles
+      // corner cells twice :once for x-boundary, once for y-boundary (in 2D)
```
jatred:
```suggestion
      // corner cells twice: once for x-boundary, once for y-boundary (in 2D)
```

jatred: I would write something like this to make it clear it is not limited to 2D.

```suggestion
      // Iterate over all entries. Multimap correctly handles corner cells
      // in higher dimensions. In 2D, a corner cell would be iterated over
      // twice: once for the x-boundary, and once for the y-boundary.
```

source/dem/particle_particle_fine_search.cc
jatred:
```suggestion
  const typename DEM::dem_data_structures<2>::particle_index_iterator_map
```

jatred (changes requested):
My review is 95% focused on layout/code aesthetic, I will focus on the code itself when the current comments from other reviewers have been resolved.

include/dem/periodic_boundaries_manipulator.h
karlwilson: Clarified the comment

include/dem/periodic_boundaries_manipulator.h
karlwilson: I meant that it is calculated from the first cell encountered on a given periodic boundary in the loop in ``map_periodic_cells``. I don't really need to mention that here, I'll remove it.

include/dem/contact_info.h
karlwilson: Added a derived struct for periodic contact info

source/dem/particle_particle_fine_search.cc
karlwilson: Modified this to have two branches in the loop, one for non-periodic contacts and another for periodic contacts

include/dem/read_mesh.h
karlwilson: You are right, there was redundancy in defining a new ``periodic_bc_indices`` here. Refactored this file to use the one from ``bc_params``

include/dem/force_chains_visualization.h
```diff
@@ -191,11 +191,11 @@ class ParticlesForceChains
    * @param[out] normal_forces_vector the vector of normal forces between each
    * touching particles.
    */
-  template <ContactType contact_type>
+  template <ContactType contact_type, typename ContactInfoContainer>
   inline void
   execute_contact_calculation(
-    typename DEM::dem_data_structures<dim>::particle_contact_info
-                          &adjacent_particles_list,
+    // typename DEM::dem_data_structures<dim>::particle_contact_info
```
OGaboriault:
```suggestion
```

include/dem/particle_particle_contact_force.h
```diff
@@ -1675,15 +1702,15 @@ class ParticleParticleContactForce
...
    * @param[in] dt DEM time step.
    * @param[out] contact_outcome Interaction outcomes.
    */
-  template <ContactType contact_type>
+  template <ContactType contact_type, typename ContactInfoContainer>
   inline void
   execute_contact_calculation(
-    typename DEM::dem_data_structures<dim>::particle_contact_info
-                                                 &adjacent_particles_list,
+    // typename DEM::dem_data_structures<dim>::particle_contact_info
```
OGaboriault:
```suggestion

```

OGaboriault (approved):
I think this is good to be merge when the tests will succeed on the CI. 

I would create an issue or something on the TODO list about what we talked about today @le-zouave .

applications_tests/lethe-fluid-particles/periodic_particles_qcm.prm
```diff
@@ -88,8 +88,9 @@ end
 
 subsection boundary conditions
   set number         = 5
+  set fix pressure constant = true
   set time dependent = true
-  set number         = 5
+  #set number         = 5
```
blaisb:
```suggestion
  set fix pressure constant = true
set time dependent = true
```

blaisb: I'm surprised you would need to fix the pressure constant here.  However, if the test results have not changed, that's a good sign. I'll try to run it on my machine to see what's up.

include/core/parameters_lagrangian.h
```diff
@@ -863,15 +868,20 @@ namespace Parameters
...
       void
       initialize_containers(
         std::unordered_map<unsigned int, Tensor<1, 3>> &boundary_trans_velocity,
         std::unordered_map<unsigned int, double>       &boundary_rot_speed,
         std::unordered_map<unsigned int, Tensor<1, 3>> &boundary_rot_vector,
         std::unordered_map<unsigned int, Point<3>>     &point_on_rot_axis,
-        std::vector<unsigned int>                      &outlet_boundaries_id,
-        std::vector<BoundaryType>                      &boundaries_types) const;
+        std::unordered_map<unsigned int, types::boundary_id> &periodic_bndry_0,
+        std::unordered_map<unsigned int, types::boundary_id> &periodic_bndry_1,
```
blaisb:
```suggestion
        std::unordered_map<unsigned int, types::boundary_id> &periodic_boundary_0,
        std::unordered_map<unsigned int, types::boundary_id> &periodic_boundary_1,
```

we try to spell out the words completely. it's a bit more verbose, but easier to re-read.

include/dem/particle_particle_contact_force.h
```diff
@@ -61,23 +63,40 @@ class ParticleParticleContactForceBase
...
 
-  void
-  set_periodic_offset(const Tensor<1, dim> &periodic_offset)
+  /**
+   * @brief Set the distance between a principal periodic boundary pn0 (with
+   * mesh ID boundary_id) and its associated periodic boundary pb1 in the
+   * periodic_offsets map.
+   *
+   * @param offset Distance between the pair of periodic boundaries.
+   * @param boundary_id Mesh ID of the principal periodic boundary of the pair.
```
blaisb:
```suggestion
   * @param[in] offset Distance between the pair of periodic boundaries.
   * @param[in] boundary_id Mesh ID of the principal periodic boundary of the pair.
```

include/dem/particle_particle_contact_force.h
```diff
@@ -61,23 +63,40 @@ class ParticleParticleContactForceBase
...
+   */
+  inline void
+  set_periodic_offset(const Tensor<1, dim>    &offset,
+                      const types::boundary_id boundary_id)
+  {
+    this->periodic_offsets[boundary_id] = offset;
+  }
+
+  inline void
+  set_combined_periodic_offsets(const std::vector<Tensor<1, dim>> &offsets)
```
blaisb, line 98: I know this is a very small function, but can you add the brief and documentation for it.

include/dem/particle_particle_contact_force.h
```diff
@@ -61,23 +63,40 @@ class ParticleParticleContactForceBase
...
   {
-    this->periodic_offset = periodic_offset;
+    this->combined_periodic_offsets = offsets;
   }
 
 protected:
-  Tensor<1, dim> periodic_offset;
+  std::unordered_map<types::boundary_id, Tensor<1, dim>> periodic_offsets;
```
blaisb, line 110: Can  you add /// comments for that member

include/dem/particle_particle_contact_force.h
```diff
@@ -61,23 +63,40 @@ class ParticleParticleContactForceBase
...
+
+  std::vector<Tensor<1, dim>> combined_periodic_offsets;
```
blaisb, line 116: Can you add /// description for that member

source/fem-dem/ib_particles_dem.cc
```diff
@@ -48,11 +48,19 @@ IBParticlesDEM<dim>::is_boundary_excluded(
...
+      if (std::ranges::any_of(pb0,
+                              [boundary_id](const auto &p) {
+                                return p.second == boundary_id;
+                              }) ||
+          std::ranges::any_of(pb1, [boundary_id](const auto &p) {
+            return p.second == boundary_id;
+          }))
+        return true;
+    }
   return false;
```
blaisb, line 64: This does not mean that the Sharp IB works with multiple periodic boundary however. But no worries. It does not work with periodics BCs at all right now if particles cross the domain

blaisb (changes requested):
Code looks good.
However, the pneumatic conveying example crashes in one of the map when using many cores in the first loading. So there's still some bug to iron out. I will start taking a look at it today and this weekend to help you debug it.

I want to be extra careful before we merge this. I think we will do release 1.1 next week before this PR is merged.

include/dem/particle_particle_contact_force.h
karlwilson: Should it be `///` comments or `/** @brief ... */` ?

include/core/parameters_lagrangian.h
karlwilson: this comment is on an outdated version of this file. I think you removed these containers in your current version?

include/dem/particle_particle_contact_force.h
blaisb: Both are fine. /// is good for single.

include/core/parameters_lagrangian.h
blaisb: Yup I fixed it while fixing the bug. Sorry about that :)

blaisb:
So removing the fix pressure node is preventing convergence again, but i really don't get why this has become necessary for this single case. I'll have to play around with it to see.

source/fem-dem/particle_projector.cc
```diff
@@ -776,7 +792,7 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
               // neighboring particle contributes to the solid volume of the
               // current reference sphere
               //***********************************************************************
-              for (unsigned int n = 0; n < active_neighbors.size(); n++)
+              for (unsigned int m = 0; m < active_neighbors.size(); m++)
```
blaisb, line 793: Why rename it from n to m ?

source/fem-dem/particle_projector.cc
```diff
@@ -735,16 +750,17 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
             LetheGridTools::find_cells_around_cell<dim>(
               vertices_to_periodic_cell, cell);
 
+
```
blaisb:
```suggestion
```

source/fem-dem/particle_projector.cc
```diff
@@ -829,17 +845,31 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
...
                       const Point<dim> particle_location =
-                        (active_periodic_neighbors[n]
-                           ->center()[periodic_direction] >
-                         cell->center()[periodic_direction]) ?
-                          particle.get_location() + periodic_offset :
-                          particle.get_location() - periodic_offset;
+                        (active_periodic_neighbors[m]->center()[neighbor_dir] >
+                         cell->center()[neighbor_dir]) ?
+                          particle.get_location() - neighbor_offset :
+                          particle.get_location() + neighbor_offset;
```
blaisb: Really not sur about this change.

blaisb (commented):
Some last small changes. I fixed the QCM issue.

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -139,31 +193,54 @@ class PeriodicBoundariesManipulator
...
   bool periodic_boundaries_enabled;
 
   /**
-   * @brief ID of the first periodic boundary. No needs to store the second one
-   * since there are linked on the triangulation, and accessible through
-   * functions on cells on the boundary condition 0.
+   * @brief Direction of the periodic boundaries, it is the perpendicular axis
+   * of the periodic boundaries.
+   */
+  std::unordered_map<unsigned int, unsigned int> directions;
```
jatred, line 216: Could be useful to specify in the brief what key the map uses, between the PB index from the .prm and the ID of the primary PB.

jatred (approved):
Everything looks good in my opinion, I changed some comments and layout stuff but aside from that really good and complete PR!

source/fem-dem/particle_projector.cc
karlwilson: For consistency with the other loops on `active_neighbors` in this file, that already had `m`. The `n` label is also being used to index quadrature weights in this file, so I thought of using separate labels for separate quantities

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
```diff
@@ -0,0 +1,139 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# This DEM test checks the simultaneous multi-pair periodic contact of a
+# moving particle with three stationary corner particles in a fully
+# periodic 2D unit square. Particle 0 starts near (0.85, 0.15) and moves
+# diagonally toward the (0.975, 0.025) corner. Just before reaching the
+# domain boundary it makes simultaneous contact with:
+#  - particle 1 at (0.025, 0.975) via the (+x, -y) corner offset,
+#  - particle 2 at (0.025, 0.025) via the +x face offset,
+#  - particle 3 at (0.975, 0.975) via the -y face offset.
```
karlwilson, line 11: I think the signs of these offsets should be inverted; they are defined in `periodic_boundaries_manipulator.cc` as being directed from pb0 to pb1. For example, the offset that brings particle 0 into contact with particle 2 is `(-x, 0)`.

karlwilson (commented):
These tests look good. I think this PR can be merged, and revisited for further improvements later.

applications_tests/lethe-particles/multiperiodic_collisions_2d.prm
blaisb: Yeah that's true I messed that up. Should be the reverse.

PR #1962, opened 2026-04-01
Skip unnecessary Sharp fluid GMRES solves below tolerance

fjbarter:
### Description

This PR improves the robustness of Sharp resolved CFD-DEM when the coupled outer nonlinear iteration remains active because of the particle residual even though the fluid subsystem is already converged to the requested nonlinear tolerance.

The motivation is that, in Sharp, the outer residual is not just the fluid residual. `FluidDynamicsSharp::get_current_residual()` returns the maximum of:

- the assembled fluid RHS norm, and
- a scaled particle-coupling residual

As a result, it is possible for the freshly assembled fluid residual to already be below the nonlinear tolerance while the outer coupled iteration still needs to continue because the particle residual has not yet converged. Previously, the Newton and inexact-Newton strategies still called `solve_linear_system()` in that situation. In the problematic coarse Sharp cases, this launched GMRES on an almost-homogeneous fluid system with a very small RHS and could fail with AztecOO `GMRES Hessenberg ill-conditioned`.

The integration is intentionally minimal, and does not affect solvers besides Sharp (although I propose that it may be worthwhile for the others - I am just not anywhere near qualified to say for sure...):

- `PhysicsSolver` now exposes an explicit internal hook, `allow_skip_linear_solve_when_residual_is_below_tolerance()`, which defaults to `false`
- `FluidDynamicsSharp` overrides that hook and returns `true`
- `NewtonNonLinearSolverStrategy::solve()` now checks the freshly assembled fluid residual immediately before the linear solve, but only for solvers that opt in through that hook
- `InexactNewtonNonLinearSolverStrategy::solve()` applies the same logic
- if the assembled fluid residual is already below the nonlinear tolerance, the code skips the fluid linear solve, zeroes the Newton update, preserves the current fluid state, and continues the outer coupled iteration so that the particle residual can still converge

This PR does not change the Sharp convergence criterion. It only avoids an unnecessary inner fluid linear solve once the fluid subsystem is already within the configured nonlinear tolerance.

### Testing

This PR adds one minimal Sharp application test:

- `applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm`
- `applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.run_only`

This test is based on the existing 2D moving Stokes Sharp case and is designed to trigger the exact case described:

- a relatively loose fluid nonlinear tolerance of `1e-4`
- a tighter Sharp particle tolerance of `1e-6`
- a short 2D run with no output files

There should be no negative impact on existing tests. The branch does not add new required parameters and only changes the behaviour when the assembled fluid residual is already below the nonlinear tolerance.

### Documentation

This PR does not introduce new simulation parameters.

No parameter documentation changes were needed. The user-facing change is instead reflected in the changelog

The below-tolerance linear-solve skip is now explicitly scoped to Sharp rather than applied generically to every Newton-based solver. This keeps the fix aligned with the original motivation:

- in Sharp, the outer nonlinear residual can remain active because of particle coupling even when the assembled fluid residual is already below tolerance
- for other solvers, the default behaviour is unchanged because the new internal hook returns `false`

Code related list:

Pull request related list:
- [ ] Labels are applied

### DIFF
--- a/applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
+++ b/applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
@@ -1,0 +1,263 @@
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
+
+set dimension = 2
+
+#---------------------------------------------------
+# Simulation Control
+#---------------------------------------------------
+
+subsection simulation control
+  set method            = bdf1 # Time-stepping method
+  set time step         = 1    # Time step
+  set number mesh adapt = 2    # If steady, nb mesh adaptation
+  set time end          = 1    # End time of simulation
+  set output frequency  = 0    # Frequency of simulation output
+end
+
+#---------------------------------------------------
+# Physical Properties
+#---------------------------------------------------
+
+subsection physical properties
+  subsection fluid 0
+    set kinematic viscosity = 1
+    set density             = 1
+  end
+end
+
+#---------------------------------------------------
+# Mesh
+#---------------------------------------------------
+
+subsection mesh
+  set type               = dealii
+  set grid type          = subdivided_hyper_rectangle
+  set grid arguments     = 2,1: -2,-1 : 2 , 1  : true
+  set initial refinement = 3
+end
+
+#---------------------------------------------------
+# FEM
+#---------------------------------------------------
+
+subsection FEM
+  set velocity order = 1
+  set pressure order = 1
+end
+
+#---------------------------------------------------
+# Boundary Conditions
+#---------------------------------------------------
+
+subsection boundary conditions
+  set number         = 4
+  set time dependent = true
+  subsection bc 0
+    set id   = 0
+    set type = function
+    subsection u
+      set Function expression = (-y^2*(0.375*(y^2 + (x - cos(t))^2)^(3/2) + 0.03125*sqrt(y^2 + (x - cos(t))^2) - (y^2 + (x - cos(t))^2)^2) + (x - cos(t))^2*(-0.75*(y^2 + (x - cos(t))^2)^(3/2) + 0.0625*sqrt(y^2 + (x - cos(t))^2) + (y^2 + (x - cos(t))^2)^2) - (y^2 + (x - cos(t))^2)^3)*sin(t)/(y^2 + (x - cos(t))^2)^3
+    end
+    subsection v
+      set Function expression = y*(x - cos(t))*(-0.375*y^2 - 0.375*(x - cos(t))^2 + 0.09375)*sin(t)/(y^2 + (x - cos(t))^2)^(5/2)
+    end
+    subsection w
+      set Function expression = 0
+    end
+  end
+  subsection bc 1
+    set id   = 3
+    set type = function
+    subsection u
+      set Function expression = (-y^2*(0.375*(y^2 + (x - cos(t))^2)^(3/2) + 0.03125*sqrt(y^2 + (x - cos(t))^2) - (y^2 + (x - cos(t))^2)^2) + (x - cos(t))^2*(-0.75*(y^2 + (x - cos(t))^2)^(3/2) + 0.0625*sqrt(y^2 + (x - cos(t))^2) + (y^2 + (x - cos(t))^2)^2) - (y^2 + (x - cos(t))^2)^3)*sin(t)/(y^2 + (x - cos(t))^2)^3
+    end
+    subsection v
+      set Function expression = y*(x - cos(t))*(-0.375*y^2 - 0.375*(x - cos(t))^2 + 0.09375)*sin(t)/(y^2 + (x - cos(t))^2)^(5/2)
+    end
+    subsection w
+      set Function expression = 0
+    end
+  end
+  subsection bc 2
+    set id   = 2
+    set type = function
+    subsection u
+      set Function expression = (-y^2*(0.375*(y^2 + (x - cos(t))^2)^(3/2) + 0.03125*sqrt(y^2 + (x - cos(t))^2) - (y^2 + (x - cos(t))^2)^2) + (x - cos(t))^2*(-0.75*(y^2 + (x - cos(t))^2)^(3/2) + 0.0625*sqrt(y^2 + (x - cos(t))^2) + (y^2 + (x - cos(t))^2)^2) - (y^2 + (x - cos(t))^2)^3)*sin(t)/(y^2 + (x - cos(t))^2)^3
+    end
+    subsection v
+      set Function expression = y*(x - cos(t))*(-0.375*y^2 - 0.375*(x - cos(t))^2 + 0.09375)*sin(t)/(y^2 + (x - cos(t))^2)^(5/2)
+    end
+    subsection w
+      set Function expression = 0
+    end
+  end
+  subsection bc 3
+    set id   = 1
+    set type = function weak
+    set beta = 10000
+
+    subsection u
+      set Function expression = (-y^2*(0.375*(y^2 + (x - cos(t))^2)^(3/2) + 0.03125*sqrt(y^2 + (x - cos(t))^2) - (y^2 + (x - cos(t))^2)^2) + (x - cos(t))^2*(-0.75*(y^2 + (x - cos(t))^2)^(3/2) + 0.0625*sqrt(y^2 + (x - cos(t))^2) + (y^2 + (x - cos(t))^2)^2) - (y^2 + (x - cos(t))^2)^3)*sin(t)/(y^2 + (x - cos(t))^2)^3
+    end
+    subsection v
+      set Function expression = y*(x - cos(t))*(-0.375*y^2 - 0.375*(x - cos(t))^2 + 0.09375)*sin(t)/(y^2 + (x - cos(t))^2)^(5/2)
+    end
+    subsection w
+      set Function expression = 0
+    end
+    subsection p
+      set Function expression = -0.75*(x - cos(t))*sin(t)/(y^2 + (x - cos(t))^2)^(3/2)
+    end
+  end
+end
+
+#---------------------------------------------------
+# Analytical Solution
+#---------------------------------------------------
+
+subsection analytical solution
+  set enable    = true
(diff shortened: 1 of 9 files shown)


applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+    set ilu preconditioner absolute tolerance     = 1e-11
+    set ilu preconditioner relative tolerance     = 1
+    set amg aggregation threshold                 = 1e-20
+    set amg n cycles                              = 1
+    set amg preconditioner ilu absolute tolerance = 1e-20
+    set amg preconditioner ilu fill               = 0
+    set amg preconditioner ilu relative tolerance = 1.00
+    set amg smoother overlap                      = 1
+    set amg smoother sweeps                       = 2
+    set amg w cycles                              = false
```
Luckabarbeau: Avoid AMG in sharp IB in general. Usually unstable.

include/core/inexact_newton_non_linear_solver_strategy.h
```diff
@@ -112,6 +112,28 @@ InexactNewtonNonLinearSolverStrategy<VectorType>::solve()
                         << "  - Residual:  " << current_res << std::endl;
         }
 
+      {
+        const double assembled_res =
+          solver->get_system_rhs().l2_norm() / rescale_metric;
+        if (solver
+              ->allow_skip_linear_solve_when_residual_is_below_tolerance() &&
+            assembled_res <= this->params.tolerance)
+          {
```
Luckabarbeau: Nice! This is the right way to do it since between two linear solutions, the residual can change significantly due to a large change in particle velocity. Re-evaluating the residual before here gives the change to the solver to have a proper update of the particle velocity taken into account.

Luckabarbeau (commented):
Hello @fjbarter.

Nice update!

This can help in some cases and even speed up stuff! I can see a case where this happens since the residual of the particle is lagging in the update by one newton iteration due to how the coupling is done. So the check of the RHS before the solve is a neat trick!

I will simply note one thing that is related but not directly linked to your update. If residual particles stagnate while the residual of the RHS solver converges in the log, it usually means that some dof are changing status between iterations. This can happen when you couple positions implicitly (default behavior). What happens is that you have particle boundary change side of dof between iterations. When this happens, your particle residual can oscillate since the cell uses to evaluate the forces change, while the RHS solution can stay almost exactly the same (especially in laminar cases). One solution is to use the explicit position integration option instead. 

Cheers,

Lucka

blaisb:
I am away this week. I will review this next week.

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+
+  subsection extrapolation function
+    set stencil order = 6
+    set length ratio  = 2
+  end
+
+  subsection input file
+    set load particles from file = false
+    set particles file           = particles.input
+  end
```
antonellasuccar: I am not sure that this subsection is needed here

include/core/inexact_newton_non_linear_solver_strategy.h
```diff
@@ -96,7 +96,9 @@ InexactNewtonNonLinearSolverStrategy<VectorType>::solve()
           solver->setup_preconditioner();
         }
 
-      if (this->params.force_rhs_calculation || outer_iteration == 0)
+      const bool rhs_was_assembled =
+        this->params.force_rhs_calculation || outer_iteration == 0;
```
antonellasuccar: ``rhs_was_assembled`` is not super suitable for its first use here. Maybe this is clearer?
```suggestion
      const bool assemble_rhs =
        this->params.force_rhs_calculation || outer_iteration == 0;
      
      if (assemble_rhs)
        solver->assemble_system_rhs();
            
      const bool rhs_was_assembled = assemble_rhs;
```

antonellasuccar (approved):
Nicely done :)

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+set dimension = 2
+
+#---------------------------------------------------
+# Simulation Control
+#---------------------------------------------------
+
+subsection simulation control
+  set method            = bdf1 # Time-stepping method
+  set time step         = 1    # Time step
+  set number mesh adapt = 2    # If steady, nb mesh adaptation
```
blaisb:
```suggestion
```
This parameter is not used in this application test.

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
blaisb:
```suggestion
```
Not used so not needed.

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+  subsection input file
+    set load particles from file = false
+    set particles file           = particles.input
+  end
+
+  subsection local mesh refinement
+    set initial refinement                = 0
+    set refine mesh inside radius factor  = 111
+    set refine mesh outside radius factor = 111
+  end
```
blaisb, line 160:
```suggestion
```

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+
+  # Maximum number of elements
+  set max number elements = 100000
+
+  # Type of mesh adaptation. Choices are <none|uniform|adaptive>.
+  set type = none
+
+  # Variable for kelly estimationChoices are <velocity|pressure>.
+  set variable = velocity
+end
```
blaisb:
```suggestion
```

If the section is not used we'd rather not put it inside the prm file.

applications_tests/lethe-fluid-sharp/moving_stokes_2d_gmres_skip.prm
```diff
@@ -0,0 +1,263 @@
...
+subsection linear solver
+  subsection fluid dynamics
+    set method                                    = gmres
+    set max iters                                 = 1000
+    set relative residual                         = 1e-6
+    set minimum residual                          = 1e-9
+    set preconditioner                            = amg
+    set ilu preconditioner fill                   = 2
+    set ilu preconditioner absolute tolerance     = 1e-11
+    set ilu preconditioner relative tolerance     = 1
```
blaisb:
```suggestion
```

When using AMG as a preconditioner, let's avoid putting ILU parameters.

include/core/inexact_newton_non_linear_solver_strategy.h
```diff
@@ -112,6 +114,34 @@ InexactNewtonNonLinearSolverStrategy<VectorType>::solve()
...
+            assembled_res <= this->params.tolerance)
+          {
+            // This shortcut is only valid when system_rhs was freshly
+            // assembled on this Newton iteration. Otherwise assembled_res may
+            // still reflect an older evaluation point, and skipping the solve
+            // from that stale residual would be incorrect.
+            //
+            // Solvers that opt into this path are expected to keep
+            // force_rhs_calculation enabled, or otherwise guarantee that the
+            // RHS was explicitly reassembled before reaching this check.
```
blaisb: If you expect a behavior like this, it's better to AssertThrow here.
For example AssertThrow(force_rhs_calculation, Message). If we expect something, we try to never take it for granted.

include/core/inexact_newton_non_linear_solver_strategy.h
```diff
@@ -112,6 +114,34 @@ InexactNewtonNonLinearSolverStrategy<VectorType>::solve()
                         << "  - Residual:  " << current_res << std::endl;
         }
 
+      {
+        const double assembled_res =
+          solver->get_system_rhs().l2_norm() / rescale_metric;
+        if (rhs_was_assembled &&
+            solver
+              ->allow_skip_linear_solve_when_residual_is_below_tolerance() &&
+            assembled_res <= this->params.tolerance)
```
blaisb: can you document better the content of the scope starting at 124. Explain a bit more clearly what is done here. Since that block is also reused in the newton_non_linear_solver then I would advice making a small function out of it that would get inlined, it would be easier to understand this way since you could give it an explicit name and document it accordingly.

include/core/newton_non_linear_solver_strategy.h
```diff
@@ -97,6 +99,34 @@ NewtonNonLinearSolverStrategy<VectorType>::solve()
...
+            // force_rhs_calculation enabled, or otherwise guarantee that the
+            // RHS was explicitly reassembled before reaching this check.
+            current_res         = assembled_res;
+            auto &newton_update = solver->get_newton_update();
+            newton_update       = 0;
+
+            global_res       = solver->get_current_residual() / rescale_metric;
+            present_solution = evaluation_point;
+            last_res         = current_res;
+            ++this->outer_iteration;
```
blaisb: Logic is identical in the other non-linear solver. I'd wrap this part in a small function and document it better.

blaisb (changes requested):
Nothing too big in terms of changes. I am not sure if that feature is extremely useful outside of the sharp solver, but I don't mind bringing it. Let's just try to make it a bit less invasive by refactoring this part into a function that is aptly named and well documented.

Could you also make sure that the test prm file is pruned of the parameters which are not altered?

Thanks for the work!

blaisb:
dear @fjbarter Could you please address the comments in the PR ? I would like this change to not go orphaned.

fjbarter:
Addressed all inline comments.
The duplicated skip logic is now factored into a shared function with better naming and documentation, and it now asserts the required force rhs calculation.
The test .prm was also tidied by removing unnecessary sections

fjbarter:
Is that failing test actually a problem? It looks to me like a build machine issue and should probably be rerun, but I might be wrong?

blaisb:
> Is that failing test actually a problem? It looks to me like a build machine issue and should probably be rerun, but I might be wrong?

It's a CI issue. We have some tests that get flaky sometimes.

blaisb (approved):
Good to go. Merging this.

PR #1967, opened 2026-04-14
Mortar: Generalize linear velocity computation for 3D case

brunaccampos:
### Description

This PR fixes the computation of cell radial distance for 3D mortar cases, and improves the indentification of the rotor/stator domains by using a material_id() flag (as opposed to comparing the radial distance of the cell center with the computed interface radius). The linear velocity computation is also generalized for 3D cases with rotation axis other than z.

### Solution

The computation of the radial distance for 3D cases is fixed, and the corresponding function has been added in `lethe_grid_tools`.

### Testing

No changes in tests.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/core/lethe_grid_tools.h
+++ b/include/core/lethe_grid_tools.h
@@ -402,6 +402,45 @@ namespace LetheGridTools
   };
 
   /**
+   * @brief Transform angular velocity into linear velocity considering that the
+   * angular velocity @p omega is constant in all directions.
+   *
+   * @param[in] omega Angular velocity value
+   * @param[in] point Coordinates of the point in which the linear velocity is
+   * being calculated
+   * @param[in] rotation_axis Rotation axis used in 3D cases, so that the
+   * angular velocity vector is given by omega_vec = omega * rotation_axis
+   *
+   * @return Linear velocity at @p point
+   */
+  template <int dim>
+  inline Tensor<1, dim>
+  angular_to_linear_velocity(
+    const double          omega,
+    const Tensor<1, dim> &point,
+    const Tensor<1, dim> &rotation_axis = Tensor<1, dim>())
+  {
+    Tensor<1, dim> result;
+
+    if constexpr (dim == 2)
+      {
+        result[0] = -omega * point[1];
+        result[1] = omega * point[0];
+      }
+
+    if constexpr (dim == 3)
+      {
+        // Store angular velocity according to unit vector that defines the
+        // rotation axis
+        const auto omega_vec = omega * rotation_axis;
+        // Compute cross product
+        result = cross_product_3d(omega_vec, point);
+      }
+
+    return result;
+  }
+
+  /**
    * @brief Rotate the mapping according to the rotation angle in a rotor-stator configuration
    *
    * @param[in] dof_handler DofHandler of the triangulation
--- a/include/core/lethe_grid_tools.h
+++ b/include/core/lethe_grid_tools.h
@@ -407,7 +407,6 @@ namespace LetheGridTools
    * @param[in] dof_handler DofHandler of the triangulation
    * @param[in,out] mapping_cache Rotated mapping
    * @param[in] mapping Current mapping
-   * @param[in] radius Radius or the rotor domain which is going to be rotated
    * @param[in] rotation_angle Rotation angle of the rotor domain
    * @param[in] center_of_rotation Center of rotation of the rotor domain.
    * Default is the origin
@@ -418,7 +417,6 @@ namespace LetheGridTools
   rotate_mapping(const DoFHandler<dim> &dof_handler,
                  MappingQCache<dim>    &mapping_cache,
                  const Mapping<dim>    &mapping,
-                 const double          &radius,
                  const double          &rotation_angle,
                  const Point<dim>      &center_of_rotation = Point<dim>(),
                  const Tensor<1, dim>  &rotation_axis      = Tensor<1, dim>());
--- a/include/core/lethe_grid_tools.h
+++ b/include/core/lethe_grid_tools.h
@@ -422,22 +422,6 @@ namespace LetheGridTools
                  const Tensor<1, dim>  &rotation_axis      = Tensor<1, dim>());
 
   /**
-   * @brief Computes the radial distance between a given point and a center of rotation
-   *
-   * @param[in] point Current point from which the radial distance is computed
-   * @param[in] rotation_axis Rotation axis of the domain
-   * @param[in] center_of_rotation Center of rotation of the domain
-   *
-   * @return Radial distance
-   */
-  template <int dim>
-  double
-  compute_radial_distance_3d(
-    const Point<dim>     &point,
-    const Tensor<1, dim> &rotation_axis,
-    const Point<dim>     &center_of_rotation = Point<dim>());
-
-  /**
    * @brief
    * A functor that provides a way to check if two cells are the same cell.
    * Used to store cells in hash sets.
--- a/include/core/lethe_grid_tools.h
+++ b/include/core/lethe_grid_tools.h
@@ -424,6 +424,22 @@ namespace LetheGridTools
                  const Tensor<1, dim>  &rotation_axis      = Tensor<1, dim>());
 
   /**
+   * @brief Computes the radial distance between a given point and a center of rotation
+   *
+   * @param[in] point Current point from which the radial distance is computed
+   * @param[in] rotation_axis Rotation axis of the domain
+   * @param[in] center_of_rotation Center of rotation of the domain
+   *
+   * @return Radial distance
+   */
+  template <int dim>
+  double
+  compute_radial_distance_3d(
+    const Point<dim>     &point,
+    const Tensor<1, dim> &rotation_axis,
+    const Point<dim>     &center_of_rotation = Point<dim>());
+
+  /**
    * @brief
    * A functor that provides a way to check if two cells are the same cell.
    * Used to store cells in hash sets.
--- a/include/core/lethe_grid_tools.h
+++ b/include/core/lethe_grid_tools.h
@@ -406,10 +406,11 @@ namespace LetheGridTools
    * angular velocity @p omega is constant in all directions.
    *
    * @param[in] omega Angular velocity value
-   * @param[in] point Coordinates of the point in which the linear velocity is
+   * @param[in] point Coordinates of the point at which the linear velocity is
    * being calculated
    * @param[in] rotation_axis Rotation axis used in 3D cases, so that the
-   * angular velocity vector is given by omega_vec = omega * rotation_axis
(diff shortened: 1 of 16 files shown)


blaisb:
You might need to update one of the test results,. I would double check what has changed.

source/core/grids.cc
```diff
@@ -363,6 +363,15 @@ read_mesh_and_manifolds_for_stator_and_rotor(
...
   attach_grid_to_triangulation(rotor_temp_tria, *mortar_parameters.rotor_mesh);
 
+  // Add a flag to cells on the rotor and stator triangulations; this will be
+  // needed when attributing them an angular velocity. We use the material_id
+  // flag because it won't be erased when the triangulations are merged.
+  for (const auto &cell : rotor_temp_tria.active_cell_iterators())
+    cell->set_material_id(1);
+
+  for (const auto &cell : stator_temp_tria.active_cell_iterators())
+    cell->set_material_id(2);
```
blaisb:
```suggestion
  for (const auto &cell : rotor_temp_tria.active_cell_iterators())
    cell->set_material_id(1);

  for (const auto &cell : stator_temp_tria.active_cell_iterators())
    cell->set_material_id(0);
```
Isnt the default material id 0?
I would maybe suggest to use 0 for the stator and 1 for the rotor? 
Right now when we do conjugate heat transfer, we say that the solid start at 1. So here if we have mortar enabled, we will just be able to say the solid start at 2 :)

source/core/lethe_grid_tools.cc
```diff
@@ -1821,20 +1821,19 @@ LetheGridTools::rotate_mapping(const DoFHandler<dim> &dof_handler,
...
-          // Compute point radial distance with respect to the rotation axis
-          const auto aux =
-            cross_product_3d((cell->center() - center_of_rotation),
-                             rotation_axis);
-          const double point_radial_distance =
-            aux.norm() / rotation_axis.norm();
-
-          if (point_radial_distance > radius)
+          // Do not rotate point if current cell is part of the stator
+          if (cell->material_id() == 2)
```
blaisb:
```suggestion
          if (cell->material_id() == 0)
```

source/core/lethe_grid_tools.cc
```diff
@@ -1800,7 +1799,8 @@ LetheGridTools::rotate_mapping(const DoFHandler<dim> &dof_handler,
         mapping,
         dof_handler.get_triangulation(),
         [&](const auto &cell, const auto &point) {
-          if ((cell->center() - center_of_rotation).norm() > radius)
+          // Do not rotate the point if current cell is part of the stator
+          if (cell->material_id() == 2)
```
blaisb:
```suggestion
          if (cell->material_id() == 0)
```

source/core/lethe_grid_tools.cc
```diff
@@ -1821,20 +1821,19 @@ LetheGridTools::rotate_mapping(const DoFHandler<dim> &dof_handler,
...
-          return static_cast<Point<dim>>(
+          // Shift point by the center of rotation
+          const auto shift_point = point - center_of_rotation;
+          // Rotate
+          const auto rotate_point =
             Physics::Transformations::Rotations::rotation_matrix_3d(
               rotation_axis, rotation_angle) *
-            point);
+            shift_point;
+          // Returnn rotated point according to the center of rotation
```
blaisb:
```suggestion
          // Return rotated point according to the center of rotation
```

source/core/lethe_grid_tools.cc
```diff
@@ -1853,11 +1851,34 @@ template void
...
 
+template <int dim>
+double
+LetheGridTools::compute_radial_distance_3d(const Point<dim>     &point,
+                                           const Tensor<1, dim> &rotation_axis,
+                                           const Point<dim> &center_of_rotation)
+{
+  // Compute distance =  ||VC x d||/||d||, where d is the rotation axis and VC
+  // is the distance between the current point and the center of rotation
+  const auto axis = rotation_axis / rotation_axis.norm();
```
blaisb:
```suggestion
  Assert(rotation_axis.norm()>0)
  const auto axis = rotation_axis / rotation_axis.norm();
```
Just to be safe.

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -1188,31 +1188,46 @@ NavierStokesOperatorBase<dim, number>::evaluate_velocity_ale(
...
-          else
-            cell_rotor_angular_velocity = rotor_angular_velocity_value;
+          // Use prescribed rotor angular velocity and verify if cells are part
+          // of the rotor (material_id = 1) or the stator (material_id = 2)
+          double omega;
+          if (matrix_free.get_cell_iterator(cell, lane)->material_id() == 1)
+            omega = rotor_angular_velocity_value;
+          else if (matrix_free.get_cell_iterator(cell, lane)->material_id() ==
+                   2)
+            omega = 0.0;
```
blaisb:
```suggestion
          double omega=0.;
          if (matrix_free.get_cell_iterator(cell, lane)->material_id() == 1)
            omega = rotor_angular_velocity_value;
            
           // The mortar implementation is not defined more than 2 materials.
           // Throw when running debug mode as a security measure.
          Assert(matrix_free.get_cell_iterator(cell, lane)->material_id()<2)
```
This way if we have more than one material it won't be undefined.
Better to have a default of zero, but still have a defensive assertion mechanism :)

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -410,38 +410,50 @@ template <int dim>
...
-    cell_rotor_angular_velocity = 0.0;
-  else
-    cell_rotor_angular_velocity = rotor_angular_velocity;
+  // Use prescribed rotor angular velocity and verify if cells are part of the
+  // the rotor (material_id = 1) or the stator (material_id = 2)
+  double omega;
+  if (cell->material_id() == 1)
+    omega = rotor_angular_velocity;
+  else if (cell->material_id() == 2)
+    omega = 0.0;
```
blaisb: ### 
```suggestion
  double omega=0;
  if (cell->material_id() == 1)
    omega = rotor_angular_velocity;
   // The mortar interface has currently not been tested for more than 2 zones
   // Defensively assert for this scenario to prevent this case.
  Assert(cell->material_id() <2)
```

blaisb (changes requested):
Very good work. This should be a lot more robust
I only have a suggestion. I think the main material id is 0 in general. I would suggest we number them 0 as the stator and 1 as the rotor and so, eventually if we have solid with mortar, they will start at 2 instead of 1 and everything would be fine.
Otherwise, great work :)!

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -1188,31 +1188,51 @@ NavierStokesOperatorBase<dim, number>::evaluate_velocity_ale(
...
+                  // defines the rotation axis
+                  const auto omega_vec = omega * rotation_axis;
+
+                  // Compute terms of u_ale = omega_vec x [x, y, z]
+                  this->velocity_ale[cell][q][0][lane] =
+                    omega_vec[1] * z - omega_vec[2] * y;
+                  this->velocity_ale[cell][q][1][lane] =
+                    omega_vec[2] * x - omega_vec[0] * z;
+                  this->velocity_ale[cell][q][2][lane] =
+                    omega_vec[0] * y - omega_vec[1] * x;
```
hepap: Could you use the `cross_product_3d` function instead here?

source/solvers/navier_stokes_scratch_data.cc
```diff
@@ -410,38 +410,56 @@ template <int dim>
...
+          const auto omega_vec = omega * mortar_parameters.rotation_axis;
+
+          // Compute terms of u_ale = omega_vec x [x, y, z]
+          rotor_linear_velocity_values[q][0] =
+            omega_vec[1] * p[2] - omega_vec[2] * p[1];
+          rotor_linear_velocity_values[q][1] =
+            omega_vec[2] * p[0] - omega_vec[0] * p[2];
+          rotor_linear_velocity_values[q][2] =
+            omega_vec[0] * p[1] - omega_vec[1] * p[0];
+        }
```
hepap: Do you want to add this as a function in lethe_grid_tools to avoid repetition?

source/core/lethe_grid_tools.cc
```diff
@@ -1853,11 +1851,39 @@ template void
...
                                MappingQCache<3>    &mapping_cache,
                                const Mapping<3>    &mapping,
-                               const double        &radius,
                                const double        &rotation_angle,
                                const Point<3>      &center_of_rotation,
                                const Tensor<1, 3>  &rotation_axis);
 
+template <int dim>
+double
+LetheGridTools::compute_radial_distance_3d(const Point<dim>     &point,
```
hepap: This could be name maybe `compute_point_to_ray_distance` or `compute_point_to_axis_distance` because it can be to other things than radial distance :)

brunaccampos: That makes sense! I saw that actually I can use the `find_point_line_distance` function that's already implemented in `lethe_grid_tools`, so I will change that

source/solvers/navier_stokes_scratch_data.cc
brunaccampos: Done!

blaisb (approved):
I like this new implementation. It is much better than the previous one.

blaisb:
I'll wait for a changelog rebase and the OK from @hepap @brunaccampos and then this is good to go on my side :)

jatred (approved):
Looks good to me!

include/core/lethe_grid_tools.h
```diff
@@ -401,13 +401,51 @@ namespace LetheGridTools
     }
   };
 
+  /**
+   * @brief Transform angular velocity into linear velocity considering that the
+   * angular velocity @p omega is constant in all directions.
+   *
+   * @param[in] omega Angular velocity value
+   * @param[in] point Coordinates of the point in which the linear velocity is
```
hepap:
```suggestion
   * @param[in] point Coordinates of the point at which the linear velocity is
```

include/core/lethe_grid_tools.h
```diff
@@ -401,13 +401,51 @@ namespace LetheGridTools
...
+   * being calculated
+   * @param[in] rotation_axis Rotation axis used in 3D cases, so that the
+   * angular velocity vector is given by omega_vec = omega * rotation_axis
```
hepap:
```suggestion
   * @param[in] rotation_axis Rotation axis used in 3D cases, so that the
   * angular velocity vector is given by omega_vec = omega * rotation_axis.
   * In 2D, we apply a rotation around the z-axis.
```

hepap (approved):
I like the new implementation also :)

PR #1968, opened 2026-04-14
Square channel mesh

OresteMarquis:
### Description

 This PR add a new mesh to simulate a channel with an object in it. Here the object is a rectangular prism with a square base. The name of this new mesh is "uniform_channel_with_meshed_square_prism". This will be used for benchmark and test the multiphysic microwave-heating capabilities of Lethe. The square prism can rotate and the size of its side can be adjusted in addition to the channel dimension.

### Testing

I produced the following meshes to test: 
- 0 degree:
Screenshot from 2026-04-14 16-32-27
- 45 degree: 
Screenshot from 2026-04-14 16-33-27
- 65 degree: 
Screenshot from 2026-04-14 16-34-01
- 65 degree, with smaller square and 2 level of refinement: 
Screenshot from 2026-04-14 16-35-26

### Documentation

The explanation of the parameters will be made in a future example, but the mesh is added to the relevant section in documentation.

Code related list:

Pull request related list:
- [x] Labels are applied


blaisb:
Could you add an application test which calculates something like the force on the square cylinder or something like that so we have a way to test the mesh? :) (or a unit test would be even better :))

OresteMarquis:
What do you want to test for the mesh specifically with the unit test? Because it could be simply use in an example to show that it's working? We don't really play with custom Lethe mesh no?

doc/source/parameters/cfd/mesh.rst
```diff
@@ -66,6 +66,7 @@ This subsection provides information of the simulation geometry and its mesh. Th
...
     * ``fichera_oven``: if this grid type is chosen a mesh for the Fichera oven time-harmonic Maxwell benchmark is created. More details on this type of mesh and its grid arguments will be provided in the future example.
     * ``uniform_channel_with_meshed_cylinder``: if this grid type is chosen a mesh for a channel (similar to the deal.II ``uniform_channel_with_cylinder``) is created. The difference is the cylinder in the channel is meshed enabling the calculation of fields inside the cylinder which is useful for multiphysics simulations. More details on this type of mesh and its grid arguments will be provided in a future example.
+    * ``uniform_channel_with_meshed_square_prism``: This mesh is similar to the ``uniform_channel_with_meshed_cylinder`` but uses a square prism, that can be rotated instead of a cylindrical obstacle.  More details on this type of mesh and its grid arguments will be provided in a future example.
```
blaisb: Can you at least provide the grid argument in the documentation. Otherwise it's not useable right now without reading the code. Same for the boundary IDs.

doc/source/parameters/cfd/mesh.rst
```diff
@@ -66,6 +66,7 @@ This subsection provides information of the simulation geometry and its mesh. Th
     * ``periodic_hills``: if this grid type is chosen a mesh for the periodic hills CFD benchmark is created. For more details on this type of mesh and its grid arguments refer to :doc:`../../examples/incompressible-flow/3d-flow-over-periodic-hills/3d-flow-over-periodic-hills`.
     * ``fichera_oven``: if this grid type is chosen a mesh for the Fichera oven time-harmonic Maxwell benchmark is created. More details on this type of mesh and its grid arguments will be provided in the future example.
     * ``uniform_channel_with_meshed_cylinder``: if this grid type is chosen a mesh for a channel (similar to the deal.II ``uniform_channel_with_cylinder``) is created. The difference is the cylinder in the channel is meshed enabling the calculation of fields inside the cylinder which is useful for multiphysics simulations. More details on this type of mesh and its grid arguments will be provided in a future example.
```
blaisb: Can you also add at least a quick 1-line description of the grid arguments here like I did for the Birgminham reactor. Otherwise, it's impossible to use the grid right now.  Even if it's a quick thing generated by AI, at least the usage of the grid will be documented. Same for the boundary IDs. .

include/core/grid_uniform_channel_with_meshed_square_prism.h
```diff
@@ -0,0 +1,190 @@
...
+   * |  4 | outer_half_side | double     | yes      | Half-side of the outer transition square                             |
+   * |  5 | rotation_deg    | double     | no       | In-plane rotation (degrees) applied to the inner region (default: 0) |
+   * |  6 | pad_bottom      | int        | no       | N cells below the transition (default: 0)                            |
+   * |  7 | pad_top         | int        | no       | N cells above the transition (default: 0)                            |
+   * |  8 | pad_left        | int        | no       | N cells left of the transition (default: 0)                          |
+   * |  9 | pad_right       | int        | no       | N cells right of the transition (default: 0)                         |
+   * | 10 | height          | double     | no       | Extrusion height in z, 3D only (default: 1.0)                        |
+   * | 11 | n_slices        | int        | no       | Number of z-layers, 3D only (default and min: 2)                     |
+   * | 12 | colorize        | true/false | no       | Assign distinct boundary IDs (default: false)                        |
```
blaisb, line 65: Wow that's a brutal number of arguments. Hahaha

tests/core/lethe_grid_generator.output
```diff
@@ -0,0 +1,261 @@
...
+DEAL::Beginning
+DEAL::Case 1: Unsupported specialization
+DEAL::  Caught expected exception: 
+--------------------------------------------------------
+An error occurred in file <uniform_channel_with_meshed_square_prism_grid.cc> in function
+    UniformChannelWithMeshedSquarePrismGrid<dim, spacedim>::UniformChannelWithMeshedSquarePrismGrid(const std::string&) [with int dim = 2; int spacedim = 3; std::string = std::__cxx11::basic_string<char>]
+Additional information: 
+    The uniform channel with a square prism obstacle is only supported in
+    3D space with 3D elements.
+--------------------------------------------------------
```
blaisb: I'm not sure it's necessary to test the exception in the unit test. I think it's better to test the expected usage and not the unexpected. :)

blaisb (changes requested):
I started the review, I will finish this weekend. Sorry about the delay.

include/core/grid_uniform_channel_with_meshed_square_prism.h
blaisb, line 1: Could you rename the file to grid_uniform_channel_with_meshed_Square_prism.h 
so that all the grids follow a similar monicker?
uniform_channel_with_meshed_cylinder_grid
to
grid_uniform_channel_with_meshed_cylinder


I think it makes it more structured if we put a grid as a prefix so that they are all at the same location. It follows the Birgminham grid pattern a bit more and I like that.

include/core/grid_uniform_channel_with_meshed_square_prism.h
```diff
@@ -0,0 +1,190 @@
...
+   * @brief Generate the 2D/3D channel mesh with a meshed square obstacle.
+   *
+   * For 2D: Delegates geometry construction to generate_2d_channel_mesh().
+   * The boundary IDs are assigned following the subdivided_hyper_rectangle
+   * convention when colorize is enabled: 0 (-x), 1 (+x), 2 (-y), 3 (+y).
+   *
+   * For 3D: Delegates geometry construction to generate_2d_channel_mesh(), then
+   * extrudes the mesh along the z-axis and attaches the new boundary IDs. The
+   * lateral boundary IDs are inherited from the 2D mesh when colorize is
+   * enabled. The extruded bottom (z = 0) gets boundary_id = 4 and the top (z =
```
blaisb, line 87:
```suggestion
   * enabled. The extruded front (z = 0) gets boundary_id = 4 and the back (z =
```

include/core/uniform_channel_with_meshed_square_prism_grid.h
```diff
@@ -0,0 +1,190 @@
...
+                           const double       outer_half_side,
+                           const double       inner_rotation_angle,
+                           const unsigned int pad_bottom,
+                           const unsigned int pad_top,
+                           const unsigned int pad_left,
+                           const unsigned int pad_right,
+                           const bool         colorize);
+
+  /// The original input string with the geometry parameters.
+  std::string grid_arguments;
```
blaisb: Do those need to be kept as a member variable? i'm not sure we need to keep a duplicate of the parameters in both the string and the parse parameters internally.

source/core/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ add_library(lethe-core
   tracer_diffusivity_model.cc
   tracer_reaction_model.cc
   uniform_channel_with_meshed_cylinder_grid.cc
+  uniform_channel_with_meshed_square_prism_grid.cc
```
blaisb: Let's rename these two files with grid_ as a the first word instead.
This way grid will be the prefix and all our grids will be at the same location. 

```suggestion
grid_uniform_channel_with_meshed_cylinder.cc
  grid_uniform_channel_with_meshed_square_prism.cc
```

source/core/CMakeLists.txt
```diff
@@ -107,6 +108,7 @@ add_library(lethe-core
   ../../include/core/tracer_diffusivity_model.h
   ../../include/core/tracer_reaction_model.h
   ../../include/core/uniform_channel_with_meshed_cylinder_grid.h
+  ../../include/core/uniform_channel_with_meshed_square_prism_grid.h
```
blaisb:
```suggestion
../../include/core/grid_uniform_channel_with_meshed_cylinder.h
  ../../include/core/grid_uniform_channel_with_meshed_square_prism.h
```

source/core/grid_uniform_channel_with_meshed_cylinder.cc
```diff
@@ -280,19 +280,25 @@ UniformChannelWithMeshedCylinderGrid<dim, spacedim>::generate_2d_channel_mesh(
...
-                                       &pad_bottom_right_corner_tria,
-                                       &pad_top_left_corner_tria,
-                                       &pad_top_right_corner_tria},
-                                      triangulation);
-
+  // Merge only non-empty triangulations
+  std::vector<const Triangulation<2> *> trias = {&cylinder_tria, &box_tria};
+  if (pad_bottom > 0)
+    trias.push_back(&pad_bottom_tria);
+  if (pad_top > 0)
```
blaisb, line 275: Nice, great idea. I had not thought about the zero padding case

tests/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,165 @@
...
+    Triangulation<dim>                                triangulation;
+    UniformChannelWithMeshedSquarePrismGrid<dim, dim> grid(grid_arguments);
+    grid.make_grid(triangulation);
+
+    if (global_refinement > 0)
+      triangulation.refine_global(global_refinement);
+
+    deallog << name << std::endl;
+    deallog << "Active cells: " << triangulation.n_active_cells() << std::endl;
+    print_faces(triangulation);
```
blaisb: Instead of printing the face, what I would suggest is instead to use a GridOut and print the grid inside of the log. it would make for a simpler and more complete test that reuses regular deal.II functionnality
You can follow the idea there which I think is a good one:
https://github.com/dealii/dealii/blob/master/tests/grid/grid_generator_04.cc#L76

The ucd format is quite human readable. this will prevent you from having to use a local function and instead it outputs the whole grid. Just make sure you won't generate too many cells for no reason. Or you could also use the VTK format which is also quite human readable (and easy to see in paraview)
In essence, try to use a grid_out and write that to the deal log :)

source/core/uniform_channel_with_meshed_square_prism_grid.cc
```diff
@@ -0,0 +1,647 @@
...
+        ExcMessage(
+          "The uniform channel with a square prism obstacle is only supported in 2D and 3D space."));
+      return;
+    }
+  else if constexpr (dim == 2 && spacedim == 3)
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "The uniform channel with a square prism obstacle is only supported in 3D space with 3D elements."));
```
blaisb:
```suggestion
        ExcMessage(
           "UniformChannelWithMeshedSquarePrismGrid is only supported for <2,2> and <3,3> <dim,spacedim> specializations."));
```

I feel this is more accurate, because it's supported in both 2D and 3D, just with dim=spacedim.

source/core/uniform_channel_with_meshed_square_prism_grid.cc
```diff
@@ -0,0 +1,647 @@
...
+template <int dim, int spacedim>
+UniformChannelWithMeshedSquarePrismGrid<dim, spacedim>::
+  UniformChannelWithMeshedSquarePrismGrid(const std::string &grid_arguments)
+{
+  if constexpr (dim == 1 || spacedim == 1)
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "The uniform channel with a square prism obstacle is only supported in 2D and 3D space."));
```
blaisb:
```suggestion
           "UniformChannelWithMeshedSquarePrismGrid is only supported for <2,2> and <3,3> <dim,spacedim> specializations."));
```

source/core/uniform_channel_with_meshed_square_prism_grid.cc
```diff
@@ -0,0 +1,647 @@
...
+  else if constexpr (dim == 2 && spacedim == 3)
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "The uniform channel with a square prism obstacle is only supported in 3D space with 3D elements."));
+      return;
+    }
+
+  this->grid_arguments = grid_arguments;
```
blaisb: Why do you need to make the grid_arguments a member of the class? :)

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+  // Parse bottom_left point
+  std::stringstream   bottom_left_stream(arguments[0]);
+  std::vector<double> bottom_left_coords;
+  std::string         coord_str;
+  while (getline(bottom_left_stream, coord_str, ','))
+    {
+      bottom_left_coords.push_back(Utilities::string_to_double(coord_str));
+    }
```
blaisb, line 53: If you want this could be made into a small inlined function in utilities (parse_string_to_point) and this would prevent you copy-pasting this block of code 7 times. We might want to reuse that feature in the future anyway.

source/core/uniform_channel_with_meshed_square_prism_grid.cc
```diff
@@ -0,0 +1,647 @@
...
+                "The outer transition square does not fit in the channel."));
+
+  // Parse optional parameters if provided
+  this->inner_rotation_angle =
+    (arguments.size() > 5) ? Utilities::string_to_double(arguments[5]) : 0.0;
+
+  AssertThrow(
+    (inner_rotation_angle >= 0.0) && (inner_rotation_angle < 90.0),
+    ExcMessage(
+      "The rotation angle needs to be in the range [0, 90) degrees. Different values will result in the same mesh pattern due to symmetry."));
```
blaisb:
```suggestion
      "The rotation angle needs to be in the range [0, 90) degrees. "));
```
Just be as brief as possible. An alert reader will understand the notion of symmetry here, but if you use more than 90 degrees you are crashing the generator, so it's not giving the same results as a symmetry, it is terminating the program :) .

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,647 @@
...
+  // obstacle by moving the vertices. The outer ring of the circular mesh is
+  // used to define the transition region and is attached to the outer square by
+  // vertex projection, the inner vertices are shaped to form the inner square.
+  // To ensure that the inner square vertices can be easily identified  even
+  // when rotated, the initial circular mesh is generated using twice the
+  // inner_half_side as radius and every vertex can be identified using a
+  // distance criterion.
+  GridGenerator::hyper_ball_balanced(obstacle_tria,
+                                     center,
+                                     2 * inner_half_side);
```
blaisb, line 197: Seems a bit involved to make an inner square, but why not I guess?

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,647 @@
...
+  // angles k*pi/4. After rotation they sit at inner_rotation_angle + k*pi/4.
+  // The projection maps each direction onto the outer square along the same
+  // line. To avoid having voids in the corners we choose the subset of the 4
+  // target points that are closest to  corners and change the target projection
+  // to attach them there instead.
+  std::array<Point<2>, 8> outer_targets;
+  for (unsigned int k = 0; k < 8; ++k)
+    {
+      double theta = inner_rotation_angle + k * std::numbers::pi / 4.0;
+
```
blaisb, line 255:
```suggestion
```

blaisb (changes requested):
Some small things to address there and there, great feature (a bit complex in the implementation with the lambda functions and all though).

That will be a very useful mesh!

tests/core/grid_uniform_channel_with_meshed_square_prism.cc
blaisb: Ok so I tested with the Birmingham grid and writing full grids as output generates quite large output files (like 160kb for Birminham). Could you just instead write volume and some basic statistics about the grid like I did here: https://github.com/chaos-polymtl/lethe/pull/1971/changes#diff-c13eaa7e88731f7b46a1231196c79382cacc97e6768a77b7cd2b62b82b6acf56R37

This would be a better approach and would make for smaller output files.

tests/core/grid_uniform_channel_with_meshed_square_prism.cc
blaisb, line 1: Can you rename the test file executable to reflect the name of the grid you are testing

include/core/grid_uniform_channel_with_meshed_square_prism.h
OresteMarquis: I disagree with that; the front of the channel, in my head, would be at the inlet, and the back at the outlet. Now the inlet is at x=x_min and the outlet x=x_max.

source/core/grid_uniform_channel_with_meshed_square_prism.cc
OresteMarquis: Maybe, it is the simplest idea I came up with because I started from the uniform_channel_with_cylinder.

include/core/grid_uniform_channel_with_meshed_square_prism.h
blaisb: Ok no worries, we can keep it as is.
I am just used to left being x-, right being x+, bottom being y-, top being y+, front being z- and back being z+. I don't it matters much anyway since you explicitly mention the axis. Let's keep it this way then :)

source/core/grid_uniform_channel_with_meshed_square_prism.cc
blaisb: Well it turns out to work well right? So it was not a bad idea either :). Your rotated meshes actually look quite good I think.

include/core/grid_uniform_channel_with_meshed_square_prism.h
OresteMarquis: I see, it is a matter of convention indeed hahaha

source/core/grid_uniform_channel_with_meshed_square_prism.cc
OresteMarquis: Yes, I am quite happy with the result! The only thing is when the transition region is about the size of the square. This creates really skewed elements, but i don't think it's avoidable.

include/core/grid_uniform_channel_with_meshed_square_prism.h
blaisb: Yeah but it's fine. Let's be flexible. It is not because i am used to something that it means this thing is right. I think you mentioning explicitly the axis (z=0, z=..) clears out the confusion.  So let's leave it like you like it :).

source/core/grid_uniform_channel_with_meshed_square_prism.cc
blaisb: Well you have skewed elements when the square is rotated and only in the vicinity of the square. The rest is very good. It's a matter of compromise I think. We will see, but anyway this is where you will refine a lot so it should be fine.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to the Lethe project will be documented in this file.
 The changelog for the previous releases of Lethe are located in the release_notes folder.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2026/04/15
+
+### Added
+
+- MINOR This PR add a new mesh to simulate a channel with an object in it. Here the object is a rectangular prism with a square base. The name of this new mesh is "uniform_channel_with_meshed_square_prism". This will be used for benchmark and test the multiphysic microwave-heating capabilities of Lethe. The square prism can rotate and the size of its side can be adjusted in addition to the channel dimension. [#1968](https://github.com/chaos-polymtl/lethe/pull/1968)
```
Julamipolymtl:
```suggestion
- MINOR This PR adds a new mesh to simulate a channel with an object in it. Here the object is a rectangular prism with a square base. The name of this new mesh is "uniform_channel_with_meshed_square_prism". This will be used for benchmark and test the multiphysic microwave-heating capabilities of Lethe. The square prism can rotate and the size of its side can be adjusted in addition to the channel dimension. [#1968](https://github.com/chaos-polymtl/lethe/pull/1968)
```

source/core/grids.cc
```diff
@@ -207,6 +208,19 @@ attach_grid_to_triangulation(Triangulation<dim, spacedim> &triangulation,
             mesh_parameters.grid_arguments);
           grid.make_grid(triangulation);
```
Julamipolymtl, line 209: I think this PR also changed the name of the previous cylinder mesh you did. This instance is missing from the renaming. I haven't seen any more.

include/core/grid_uniform_channel_with_meshed_square_prism.h
```diff
@@ -0,0 +1,188 @@
...
+   *   @p inner_rotation_angle while only half of the outer vertices are rotated
+   *   and projected to the closest point on the outer square. The other half
+   *   are attached to the outer square corners. The choice of which outer
+   *   vertices are rotated and which are attached to the corners is made on a
+   *   closest-angle basis to minimize the distortion of the transition cells.
+   * - Up to 8 rectangular padding sub-meshes filling the gap between the
+   *   transition square and the channel boundary. Those are set up in a way
+   *   that their vertices lines up with the vertices projected from the
+   *   transition region to avoid the need of introducing new vertices when
+   *   merging the triangulations.
```
Julamipolymtl, line 113:
```suggestion
   * This private helper creates the composite 2D triangulation by merging:
   * - A modified hyper_ball_balanced from deal.ii so the inner vertex are
   *   placed to form the inner square of half-side @p inner_half_side and the
   *   outer vertices are placed to form the outer square of half-side
   *   @p outer_half_side. When rotated, the inner vertices are rotated by
   *   @p inner_rotation_angle while only half of the outer vertices are rotated
   *   and projected to the closest point on the outer square. The other half
   *   are attached to the outer square corners. The choice of which outer
   *   vertices are rotated and which are attached to the corners is made on a
   *   closest-angle basis to minimize the distortion of the transition cells.
   * - Up to 8 rectangular padding sub-meshes filling the gap between the
   *   transition square and the channel boundary. Those are set up in a way
   *   that their vertices line up with the vertices projected from the
   *   transition region to avoid the need of introducing new vertices when
   *   merging the triangulations.
```

tests/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  template <int dim>
+  void
+  run_test(const std::string &name,
+           const std::string &grid_arguments,
+           const unsigned int global_refinement = 0)
+  {
+    Triangulation<dim>                                triangulation;
+    UniformChannelWithMeshedSquarePrismGrid<dim, dim> grid(grid_arguments);
+    grid.make_grid(triangulation);
```
Julamipolymtl: Same as with the renaming of the meshed cylinder.

```suggestion
    Triangulation<dim>                                triangulation;
    GridUniformChannelWithMeshedSquarePrism<dim, dim> grid(grid_arguments);
    grid.make_grid(triangulation);
```

source/core/grids.cc
```diff
@@ -207,6 +208,19 @@ attach_grid_to_triangulation(Triangulation<dim, spacedim> &triangulation,
...
+      else if (grid_type == "uniform_channel_with_meshed_square_prism")
+        {
+          AssertThrow(
+            !mesh_parameters.simplex,
+            ExcMessage(
+              "Unsupported mesh type - uniform channel with rotated meshed rectangle mesh with simplex is not supported"));
+
+          UniformChannelWithMeshedSquarePrismGrid<dim, spacedim> grid(
+            mesh_parameters.grid_arguments);
+          grid.make_grid(triangulation);
```
Julamipolymtl:
```suggestion
          GridUniformChannelWithMeshedSquarePrism<dim, spacedim> grid(
            mesh_parameters.grid_arguments);
          grid.make_grid(triangulation);
```

include/core/grid_uniform_channel_with_meshed_square_prism.h
```diff
@@ -0,0 +1,188 @@
...
+  /// Point that define the top-right of the channel in the xy-plane.
+  Point<dim> top_right;
+  /// Point that define the center of the obstacle in the xy-plane.
+  Point<dim> center;
+  /// Half-side of the inner square obstacle.
+  double inner_half_side;
+  /// Half-side of the outer transition square.
+  double outer_half_side;
+  /// In-plane rotation angle in degrees applied to the inner mesh pattern.
+  double inner_rotation_angle;
```
Julamipolymtl: Correct me if I'm wrong but I think its the other way around. The param in is in degrees, and the member variable converts it from degrees to radians. 
```suggestion
   * @param[in]  inner_rotation_angle In-plane rotation angle in degrees.
   * @param[in]  pad_bottom     Number of subdivisions in the bottom padding.
   * @param[in]  pad_top        Number of subdivisions in the top padding.
   * @param[in]  pad_left       Number of subdivisions in the left padding.
   * @param[in]  pad_right      Number of subdivisions in the right padding.
   * @param[in]  colorize       Whether to assign distinct boundary IDs.
   */
  void
  generate_2d_channel_mesh(Triangulation<2>  &triangulation,
                           const Point<2>    &bottom_left,
                           const Point<2>    &top_right,
                           const Point<2>    &center,
                           const double       inner_half_side,
                           const double       outer_half_side,
                           const double       inner_rotation_angle,
                           const unsigned int pad_bottom,
                           const unsigned int pad_top,
                           const unsigned int pad_left,
                           const unsigned int pad_right,
                           const bool         colorize);

  /// Point that define the bottom-left of the channel in the xy-plane.
  Point<dim> bottom_left;
  /// Point that define the top-right of the channel in the xy-plane.
  Point<dim> top_right;
  /// Point that define the center of the obstacle in the xy-plane.
  Point<dim> center;
  /// Half-side of the inner square obstacle.
  double inner_half_side;
  /// Half-side of the outer transition square.
  double outer_half_side;
  /// In-plane rotation angle in radians applied to the inner mesh pattern.
  double inner_rotation_angle;
```

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,645 @@
...
+    // Create the steps list
+    std::vector<double> steps;
+    for (std::size_t i = 1; i < coords.size(); ++i)
+      steps.push_back(coords[i] - coords[i - 1]);
+    return steps;
+  };
+
+  // Similar as above, this lambda function create a vector of uniform step
+  // sizes given the number of steps n and the total length to cover.
```
Julamipolymtl:
```suggestion
  // Similar as above, this lambda function creates a vector of uniform step
  // sizes given the number of steps n and the total length to cover.
```

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,645 @@
...
+
+  AssertThrow(
+    inner_half_side > 0.0 && outer_half_side >= inner_half_side,
+    ExcMessage(
+      " The inner_half_side must be greater than 0 to have an object in the channel."));
+
+  AssertThrow(
+    outer_half_side >= 1.5 * inner_half_side - 1e-12 * inner_half_side,
+    ExcMessage(
+      "The outer half-side must be at least 1.5 times the inner half-side, if not the elements in the transition region will be too distorted."));
```
Julamipolymtl, line 108: Very nitpicky, but `outer_half_side >= inner_half_side` is now redundant with `outer_half_side >= 1.5 * inner_half_side - 1e-12 * inner_half_side` which is much stricter. Unless there was a situation where it was genuinely relevant?

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,645 @@
...
+      const double cx = std::cos(theta);
+      const double cy = std::sin(theta);
+      const double scale =
+        outer_half_side / std::max(std::abs(cx), std::abs(cy));
+      outer_targets[k] = center + Point<2>(scale * cx, scale * cy);
+    }
+
+  // Attach each outer ring vertex to its analytical target (closest by angle).
+  // Using angle difference avoids the Euclidean distance ambiguity and ensures
+  // exact vertex matching with the padding meshes.
```
Julamipolymtl, line 298: Nice way to do it! Well explained.

source/core/grid_uniform_channel_with_meshed_square_prism.cc
```diff
@@ -0,0 +1,645 @@
...
+  if (pad_top > 0 && pad_right > 0)
+    trias.push_back(&pad_top_right_corner_tria);
+  GridGenerator::merge_triangulations(trias, triangulation);
+
+  triangulation.reset_all_manifolds();
+  triangulation.set_all_manifold_ids(0);
+
+  // Here we mark obstacle cells with the obstacle material ID. This is done in
+  // the unrotated frame by rotating the cell center back to the axis-aligned
+  // frame temporarily and checking if it is within the inner square dimensions.
```
Julamipolymtl, line 513: So simple! Nicely done.

source/core/grid_uniform_channel_with_meshed_square_prism.cc
OresteMarquis: Ca existe deja je crois: `inline Tensor<1, spacedim> value_string_to_tensor(const std::string &value_string)`

Julamipolymtl (changes requested):
Nice new mesh you made available! I had to look into lambda functions it had been a while:') I found a few renamings missing, but other than that mostly typos!

You did well to create a bunch of cases so that vertices jump to the next element to limit element skewness or edge alignment. 

This is more of an open question, as meshing kind be done a 1000 ways, but I wonder if a sort of transfinite interpolation between nodes in the padding region could help with the skewness of a few elements. That way the line connecting nodes gradually curves in the padding section instead of being straight. But that is much more involved and I guess a better use case for gmsh!

blaisb (approved):
Great work!

Julamipolymtl (approved):

PR #1971, opened 2026-04-20
Add a unit test for the Birmingham mesh

blaisb:
### Description

The Birmingham reactor mesh had no unit test and it's class name did not fit the file name. This PR adds a unit test for all possible configurations of the reactor and refactor the name of the class to be coherent with the file name.

### Testing

There is a new unit test for that grid. It generates four configurations of the grid and write statistics about the grid to the output file.

### Documentation

Nothing to document here.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/grid_birmingham_fluidized_bed.h
hepap, line 1: I don't mind the naming convention we chose, but this one starts with grid and the periodic hill and cylinder ones end with grid (`cylinder_grid.h` and `uniform_channel_with_meshed_cylinder_grid.h`). Maybe we can uniformize it :)

tests/core/grid_birmingham_fluidized_bed.cc
```diff
@@ -0,0 +1,110 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+/**
+ * @brief Tests the generation of the Birmingham fluidized bed mesh. The
+ * grid is only implemented in 3D, so only dim = spacedim = 3 is exercised.
```
hepap:
```suggestion
 * grid is only implemented in 3D, so only dim = spacedim = 3 is tested.
```

hepap (approved):
I would keep the previous name of the class and change the file name to match the already existing `PeriodicHillsGrid`, `UniformChannelWithMeshedCylinderGrid` and `CylinderGrid` naming convention.

Other than that, all good :)

include/core/grid_birmingham_fluidized_bed.h
OresteMarquis: Maybe naming them "Grid..." at the starts would indeed be better because they would be together in the include and source files!

tests/core/grid_birmingham_fluidized_bed.cc
```diff
@@ -0,0 +1,110 @@
...
+            << " face count : " << count << std::endl;
+
+  // Write in VTK format to a separate file so that it can
+  // be opened in ParaView as a visual double-check.
+  GridOut           go;
+  const std::string vtk_filename =
+    "grid_birmingham_fluidized_bed_" + case_name + ".vtk";
+  std::ofstream vtk_out(vtk_filename);
+  go.write_vtk(triangulation, vtk_out);
+}
```
OresteMarquis, line 62: Do we still want to output this? From the comment in #1968, I understood that this creates quite large files.

OresteMarquis (changes requested):

include/core/grid_birmingham_fluidized_bed.h
blaisb: Yeah that was my idea. Rename them one by one while adding the unit test :)

tests/core/grid_birmingham_fluidized_bed.cc
blaisb: the comment in #1968 was referring to outputting the whole grid into the .output file. This generated a huge output file (150kb or so)

Here, we are writing the vtk files which are like 40-50 kb each, but they are not kept track of. They just allow a rapid look / debug of the mesh. If the test every fails, somebody could recheck that the mesh makes sense.
I am not sure if it is necessary. If you prefer I could just keep that line commented out and if someone needs to output the mesh to see them, they could uncomment them. What do you think? Maybe that would be a better solution :)

OresteMarquis: If the VTK is not heavy, I would keep it. It is easier to debug when we see the mesh! I will do the same for the other test.

OresteMarquis (approved):
Good with me also!

PR #1972, opened 2026-04-20
Compute melt volume

AmishgaAlphonius:
### Description

This PR adds the postprocessing of the melt volume for cases with phase change. Setting `calculate geometric melt volume` to true enables the computation of the geometric melt volume. The region where the temperature of the fluid with phase change is above the `melting temperature` describes the melt subdomain. The implementation is compatible with CLS multiphase flows as long as only 1 of the 2 fluids has phase change. `monitored fluid with phase change` set the fluid monitored. `geometric melt volume name` sets the name of the file where computed melt volume values are saved.

  This PR also changes the computation of the liquid fraction with `calculate liquid fraction` to `calculate algebraic melt volume`. The liquid fraction in the domain is no longer outputted. Instead, a liquid fraction weighted melt volume is outputted in a file named with `algebraic melt volume name`. In CLS simulations, the fluid of interest must be specified with `monitored fluid with phase change`. 

### Testing

- `applications_tests/lethe-fluid/heat_transfer_stefan.prm` was adapted to include melt volume outputs.
- `applications_tests/lethe-fluid/heat_transfer_cls_melt_volume.prm` was added to track melt volume in CLS flows.
Melt volume computation was confirmed through visual inspection in Paraview and mesh convergence analysis below.

### Documentation

- The new parameters have been added to the post-processing section of the documentation.

-  The possibility of adding a weighted contribution of the volume for the mushy zone can be added in a future PR.

### Contour length bug - Deal.II versions

- Current master version and v7.9.1 do not output the same contour length in the new test. This is probably due to the change in classification of cells where the 0-isocontour is aligned with a cell face [dealii PR#16950](https://github.com/dealii/dealii/commit/f61e44e09d1c79a4fd44706ddfefd7087495457e). The test was changed so that the 0-isocontour is no more aligned with cell faces and results were checked with mesh convergence:  

  - Expected value for the contour length: 0.3725 + 0.1125 = 0.485
    contour-length-convergence
    00-contour-length-convergence

  - Expected value for the melt surface area: 0.3725 * 0.1125 = 0.4190625
    melt-surface-convergence
    01-melt-surface-convergence

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/heat_transfer.h
```diff
@@ -32,6 +32,9 @@
 #include <deal.II/lac/trilinos_sparse_matrix.h>
 #include <deal.II/lac/trilinos_vector.h>
 
+#include <deal.II/non_matching/fe_values.h>
+#include <deal.II/non_matching/mesh_classifier.h>
+
```
hepap, line 37:
```suggestion
```

source/solvers/heat_transfer.cc
```diff
@@ -1189,6 +1189,22 @@ HeatTransfer<dim>::postprocess(bool first_iteration)
         this->write_liquid_fraction();
     }
 
+  // Melt volume
+  if (simulation_parameters.post_processing.calculate_melt_volume)
+    {
+      AssertThrow(
+        simulation_parameters.physical_properties_manager.has_phase_change(),
+        ExcMessage(
+          "Calculation of the melt volume requires that a fluid has a 'phase_change' specific heat model"));
```
hepap: I don't think this assert is required, because we could be interested in a case where there is no difference in the cp of the liquid and solid (I do it in the benchmark).

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,190 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+          liquidus_temperature = physical_properties_parameters.fluids[0]
+                                   .phase_change_parameters.T_liquidus;
+        }
+      else // Fluid 1 has phase change
+        {
+          liquidus_temperature = physical_properties_parameters.fluids[1]
+                                   .phase_change_parameters.T_liquidus;
+          phase_indicator_vector_owned_copy->operator*=(
+            -1); // Make fluid 1 'inside' (negative values)
+        }
```
hepap: I think we could instead add a parameter for which fluid we want to postprocess the volume (like in the cls mass conservation info).

Like this, it could work for any fluid, independent of the cp model. It could even be more general if the temperature iso-contour is provided as a parameter :)

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,190 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+
+  // Transpose temperature to get volume of the region where the temperature is
+  // over the liquidus temperature
+  GlobalVectorType temperature_vector_owned_copy(
+    this->dof_handler->locally_owned_dofs(), mpi_communicator);
+  temperature_vector_owned_copy = *this->present_solution;
+  temperature_vector_owned_copy.add(-liquidus_temperature);
+  temperature_vector_owned_copy.operator*=(
+    -1); // Make the liquid region 'inside' (negative values)
+
```
hepap, line 2251: I think this copy could be avoided (if it is expensive to do) by doing the operation on the dof loop at line 2235.

applications_tests/lethe-fluid/heat_transfer_cls_melt_volume.prm
```diff
@@ -0,0 +1,196 @@
...
+    set preconditioner                        = ilu
+    set ilu preconditioner fill               = 0
+    set ilu preconditioner absolute tolerance = 1e-14
+    set ilu preconditioner relative tolerance = 1.00
+  end
+  subsection CLS
+    set verbosity               = quiet
+    set relative residual       = 1e-3
+    set minimum residual        = 1e-10
+    set ilu preconditioner fill = 1
```
blaisb: Do you need a high fill level like this here? I'm curious

applications_tests/lethe-fluid/heat_transfer_stefan.prm
```diff
@@ -1,6 +1,14 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023 The Lethe Authors
+# SPDX-FileCopyrightText: Copyright (c) 2022-2024, 2026 The Lethe Authors
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
 
+#-------------------------------------------------------------------------------
```
blaisb, line 4: Thanks for adding that!

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +256,23 @@ This subsection controls the post-processing other than the forces and torque on
...
+* ``calculate liquid fraction``: calculates the liquid fraction in simulations with phase change. The liquid fraction corresponds to the volume (area) integral of the liquid divided by the volume (area) of the domain.
+
+  * ``liquid fraction name``: name of the output file containing the time evolution of the liquid fraction.
+
+* ``calculate melt volume``: calculates the melt volume in simulations with phase change. The melt volume corresponds to the volume (area) integral of the liquid phase of the fluid with phase change. The melt region (:math:`\Omega_\mathrm{melt}`) is described by temperatures above the fluid's liquidus temperature (:math:`T_\mathrm{liquidus}`):
+
+  .. math::
+    \Omega_\mathrm{0,melt} = \{\vec{x} \in \Omega_0 | T > T_\mathrm{liquidus} \}
+
+  here, fluid 0 is a fluid a with phase change and :math:`\Omega_0` corresponds to the subdomain occupied by fluid 0.
```
blaisb: So if I understand correctly, one is the volume ABOVE the liquidus and the other one is the volume taking into account the mushy zone. Am I correct?

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,190 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+      // Get the intersection region between the correct fluid and the liquidus
+      // temperature isocurve
+      intersection_vector_relevant_copy = std::make_shared<GlobalVectorType>(
+        this->dof_handler->locally_owned_dofs(), mpi_communicator);
+
+      for (const auto dof_id : this->dof_handler->locally_owned_dofs())
+        {
+          (*intersection_vector_relevant_copy)[dof_id] =
+            std::max(temperature_vector_owned_copy[dof_id],
+                     (*phase_indicator_vector_owned_copy)[dof_id]);
```
hepap: This could become (with the previous comment):


```suggestion
            std::max(-1.0*(present_solution[dof_id]-liquidus_temperature),
                     (*phase_indicator_vector_owned_copy)[dof_id]);
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +256,23 @@ This subsection controls the post-processing other than the forces and torque on
...
+
+  .. warning::
+
+    This is compatible with CLS two-phase flows. A geometrical volume is computed delimited by the :math:`0.5` phase indicator iso-curve. However, only one of the two fluids can have a ``phase_change`` specific heat model.
```
blaisb: Is it always fluid 0 then? or it can be fluid 1? or this is determined at run time and it is the one that has a phase change? I would add more precisions here.
Does this only work with the CLS model,  or it also works in single phase? I would think it works also on single phase?

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,190 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+          if (inside_fe_values)
+            {
+              for (const unsigned int q :
+                   inside_fe_values->quadrature_point_indices())
+                {
+                  melt_volume += inside_fe_values->JxW(q);
+                }
+            }
+        }
+    }
```
hepap: If you want, you can reuse the `compute_surface_and_volume` of interface_tools.cc and just keep the volume that it returns :) It would avoid code duplication.

include/solvers/heat_transfer.h
```diff
@@ -569,17 +572,35 @@ class HeatTransfer : public AuxiliaryPhysics<dim, GlobalVectorType>
...
   /**
    * @brief Post-processing. Write the liquid fraction to an output file.
    */
-
   void
   write_liquid_fraction();
 
+  /**
+   * @brief Compute the melt volume (3D) or surface (2D) in a domain. The mushy
+   * zone is excluded from the melt region.
```
blaisb: That's a very good comment. I would add this information also phrased like this in the documentation.

hepap (commented):
Great addition! I like it :)

source/solvers/heat_transfer.cc
AmishgaAlphonius: Indeed, like we discussed, the choice of the iso-contour (T_liquidus or T_solidus) can vary from one case to another and this can make it more user-friendly :)

source/solvers/heat_transfer.cc
AmishgaAlphonius: As discussed, I'll do a sub-function in `InterfaceTools`, where we feed the relevant copy to make the non-matching etc. so that I can use it here also :)

source/solvers/heat_transfer.cc
blaisb: In that case how is the liquidus temperature established to calculate the melt volume?

source/solvers/heat_transfer.cc
blaisb: Yeah that makes sense. It also enables you to use an iso contour value that is neither T_Solidus or T_liquidus (for example the average between the two)
It's a good idea.

source/solvers/heat_transfer.cc
blaisb: The vector copy won't be very expensive. We already do that many times per iteration. I would not worry about this too much unless the post-processing start to be a significant % of the simulation.

source/solvers/heat_transfer.cc
blaisb: Good idea nonetheless (cost or not) to do it this way. It's easier to read and introduces less variables in my opinion.

blaisb (changes requested):
Some slight changes (in addition to the very good suggestions of @hepap). On my side is mostly documentations elements.

applications_tests/lethe-fluid/heat_transfer_cls_melt_volume.prm
AmishgaAlphonius: I don't I've changed it back to 0, it is a bad copy-paste case.

doc/source/parameters/cfd/post_processing.rst
AmishgaAlphonius: I've changed the description. I hope it is more clear now with the user-set temperature limit.

doc/source/parameters/cfd/post_processing.rst
AmishgaAlphonius: It works for both CLS and single-fluid flows. Now the fluid is chosen with `monitored fluid with phase change`, before it was checked internally with the specific heat model. :)

include/solvers/heat_transfer.h
AmishgaAlphonius: I've changed a bit the description in the user documentation. I hope it is more clear.

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,25 @@ This subsection controls the post-processing other than the forces and torque on
 		0.0000          0.0000               0.0000               0.0000            1000.0000 
 		1.0000         -0.9732               0.0000               1.4856               0.9732 
 
+* ``calculate liquid fraction``: calculates the liquid fraction in simulations with phase change. The liquid fraction corresponds to the volume (area) integral of the liquid divided by the volume (area) of the domain.
+
+  * ``liquid fraction name``: name of the output file containing the time evolution of the liquid fraction.
+
+* ``calculate melt volume``: computes the melt volume (surface, in 2D) and contour area (length, in 2D) in simulations with phase change. The melt volume corresponds to the volume integral of the melted phase of the fluid with phase change. The melt region (:math:`\Omega_\mathrm{i,melt}` with :math:`i \in \{0,1\}` depending on the fluid) is characterized by temperatures above the fluid's melting temperature (:math:`T_\mathrm{melt}`):
+
```
blaisb: Can you explain clearly the difference between the melt volume and the liquid fraction other than the fact that they are normalized?
one is geometric the other one is calculated by integrating the mushy zone also? Try to make it explicit what is the difference.
Maybe then we should deprecate the liquid fraction parameter and calculation if this new feature is more "adequate" than the previous one. that's also something I would consider. In that case, we should adapt the melting cavity example to use this new feature instead.

I am not saying this needs to be done, but the parameter description should clearly explain the difference between the two calculation methods and when one wants to use one or the other.

include/core/interface_tools.h
```diff
@@ -10,6 +10,8 @@
 #include <core/utilities.h>
 #include <core/vector.h>
 
+#include "solvers/time_harmonic_maxwell.h"
+
```
blaisb: Why is this include added here?

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,25 @@ This subsection controls the post-processing other than the forces and torque on
 		0.0000          0.0000               0.0000               0.0000            1000.0000 
 		1.0000         -0.9732               0.0000               1.4856               0.9732 
 
+* ``calculate liquid fraction``: calculates the liquid fraction in simulations with phase change. The liquid fraction corresponds to the volume (area) integral of the liquid divided by the volume (area) of the domain.
```
blaisb: This one is normalized by the volume and the new one is not. Maybe it would be better to refactor this one so that it is not normalized also, to ensure that at least both definitions are consistent with one another.

include/solvers/heat_transfer.h
```diff
@@ -569,17 +571,39 @@ class HeatTransfer : public AuxiliaryPhysics<dim, GlobalVectorType>
...
+   *
+   * @param[in] gather_cls Boolean indicating when a CLS multiphase flow is
+   * being simulated.
+   *
+   * @remark For CLS multiphase flow, only one of the fluids can be monitored.
+   * The volume is computed as the geometrical volume with respect to
+   * the phase indicator isocurve (isosurface) of value 0.5.
+   */
+  void
+  postprocess_melt_volume_and_surface(const bool gather_cls);
```
blaisb: The heat transfer solver can already check if cls is enabled from the parameters. You do not need to pass that boolean here, since that information is already available due to it being a class member. There is no cases where cls is enabled AND you would not want to have gather cls

source/core/interface_tools.cc
```diff
@@ -141,14 +127,56 @@ InterfaceTools::compute_surface_and_volume(const DoFHandler<dim> &dof_handler,
...
+  const FiniteElement<3> &fe,
+  const MPI_Comm         &mpi_communicator);
+
+template <int dim, typename VectorType>
+std::pair<double, double>
+InterfaceTools::compute_volume_and_surface(const DoFHandler<dim> &dof_handler,
+                                           const FiniteElement<dim> &fe,
+                                           const VectorType &level_set_vector,
+                                           const double      iso_level,
+                                           const MPI_Comm   &mpi_communicator)
```
blaisb: I am not sure this is the right name. The other one integrates the volume and surface for a vector field. Here the only thing that this does is integrate the volume and surface for a vector field shifted by a scalar.
Why not have the same function name, but one that takes an extra argument which is the iso_level. value. This way you have only one function name, but two implementation. One with an iso value and one without.  You can call the version without the isovalue from the version with it.

blaisb: You don't need the communicator. You can get it from the dof_handler
https://dealii.org/current/doxygen/deal.II/classDoFHandler.html#acb38b25dc7a7d64a452b86ff6ca8d116

source/core/interface_tools.cc
```diff
@@ -141,14 +127,56 @@ InterfaceTools::compute_surface_and_volume(const DoFHandler<dim> &dof_handler,
...
+                                      mpi_communicator);
+}
+
+template std::pair<double, double>
+InterfaceTools::compute_volume_and_surface(
   const DoFHandler<2>    &dof_handler,
   const FiniteElement<2> &fe,
   const GlobalVectorType &level_set_vector,
   const double            iso_level,
   const MPI_Comm         &mpi_communicator);
```
blaisb: You don't need the communicator. You can get it from the dof_handler
https://dealii.org/current/doxygen/deal.II/classDoFHandler.html#acb38b25dc7a7d64a452b86ff6ca8d116

source/core/interface_tools.cc
```diff
@@ -141,14 +127,56 @@ InterfaceTools::compute_surface_and_volume(const DoFHandler<dim> &dof_handler,
 }
 
 template std::pair<double, double>
-InterfaceTools::compute_surface_and_volume(
+InterfaceTools::integrate_volume_and_surface(
+  const GlobalVectorType &level_set_vector,
+  const DoFHandler<2>    &dof_handler,
+  const FiniteElement<2> &fe,
+  const MPI_Comm         &mpi_communicator);
```
blaisb: No need for communiator. yOu can get it from the dof handler
https://dealii.org/current/doxygen/deal.II/classDoFHandler.html#acb38b25dc7a7d64a452b86ff6ca8d116

source/core/interface_tools.cc
```diff
@@ -141,14 +127,56 @@ InterfaceTools::compute_surface_and_volume(const DoFHandler<dim> &dof_handler,
...
+
+template std::pair<double, double>
+InterfaceTools::integrate_volume_and_surface(
+  const GlobalVectorType &level_set_vector,
+  const DoFHandler<3>    &dof_handler,
+  const FiniteElement<3> &fe,
+  const MPI_Comm         &mpi_communicator);
```
blaisb: No need for communicator

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,147 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+    this->simulation_parameters.post_processing.melting_temperature;
+
+  // Initialize CLS related objects
+  std::shared_ptr<const DoFHandler<dim>>      dof_handler_cls;
+  std::shared_ptr<NonMatching::FEValues<dim>> non_matching_fe_values_cls;
+  std::shared_ptr<GlobalVectorType>           phase_indicator_vector_owned_copy;
+  std::shared_ptr<GlobalVectorType>           intersection_vector_relevant_copy;
+  const double phase_indicator_interface_value = 0.5;
+
+  if (gather_cls)
```
blaisb, line 2199: Can be parsed directly from the parameters, no need for a boolean here since this is already a member of the heat transfer class.

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,147 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+      // Local copy to shift phase indicator solution to make the region of
+      // interest 'inside' (negative values)
+      phase_indicator_vector_owned_copy = std::make_shared<GlobalVectorType>();
+      phase_indicator_vector_owned_copy->reinit(
+        this->dof_handler->locally_owned_dofs(), mpi_communicator);
+      VectorTools::interpolate_to_different_mesh(
+        *dof_handler_cls,
+        this->multiphysics->get_solution(PhysicsID::CLS),
+        *this->dof_handler,
+        *phase_indicator_vector_owned_copy); // Ensure DoF correspondence for
```
blaisb: It's not exactly interpolating to a different mesh, but I guess that function works here also. It's not a cheap thing to call though.

source/solvers/heat_transfer.cc
```diff
@@ -2121,7 +2144,147 @@ HeatTransfer<dim>::write_liquid_fraction()
...
+      melt_volume_column_name  = "melt surface";
+      melt_contour_column_name = "melt contour length";
+    }
+  else if constexpr (dim == 3)
+    {
+      melt_volume_column_name  = "melt volume";
+      melt_contour_column_name = "melt contour surface";
+    }
+
+  // Consol output
```
blaisb:
```suggestion
  // Console output
```

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to the Lethe project will be documented in this file.
 The changelog for the previous releases of Lethe are located in the release_notes folder.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2026/04/23
+
+### Added
+
+- MINOR This PR adds the postprocessing of the melt volume for cases with phase change. Setting `calculate melt volume` to true enables the computation of the melt volume. The region where the temperature of the fluid with phase change is above the `melting temperature` describes the melt subdomain. The implementation is compatible with CLS multiphase flows as long as only 1 of the 2 fluids has phase change. `monitored fluid with phase change` set the fluid monitored. `melt volume name` sets the name of the file where computed melt volume values are saved. [#1972](https://github.com/chaos-polymtl/lethe/pull/1972)
```
```suggestion
- MAJOR This PR adds the postprocessing of the melt volume for cases with phase change. Setting `calculate melt volume` to true enables the computation of the melt volume. The region where the temperature of the fluid with phase change is above the `melting temperature` describes the melt subdomain. The implementation is compatible with CLS multiphase flows as long as only 1 of the 2 fluids has phase change. `monitored fluid with phase change` set the fluid monitored. `melt volume name` sets the name of the file where computed melt volume values are saved. [#1972](https://github.com/chaos-polymtl/lethe/pull/1972)
```

blaisb (changes requested):
Some changes there and there. I would suggest you do not add a new function in the interface tools, but that instead you have two versions of the same function with different number of arguments. One with the iso value and one without and can you can call the version without the iso value from the one with the iso value. This will make the documentation and the workflow a lot easier to understand.
You can also get the communicator from the dof_hander, no need to pass it as an argument. This reduces the number of argument and makesthe code a lot more robust.

tests/core/interface_tools_sphere_volume.cc
blaisb, line 1: Since there was already a test for this utility, maybe you can extend this test (or add a new one) to test the new feature you have added. It would be quite useful and this is a more robust way to test the features of the code.

include/core/interface_tools.h
AmishgaAlphonius: It's definitely a mistake 😅

blaisb: It happens to all of us.

source/solvers/heat_transfer.cc
AmishgaAlphonius: Do you know another alternative? This was I also used in the PDE-based reinit even if we have the same mesh ^^'

doc/source/examples/multiphysics/melting-cavity/melting-cavity.rst
```diff
@@ -202,6 +206,24 @@ The solid block melts into liquid in this example, hence in the ``physical prope
     end
 
 
+Post-processing
+~~~~~~~~~~~~~~~~~~~
+
+In the ``post-processing`` subsection the algebraic and geometric melt volume computation are enabled. For the geometric approach the ``melting temperature`` is set to the liquidus temperature.
```
hepap:
```suggestion
In the ``post-processing`` subsection the algebraic and geometric melt volume computations are enabled. For the geometric approach the ``melting temperature`` is set to the liquidus temperature.
```

doc/source/examples/multiphysics/melting-cavity/melting-cavity.rst
```diff
@@ -238,8 +260,7 @@ A python post-processing code `(melting-cavity.py)`
...
-
+contains the simulation results. In post-processing, the position of the solid-liquid interface at the top, center and bottom of the cavity, as well as the melted volume fraction are plotted and compared with experiments of Gau and Viskanta [#gau1986]_. Note that the discrepancies in the interfaces are attributed to the two-dimensional simulations and they were also observed and reported by Blais and Ilinca [#blais2018]_. When comparing the current simulation results, both the algebraic and geometric melt volume results (indicated respectively by :math:`V_\mathrm{melt,alge}` and :math:`V_\mathrm{melt,geo}`) give similar liquid fractions, as expected, since the mushy zone is narrow. For the following studies, we will maintain only the algebraic melt volume computation.
```
hepap:
```suggestion
contains the simulation results. In post-processing, the position of the solid-liquid interface at the top, center and bottom of the cavity, as well as the melted volume fraction are plotted and compared with experiments of Gau and Viskanta [#gau1986]_. Note that the discrepancies in the interfaces are attributed to the two-dimensional simulations and they were also observed and reported by Blais and Ilinca [#blais2018]_. When comparing the current simulation results, both the algebraic and geometric melt volume results (indicated respectively by :math:`V_\mathrm{melt,alge}` and :math:`V_\mathrm{melt,geo}`) give similar liquid fractions, as expected, since the mushy zone is narrow. For the following studies, we will use only the algebraic melt volume computation.
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,79 @@ This subsection controls the post-processing other than the forces and torque on
 		0.0000          0.0000               0.0000               0.0000            1000.0000 
 		1.0000         -0.9732               0.0000               1.4856               0.9732 
 
+* ``monitored fluid with phase change``:  fluid with phase change properties, either ``fluid 0`` or ``fluid 1`` (for CLS multiphase flows). The current implementation does not track two melt volumes simultaneously.
+
+* ``calculate algebraic melt volume``: calculates the algebraic melt volume in simulations with phase change. The algebraic melt volume (:math:`V_\mathrm{i,melt,alge}` with :math:`i \in \{0,1\}` depending on the fluid) corresponds to the volume (area, in 2D) integral of the liquid of the specified fluid (with ``monitored fluid with phase change``, by default: ``fluid 0``) in the domain.
```
hepap:
```suggestion
* ``calculate algebraic melt volume``: calculates the algebraic melt volume in simulations with phase change. The algebraic melt volume (:math:`V_\mathrm{i,melt,alge}` with :math:`i \in \{0,1\}` depending on the monitored fluid) corresponds to the volume (surface, in 2D) integral of the liquid fraction in the specified ``monitored fluid with phase change`` domain.
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,79 @@ This subsection controls the post-processing other than the forces and torque on
...
+        V_\mathrm{0,melt,alge} = \int_{\Omega_0} (1-\phi) \, \alpha_\mathrm{0,liquid} \, \mathrm{d}\Omega
+
+    where :math:`\phi` is the phase indicator.
+
+
+    * If ``fluid 1`` is monitored:
+
+      .. math::
+        V_\mathrm{1,melt,alge} = \int_{\Omega_1} \phi \, \alpha_\mathrm{1,liquid} \, \mathrm{d}\Omega
+
```
hepap: I think we could simplify this:


```suggestion

    .. math::
      V_\mathrm{i,melt,alge} = \int_{\Omega_i} \alpha_\mathrm{i,liquid} \, \mathrm{d}\Omega

    where :math:`\alpha_\mathrm{i,liquid} = \frac{T-T_\mathrm{i,solidus}}{T_\mathrm{i,liquidus} - T_\mathrm{i,solidus}}` is the liquid fraction in the fluid :math:`i`,  with :math:`T_\mathrm{i,solidus}` and :math:`T_\mathrm{i,liquidus}` respectively the solidus and liquidus temperatures of fluid :math:`i`, and :math:`T` is the temperature.

    .. warning::

      This is compatible with CLS two-phase flows. However, only one of the two fluids can be a fluid identified with phase change using the parameter ``monitored fluid with phase change``.

```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,79 @@ This subsection controls the post-processing other than the forces and torque on
...
+    .. math::
+        V_\mathrm{0,melt,geo} = \int_{\Omega_\mathrm{0,melt,geo}} \mathrm{d}\Omega
+
+  * If ``fluid 1`` is monitored:
+
+    .. math::
+      \Omega_\mathrm{1,melt,geo} = \{\vec{x} \in \Omega | T > T_\mathrm{melt}, \, \phi > 0.5 \}
+
+    .. math::
+        V_\mathrm{1,melt,geo} = \int_{\Omega_\mathrm{1,melt,geo}} \mathrm{d}\Omega
```
hepap: Same simplification here :)

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -254,6 +258,79 @@ This subsection controls the post-processing other than the forces and torque on
...
+
+  * ``geometric melt volume name``: name of the output file containing the time evolution of the geometric melt volume.
+  * ``melting temperature``: temperature in Kelvin above which the fluid is considered melted.
```
hepap:
```suggestion
  * ``melting temperature``: temperature above which the fluid is considered melted.
```

include/core/interface_tools.h
```diff
@@ -211,37 +211,53 @@ namespace InterfaceTools
...
-   * @brief
-   * Compute the surface area of a given level of a level set field and the
+   * @brief Integrate the surface area of a given level of a level set field and the
    * volume enclosed by it. The inside volume is computed, defined by negative
    * values of the level-set field.
    *
    * @tparam dim An integer that denotes the dimension of the space in which
    * the problem is solved.
    *
+   * @param[in] dof_handler DofHandler associated to the triangulation on which
```
hepap, line 222:
```suggestion
   * @tparam VectorType The vector type of the solution vector.
   
   * @param[in] dof_handler DofHandler associated to the triangulation on which
```

include/core/interface_tools.h
```diff
@@ -211,37 +211,53 @@ namespace InterfaceTools
...
+   * the volume is computed
+   * @param[in] fe Finite element
+   * @param[in] level_set_vector_relevant_copy Level-set vector
+   *
+   * @return Surface area of the specified level and the volume enclosed by it.
```
hepap:
```suggestion
   * @return Volume enclosed by the specified level and its surface area.
```

include/core/interface_tools.h
```diff
@@ -211,37 +211,53 @@ namespace InterfaceTools
...
-   *
    * @param[in] level_set_vector Level-set vector
-   *
    * @param[in] iso_level Given level of the level-set field enclosing the
    * volume of interest
    *
-   * @param[in] mpi_communicator MPI communicator
-   *
-   * @return Surface area of the specified level and the volume enclosed by it
+   * @return Surface area of the specified level and the volume enclosed by it.
```
hepap:
```suggestion
   * @return Volume enclosed by the specified level and its surface area.
```

include/solvers/heat_transfer.h
```diff
@@ -563,22 +565,43 @@ class HeatTransfer : public AuxiliaryPhysics<dim, GlobalVectorType>
   write_temperature_statistics(const std::string domain_name);
 
   /**
-   * @brief Post-processing.
-   * Calculate liquid fraction on the domain.
+   * @brief Compute the algebraic (phase indicator and liquid fraction weighted)
+   * melt volume (3D) or surface (2D) in a domain. The phase evolving fluid must
+   * be selected with
+   * Parameters::PostProcessing::monitored_fluid_with_phase_change.
```
hepap:
```suggestion
   * melt volume (3D) or surface (2D) in a domain. The monitored fluid must
   * be selected with
   * Parameters::PostProcessing::monitored_fluid_with_phase_change.
```

include/solvers/heat_transfer.h
```diff
@@ -563,22 +565,43 @@ class HeatTransfer : public AuxiliaryPhysics<dim, GlobalVectorType>
...
    *
-   * @param gather_cls boolean true when CLS=true (multiphase flow), used to gather
-   * CLS information
+   * @remark For CLS multiphase flow, only one of the fluids can be monitored.
    */
+  void
+  postprocess_algebraic_melt_volume();
 
+  /**
+   * @brief Post-processing. Write the algebraic (phase-weighted) melt volume to
```
hepap:
```suggestion
   * @brief Post-processing. Write the algebraic melt volume to
```

source/core/parameters.cc
```diff
@@ -2322,17 +2322,40 @@ namespace Parameters
...
+      prm.declare_entry("monitored fluid with phase change",
+                        "fluid 0",
+                        Patterns::Selection("fluid 0|fluid 1"),
+                        "Fluid with phase change properties <fluid 0|fluid 1>");
+
+      prm.declare_entry(
+        "calculate algebraic melt volume",
+        "false",
+        Patterns::Bool(),
+        "Enable calculation of the algebraic (phase indicator and liquid fraction weighted) melt volume in the domain. In the case of CLS simulations, the fluid of interest must be selected with 'monitored fluid with phase change.'");
```
hepap:
```suggestion
        "Enable calculation of the algebraic (phase indicator and liquid fraction weighted) melt volume in the domain. In the case of CLS simulations, the fluid of interest is selected with 'monitored fluid with phase change.'");
```

doc/source/parameters/cfd/post_processing.rst
hepap: Can you add a warning that phase change as to be enable?

source/solvers/heat_transfer.cc
blaisb: Not that "that simply" no. If it works I would keep like that for now.

applications_tests/lethe-fluid/heat_transfer_stefan.output
```diff
@@ -3,1506 +3,2259 @@ Running on 1 MPI rank(s)...
    Number of degrees of freedom: 606
    Volume of triangulation:      1.000000e-01
    Number of thermal degrees of freedom: 202
+melt_surface_alge: 1.000000e-03
+melt_surface_geo: 9.800000e-04
+melt_contour_length_geo: 1.000000e-01
```
blaisb: We generally use complete words here with spaces.
I would say melt surface algebraic without the "_" 
I am not sure if "melt" is the right word since the phase could have been solid and melted or could have been melt from the beggining, but I think this is a satisfactory way of providing the information

blaisb (changes requested):
I wll continue today. Just a temporary checkpoint.

source/solvers/heat_transfer.cc
hepap: I think FETools::interpolate is maybe the one your looking for, that's what we use in the geometric reinitialization.

source/solvers/heat_transfer.cc
```diff
@@ -2074,54 +2104,212 @@ HeatTransfer<dim>::postprocess_liquid_fraction(const bool gather_cls)
...
+        *phase_indicator_vector_owned_copy); // Ensure DoF correspondence for
+      // intersection vector later
+      phase_indicator_vector_owned_copy->add(-phase_indicator_interface_value);
+      if (this->simulation_parameters.post_processing
+            .monitored_fluid_with_phase_change ==
+          Parameters::FluidIndicator::fluid1)
+        phase_indicator_vector_owned_copy->operator*=(-1);
+
+      // Get the intersection region between the correct fluid and the liquidus
+      // temperature isocurve
```
hepap: I think it is more the intersection between the  monitored fluid and the new temperature-based level-set field (with the level 0 corresponding to the liquidus isosurface). But, the comment is clear enough to understand.

```suggestion
      // Get the intersection region between the monitored fluid and the liquidus
      // temperature isocurve
```

source/solvers/heat_transfer.cc
```diff
@@ -2074,54 +2104,212 @@ HeatTransfer<dim>::postprocess_liquid_fraction(const bool gather_cls)
...
+  else // Single-fluid flow
+    {
+      // Transpose temperature to get volume of the region where the temperature
+      // is over the melting temperature
+      temperature_vector_owned_copy = std::make_shared<GlobalVectorType>(
+        this->dof_handler->locally_owned_dofs(), mpi_communicator);
+      *temperature_vector_owned_copy = *this->present_solution;
+      temperature_vector_owned_copy->add(-melting_temperature);
+      temperature_vector_owned_copy->operator*=(
+        -1); // Make the liquid region 'inside' (negative values)
```
hepap:
```suggestion
        -1); // Make the liquid region 'inside' (negative values) for the NonMatching MeshClassifier.
```

source/solvers/heat_transfer.cc
```diff
@@ -2074,54 +2104,212 @@ HeatTransfer<dim>::postprocess_liquid_fraction(const bool gather_cls)
...
+                                      melting_temperature)),
+                             (*phase_indicator_vector_owned_copy)[dof_id]);
+        }
+    }
+  else // Single-fluid flow
+    {
+      // Transpose temperature to get volume of the region where the temperature
+      // is over the melting temperature
+      temperature_vector_owned_copy = std::make_shared<GlobalVectorType>(
+        this->dof_handler->locally_owned_dofs(), mpi_communicator);
```
hepap: I think `intersection_vector_relevant_copy` and `temperature_vector_owned_copy` and `melt_indicator_vector_relevant_copy` could be only one vector, initializes before the if/else, like this you fill it accordingly in the if/else and you avoid the selection at line 2250.

source/solvers/heat_transfer.cc
```diff
@@ -2074,54 +2104,212 @@ HeatTransfer<dim>::postprocess_liquid_fraction(const bool gather_cls)
...
+      *temperature_vector_owned_copy = *this->present_solution;
+      temperature_vector_owned_copy->add(-melting_temperature);
+      temperature_vector_owned_copy->operator*=(
+        -1); // Make the liquid region 'inside' (negative values)
+    }
+
+  // Classify active cells as being inside/outside/intersected
```
hepap:
```suggestion
```

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Added
 
+- MAJOR This PR adds the postprocessing of the melt volume for cases with phase change. Setting `calculate geometric melt volume` to true enables the computation of the geometric melt volume. The region where the temperature of the fluid with phase change is above the `melting temperature` describes the melt subdomain. The implementation is compatible with CLS multiphase flows as long as only 1 of the 2 fluids has phase change. `monitored fluid with phase change` set the fluid monitored. `geometric melt volume name` sets the name of the file where computed melt volume values are saved.
```
hepap, line 10: The only thing I think you could consider in a next PR is that for the `calculate geometric melt volume` we don't need the fluid to have phase change since the iso-contour temperature is prescribed in the parameters.

But I think for now, this PR is enough for what we need :)

hepap (changes requested):

applications_tests/lethe-fluid/heat_transfer_stefan.output
AmishgaAlphonius: Should just change it for "liquid" instead ?

CHANGELOG.md
AmishgaAlphonius: Indeed, I've removed the assertion on that matter :)

include/core/interface_tools.h
```diff
@@ -211,37 +211,54 @@ namespace InterfaceTools
...
+   * the volume is computed
+   * @param[in] fe Finite element
+   * @param[in] level_set_vector Level-set vector
    * @param[in] iso_level Given level of the level-set field enclosing the
    * volume of interest
    *
-   * @param[in] mpi_communicator MPI communicator
-   *
-   * @return Surface area of the specified level and the volume enclosed by it
+   * @return Volume enclosed by specified level and its surface area.
```
blaisb:
```suggestion
   * @return Volume and surface area enclosed by specified iso_level.
```

include/solvers/heat_transfer.h
```diff
@@ -867,7 +888,12 @@ class HeatTransfer : public AuxiliaryPhysics<dim, GlobalVectorType>
   /**
    * @brief Liquid fraction in the domain.
    */
-  TableHandler liquid_fraction_table;
+  TableHandler melt_volume_alge_table;
+
+  /**
+   * @brief Evolution of the melt volume in the simulated domain.
+   */
+  TableHandler melt_volume_geo_table;
```
blaisb, line 896: I have not seen it, but are these two tables serialized correctly right now for simulation restart?

blaisb: They don't seem to be from what I have seen. It's quick two lines, I would take the opportunity to fix it before it blows up in your face later.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
...
+
+  This PR also changes the computation of the liquid fraction with `calculate liquid fraction` to `calculate algebraic melt volume`. The liquid fraction in the domain is no longer outputted. Instead, a liquid fraction weighted melt volume is outputted in a file named with `algebraic melt volume name`. In CLS simulations, the fluid of interest must be specified with `monitored fluid with phase change`. [#1972](https://github.com/chaos-polymtl/lethe/pull/1972)
```
blaisb, line 12:
```suggestion
This PR also changes the computation of the liquid fraction with `calculate liquid fraction` to `calculate algebraic melt volume`. The liquid fraction in the domain is no longer outputted. Instead, a liquid fraction weighted melt volume is outputted in a file named with `algebraic melt volume name`. In CLS simulations, the fluid of interest must be specified with `monitored fluid with phase change`. [#1972](https://github.com/chaos-polymtl/lethe/pull/1972)
```

Just a weird indent here.

blaisb (changes requested):
Other than the checkpoint / reading of the table, this PR is good to go on my side. Please do add the table checkpoint/reading though, this will save you headache down the road.

include/solvers/heat_transfer.h
AmishgaAlphonius: Indeed, I completely forgot the checkpointing! Good catch!

CHANGELOG.md
AmishgaAlphonius: I need to leave it. Otherwise it, does not align correctly below the other paragraph in the Markdown rendering :)

include/solvers/heat_transfer.h
AmishgaAlphonius: Nevermind, it is in the `gather_tables` function already :)

hepap (approved):

blaisb (approved):

PR #1975, opened 2026-04-23
Mortar: allow domain shift along rotation axis

jatred:
### Description

This PR removes the obligation to start the domain at z=0 when using the mortar method. The domain can now be shifted along the rotation axis without affecting the results.

### Solution

The minimum cell center height is now computed in 3D with its corresponding function and is used to ID the mortar cells along the rotation axis and to calculate the height of the quadrature points.

### Testing

Some tests had to be modified to take into account the new structure of the mortar manager class.

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/mortar_coupling_manager.h
```diff
@@ -278,6 +282,19 @@ Quadrature<dim>
 construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
+/**
+ * @brief Sets the minimum z value of cell centers at the mortar interface
+ * @param[in] triangulation The triangulation object
+ * @param[in] stator_boundary_id Boundary ID corresponding to the stator side
```
```suggestion
 * @param[in] mortar_parameters The information about the mortar method
 * control, including the rotor mesh parameters
```

include/core/mortar_coupling_manager.h
```diff
@@ -278,6 +282,19 @@ Quadrature<dim>
...
+ * of the mortar interface
+ *
+ * @return Minimum z value of cell centers at the mortar interface
```
```suggestion
 * @return Minimum cell height in the rotation axis direction. At the moment it is assumed that the rotation axis is in z
```

include/core/mortar_coupling_manager.h
```diff
@@ -278,6 +282,19 @@ Quadrature<dim>
 construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
+/**
+ * @brief Sets the minimum z value of cell centers at the mortar interface
```
```suggestion
 * @brief Computes the minimum cell center height in the direction of the rotation axis at the mortar interface
```

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to the Lethe project will be documented in this file.
 The changelog for the previous releases of Lethe are located in the release_notes folder.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2026/04/23
+
+### Fixed
+
+- MINOR This PR removes the obligation to start the domain at z=0 when using the mortar method. The domain can now be shifted along the rotation axis without affecting the results. [#1975](https://github.com/chaos-polymtl/lethe/pull/1975)
```
```suggestion
- MINOR This PR removes the obligation to start the domain at z=0 when using the mortar method in 3D cases. The domain can now be shifted along the rotation axis without affecting the results. [#1975](https://github.com/chaos-polymtl/lethe/pull/1975)
```

brunaccampos (approved):
Small comments only. Nice work :)

source/core/mortar_coupling_manager.cc
```diff
@@ -574,7 +584,7 @@ compute_interface_dimensions_circular(
   // Min and max vertex coordinates for length computation in the axial
   // direction. Used in 3D case
   double vertex_min = std::numeric_limits<double>::max();
-  double vertex_max = 0;
+  double vertex_max = std::numeric_limits<double>::max() * -1;
```
Julamipolymtl:
```suggestion
  double vertex_max = std::numeric_limits<double>::lowest();
```

source/core/mortar_coupling_manager.cc
```diff
@@ -171,7 +171,10 @@ MortarManagerBase<dim>::get_points(const Point<dim> &face_center,
 
           if constexpr (dim == 3)
             points.emplace_back(
-              x[0], x[1], (id_out_plane + quadrature.point(q)[1]) * delta_1);
+              x[0],
+              x[1],
+              height_min + (id_out_plane + quadrature.point(q)[1]) *
+                             delta_1); // TODO Generalize for x and y directions
```
Julamipolymtl: Is height_min corresponding to a cell_center or a cells edge/face? Depending on the definition this value could differ by delta_1/2 from what is expected.

include/core/mortar_coupling_manager.h
```diff
@@ -211,6 +213,8 @@ class MortarManagerBase
   const unsigned int n_quadrature_points;
   /// Rotation angle for the inner domain
   const double rotation_angle;
+  /// Minimum z value in 3D problems
+  const double height_min;
```
Julamipolymtl:
```suggestion
  /// Minimum height along the rotation axis at the mortar interface. Currently limited to z axis.
  const double height_min;
```

Julamipolymtl (changes requested):
Very nice! Do we currently have a test that runs with a non-zero `height_min`?

source/core/mortar_coupling_manager.cc
jatred: It corresponds to the lowest cell center, not the lowest vertex

Julamipolymtl: So from what I understand, if we currently tried to do the mesh the way we used to do it (from z=0) it would offset by half a cell size since `height_min` would find the first cell center. Is that right?

jatred: Yep nice catch, I changed it so it works now.

include/core/mortar_coupling_manager.h
```diff
@@ -56,7 +57,8 @@ class MortarManagerBase
   MortarManagerBase(const std::vector<unsigned int> &n_subdivisions,
                     const std::vector<double>       &radius,
                     const Quadrature<dim2>          &quadrature,
-                    const double                     rotation_angle);
+                    const double                     rotation_angle,
+                    const double                     height_min = 0.0);
```
blaisb:
```suggestion
                    const double                     minimum_height = 0.0);
```
I think it's better to spell out the words completely

include/core/mortar_coupling_manager.h
```diff
@@ -211,6 +213,9 @@ class MortarManagerBase
   const unsigned int n_quadrature_points;
   /// Rotation angle for the inner domain
   const double rotation_angle;
+  /// Minimum cell center height at the mortar interface along the rotation axis
+  /// value in 3D problems
+  const double height_min;
```
blaisb:
```suggestion
  const double minimum_height;
```

Could also be "height_first_cell". Whatever you prefer, but it's better to be more explicit in the variable names.

include/core/mortar_coupling_manager.h
```diff
@@ -278,6 +283,21 @@ Quadrature<dim>
...
+ * control, including the rotor mesh parameters
+ *
+ * @return Minimum cell height in the rotation axis direction. At the moment,
+ * it is assumed that the rotation axis is in z
+ */
+template <int dim>
+double
+compute_height_min(const Triangulation<dim>      &triangulation,
+                   const Parameters::Mortar<dim> &mortar_parameters);
+
```
blaisb:
```suggestion
compute_minimum_height(const Triangulation<dim>      &triangulation,
                   const Parameters::Mortar<dim> &mortar_parameters);

```
Same comment as for the parameter name.

include/core/mortar_coupling_manager.h
```diff
@@ -320,6 +340,7 @@ class MortarManagerCircle : public MortarManagerBase<dim>
                       double                  radius,
                       const Quadrature<dim2> &quadrature,
                       const double            rotation_angle,
+                      const double            height_min         = 0.0,
```
blaisb: See parameter name. I won't comment on the other same comments.

include/core/mortar_coupling_manager.h
```diff
@@ -278,6 +283,21 @@ Quadrature<dim>
...
+ * @param[in] mortar_parameters The information about the mortar method
+ * control, including the rotor mesh parameters
+ *
+ * @return Minimum cell height in the rotation axis direction. At the moment,
+ * it is assumed that the rotation axis is in z
+ */
+template <int dim>
+double
+compute_height_min(const Triangulation<dim>      &triangulation,
+                   const Parameters::Mortar<dim> &mortar_parameters);
```
blaisb: I have a question here. Instead of calling this function from the outside and passing the height min as a constructor parameter for the mortar mesh manager, couldnt you call that function from inside the constructor of the mortar manager it self?
I just want to understand what's the advantage of doing the call "outside" of the constructor instead of doing it inside? Is it because you want to minimize the number of time you do that calculation?

include/core/mortar_coupling_manager.h
```diff
@@ -520,7 +557,8 @@ MortarManagerLinear<dim>::MortarManagerLinear(
                                              mortar_parameters))) /
         (2.0 * numbers::PI),
       quadrature,
-      0.0)
+      0.0,
+      compute_height_min(dof_handler.get_triangulation(), mortar_parameters))
```
blaisb: I think it is a good choice to use the triangulation here, since this is exactly what we are aiming at
Just be careful that this means that if there is a mapping at the bottom it won't be taken into account (like if the bottom was curved or something) but I don't think it is an issue.

source/core/mortar_coupling_manager.cc
```diff
@@ -769,6 +782,50 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
+{
+  if constexpr (dim == 3)
+    {
+      // Direction of the rotation axis
+      unsigned int direction = 0;
+      for (int d = 0; d < dim; d++)
+        if (mortar_parameters.rotation_axis[d] != 0.0)
+          direction = d;
+      // Minimum coordinate in the direction of the rotation axis
+      double height_min_local = std::numeric_limits<double>::max();
```
blaisb:
```suggestion
      // Direction of the rotation axis
     unsigned int direction = 0;
    for (unsigned int d = 1; d < dim; ++d)
    if (std::abs(axis[d]) > std::abs(axis[direction]))
       direction = d;
       
      AssertThrow(std::abs(axis[direction]) > 0.99,
       ExcMessage("Rotation axis is not aligned with a coordinate direction."));
       
      // Minimum coordinate in the direction of the rotation axis
      double height_min_local = std::numeric_limits<double>::max();
```

Just so you add a security check to ensure also that it's aligned with a direction.
This approach is both more robust and more defensive since right now we don't allow for axis which are not aligned with a unit vector anyway.

source/core/mortar_coupling_manager.cc
```diff
@@ -769,6 +782,50 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
+
+      // Loop over the cells to find the minimum coordinate in the rotation axis
+      // direction
```
blaisb, line 808: This is a nice approach. It's robust. I like this. Great work.

source/core/mortar_coupling_manager.cc
```diff
@@ -769,6 +782,50 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
+                      face->boundary_id() ==
+                        mortar_parameters.stator_boundary_id)
+                    height_min_local =
+                      std::min(height_min_local, cell->center()[direction]);
+                }
+            }
+        }
+      // Return the minimum value across all processes
+      return Utilities::MPI::min(height_min_local,
+                                 triangulation.get_mpi_communicator());
```
blaisb, line 830: Yes. very smart to get the mpi_communicator from the triangulation. Good job.

tests/mortar/poisson_3d_01.cc
```diff
@@ -104,7 +104,8 @@ main(int argc, char **argv)
                               2 * Utilities::pow(2, n_global_refinements)},
     std::vector<double>{0.5 * radius, radius},
     construct_quadrature(quadrature),
-    0.0);
+    0.0,
+    radius / (2 * Utilities::pow(2, n_global_refinements)) * 0.5);
```
blaisb, line 112: Can you add documentation as to why you give this value here?

tests/mortar/poisson_3d_02.cc
```diff
@@ -115,7 +115,8 @@ main(int argc, char **argv)
                               1 * Utilities::pow(2, n_global_refinements)},
     std::vector<double>{radius, radius},
     construct_quadrature(quadrature),
-    rotate_pi);
+    rotate_pi,
+    radius / Utilities::pow(2, n_global_refinements) * 0.5);
```
blaisb, line 123: Can you add documentation as to why you give this value here?

blaisb (changes requested):
Nothing major to change. Just some naming convention and some small changes there and there. This is very good work. You are catching on quickly, bravo!

Julamipolymtl (approved):
On my side once #1976 is merged this is good to go. Great job!

include/core/mortar_coupling_manager.h
jatred: Yeah right now since we can't really have a bottom in our mortar geometry it won't be a problem, but I'll definitely have to take it into account whenever we want to add that feature.

source/core/mortar_coupling_manager.cc
jatred: The direction is calculated in 3 different functions in a slightly different way and with different ExcMessage, should I make them all the same for standardization?

include/core/mortar_coupling_manager.h
jatred: I could have done it that way, but all of the MortarManager constructors (Base, Linear and Circular) are empty aside from a single line in Linear, so I wanted to keep it that way. There was no other reason behind it so if it's better to put it in inside the constructor I can change it to that.

tests/mortar/poisson_3d_01.cc
jatred: I added a comment to explain where it comes from but it might be overkill, let me know how to simplify it if need be.

blaisb:
image
 You need to explicitely cast from unsigned to integer or vice-versa, but you can't do such a comparison directly :) .

source/core/mortar_coupling_manager.cc
brunaccampos: That's a good point, this calculation is repeated in a few functions now. It would be nice to make a separate function for this, but it can be done in a separate PR if it's easier :)

brunaccampos:
@jatred Last thing to do would be to shift the mesh for the 3d mms case

tests/mortar/poisson_3d_01.cc
blaisb: Nah it's a good comment.

blaisb (approved):
I think it's ok to keep the minimum height calculation outside for now if you want if @brunaccampos is ok with that :). Anyway this is easy to refactor in the future.

blaisb:
This is good from my point of view . Just mark it as ready to merge when you and @brunaccampos are satisfied with it and then I can merge :).

PR #1978, opened 2026-05-01
Mortar: Variable mortar cell height along rotation axis

jatred:
### Description

This PR enables the mortar method to handle varying mortar cell heights in the rotation-axis direction. The rotation axis is still assumed to be the z-axis. The next step is to support mortar interfaces with shapes other than a cylinder, making it possible to use solids of revolution instead.

### Solution

Instead of only storing the minimum domain height, we now store the height of all unique stages. With this information, we can calculate the ID of each cell and the position of the quadrature points without relying on the assumption that the cell height is constant along the rotation axis.

### Testing

A new test has been added for the new feature. Some tests had to be modified to take into account the new structure of the mortar manager class. 

Code related list:

Pull request related list:
- [x] Labels are applied


include/core/mortar_coupling_manager.h
brunaccampos, line 36: Add `stage_heights` in the description

brunaccampos, line 56: Same here for `stage heights`

include/core/mortar_coupling_manager.h
```diff
@@ -213,9 +213,9 @@ class MortarManagerBase
   const unsigned int n_quadrature_points;
   /// Rotation angle for the inner domain
   const double rotation_angle;
-  /// Minimum cell center height at the mortar interface along the rotation axis
-  /// value in 3D problems
-  const double minimum_height;
+  /// Stage heights at the mortar interface along the rotation axis
```
brunaccampos, line 220: I think this needs a better description. What is a stage? Is it defined at the center of the cell or at the edge?

include/core/mortar_coupling_manager.h
```diff
@@ -284,19 +284,18 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
 /**
- * @brief Computes the minimum cell center height in the direction of the
+ * @brief Compute the stage heights in the direction of the
```
brunaccampos, line 292: Same here: mention the position where each stage is computed

include/core/mortar_coupling_manager.h
```diff
@@ -284,19 +284,18 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
  * rotation axis at the mortar interface
  * @param[in] triangulation The triangulation object
  * @param[in] mortar_parameters The information about the mortar method
  * control, including the rotor mesh parameters
  *
- * @return Minimum cell height in the rotation axis direction. At the moment,
- * it is assumed that the rotation axis is in z
+ * @return Vector of stage heights in the rotation axis direction.
```
brunaccampos, line 301:
```suggestion
 * @return Vector of stage heights in the rotation axis direction. At the moment, it is assumed that the rotation axis is in z.
```
I think we still assume rotation in z?

include/core/mortar_coupling_manager.h
```diff
@@ -448,12 +447,12 @@ MortarManagerBase<dim>::MortarManagerBase(unsigned int n_subdivisions,
...
                                           const Quadrature<dim2> &quadrature_in,
                                           const double rotation_angle,
-                                          const double minimum_height)
+                                          const double stage_heights)
   : MortarManagerBase(std::vector<unsigned int>{n_subdivisions, 1},
                       std::vector<double>{radius, 1.0},
                       quadrature_in,
                       rotation_angle,
-                      minimum_height)
+                      std::vector<double>{stage_heights, stage_heights + 1.0})
```
brunaccampos, line 467: Why stage_heights +1?

source/core/mortar_coupling_manager.cc
```diff
@@ -158,7 +160,7 @@ MortarManagerBase<dim>::get_points(const Point<dim> &face_center,
   double       delta_1 = 1.0;
 
   if constexpr (dim == 3)
-    delta_1 = radius[1] / n_subdivisions[1];
+    delta_1 = stage_heights[id_out_plane + 1] - stage_heights[id_out_plane];
```
brunaccampos, line 164: Can you add a comment on what is the definition of delta_1?

source/core/mortar_coupling_manager.cc
```diff
@@ -801,11 +810,15 @@ compute_minimum_height(const Triangulation<dim>      &triangulation,
...
           "Rotation axis is not aligned with a coordinate direction."));
 
-      // Minimum coordinate in the direction of the rotation axis
-      double minimum_height_local = std::numeric_limits<double>::max();
+      // Container storing all vertex coordinates in the rotation axis direction
+      // for the local cells at the mortar boundary
+      std::vector<double> stage_heights_local;
+      // Smallest mortar cell face diameter, used to set the tolerance for
+      // height comparison
+      double minimum_face_size_local = std::numeric_limits<double>::max();
```
brunaccampos: Maybe name it minimum_face_diameter_local instead? To match the variable description

source/core/mortar_coupling_manager.cc
```diff
@@ -801,11 +810,15 @@ compute_minimum_height(const Triangulation<dim>      &triangulation,
...
 
-      // Loop over the cells to find the minimum coordinate in the rotation axis
-      // direction
+      // Loop over the cells to collect the vertex coordinates in the rotation
```
brunaccampos:
```suggestion
      // Loop over the cells to store the vertex coordinates in the rotation
```

source/core/mortar_coupling_manager.cc
```diff
@@ -820,17 +833,49 @@ compute_minimum_height(const Triangulation<dim>      &triangulation,
...
+                          face->vertex(vertex_no)[direction]);
+                    }
                 }
             }
         }
-      // Return the minimum value across all processes
-      return Utilities::MPI::min(minimum_height_local,
-                                 triangulation.get_mpi_communicator());
+      // Set the minimum face size over all processes
+      double minimum_face_size =
```
brunaccampos:
```suggestion
      double minimum_face_diameter =
```

source/core/mortar_coupling_manager.cc
```diff
@@ -820,17 +833,49 @@ compute_minimum_height(const Triangulation<dim>      &triangulation,
...
+        Utilities::MPI::min(minimum_face_size_local,
+                            triangulation.get_mpi_communicator());
+
+      // Collect the vertex coordinates in the rotation axis direction from all
```
brunaccampos:
```suggestion
      // Store the vertex coordinates in the rotation axis direction from all
```

brunaccampos (changes requested):
Great work! This is a nice way to generalize the cell height. Just some small comments to improve description of variables

include/core/mortar_coupling_manager.h
jatred: We do in the code, but this method is still generalized for all main axis directions. That's why I didn't add this part to the description.

include/core/mortar_coupling_manager.h
jatred: That MortarManagerBase is when we feed 2D inputs and it "transfers" it to 3D; we give it 1 n_subdivsions, 1 radius and 1 stage_heights (which should be 0 but still works if it's not).

When we transform these inputs to 3D, we say we have 1 subdivision parallel to the rotation axis and that the height of this subdivision is 1.0 (radius[1]). So to keep the same logic, I say that the 3D stage heights are "stage_heigts" and "stage_heights + radius[1] = stage_heights + 1.0"

Let me know if I should change it or clarify it in the code.

include/core/mortar_coupling_manager.h
```diff
@@ -213,9 +217,10 @@ class MortarManagerBase
   const unsigned int n_quadrature_points;
   /// Rotation angle for the inner domain
   const double rotation_angle;
-  /// Minimum cell center height at the mortar interface along the rotation axis
-  /// value in 3D problems
-  const double minimum_height;
+  /// Stage heights at the mortar interface along the rotation axis
+  /// in 3D problems. A stage is defined as a layer of vertices with the same
+  /// height in the rotation axis direction.
```
```suggestion
  /// in 3D problems.  A stage is defined as a plane perpendicular to the rotation axis containing vertices of the same height.
```

include/core/mortar_coupling_manager.h
```diff
@@ -284,19 +289,21 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
 /**
- * @brief Computes the minimum cell center height in the direction of the
- * rotation axis at the mortar interface
+ * @brief Compute the stage heights in the direction of the
+ * rotation axis at the mortar interface. Stages are defined as layers of
+ * vertices with the same height in the rotation axis direction, so there will
+ * always be one more stage than the number of cells in the rotation axis
+ * direction.
```
```suggestion
 * rotation axis at the mortar interface. Stages are defined as planes 
 * perpendicular to the rotation axis containing vertices of the same 
 * height. There will always be one more stage than the number of 
 * cells in the rotation axis direction.
```

include/core/mortar_coupling_manager.h
brunaccampos: Okay, makes sense!

include/core/mortar_coupling_manager.h
brunaccampos: Oh got it! Thanks. Yes it does follow the same logic for the subdivisions so I think it's good like this

brunaccampos (approved):
Looks good to me :)

applications_tests/lethe-fluid/mortar_asymmetric_height.prm
```diff
@@ -0,0 +1,147 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+# Checks the 3D matrix-based mortar implementation with a varying cell height along the rotation axis
+
+# Listing of Parameters
+#----------------------
+
```
blaisb, line 8: Coul you add a description of what the test checks and does explicitely?
We try to always have a short piece of text at the top that explains what the test actually tests so that someone can easily understand what's being tested here :)

blaisb (changes requested):
Just come comments. I am continuing the review tonight

applications_tests/lethe-fluid/mortar_asymmetric_height.prm
jatred: I already have a small description on line 3, should it be more detailed?

antonellasuccar (approved):

applications_tests/lethe-fluid/mortar_asymmetric_height.mpirun=3.output
```diff
@@ -0,0 +1,46 @@
...
+*****************************
+Steady iteration:        3/3
+*****************************
+-----------------
+Mortar Interface
+-----------------
+Rotor grid angle:    1.670000e-01 rad 
+Rotor grid velocity: 0.000000e+00 rad/time 
+
+   Number of active cells:       1152
```
blaisb, line 44: This test does not compare any analytical solution or something. It would be good to have some analytical solution to compare to, even if it's just a pressure profile derived from a source term like we had looked at together. Otherwise it is very difficult to check if the test is "working or not". the goal is really to make the test as sensitive as possible

applications_tests/lethe-fluid/mortar_asymmetric_height.prm
blaisb: Can you just separate it from the license? :)

applications_tests/lethe-fluid/mortar_asymmetric_height.prm
```diff
@@ -0,0 +1,147 @@
...
+#---------------------------------------------------
+# Source term
+#---------------------------------------------------
+
+subsection source term
+  subsection fluid dynamics
+    set Function expression = 1;0;0;0
+  end
+end
```
blaisb, line 125: You have a source term, so I think you can provide an analytical solution for profile and it should work :). Otherwise maybe use an MMS or something like this? just so we have something to check that it's working.

include/core/mortar_coupling_manager.h
```diff
@@ -34,13 +34,15 @@ class MortarManagerBase
...
+   * @param[in] stage_heights Vector containing the unique vertex heights at the
+   * mortar interface
    */
   template <int dim2>
   MortarManagerBase(unsigned int            n_subdivisions,
                     double                  radius,
                     const Quadrature<dim2> &quadrature,
                     const double            rotation_angle,
-                    const double            minimum_height = 0.0);
+                    const double            stage_heights = 0.0);
```
blaisb, line 45: You document it as a vector, but this is a value?
I am not sure what's happening here :)

include/core/mortar_coupling_manager.h
```diff
@@ -52,13 +54,15 @@ class MortarManagerBase
...
+   * @param[in] stage_heights Vector containing the unique vertex heights at the
+   * mortar interface
    */
   template <int dim2>
   MortarManagerBase(const std::vector<unsigned int> &n_subdivisions,
                     const std::vector<double>       &radius,
                     const Quadrature<dim2>          &quadrature,
                     const double                     rotation_angle,
-                    const double                     minimum_height = 0.0);
+                    const std::vector<double> stage_heights = {0.0, 1.0});
```
blaisb, line 65: OK here it is a vector. So you just have two constructors. You just need to document the two constructors differently :)

include/core/mortar_coupling_manager.h
```diff
@@ -330,6 +337,8 @@ class MortarManagerCircle : public MortarManagerBase<dim>
    * @param[in] radius Radius at the mortar interface
    * @param[in] quadrature Quadrature for local cell operations
    * @param[in] rotation_angle Rotation angle for the inner domain
+   * @param[in] stage_heights Vector containing the unique vertex heights at the
```
blaisb: Here also documentation does not match constructor

source/core/mortar_coupling_manager.cc
```diff
@@ -421,10 +425,16 @@ MortarManagerBase<dim>::get_config(const Point<dim> &face_center,
...
+      // Return the iterator of the first element in stage_heights that is
+      // greater than the face center height
+      auto upper_height_iterator = std::upper_bound(
+        stage_heights.begin(),
+        stage_heights.end(),
+        face_center[2]); // TODO Generalize for x and y directions
+      // The id_out_plane of the cell can be obtained with the distance between
+      // the iterator obtained and the lowest stage height iterator
+      id_out_plane =
+        std::distance(stage_heights.begin(), upper_height_iterator) - 1;
```
blaisb, line 437: If this only works for z, can you at least assert at one point in the code that if the direction is not z, then this throws something so that someone trying to use it in the "non correct" way would get an error? I'm not sure if this is not already done, in that case don't worry then.

blaisb (changes requested):
Some changes. Nothing big, this is great work.

include/core/mortar_coupling_manager.h
jatred: Yep my mistake, I forgot to adapt the description for the 2D constructors

source/core/mortar_coupling_manager.cc
jatred: We have an AssertThrow if it's not aligned with x, y or z even if x and y are not supported. I added another one right after it that we can remove when x and y are supported; it's a bit redundant but it limits the amount of changes we'll need to make when we expand to x and y

blaisb: That's great then :)

blaisb (approved):
All good to merge. Good work.

PR #1984, opened 2026-05-07
Solid in mf

OresteMarquis:
### Description

When using the matrix free solver with solid in the domain, the GMG preconditioner was not taking into the solid domain in the coarser level. It follows that the solver could not converge.

Additionally, the transfinite manifold is not supported for GMG, so it is now possible to disable it when necessary in the `grid_uniform_channel_with_meshed_cylinder` and `grid_uniform_channel_with_meshed_square_prism`. A test for the grid with the cylinder has also been added to the test suit of Lethe since it was missing.

### Solution

Bruno refactored the solid constraints so it can now apply the constraints on all levels of multi-grid preconditioner. The helper function is now on its own in `solid_constraints` instead of of being own by  NavierStokesBase. This way the matrix-free operator can call the function to update the different multi-grid level constraints. 

### Testing

This bug fix has been tested using the concentric heat exchanger example adapted from the examples of Lethe to run in matrix free. A custom PRM using the `grid_uniform_channel_with_meshed_cylinder` mesh and the matrix free solver has also been use to validate the solver. 

The changes in the custom Lethe mesh also now has a test to keep track of it.

### Documentation

No new documentation required.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to the Lethe project will be documented in this file.
 The changelog for the previous releases of Lethe are located in the release_notes folder.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Master] - 2026/05/11
+
+### Fixed
+
+- MAJOR Fixed the usage of matrix free solver with solid material id in the domain for both the GCMG and LSMG multigrid preconditioners. This has been tested with the `grid_uniform_channel_with_meshed_cylinder` and required the mesh to have an additional option to disable the transfinite manifold has it does not work for multi-grid which have also been added to this PR. Finally, a test for the mesh `grid_uniform_channel_with_meshed_cylinder` has also been created to keep track of its changes in the future. [#1984](https://github.com/chaos-polymtl/lethe/pull/1984)
 
 ## [Sample] - YYYY/MM/DD
 
--- a/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
+++ b/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
@@ -318,8 +318,7 @@ VANSOperator<dim, number>::do_cell_integral_local(
   VectorizedArray<number> cell_mask(1.0);
   if (this->has_solids)
     {
-      Assert(cell < this->solid_cell_mask.size(),
-             ExcDimensionMismatch(cell, this->solid_cell_mask.size()));
+      AssertIndexRange(cell, this->solid_cell_mask.size());
       cell_mask = this->solid_cell_mask[cell];
     }
 
--- a/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
+++ b/source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
@@ -314,6 +314,15 @@ VANSOperator<dim, number>::do_cell_integral_local(
 
   const unsigned int cell = integrator.get_current_cell_index();
 
+  // Solid-cell mask (1.0 for fluid lanes, 0.0 for solid lanes).
+  VectorizedArray<number> cell_mask(1.0);
+  if (this->has_solids)
+    {
+      Assert(cell < this->solid_cell_mask.size(),
+             ExcDimensionMismatch(cell, this->solid_cell_mask.size()));
+      cell_mask = this->solid_cell_mask[cell];
+    }
+
   // To identify whether the problem is transient or steady
   bool transient =
     (time_stepping_is_bdf(this->simulation_control->get_assembly_method())) ?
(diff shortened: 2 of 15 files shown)


source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
```diff
@@ -314,6 +314,15 @@ VANSOperator<dim, number>::do_cell_integral_local(
 
   const unsigned int cell = integrator.get_current_cell_index();
 
+  // Solid-cell mask (1.0 for fluid lanes, 0.0 for solid lanes).
+  VectorizedArray<number> cell_mask(1.0);
+  if (this->has_solids)
+    {
+      Assert(cell < this->solid_cell_mask.size(),
+             ExcDimensionMismatch(cell, this->solid_cell_mask.size()));
```
OresteMarquis: Not sure the message in  `ExcDimensionMismatch` is appropriate here?

source/solvers/fluid_dynamics_matrix_free_operators.cc
```diff
@@ -193,6 +193,13 @@ NavierStokesOperatorBase<dim, number>::reinit(
 
   this->compute_element_size();
 
+  this->has_solids = (this->properties_manager &&
```
OresteMarquis, line 196: Why do we need both conditions? I would have thought that `this->has_solids =  this->properties_manager->get_number_of_solids() > 0)` would be enough.

OresteMarquis (commented):
J'aime ca comme fixe! C'est bien aussi tous les nouveaux assert :) Pour ma part c'est beau!

blaisb:
@OresteMarquis maybe you changed the arguments the grid hence your test breaking?

OresteMarquis:
> @OresteMarquis maybe you changed the arguments the grid hence your test breaking?

There were no test before, I added it in this PR hahaha, its just the debug was  complaining about trying to assign a manifold to the default `numbers::flat_manifold_id`

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
blaisb: Dimension mismatch is to indicate that they both have the same size (not dimension as dim, but dimension as number of elements), so it works no?

source/solvers/fluid_dynamics_matrix_free_operators.cc
blaisb: I'll need to double check. I think the properties_manager is not always instantiated, but I need to double check that actually. It's a good point.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## [Master] - 2026/05/08
 
+### Fixed
+
+- MINOR Fixed the usage of matrix free solver with solid material id in the domain. This has been tested with the `grid_uniform_channel_with_meshed_cylinder` and required the mesh to have an additional option to disable the transfinite manifold has it does not work for multi-grid. A test for the mesh has also been added to keep track of its changes in the future. [#1984](https://github.com/chaos-polymtl/lethe/pull/1984)
```
blaisb: Can you update the Changelog entry @OresteMarquis  to include the other changes?

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
OresteMarquis: Its just that the condition is `cell < this->solid_cell_mask.size()`, which in my head means the cell index needs to be within the size of the mask. But the error message of ExcDimensionMismatch is : `"Two sizes or dimensions were supposed to be equal, but aren't. They are " << arg1 << " and " << arg2 << '.'` which doesn't convey the same information no?

CHANGELOG.md
OresteMarquis: I tried to make it clearer, but I am not sure of what other changes you are referring to?

source/fem-dem/fluid_dynamics_vans_matrix_free_operators.cc
blaisb: Yeah you are right. That's a bit of an abuse because the "cell" here won't be a size per se. I will fix that.

blaisb: Yeah the right approach would be to use AssertIndexRange(value,range). I have fixed it

blaisb (approved):
Good on my side. I have puished some last changes.

OresteMarquis:
Good for me also! Do we want to address the TODO in NS_base now?

PR #2012, opened 2026-06-03
Add gaussian filter for void fraction

blaisb:
### Description

The QCM filtering method only supported a Heaviside filter. Another popular alternative in the literature is to use a Gaussian filter. This PR adds the capability to support both a Gaussian or a Heaviside filter within the QCM framework. For both approaches, the classical normalisation is used to ensure that the volume of the particles is conserved no matter what.

### Testing

A unit test has been added for the projection functionalities. I need to test the code a bit more with a few different use case to see if there's a difference in the results or not.

### Documentation

The qcm sphere diameter has been changed to qcm filtering length. For a spherical heaviside, this length if the diameter. For a Gaussian, it is equal to two standard deviations. A new parameter (qcm_filter_type) has been added to allow the use to choose the type of filter.

Code related list:

Pull request related list:
- [x] Labels are applied


OresteMarquis:
I have not reviewed it yet, but something came to my mind while reading the description. Why is the filtering length of the Gaussian only 2 sigma and not 3?

blaisb:
> I have not reviewed it yet, but something came to my mind while reading the description. Why is the filtering length of the Gaussian only 2 sigma and not 3?

I wanted something that was equivalent to the diameter of the sphere as a metric of length. It could be 2 or 3 sigma, it's a bit of an arbitrary choice. At 2 sigma you have 95% of the volume, which is already not too bad, but not too steep.
It's debatable i guess

OresteMarquis:
I guess it doesn't change much because it is not required to be a compact support kernel and it is normalized in the end. I was just curious of the choice, maybe at some point it could be interesting to see the effect.

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -38,8 +39,28 @@ If the ``mode`` chosen is ``pcm``, then the void fraction is calculated using th
...
+* The ``qcm sphere equal cell volume`` determines whether or not we want to use a reference sphere with the same volume as the element in which it is located. If it is disabled, then each sphere will have a radius equal to the size of the element in which it is located. This parameter is important only when the ``qcm smoothing length`` is not used or is set to 0.
+* The ``qcm filter type`` selects the kernel used to weigh particle contributions onto the mesh in the ``qcm`` method. Two options are available: ``spherical`` (default) and ``gaussian``. The user always sets ``qcm smoothing length`` to the **diameter** of the averaging sphere (spherical filter) or to **2σ** (Gaussian filter); the code halves this value internally to obtain the kernel characteristic length. For each particle of radius :math:`r_p` and volume :math:`V_p`, the kernel value at distance :math:`d` from the particle centre to a quadrature point is evaluated and then normalised by the sum of all kernel values over the neighbour-cell stencil so that the total particle volume is conserved.
```
OresteMarquis:
```suggestion
* The ``qcm filter type`` selects the kernel used to weigh particle contributions onto the mesh in the ``qcm`` method. Two options are available: ``spherical`` (default) and ``gaussian``. The user always sets ``qcm smoothing length`` to the **diameter** of the averaging sphere (spherical filter) or to **2σ** (Gaussian filter); the code halves this value internally to obtain the kernel characteristic length (i.e., the radius or the standard deviation respectively). For each particle of radius :math:`r_p` and volume :math:`V_p`, the kernel value at distance :math:`d` from the particle centre to a quadrature point is evaluated and then normalised by the sum of all kernel values over the neighbour-cell stencil so that the total particle volume is conserved.
```

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -38,8 +39,28 @@ If the ``mode`` chosen is ``pcm``, then the void fraction is calculated using th
...
+
+  * **Spherical** (``spherical``): the kernel is the intersection volume between the particle sphere (radius :math:`r_p`) and an averaging sphere of radius :math:`\ell = \tfrac{1}{2}\,d_{\text{filter}}` centred at the quadrature point. In 3-D, for an overlap distance :math:`0 < d < r_p + \ell`, this is
```
OresteMarquis:
```suggestion
  * **Spherical** (``spherical``): the kernel is the intersection volume between the particle sphere (radius :math:`r_p`) and an averaging sphere of radius :math:`r_{\mathrm{avg}} = \tfrac{1}{2}\,d_{\text{filter}}` centred at the quadrature point. In 3-D, for an overlap distance :math:`0 < d < r_p +r_{\mathrm{avg}}`, this is
```

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -38,8 +39,28 @@ If the ``mode`` chosen is ``pcm``, then the void fraction is calculated using th
...
+
+    .. math::
+
+       K_{\text{sph}}(r_p,\ell,d) = \frac{\pi\,(r_p + \ell - d)^2
```
OresteMarquis:
```suggestion
       K_{\text{sph}}(r_p,r_{\mathrm{avg}},d) = \frac{\pi\,(r_p + r_{\mathrm{avg}}- d)^2
```

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -38,8 +39,28 @@ If the ``mode`` chosen is ``pcm``, then the void fraction is calculated using th
...
+         \left(d^2 + 2d\,r_p - 3r_p^2 + 2d\,\ell + 6\,r_p\,\ell - 3\ell^2\right)}
```
OresteMarquis:
```suggestion
         \left(d^2 + 2d\,r_p - 3r_p^2 + 2d\,r_{\mathrm{avg}} + 6\,r_p\,r_{\mathrm{avg}} - 3r_{\mathrm{avg}}^2\right)}
```

doc/source/parameters/unresolved-cfd-dem/void-fraction.rst
```diff
@@ -38,8 +39,28 @@ If the ``mode`` chosen is ``pcm``, then the void fraction is calculated using th
...
+         {12\,d}
+
+    and zero when :math:`d \geq r_p + \ell`, or :math:`V_p` when :math:`d \leq \ell - r_p`.
```
OresteMarquis:
```suggestion
    and zero when :math:`d \geq r_p +r_{\mathrm{avg}}`, or :math:`V_p` when :math:`d \leq r_{\mathrm{avg}} - r_p`.
```

OresteMarquis (approved):
Good with me as it is. Just suggested some personal preferences :)

source/fem-dem/particle_projector.cc
antonellasuccar, line 819:
```suggestion
// Loop over neighboring cells to determine if a given
// neighboring particle intersects the current reference
// kernel
```

source/fem-dem/particle_projector.cc
```diff
@@ -801,9 +825,9 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
                   // diameter was given by the user the value is already
                   // defined since it is not dependent on any measure of the
                   // active cell
```
antonellasuccar, line 831:
```suggestion
// Define the length of the reference kernel to be used as
// the averaging volume for the QCM. If the reference kernel
// length was given by the user the value is already
// defined since it is not dependent on any measure of the
// active cell
```

source/fem-dem/particle_projector.cc
antonellasuccar: Should the comment here be:
``
// Loop over neighboring cells to determine if the current
// particle intersects the kernels centred on quadrature
// points in neighboring cells
`` instead?

antonellasuccar, line 859: Same thing here

Should the comment here be:
``// Loop over periodic neighboring cells to determine if the current // particle intersects the kernels centred on quadrature // points in neighboring cells instead?``

source/fem-dem/particle_projector.cc
```diff
@@ -970,9 +995,9 @@ ParticleProjector<dim>::calculate_void_fraction_quadrature_centered_method()
                       // Calculate the normalized particle contribution
                       particles_volume_in_sphere +=
```
antonellasuccar: Is it better to rename ``particles_volume_in_sphere`` ``particles_volume_in_kernel``?

tests/fem-dem/particle_projector_04.cc
```diff
@@ -0,0 +1,259 @@
...
+  constexpr unsigned int n_refinements = 1;
+
+  // We make a background triangulation which consists in a 1x1x1 cube.
+  parallel::distributed::Triangulation<3> domain_triangulation(MPI_COMM_WORLD);
+  GridGenerator::hyper_cube(domain_triangulation, 0, 1, false);
+
+  // We refine this triangulation for a given number of refinement
+  domain_triangulation.refine_global(n_refinements);
+
+  // We initialze the particle handler
```
antonellasuccar:
```suggestion
  // We initialize the particle handler
```

tests/fem-dem/particle_projector_04.cc
```diff
@@ -0,0 +1,259 @@
...
+      fe_values_void_fraction.reinit((cell));
+      fe_values_void_fraction.get_function_values(
+        particle_projector.void_fraction_solution, void_fraction_values);
+      for (unsigned int q = 0; q < fe_values_void_fraction.n_quadrature_points;
+           ++q)
+        total_particle_volume += fe_values_void_fraction.get_JxW_values()[q] *
+                                 (1. - void_fraction_values[q]);
+    }
+  deallog << "Total particle volume " << total_particle_volume << std::endl;
+}
```
antonellasuccar, line 209: I would maybe add an empty line here for a clearer output
```suggestion
deallog.get_file_stream() << std::endl;
}
```

antonellasuccar (approved):
Nice PR :) Not much to add. I think we just need to update the Theory section to add the kernel corresponding to the gaussian filter, but I can do that in a subsequent PR

source/fem-dem/particle_projector.cc
blaisb: I'm not sure this adds much. I prefer the current comment to be honest.

blaisb: Same as above

source/fem-dem/particle_projector.cc
blaisb: Done

PR #2016, opened 2026-06-09
Mortar method rotation axis from z to x, y or z

jatred:
### Description

The mortar method now work with rotation axis x or y, and not just z. The same case rotated around x, y and z gives the same results.

### Testing

Tests have been modified to work with the modified MortarManager, a test will be added or modified with a x or y rotation axis

Code related list:

Pull request related list:
- [x] Labels are applied


source/core/mortar_coupling_manager.cc
blaisb, line 637: This is where your bug is I think.

include/core/mortar_coupling_manager.h
```diff
@@ -42,6 +43,7 @@ class MortarManagerBase
                     double                  radius,
                     const Quadrature<dim2> &quadrature,
                     const double            rotation_angle,
+                    const unsigned int      rotation_axis = 0,
```
```suggestion
                    const unsigned int      rotation_axis_direction = 0,
```
`rotation_axis` seems like a Tensor<dim>; since this is returning the direction, I'd change the variable name

include/core/mortar_coupling_manager.h
```diff
@@ -288,6 +294,18 @@ Quadrature<dim>
 construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
+/**
+ * @brief Gets the direction of the rotation axis
```
```suggestion
 * @brief Return the rotation axis direction
```

include/core/mortar_coupling_manager.h
```diff
@@ -288,6 +294,18 @@ Quadrature<dim>
...
+ *
+ * @param[in] mortar_parameters The information about the mortar method
+ * control, including the rotor mesh parameters
+ *
+ * @return Direction of the rotation axis
+ */
+template <int dim>
+unsigned int
+get_rotation_axis(const Parameters::Mortar<dim> &mortar_parameters);
```
```suggestion
get_rotation_axis_direction(const Parameters::Mortar<dim> &mortar_parameters);
```

include/core/utilities.h
```diff
@@ -1001,6 +1001,39 @@ point_to_angle(const Point<dim> &point,
...
+ * @param[in] center_of_rotation Center of rotation
+ * @param[in] rotation_axis Direction of rotation axis (0=x, 1=y, 2=z)
+ *
+ * @return Angle in the plane perpendicular to the rotation axis
+ */
+template <int dim>
+inline double
+point_to_angle(const Point<dim>  &point,
+               const Point<dim>  &center_of_rotation,
+               const unsigned int rotation_axis)
```
brunaccampos: You can just change the existing function to have a `rotation_axis_direction` parameter with a default value, and use only the generalized version of the function

release_notes/current/2026_06_26_Tremblay.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/06/25
+
+### Added
+
+- MINOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. The code is still limited to an x, y or z rotation axis prassing through the origin. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
brunaccampos:
```suggestion
- MINOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. The code is still limited to an x, y or z rotation axis passing through the origin. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```

source/core/mortar_coupling_manager.cc
```diff
@@ -173,12 +173,17 @@ MortarManagerBase<dim>::get_points(const Point<dim> &face_center,
...
-                quadrature.point(q)[1] *
-                  delta_1); // TODO Generalize for x and y directions
+            {
+              Point<3> coordinates;
+              coordinates[this->rotation_axis] =
+                stage_heights[id_out_plane] + quadrature.point(q)[1] * delta_1;
+              coordinates[(this->rotation_axis + 1) % 3] = x[0];
+              coordinates[(this->rotation_axis + 2) % 3] = x[1];
+
+              points.emplace_back(coordinates[0],
```
brunaccampos: Is there a way to modify the `angle_to_point()` function inside `from_1D()` so that this choice of the rotation axis is done inside there? This way you wouldn't need to have this intermediate step here

source/core/mortar_coupling_manager.cc
```diff
@@ -473,45 +488,15 @@ compute_number_interface_cells(const Triangulation<dim>      &triangulation,
   unsigned int n_subdivisions_plane_local = 0;
   // Tolerance for rotor radius computation
   const double radius_tolerance = mortar_parameters.radius_tolerance;
+  // Direction of the rotation axis
+  const unsigned int rotation_axis = get_rotation_axis(mortar_parameters);
```
brunaccampos:
```suggestion
  const unsigned int rotation_axis_direction = get_rotation_axis_direction(mortar_parameters);
```

source/core/mortar_coupling_manager.cc
```diff
@@ -600,42 +585,13 @@ compute_interface_dimensions_circular(
   // Minimum rotation angle in initial mesh configuration
   double pre_rotation_min = std::numeric_limits<double>::max();
 
-  // Rotation axis direction. Used in 3D case
-  unsigned int direction = 0;
+  // Direction of the rotation axis
+  const unsigned int rotation_axis = get_rotation_axis(mortar_parameters);
```
brunaccampos:
```suggestion
  const unsigned int rotation_axis_direction = get_rotation_axis_direction(mortar_parameters);
```

source/core/mortar_coupling_manager.cc
```diff
@@ -948,9 +927,23 @@ template <int dim>
 double
 MortarManagerCircle<dim>::to_1D(const Point<dim> &point) const
 {
-  return std::fmod(point_to_angle(point, this->center_of_rotation) -
-                     pre_rotation_angle + 2 * numbers::PI,
-                   2 * numbers::PI);
+  if constexpr (dim == 3)
+    {
+      Point<3> p_proj(point[(this->rotation_axis + 1) % 3],
```
brunaccampos: Can you use more descriptive names for `p_proj` and `c_proj`? Maybe add a comment description if needed

source/core/mortar_coupling_manager.cc
```diff
@@ -1092,15 +1085,13 @@ CouplingOperator<dim, Number>::CouplingOperator(
...
-                    Point<dim - 1> temp;
-                    for (int i = 0, j = 0; i < dim; ++i)
-                      if ((face_no / 2) != static_cast<unsigned int>(i))
-                        temp[j++] = p[i];
+                    Point<dim - 1>     temp;
+                    const unsigned int axis = face_no / 2;
 
-                    if ((dim == 3) && ((face_no / 2) == 1))
-                      std::swap(temp[0], temp[1]);
+                    for (int t = 0; t < dim - 1; ++t)
```
brunaccampos, line 1073: Can you also add a comment here on what is this doing

brunaccampos (changes requested):
Nice generalization! Just a few comments

source/core/mortar_coupling_manager.cc
jatred: With a generalized point_to_angle function I can just scrap this part, it's a lot cleaner now

include/core/utilities.h
```diff
@@ -990,33 +990,80 @@ delete_vtu_and_pvd_files(const std::string &output_path);
...
+ * @param[in] rotation_axis Direction of rotation axis (0=x, 1=y, 2=z)
+ *
+ * @return Angle in the plane perpendicular to the rotation axis
+ */
 template <int dim>
 inline double
-point_to_angle(const Point<dim> &point,
-               const Point<dim> &center_of_rotation = Point<dim>())
+point_to_angle(const Point<dim>  &point,
+               const Point<dim>  &center_of_rotation,
```
brunaccampos:
```suggestion
               const Point<dim>  &center_of_rotation = Point<dim>(),
```
I think we need a default value here, maybe this is why some tests are failing

jatred: Yep, it fixed one but there is still a last one failing

include/core/mortar_coupling_manager.h
```diff
@@ -288,6 +294,18 @@ Quadrature<dim>
 construct_quadrature(const Quadrature<dim>         &quadrature,
                      const Parameters::Mortar<dim> &mortar_parameters);
 
+/**
+ * @brief Return the rotation axis direction for the mortar interface
+ *
+ * @param[in] mortar_parameters The information about the mortar method
+ * control, including the rotor mesh parameters
+ *
+ * @return Direction of the rotation axis
```
```suggestion
 * @return Direction of the rotation axis (0=x, 1=y, or 2=z)
```

include/core/utilities.h
```diff
@@ -984,39 +984,68 @@ void
 delete_vtu_and_pvd_files(const std::string &output_path);
 
 /**
- * @brief Converts point to angle (in radians)
- * @param[in] point Point in cartesian coordinates (x-y plane)
+ * @brief Converts point to angle (in radians) based on rotation axis
+ * @param[in] point Point in cartesian coordinates
  * @param[in] center_of_rotation Center of rotation
+ * @param[in] rotation_axis Direction of rotation axis (0=x, 1=y, 2=z)
```
brunaccampos:
```suggestion
 * @param[in] rotation_axis_direction Direction of rotation axis (0=x, 1=y, 2=z)
```

include/core/utilities.h
```diff
@@ -984,39 +984,68 @@ void
...
- * @return Angle between point and the x-axis
+ * @return Angle in the plane perpendicular to the rotation axis
  */
 template <int dim>
 inline double
-point_to_angle(const Point<dim> &point,
-               const Point<dim> &center_of_rotation = Point<dim>())
+point_to_angle(const Point<dim>  &point,
+               const Point<dim>  &center_of_rotation = Point<dim>(),
+               const unsigned int rotation_axis      = 2)
```
brunaccampos:
```suggestion
               const unsigned int rotation_axis_direction      = 2)
```

include/core/utilities.h
```diff
@@ -984,39 +984,68 @@ void
...
+    {
+      const unsigned int coord1 = (rotation_axis + 1) % 3;
+      const unsigned int coord2 = (rotation_axis + 2) % 3;
+
+      return std::fmod(std::atan2(point[coord2] - center_of_rotation[coord2],
+                                  point[coord1] - center_of_rotation[coord1]) +
+                         2 * numbers::PI,
+                       2 * numbers::PI);
+    }
+  else
```
brunaccampos, line 1008: I'm not sure but I think we can assume that 2D cases will always use x,y coordinates, and the default rotation axis direction will be 2. Meaning that we could remove the `if dim==3` and use the general expression only
This wouldn't change many things though so maybe just leave like this, it still makes sense :)

include/core/utilities.h
```diff
@@ -984,39 +984,68 @@ void
...
+                     2 * numbers::PI);
 }
 
 /**
- * @brief Converts radius to point in cartesian coordinates (in the x-y plane)
+ * @brief Converts radius to point in cartesian coordinates based on rotation axis
  * @param[in] radius Radial distance
  * @param[in] angle_rad Angle (in radians)
+ * @param[in] center_of_rotation Center of rotation
+ * @param[in] rotation_axis Direction of rotation axis (0=x, 1=y, 2=z)
```
brunaccampos:
```suggestion
 * @param[in] rotation_axis_direction Direction of rotation axis (0=x, 1=y, 2=z)
```

include/core/utilities.h
```diff
@@ -984,39 +984,68 @@ void
...
- * @return point Point cartesian coordinates
+ * @return point Point in cartesian coordinates
  */
 template <int dim>
 inline Point<dim>
-radius_to_point(const double radius, const double angle_rad)
+radius_to_point(const double       radius,
+                const double       angle_rad,
+                const Point<dim>  &center_of_rotation = Point<dim>(),
+                const unsigned int rotation_axis      = 2)
```
brunaccampos:
```suggestion
                const unsigned int rotation_axis_direction      = 2)
```

source/core/mortar_coupling_manager.cc
```diff
@@ -941,16 +897,26 @@ template <int dim>
...
-                     pre_rotation_angle + 2 * numbers::PI,
-                   2 * numbers::PI);
+  if constexpr (dim == 3)
+    return std::fmod(point_to_angle(point,
+                                    this->center_of_rotation,
+                                    this->rotation_axis_direction) -
+                       pre_rotation_angle + 2 * numbers::PI,
+                     2 * numbers::PI);
+  else
+    return std::fmod(point_to_angle(point, this->center_of_rotation) -
```
brunaccampos: I think you don't need the `if` statement here, just call it as it would be for the 3d case. If it's 2D, the `point_to_angle` function will adjust itself

applications_tests/lethe-fluid/mortar_mms_gmsh.mpirun=3.output
```diff
@@ -21,6 +21,6 @@ Steady iteration:        3/3
    Number of degrees of freedom: 62595
    Volume of triangulation:      3.600000e+01
 cells  error_velocity    error_pressure   
-  320 5.132546e-01    - 2.262208e+00    - 
- 1280 7.585448e-02 2.76 5.601373e-01 2.01 
- 5120 1.015784e-02 2.90 1.445052e-01 1.95 
+  320 5.145033e-01    - 2.329421e+00    - 
+ 1280 7.733195e-02 2.73 5.796338e-01 2.01 
+ 5120 1.261081e-02 2.62 1.725119e-01 1.75
```
brunaccampos: We discussed that this test changed because of the `radius_to_point` angle that was not accounting for the center of rotation, which makes sense. The convergence rates deteriorate a bit though. 
I was checking the output files and, after fixing the center of rotation issue, there's some pressure peaks at the mortar interface that weren't there before. Turns out that increasing the penalty factor to 10 makes this pressure discontinuity improve and recover the right convergence

brunaccampos (changes requested):
A few other minor comments :) I think the penalty factor on the `mortar_mms_gmsh` example needs to be increased so that the right convergence rates are restored

applications_tests/lethe-fluid/mortar_mms_gmsh.mpirun=3.output
blaisb: But in this case the center of rotation is zero is it not? It is a bit odd that the results change like that... Worth discussing together I think.

include/core/utilities.h
blaisb: yeah, nonetheless I think this is fine right now. Prevents any sort of ambiguity. In 2D it can only be (2) the rotation axis. Otherwise we would need to assert for the axis and crash the code voluntarily if the axis is 2.

blaisb (commented):
I just have one file left

applications_tests/lethe-fluid/mortar_mms_gmsh.mpirun=3.output
brunaccampos: The center of rotation in this example is (1,1)

brunaccampos: But yes, not sure why the penalty factor had a role here to recover the convergence

blaisb (changes requested):
I think there's still a major issue. 

MortarManagerCircle::get_normal still assumes an origin for the circle, and this PR breaks that assumption without updating it.

 Previously from_1D returned R(cosθ, sinθ) about the origin, so point/point.norm() gave  (cosθ, sinθ) . The two origin assumptions cancelled out. Now that from_1D passes center_of_rotation into radius_to_point, the points come back as c + R·(cosθ, sinθ) but the normalization is still about the origin, so the normal is no longer radial whenever center_of_rotation != 0.

Get normal needs to follow this however. Something like this would work
```
 template <int dim>
Tensor<1, dim, double>
MortarManagerCircle<dim>::get_normal(const Point<dim> &point) const
{
 Tensor<1, dim, double> radial = point - this->center_of_rotation;
if constexpr (dim == 3)
  radial[this->rotation_axis_direction] = 0.0;

  return radial / radial.norm();
}
```

The axial zeroing is what makes this correct for an x or y axis, and it mirrors what the test suite's own reference implementation already does (tests/mortar/tests.h:2441 zeroes point[2] before normalizing). It's inert today because the dim == 3 path takes its normals from FEFaceValues (line 1046) rather than from get_normal, but leaving it out would bake the z-axis assumption right back into the class this PR is generalizing.


@brunaccampos  I think this is what the mortar_mms_gmsh.mpirun=3.output change is: that test is 2D with set center of rotation = 1, 1. A wrong normal breaks the SIPG consistency term while the jump-based penalty term still glues the solution together, which matches the symptom  order reduction rather than a blow-up. Could you re-run that test after fixing get_normal? I'd expect it back at ~3/~2 and the .output change to drop out of the diff entirely.

source/core/mortar_coupling_manager.cc
```diff
@@ -696,7 +630,8 @@ compute_interface_dimensions_circular(
                           pre_rotation_min = std::min(
                             pre_rotation_min,
                             point_to_angle(mortar_parameters.center_of_rotation,
-                                           v));
+                                           v,
+                                           rotation_axis_direction));
```
blaisb: I think you have switched the argument in the wrong order here In a previous call center_of_rotation is the second argument.

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,45 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
     {
-      // Direction of the rotation axis
-      int            direction = 0;
-      Tensor<1, dim> axis      = mortar_parameters.rotation_axis;
-      for (int d = 1; d < dim; ++d)
-        if (std::abs(axis[d]) > std::abs(axis[direction]))
-          direction = d;
+      // First check if the vector has a unit norm
+      bool is_unit_axis =
+        mortar_parameters.rotation_axis.norm() == 1 ? true : false;
```
blaisb: I would avoid using an equality here. maybe use something like > 0.999999999 or something similar to be more robust.

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,45 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
+
+      // Now check if the vector represents the x, y, or z directions
+      // specifically (we assume those are the only options for now)
+      for (int i = 0; i < dim; i++)
+        if (mortar_parameters.rotation_axis[i] != 0. &&
+            mortar_parameters.rotation_axis[i] != 1.)
+          is_unit_axis = false;
```
blaisb, line 754: What happens if somebody specifiies a negative axis? (it would still be viable). I would instead assert that the absolute value is equal to 1.

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,45 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
         ExcMessage(
-          "Rotation axis is not aligned with a coordinate direction."));
+          "The rotation axis must be a unit vector in x, y, or z direction."));
+
+      // Find the direction of the rotation axis
+      for (int d = 0; d < dim; d++)
+        if (mortar_parameters.rotation_axis[d] != 0.0)
+          return d;
+    }
+  return 0;
```
blaisb: Wait
If you are in 2D (which is the else case here) you should be return 2 instead of 0 right. A rotation in 2D is done along the z axis

include/core/mortar_coupling_manager.h
blaisb, line 388: The rotation_axis_direction is missing in the param here.

source/core/mortar_coupling_manager.cc
blaisb, line 929: This is what needs to be modified to take into account the center of the mortar @jatred

source/core/mortar_coupling_manager.cc
jatred: Oh yeah, I default to returning 0 but it should totally be 2 for consistency

release_notes/current/2026_07_16_Tremblay.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/07/16
+
+### Added
+
+- MINOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. The code is still limited to an x, y or z rotation axis passing through the origin. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
Julamipolymtl: It seems to me that the code no longer limits to the center of rotation to pass by the center of origin. On that point we could have a test added/modified to check that behavior.

include/core/mortar_coupling_manager.h
```diff
@@ -217,6 +221,8 @@ class MortarManagerBase
   const unsigned int n_quadrature_points;
   /// Rotation angle for the inner domain
   const double rotation_angle;
+  /// Direction of the rotation axis
```
Julamipolymtl: We could have the convenion explicitely defined in the doc
```suggestion
  /// Direction of the rotation axis (x=0, y=1, z=2)
```

include/core/utilities.h
```diff
@@ -982,39 +982,69 @@ void
 delete_vtu_and_pvd_files(const std::string &output_path);
 
 /**
- * @brief Converts point to angle (in radians)
- * @param[in] point Point in cartesian coordinates (x-y plane)
+ * @brief Converts point to angle (in radians) based on rotation axis
+ * @param[in] point Point in cartesian coordinates
  * @param[in] center_of_rotation Center of rotation
+ * @param[in] rotation_axis_direction Direction of rotation axis (0=x, 1=y, 2=z)
```
Julamipolymtl: Uniformity with `mortar_coupling_manager.h`
```suggestion
 * @param[in] rotation_axis_direction Direction of rotation axis (x=0, y=1, z=2)
```

include/core/utilities.h
```diff
@@ -982,39 +982,69 @@ void
...
+                     2 * numbers::PI);
 }
 
 /**
- * @brief Converts radius to point in cartesian coordinates (in the x-y plane)
+ * @brief Converts radius to point in cartesian coordinates based on rotation axis
  * @param[in] radius Radial distance
  * @param[in] angle_rad Angle (in radians)
+ * @param[in] center_of_rotation Center of rotation
+ * @param[in] rotation_axis_direction Direction of rotation axis (0=x, 1=y, 2=z)
```
Julamipolymtl: Same as above
```suggestion
 * @param[in] rotation_axis_direction Direction of rotation axis (x=0, y=1, z=2)
```

source/core/mortar_coupling_manager.cc
```diff
@@ -1092,15 +1058,18 @@ CouplingOperator<dim, Number>::CouplingOperator(
...
                 std::vector<Point<dim - 1>> quad;
 
-                for (const auto p : points_ref)
+                for (const auto &p : points_ref)
                   {
                     Point<dim - 1> temp;
-                    for (int i = 0, j = 0; i < dim; ++i)
-                      if ((face_no / 2) != static_cast<unsigned int>(i))
-                        temp[j++] = p[i];
+                    // Axis perpendicular to the face current in the reference
```
Julamipolymtl:
```suggestion
                    // Axis perpendicular to the current face in the reference
```

source/core/mortar_coupling_manager.cc
```diff
@@ -1092,15 +1058,18 @@ CouplingOperator<dim, Number>::CouplingOperator(
...
+                    // cell
+                    const unsigned int axis = face_no / 2;
 
-                    if ((dim == 3) && ((face_no / 2) == 1))
-                      std::swap(temp[0], temp[1]);
+                    // 'temp' stores the coordinates of the quadrature point
+                    // on the face, hence the loop over the remaining 2 axes of
+                    // the reference cell
```
Julamipolymtl, line 1072:
```suggestion
                    // 'temp' stores the coordinates of the quadrature point
                    // on the face, hence the loop over the remaining dim-1 axes of
                    // the reference cell
```

tests/mortar/tests.h
```diff
@@ -2411,11 +2412,13 @@ MyMortarManagerCircle<dim>::MyMortarManagerCircle(
   const std::vector<double>       &radius,
   const Quadrature<dim2>          &quadrature,
   const double                     rotation_angle,
+  const unsigned int               rotation_axis,
   const std::vector<double>       &stage_heights)
   : MortarManagerBase<dim>(n_subdivisions,
                            radius,
                            quadrature,
                            rotation_angle,
+                           rotation_axis,
```
Julamipolymtl: It seems like while MyMortarManagerCircle is generalised to take any axis, its member functions overriden below disregard the rotation axis definition entirely. There are no tests making use of this right now, but when it does it may be a issue.

release_notes/current/2026_07_16_Tremblay.md
Julamipolymtl: Though checking back into this, it does limit it to be **axis aligned**.

Julamipolymtl (commented):
Great work on the generalisation of the rotation axis! A few things here and there, but what I think is lacking right now that you touched on in your PR note is one or more test to check that it does indeed work. Either modifiying an existing one or adding one with an x-aligned axis off origin.

blaisb:
Yeah in my opinion all that is missing is a quick test here. That would seal the deal. Like a taylor-couette where the rotation axis is X for example.

source/core/mortar_coupling_manager.cc
jatred: This loop can only be reached if dim=3, so generalizing the 2 to dim-1 is true but might be confusing

tests/mortar/tests.h
jatred: Good point, I'll change them now so it doesn't cause any problem later on.

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,46 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
 }
 
 template <int dim>
-std::vector<double>
-compute_stage_heights(const Triangulation<dim>      &triangulation,
-                      const Parameters::Mortar<dim> &mortar_parameters)
+unsigned int
+get_rotation_axis_direction(const Parameters::Mortar<dim> &mortar_parameters)
 {
   if constexpr (dim == 3)
```
blaisb, line 742: I think the fact that we accept negative values right now is a bit special. like right now if you put a negative value here,  you will get a "positive value outcome" whereas someone would expect rotation in the reverse direction. I would assert that the axis needs to have positive components here for the moment.

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,46 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
         ExcMessage(
-          "Rotation axis is not aligned with a coordinate direction."));
+          "The rotation axis must be a unit vector in x, y, or z direction."));
 
+      // Find the direction of the rotation axis
+      for (int d = 0; d < dim; d++)
+        if (mortar_parameters.rotation_axis[d] != 0.0)
+          return d;
+    }
+  return 2;
```
blaisb, line 768: This is a bit tricky right, because we never use the value afterwards (this 2) if i t's in 2D because it's unused. I would at least add a comment here regarding the fact that this value of 2 is not used later in the code when dim==2.

include/core/mortar_coupling_manager.h
```diff
@@ -42,7 +43,8 @@ class MortarManagerBase
                     double                  radius,
                     const Quadrature<dim2> &quadrature,
                     const double            rotation_angle,
-                    const double            stage_heights = 0.0);
+                    const unsigned int      rotation_axis_direction = 2,
+                    const double            stage_heights           = 0.0);
```
blaisb, line 47:
```suggestion
                    const unsigned int      rotation_axis_direction ,
                    const double            stage_heights         );
```
I would drop the default arguements here because otherwise there's a big ambiguity between this function and the other one (in my opinion).

release_notes/current/2026_07_17_Tremblay.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/07/17
+
+### Added
+
+- MINOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. Also makes it so that the rotation axis doesn't need to pass through the origin. The code is still limited to a rotation axis parallel to the main x, y or z axis. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
blaisb:
```suggestion
- MAJOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. Also makes it so that the rotation axis doesn't need to pass through the origin. The code is still limited to a rotation axis parallel to the main x, y or z axis. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```

You also fixed a bug for 2D when the center was not in (0,0). I would describe that. IMO this is a major change.

include/core/mortar_coupling_manager.h
jatred: If I drop them I would need to go modify pretty much all the 2D mortar tests, since they call directly the MortarManagerBase. I think it looks a bit strange to have to add 2 superfluous arguments in all the 2D tests. 

I am not sure I understand where the ambiguity comes from since these 2 functions have always had the same arguments, but with different types. Also, "stage_heights" should disappear in a future PR and be combined into "radius", so it will be removed from both at the same time.

applications_tests/lethe-fluid/mortar_taylorcouette3d_x_rotation.output
```diff
@@ -0,0 +1,10 @@
+Running on 1 MPI rank(s)...
+   Number of active cells:       64
+   Number of degrees of freedom: 960
+   Volume of triangulation:      5.740251e+00
+
+*****************************
+Steady iteration:        1/1
+*****************************
+cells error_velocity error_pressure
```
brunaccampos, line 9: could you do at least a couple of refinements to show the convergence here?

tests/mortar/manager_05.cc
```diff
@@ -17,6 +17,7 @@ main()
   const std::vector<unsigned int> n_subdivisions      = {10, 3};
   const unsigned int              n_quadrature_points = 3;
   const std::vector<double>       radius              = {1.2, 0.4};
+  const unsigned int              rotation_axis       = 2;
```
brunaccampos:
```suggestion
  const unsigned int              rotation_axis_direction       = 2;
```

tests/mortar/manager_05.cc
```diff
@@ -32,6 +33,7 @@ main()
                                                radius,
                                                QGauss<dim>(n_quadrature_points),
                                                rotate,
+                                               rotation_axis,
```
brunaccampos:
```suggestion
                                               rotation_axis_direction,
```

release_notes/current/2026_07_22_Tremblay.md
```diff
@@ -0,0 +1,9 @@
+## [Master] - 2026/07/22
+
+### Added
+
+- MAJOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. Also makes it so that the rotation axis doesn't need to pass through the origin. The code is still limited to a rotation axis parallel to the main x, y or z axis. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
brunaccampos:
```suggestion
- MAJOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. The code is still limited to a rotation axis parallel to the main x, y or z axis. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
We were prescribing the center of rotation before, so we could already have a problem centered outside the origin

brunaccampos (approved):
Last comments :)

applications_tests/lethe-fluid/mortar_taylorcouette3d_x_rotation.output
jatred: Sadly no, I can't refine a deal.ii mesh after rotating it since the deal.ii manifolds don't get rotated (wish I knew sooner). I could change it to use a gmsh file instead if you feel it's needed to have more than 1 iteration.

release_notes/current/2026_07_22_Tremblay.md
jatred: True, I moved it to "Fixed" but forgot to remove it from there

source/core/mortar_coupling_manager.cc
```diff
@@ -801,24 +736,48 @@ construct_quadrature(const Quadrature<dim>         &quadrature,
...
+          "The rotation axis must be a unit vector in the positive x, y, or z direction."));
 
+      // Find the direction of the rotation axis
+      for (int d = 0; d < dim; d++)
+        if (mortar_parameters.rotation_axis[d] != 0.0)
+          return d;
+    }
+  // In 2D, the rotation_axis_direction is not used, but we return 2 to indicate
+  // that the rotation axis is perpendicular to the plane
+  return 2;
```
Julamipolymtl, line 768: I would maybe like to have this in an else statement for readability so the 2D behavior is a clear logic choice and not just fall-through.

applications_tests/lethe-fluid/mortar_taylorcouette3d_x_rotation.prm
```diff
@@ -0,0 +1,174 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Checks the mortar implementation with a x-aligned rotation axis for a Taylor-Couette example.
```
Julamipolymtl:
```suggestion
# Checks the mortar implementation with an x-aligned rotation axis for a Taylor-Couette example.
```

release_notes/current/2026_07_22_Tremblay.md
```diff
@@ -0,0 +1,9 @@
+## [Master] - 2026/07/22
+
+### Added
+
+- MAJOR Implements the ability to use a different rotation axis than the z-axis when using the mortar method. The code is still limited to a rotation axis parallel to the main x, y or z axis. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
+
+### Fixed
+
+- MINOR Fixes a bug when calculating the normal vectors at the mortar interface when the rotation axis is not passing through the origin. The code would still assume the center of rotation to be at (0,0), which would affect the results. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```
Julamipolymtl:
```suggestion
- MINOR Fixes a bug when calculating the normal vectors at the mortar interface when the rotation axis does not pass through the origin. The previous code would assume the center of rotation to be at (0,0) regardless, which would affect the results. [#2016](https://github.com/chaos-polymtl/lethe/pull/2016)
```

Julamipolymtl: So does this PR fix non-origin centered axes (in both 2D and 3D)? If so I don't think we test for that yet in 3D.

tests/mortar/tests.h
```diff
@@ -2411,36 +2417,48 @@ MyMortarManagerCircle<dim>::MyMortarManagerCircle(
...
-
-  if (dim == 3)
-    point[2] = 0.0;
+  Tensor<1, dim, double> radial = point_in - center_of_rotation;
+  if constexpr (dim == 3)
+    radial[this->rotation_axis_direction] = 0.0;
 
-  return point / point.norm();
+  return radial / radial.norm();
 }
```
Julamipolymtl, line 2464: I get the need for the `MyMortarManagerCircle` class for tests, but with the recent changes these overrides seem to have the same definition from the base `MortarManagerCircle` class. Is there any way we could easily fix this code duplication? I feel it is one more thing to track for if ever we make future changes to any of these 3 methods in `MortarManagerCircle` class. 

This feels outside the scope of this PR though, just wanted to have your opinion on this.

Julamipolymtl (approved):
Looks good to me! Just a few comments.

release_notes/current/2026_07_22_Tremblay.md
jatred: It does, and yes we could modify an existing 3D test by shifting the rotation axis, we only test the feature in 2D right now.

tests/mortar/tests.h
jatred: Yeah it could be useful to check the mortar tests, since a lot of new features have been added since this structure was created. I might take a look at this file in the next PR to clean it up a bit.

blaisb (approved):

PR #2033, opened 2026-06-16
Packed insertion method

OGaboriault:
### Description
Inserting particles can be finicky and can take a lot of time depending on the situation. This PR introduce an experimental insertion method called the "packed" method. This method work as follows: 

Particles are randomly generated inside an insertion box without guarantying that there is no overlap between them. Particle that are overlapping with each other are shifted in the direction of their normal overlap up until their is no longer overlap between the particles and between the particles and the walls. 

So far, the method works. I can provide an prm to test it. There is still a lot of work/modification to be done to merge this pull request IMO. For example, I don't think that the added function in the dem.cc file should go be there, but I need help/suggestion on how to proceed.  

Code related list:

Pull request related list:
- [x] Labels are applied


include/dem/particle_wall_contact_force.h
```diff
@@ -69,6 +69,29 @@ class ParticleWallContactForceBase
     const double dt,
     const std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solids,
     ParticleInteractionOutcomes<PropertiesIndex> &contact_outcome) = 0;
+
+
+  /**
+   * @brief Return the number of contact that occurred in the
```
karlwilson:
```suggestion
   * @brief Return the number of contacts that occurred in the
```

include/dem/particle_wall_contact_force.h
```diff
@@ -69,6 +69,29 @@ class ParticleWallContactForceBase
...
+   * present pseudo-time step when using the shift insertion method.
+   */
+  unsigned int
+  get_number_of_contacts() const
+  {
+    return n_contact;
```
karlwilson:
```suggestion
    return n_contacts;
```

include/dem/particle_wall_contact_force.h
```diff
@@ -69,6 +69,29 @@ class ParticleWallContactForceBase
...
+  }
+
+  /**
+   * @brief Reset the counter of number of contact for the next pseudo-time step.
```
karlwilson:
```suggestion
   * @brief Reset the counter of number of contacts for the next pseudo-time step.
```

include/dem/particle_wall_contact_force.h
```diff
@@ -69,6 +69,29 @@ class ParticleWallContactForceBase
...
+   */
+  void
+  reset_number_of_contacts()
+  {
+    n_contact = 0;
```
karlwilson:
```suggestion
    n_contacts = 0;
```

include/dem/particle_wall_contact_force.h
```diff
@@ -1021,6 +1049,49 @@ class ParticleWallContactForce
...
+
+  inline void
+  shift_particle_using_normal_overlap(
+    particle_wall_contact_info<dim> &contact_info,
+    const double                     normal_overlap)
+  {
+    // Contact particle-wall + constant cohesive force.
+    if (normal_overlap > 0.)
+      {
+        ++this->n_contact;
```
karlwilson:
```suggestion
        ++this->n_contacts;
```

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1538,88 @@ class ParticleParticleContactForce
...
+  inline void
+  shift_particle_using_normal_overlap(
+    particle_particle_contact_info<dim> &contact_info,
+    const Tensor<1, 3>                  &normal_unit_vector,
+    const double                         normal_overlap,
+    const ArrayView<const double>       &particle_one_properties,
+    const ArrayView<const double>       &particle_two_properties)
+  {
+    ++this->n_contact;
+
```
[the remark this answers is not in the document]
OGaboriault: This is wrong, the normal overlap check is perform before using this method. Thus the normal_overlap is always positive here.

karlwilson (commented):
Partial review, will carry on when I find time.

karlwilson (commented):
Good work on this! I agree about moving the new functions from `dem.cc` to somewhere else. Why not in the `InsertionPacked` class?

Some tests should also be added.

OGaboriault:
> Good work on this! I agree about moving the new functions from `dem.cc` to somewhere else. Why not in the `InsertionPacked` class?
> 
> Some tests should also be added.

The thing is that the ``DEMSolver`` class stores the ``insertion_object`` as a shared pointer, thus I cannot use functions that are not declared as virtual void in the base ``Insertion`` class.

include/core/parameters_lagrangian.h
```diff
@@ -41,7 +41,10 @@ namespace Parameters
       /// Hertz model with Johnson-Kendall-Roberts cohesive force
       hertz_JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift the particles using their overlap. Used for the packed
+      /// insertion method
```
blaisb:
```suggestion
      /// Shift collision model. This model shift the particles using their overlap. It is only used for the packed
      /// insertion method and does not constitute a real collision model.
```

include/core/parameters_lagrangian.h
```diff
@@ -57,7 +60,10 @@ namespace Parameters
       /// Johnson-Kendall-Roberts model
       JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift the particle in contact with a wall using their overlap. Used
+      /// for the packed insertion method
```
blaisb: Same as above.

include/dem/dem.h
```diff
@@ -57,6 +57,12 @@ class DEMSolver
   void
   solve();
 
+  void
+  update_previous_position();
+
+  void
+  clamp_displacement();
```
blaisb: Could you add the brief to these three functions :)

include/dem/dem.h
```diff
@@ -524,6 +530,11 @@ class DEMSolver
    * with a packed bed, loaded with another prm.
    */
   bool disable_position_integration;
+
+  /**
+   * @brief Uses the packed insertion method.
+   */
+  bool is_packed_insertion_method = false;
```
blaisb: I'm not a fan of putting constructor values in the .h. Let's put them in the constructor (the initial value).

blaisb (changes requested):
Just begun. will tonight yo.

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * This method iterates through all active cells in the triangulation,
+   * identifying locally owned cells. For each particle in these cells, it
+   * retrieves the current particle position and stores it into the particle's
+   * properties. The position is stored within certain indices that are
+   * conventionally designated as velocity components (v_x, v_y, and v_z for 3D
+   * configurations). This allows the tracking of previous particle positions
+   * without additional data structures. This tracking is required for the
+   * packed insertion method.
+   */
```
blaisb: Can you describe the two arguments of the method

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,140 @@
...
+   * This method iterates over all particles and checks their displacement
+   * norm relative to their previous positions. If the displacement exceeds
+   * the maximum particle diameter, their positions are clamped such that
+   * the displacement does not exceed this threshold. The displacement
+   * values are updated accordingly for each particle.
+   */
+  static void
+  clamp_displacement(Particles::ParticleHandler<dim> &particle_handler,
+                     const double                     max_disp,
+                     std::vector<double>             &displacement);
```
blaisb: Can you add the param for the three arguments of the method.

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * @brief Generate an insertion point inside an insertion box.
+   *
+   * @param insertion_location Generated insertion location.
+   * @param rng Random number generator.
+   * @param insertion_information Insertion parameters read from the .prm file.
+   */
+  void
+  generate_insertion_location(
+
```
blaisb:
```suggestion
```

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,140 @@
...
+    Point<dim>               &insertion_location,
+    std::mt19937             &rng,
+    const InsertionInfo<dim> &insertion_information);
+
+  unsigned int current_inserting_particle_type = 0;
```
blaisb:
```suggestion
/// Current particle type being inserted
  unsigned int current_inserting_particle_type = 0;
```

include/dem/particle_particle_contact_force.h
```diff
@@ -100,7 +100,29 @@ class ParticleParticleContactForceBase
     this->combined_periodic_offsets = offsets;
   }
 
+  /**
+   * @brief Return the number of contact that occurred in the
+   * present pseudo-time step when using the shift insertion method.
+   */
+  unsigned int
```
blaisb:
```suggestion
  inline unsigned int
```

blaisb: Add @return please

include/dem/particle_particle_contact_force.h
```diff
@@ -345,7 +371,8 @@ class ParticleParticleContactForce
                     Tensor<1, 3>                  &tangential_force,
                     Tensor<1, 3> &particle_one_tangential_torque,
                     Tensor<1, 3> &particle_two_tangential_torque,
-                    Tensor<1, 3> &rolling_resistance_torque)
+                    Tensor<1, 3> &rolling_resistance_torque,
+                    Particles::ParticleIterator<dim> particle_one)
```
blaisb: I understand why you are adding the parameter. But please put particle_one BEFORE contact info instead of after since there's no default argument here. It will make more sense logically (particle one, contact info, etc.)
It feels a bit weird to have to refactor like that, but so be it.

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1543,89 @@ class ParticleParticleContactForce
...
+   * @brief
+   *
+   * @param[in,out] contact_info A container that contains the required
+   * information for calculation of the contact force for a particle pair in
+   * contact.
+   * @param normal_unit_vector
+   * @param normal_overlap
+   * @param particle_one_properties
+   * @param particle_two_properties
+   * @param particle_one
```
blaisb: Are those parameters in or in out or just out? Please add the description of them.

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1543,89 @@ class ParticleParticleContactForce
...
+    if constexpr (dim == 3)
+      {
+        particle_one_new_position =
+          particle_one_position -
+          (0.501 * normal_overlap + 0.01 * diameter_one) * normal_unit_vector;
+
+        particle_two_new_position =
+          particle_two_position +
+          (0.501 * normal_overlap + 0.01 * diameter_two) * normal_unit_vector;
+      }
```
blaisb, line 1604: Why do you have a displacement factor parameter in if you actually directly displace using the diameter?

Does this rule always work?

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
 .. warning::
     The critical Rayleigh time step is computed from the parameters in the ``particle type`` subsections, not the ``insertion info`` subsection. It is the user's responsibility to fill the ``particle type`` subsections correctly according to the diameter values stored in the insertion input file, otherwise Rayleigh time percentage displayed at the start of every DEM simulation may not be accurate.
 
+--------------------
+Packed
+--------------------
+The ``packed`` insertion method insert particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions as is for the initial configuration of a DEM simulation would result in a unstable simulation. To solve this problem, overlapping particles are shift apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particles are not solved.
```
blaisb:
```suggestion
The ``packed`` insertion method insert particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions as is for the initial configuration of a DEM simulation would result in a unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particles are not solved.
```

include/dem/particle_wall_contact_force.h
```diff
@@ -1053,6 +1081,49 @@ class ParticleWallContactForce
     normal_force += cohesive_term * normal_vector;
   }
 
+  /**
+   * @brief Shift particles that are in contact using their normal overlap. This
+   * contact model is used for the packed insertion method.
+   *
+   * @param[in,out] contact_info A container that contains the required
+   * information for calculation of the contact force for a particle-wall pair.
+   * @param[in] normal_overlap Contact normal overlap.
```
blaisb: I should shift the parameters order to put the [in,out] as the last parameters. We try (progressively) to have the in only parameters as first input and the out as the lasts. At least when we can :)

include/dem/set_insertion_method.h
```diff
@@ -55,8 +58,59 @@ set_insertion_type(std::vector<std::shared_ptr<Distribution>>
           triangulation,
           dem_parameters,
           maximum_particle_diameter);
+      case InsertionInfo<dim>::InsertionMethod::packed:
+        {
+          if constexpr (std::is_same_v<PropertiesIndex,
+                                       DEM::DEMProperties::PropertiesIndex>)
+            {
+              packing_method = true;
```
blaisb: I would set packing_method to false in the other cases no? Because right now it's a bit undefined what happens.

source/dem/insertion_packed.cc
```diff
@@ -0,0 +1,279 @@
...
+        previous_position[2] = particle_properties[PropertiesIndex::v_z];
+
+      const Tensor<1, dim> displacement_tensor =
+        particle.get_location() - previous_position;
+
+      const double disp_norm = displacement_tensor.norm();
+
+      // No movement
+      if (std::isnan(disp_norm))
+        continue;
```
blaisb: Can you explain a bit more. You use isnan as an indicator of no-displacement or it's just that a nan can appear sometimes and you want to avoid that?
Seems a bit tricky to rely on NANS no?

source/dem/visualization.cc
```diff
@@ -94,63 +94,52 @@ Visualization<dim, PropertiesIndex>::print_xyz(
...
-  // Mapping of all particles & find the max id on current processor
   for (auto particle = particle_handler.begin();
        particle != particle_handler.end();
        ++particle)
     {
-      current_id     = particle->get_id();
-      current_id_max = std::max(current_id, current_id_max);
+      const unsigned int id                  = particle->get_id();
+      auto               particle_properties = particle->get_properties();
+      auto               particle_location   = particle->get_location();
```
blaisb:
```suggestion
      const auto               &particle_properties = particle->get_properties();
      const auto               &particle_location   = particle->get_location();
```
No need to copy these, they can be const auto reference.

source/dem/visualization.cc
```diff
@@ -94,63 +94,52 @@ Visualization<dim, PropertiesIndex>::print_xyz(
...
-  // Aggressively force synchronization of the header line
-  usleep(500);
-  MPI_Barrier(mpi_communicator);
-  usleep(500);
-  MPI_Barrier(mpi_communicator);
 
-  std::map<int, Particles::ParticleIterator<dim>> global_particles;
-  unsigned int current_id, current_id_max = 0;
+  // Build this rank's particles as id -> formatted line
+  std::map<unsigned int, std::string> local_lines;
```
blaisb: I am not sure I understand that changes. Can you explain?

source/dem/visualization.cc
```diff
@@ -94,63 +94,52 @@ Visualization<dim, PropertiesIndex>::print_xyz(
...
 
-      global_particles.insert({current_id, particle});
+      local_lines[id] = oss.str();
     }
 
-  // Find global max particle index
-  unsigned int id_max = Utilities::MPI::max(current_id_max, mpi_communicator);
+  // Gather every rank's map to rank 0 (single collective, no polling loop)
+  const auto gathered =
+    Utilities::MPI::gather(mpi_communicator, local_lines, 0);
```
blaisb: So you are gathering all particles information on processor 0 before printing them. This is ok since this is only used for test purposed, but nonetheless I would explain a bit more in the code why you are doing this.

blaisb (changes requested):
Nothing too big, but some things to document a bit more.
Great work.

include/dem/particle_particle_contact_force.h
```diff
@@ -1873,7 +1980,8 @@ class ParticleParticleContactForce
                                         tangential_force,
                                         particle_two_tangential_torque,
                                         particle_one_tangential_torque,
-                                        rolling_resistance_torque);
+                                        rolling_resistance_torque,
+                                        particle_one);
```
Julamipolymtl: My good friend Claude flagged a possible issue with the `ghost_local_periodic` shift. `particle_one`/`particle_two` are (correctly) swapped when calling `update_contact_information`/`calculate_contact` so the local particle goes in first (so the normal points local $\to$ ghost). But `shift_particle_using_normal_overlap` assumes `particle_one` is the tail of the normal. Here the `particle_one` iterator wasn't swapped along with the rest, so the shift is applied with the ghost as `particle_one` (the head) and would actually bring particles together rather than appart.

source/dem/insertion_packed.cc
```diff
@@ -0,0 +1,279 @@
...
+          // Pull a fresh 0.0 to 1.0 random value
+          double random_value = distribution(rng);
+
+          // Map it uniformly across the specific axis span
+          insertion_location[axis] =
+            this->axis_min[axis] +
+            random_value * (this->axis_max[axis] - this->axis_min[axis]);
+        }
+      // Reject the point if the value is negative.
+      accepted = acceptance_fct->value(insertion_location) > 0.;
```
Julamipolymtl, line 271: If the `acceptance_fct` is defined negative on the whole insertion domain (user error), this while loop would silently hang. We could have a max amount of tries and a warning so it does not happen.

doc/source/parameters/dem/insertion_info.rst
Julamipolymtl:
```suggestion
The ``packed`` insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions as is for the initial configuration of a DEM simulation would result in an unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particle are not solved.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
...
+
+.. note::
+    Currently, when using this insertion method, the simulation ends once no contacts is detected, thus the particle positions needs to be extracted in an insertion file and the simulation must be relaunched with the ``file`` insertion method.
```
Julamipolymtl:
```suggestion
    Currently, when using this insertion method, the simulation ends once no contact is detected, thus the particle positions need to be extracted in an insertion file and the simulation must be relaunched with the ``file`` insertion method.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
...
+
+.. note::
+    Due to the use of pseudo time-stepping and the functioning of this insertion method, if multiple particle types are used, every particles types will be inserted on the first insertion time-step.
```
Julamipolymtl:
```suggestion
    Due to the use of pseudo time-stepping and the functioning of this insertion method, if multiple particle types are used, every particle type will be inserted on the first insertion time-step.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
...
+
+.. warning::
+    The ``packed`` insertion method is experimental. It does not guarantee that the particles will reach a non-overlapping configuration is a reasonable amount of time, especially when total inserted particle volume over the total triangulation volume is over 50%.
```
Julamipolymtl:
```suggestion
    The ``packed`` insertion method is experimental. It does not guarantee that the particles will reach a non-overlapping configuration in a reasonable amount of time, especially when total inserted particle volume over the total triangulation volume is over 50%.
```

release_notes/current/2026_07_08_Gaboriault.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/07/08
+
+### Added
+
+- MAJOR This PR adds a new insertion method called the ``packed`` method. This insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Overlapping particles are then shifted apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particle is not solved. This method should be useful when dealing with a large number of particles in a tight triangulation, which forces the use of multiple insertion boxes when using the ``volume`` insertion method. [#2033](https://github.com/chaos-polymtl/lethe/pull/2033)
```
Julamipolymtl:
```suggestion
- MAJOR This PR adds a new insertion method called the ``packed`` method. This insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Overlapping particles are then shifted apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particle are not solved. This method should be useful when dealing with a large number of particles in a tight triangulation, which forces the use of multiple insertion boxes when using the ``volume`` insertion method. [#2033](https://github.com/chaos-polymtl/lethe/pull/2033)
```

include/core/parameters_lagrangian.h
```diff
@@ -43,7 +43,11 @@ namespace Parameters
       /// Hertz model with Johnson-Kendall-Roberts cohesive force
       hertz_JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift collision mode. This model shift the particles using their
```
Julamipolymtl:
```suggestion
      /// Shift collision mode. This model shifts the particles using their
```

include/core/parameters_lagrangian.h
```diff
@@ -59,7 +63,11 @@ namespace Parameters
       /// Johnson-Kendall-Roberts model
       JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift collision mode. This model shift the particles using their
```
Julamipolymtl:
```suggestion
      /// Shift collision mode. This model shifts the particles using their
```

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,140 @@
...
+  // Number of particles of each type that remain to be inserted in the
+  // upcoming insertion steps
+  unsigned int particles_of_each_type_remaining;
+
+  // Minimum and maximum boundaries of the insertion box in the direction order
+  // It means that axis 0 is not necessarily x, since it depends on the order
+  // of the insertion direction.
+  std::vector<double> axis_min, axis_max;
+
+  /// Function used to insert particle inside our outside an arbitrary shape.
```
Julamipolymtl:
```suggestion
  /// Function used to insert particles inside or outside an arbitrary shape.
```

Julamipolymtl (commented):
Very nice insertion method. There is a possible bug when handling periodic ghost local with this insertion method that brings particles together rather than appart.

include/dem/particle_particle_contact_force.h
blaisb: Hmm yeah something is indeed odd there.

include/core/parameters_lagrangian.h
```diff
@@ -495,6 +505,9 @@ namespace Parameters
       /// using the volume insertion.
       std::shared_ptr<Function<dim>> insertion_acceptance_fct;
 
+      /// Displacement factor used for the packed insertion method.
+      double displacement_factor;
+
```
OGaboriault: This need to be remove.

include/dem/particle_particle_contact_force.h
OGaboriault: This is for periodic contact. Not gonna lie, I did not think about using this feature with periodic BC...

include/dem/set_insertion_method.h
OGaboriault: it used to be set at false in the .h by default in the past, that is why.

source/dem/insertion_packed.cc
OGaboriault: The nan is not required anymore.

source/dem/insertion_packed.cc
OGaboriault: I feel that this is dangerous. It would "silently hang" but you would clearly see in the output that the simulation is stuck at inserting particles. 

On the other hand, if we put a max amount of tries, I feel that some simulation would throw when nothing is wrong with the parameter file. It would be possible to insert inside a really weird looking geometry (like an off axis elongated geometry). If you want to insert inside this entire weird geometry, you would use a big insertion box and use an acceptance function that is positive inside that geometry and a lot of value would be rejected...

What I mean is that it is hard to have a default value that is adequate for every case scenario and I feel that it is better to have a code that is buggy due to user error than a code that is buggy due to a protection in the code. (Personally, I prefer to hate myself than a protection in the code.)

source/dem/visualization.cc
OGaboriault: I was having test that were falling because the particle weren't being printed in order... the value were still the same, juste not in the same order.

include/dem/particle_particle_contact_force.h
Julamipolymtl: I mean fair enough. From what you said this insertion method is pretty nice and I don't see why its use couldn't be extended to periodic BCs. Though no need to have it working right away in this PR, maybe for now it could simply throw or we could have an open issue about it so we know of this pulling behavior when used with periodic BCs.

source/dem/insertion_packed.cc
Julamipolymtl: Fair point about hating yourself rather than the code :) Maybe something less invasive would be a simple output warning, (no max cap per say) saying "It takes more than 1e12 tries to fit your particles in the domain, are you sure it's properly set up?" And then you get to hate yourself with a little more details on the why!

But I'm no DEM user so maybe simply getting stuck on the insertion step is in itself a dead giveaway of were the issue is in the prm file. This is very nitpicky so all good if you leave it as is!

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1546,93 @@ class ParticleParticleContactForce
...
+      }
+    if constexpr (dim == 3)
+      {
+        particle_one_new_position =
+          particle_one_position -
+          (0.501 * normal_overlap + 0.01 * diameter_one) * normal_unit_vector;
+
+        particle_two_new_position =
+          particle_two_position +
+          (0.501 * normal_overlap + 0.01 * diameter_two) * normal_unit_vector;
```
Julamipolymtl: Is there a reason why 2D and 3D don't displace the particles by the same factor of the particle's diameter (2D: 1e-3 * $D_p$ , 3D: 1e-2 * $D_p$ )

include/dem/particle_wall_contact_force.h
```diff
@@ -1053,6 +1081,48 @@ class ParticleWallContactForce
...
+        auto particle = contact_info.particle;
+
+        Point<dim> particle_position = particle->get_location();
+        Point<dim> particle_new_position;
+
+        if constexpr (dim == 2)
+          {
+            particle_new_position =
+              particle_position -
+              1.25 * normal_overlap * tensor_nd_to_2d(normal_vector);
```
Julamipolymtl: Where particle-particle contact displace with regards to overlap AND diameter, here its only displaced according to the overlap. Plus the normal overlap factor (1.25) is much meaner than the one from p-p contact (2*0.501=1.02). What is the reason for this different treatment with wall contacts?

include/dem/insertion_packed.h
```diff
@@ -0,0 +1,148 @@
...
+  generate_insertion_location(Point<dim>               &insertion_location,
+                              std::mt19937             &rng,
+                              const InsertionInfo<dim> &insertion_information);
+
+  /// Current particle type being inserted
+  unsigned int current_inserting_particle_type = 0;
+
+  // Number of particles of each type that remain to be inserted in the
+  // upcoming insertion steps
+  unsigned int particles_of_each_type_remaining;
```
Julamipolymtl: Both of these seem to come from `InsertionVolume`/`InsertionPlane` where we ahve multi-step insertion. From my understanding this method inserts all particles at once, therefore these two members are not needed.

applications_tests/lethe-particles/insertion_packed.prm
```diff
@@ -0,0 +1,101 @@
...
+
+#---------------------------------------------------
+# Insertion Info
+#---------------------------------------------------
+
+subsection insertion info
+  set insertion method                               = packed
+  set inserted number of particles at each time step = 200
+  set insertion frequency                            = 2000000
+  set insertion box points coordinates               = -0.02, -0.02, -0.02 : 0.02, 0.02, -0.02
```
Julamipolymtl: Is there a reason the insertion "box" is effectively a plane (z_min = z_max = -0.02)

Julamipolymtl: I checked and it actually makes it so all particles are on rank 0, so nothing cross-rank related is actually tested here.

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1546,93 @@ class ParticleParticleContactForce
...
+  inline void
+  shift_particle_using_normal_overlap(
+    Particles::ParticleIterator<dim>     particle_one,
+    particle_particle_contact_info<dim> &contact_info,
+    const Tensor<1, 3>                  &normal_unit_vector,
+    const double                         normal_overlap,
+    const ArrayView<const double>       &particle_one_properties,
+    const ArrayView<const double>       &particle_two_properties)
+  {
+    ++this->n_contacts;
```
Julamipolymtl: It looks as though cross-rank contacts can be counted twice in the `n_contacts`. The end condition `n_contacts == 0` remains valid even with this double counting, but its only really important if you are interested in the real evolution of the number of contacts.

source/dem/insertion_packed.cc
```diff
@@ -0,0 +1,284 @@
...
+    {
+      auto particle_properties = particle.get_properties();
+
+      const Point<dim> particle_previous_position = particle.get_location();
+
+      particle_properties[PropertiesIndex::v_x] = particle_previous_position[0];
+      particle_properties[PropertiesIndex::v_y] = particle_previous_position[1];
+      if constexpr (dim == 3)
+        particle_properties[PropertiesIndex::v_z] =
+          particle_previous_position[2];
```
Julamipolymtl, line 198: This is a weirdly funny hack to preserve the previous position, maybe worth a comment to document the velocity property is used here.

Julamipolymtl (commented):
Few things to comment, I'll have a final look a bit later.

include/dem/particle_particle_contact_force.h
blaisb: Yeah I would either make it work (it's a very simple change that Justin is proposing) or at the very least assertthrow. I would suggest the former.

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
 .. warning::
     The critical Rayleigh time step is computed from the parameters in the ``particle type`` subsections, not the ``insertion info`` subsection. It is the user's responsibility to fill the ``particle type`` subsections correctly according to the diameter values stored in the insertion input file, otherwise Rayleigh time percentage displayed at the start of every DEM simulation may not be accurate.
 
+--------------------
+Packed
+--------------------
+The ``packed`` insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions as is for the initial configuration of a DEM simulation would result in an unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching each other or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion on every particle are not solved.
```
blaisb:
```suggestion
The ``packed`` insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions **as is** for the initial configuration of a DEM simulation would result in an unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching other particles or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motions are not solved.
```

applications_tests/lethe-particles/insertion_packed.prm
OGaboriault: z_min = z_max = -0.02 is indeed a mistake, but the fact that all the particles are in the subdomain of rank 0 is not a problem in itself. 

Every rank generates the same number of particle locations and properties, The insert_global_particles function makes sure that the right rank receive the right particle.

include/dem/set_insertion_method.h
```diff
@@ -32,7 +34,8 @@ set_insertion_type(std::vector<std::shared_ptr<Distribution>>
                      &size_distribution_object_container,
                    parallel::distributed::Triangulation<dim> &triangulation,
                    DEMSolverParameters<dim>                  &dem_parameters,
-                   const double &maximum_particle_diameter)
+                   const double &maximum_particle_diameter,
+                   bool         &packing_method)
```
blaisb: I am not sure i understand why this needs to be passed as an argument and returned. You can always assess if it's a packing insertion method by doing  a check if it's equal to InsertionInfo<dim>::InsertionMethod::packed elsewhere in the code. If that's not the case, can you document why this parameter is needed.

include/dem/particle_particle_contact_force.h
blaisb: Yeah that's weird. I would make that consistent.

include/dem/particle_particle_contact_force.h
```diff
@@ -1860,8 +1974,9 @@ class ParticleParticleContactForce
                                                  particle_one_location,
                                                  dt);
 
-                // Calculation the contact force
-                this->calculate_contact(contact_info,
+                // Calculation of the contact force
+                this->calculate_contact(particle_one,
```
blaisb: Should be particle_two here. Particle one is a ghost.

source/dem/dem.cc
```diff
@@ -198,13 +199,31 @@ DEMSolver<dim, PropertiesIndex>::setup_functions_and_pointers()
...
+        ParticleWallContactForceModel::shift;
+
+      if (parameters.model_parameters.sparse_particle_contacts)
+        {
+          sparse_contacts_object =
+            AdaptiveSparseContacts<dim, PropertiesIndex>();
+          DEMActionManager::get_action_manager()
+            ->set_sparse_contacts_disabled();
+        }
+    }
```
blaisb: This is ok (to override everything). Nonetheless, it would be really worth it to print a message. Something like " You are using the packed insertion method, this method overloads the regular contact model. Please consult the documentation for information on how to appropriately use it".

Something like that. Just so that there is some information that is user-facing.

blaisb (changes requested):
Some further changes. Could you add a test with walls also?

include/dem/particle_particle_contact_force.h
blaisb: True, but I am absolutely not sure on how to remedy this to be honest (except counting particle-ghost and particle-particle contacts) differently. I think the best way to approach this here is to document that the number of contacts is given as an approximation.

source/dem/insertion_packed.cc
```diff
@@ -0,0 +1,284 @@
...
+  // We only insert once with this method. Thus, if the particle handler already
+  // possesses particles, we should not insert again.
+  if (particle_handler.n_global_particles() != 0)
+    return;
+
+
+  // Message passing interface requirements.
+  MPI_Comm           communicator = triangulation.get_mpi_communicator();
+  ConditionalOStream pcout(std::cout,
+                           Utilities::MPI::this_mpi_process(MPI_COMM_WORLD) ==
```
blaisb:
```suggestion
                           Utilities::MPI::this_mpi_process(communicator) ==
```
Instead of using the raw MPI_COMM_WORLD

include/dem/insertion_packed.h
blaisb: Yeah I could not find any usage of these parameters anywhere.

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1546,93 @@ class ParticleParticleContactForce
   }
 
   /**
-   * @brief Return the index for accessing the properties vectors for a given
-   * combinations of particle types.
+   * @brief Shift the particles using the normal overlap and the normal unit
+   * vector. Pparticles in contact are shifted in opposite directions to resolve
+   * the overlap. The shift is proportional to the normal overlap and the
+   * particle diameters. This \"contact force\" model is use for the packed
```
blaisb:
```suggestion
   * particle diameters. This \"contact force\" model is used for the packed
```

include/dem/set_insertion_method.h
OGaboriault: I could do std::is_same_v<insertion_object,InsertionPacked> instead of storing the bool packing_method in the DEMSolver class.

include/dem/particle_particle_contact_force.h
Julamipolymtl: Fair point! So long as the actual value holds no importance and all that matters is that `n_contacts` goes down this is good for me.

source/dem/insertion_packed.cc
OGaboriault: it is documented in the function brief.

blaisb (approved):

blaisb:
@Julamipolymtl I'll wait for your OK and I will merge.

include/dem/particle_wall_contact_force.h
```diff
@@ -1053,6 +1081,52 @@ class ParticleWallContactForce
...
+               0.01 * particle->get_properties()[PropertiesIndex::dp]) *
+                tensor_nd_to_2d(normal_vector);
+          }
+        if constexpr (dim == 3)
+          {
+            particle_new_position =
+              (1.02 * normal_overlap +
+               0.01 * particle->get_properties()[PropertiesIndex::dp]) *
+              normal_vector;
+          }
```
Julamipolymtl, line 1125: Looks like in 3D the new particle position does not take into account the old particle position anymore.
```suggestion
            particle_new_position =
              particle_position -
              (1.02 * normal_overlap +
               0.01 * particle->get_properties()[PropertiesIndex::dp]) *
                tensor_nd_to_2d(normal_vector);
          }
        if constexpr (dim == 3)
          {
            particle_new_position =
              particle_position -
              (1.02 * normal_overlap +
               0.01 * particle->get_properties()[PropertiesIndex::dp]) *
              normal_vector;
          }
```

include/dem/set_insertion_method.h
blaisb: Yeah it would be better than to carry an bool around.

include/dem/particle_particle_contact_force.h
```diff
@@ -1505,21 +1546,94 @@ class ParticleParticleContactForce
   }
 
   /**
-   * @brief Return the index for accessing the properties vectors for a given
-   * combinations of particle types.
+   * @brief Shift the particles using the normal overlap and the normal unit
+   * vector. Pparticles in contact are shifted in opposite directions to resolve
+   * the overlap. The shift is proportional to the normal overlap and the
+   * particle diameters. This \"contact force\" model is used for the packed
```
Julamipolymtl:
```suggestion
   * particle diameters. This "contact force" model is used for the packed
```

include/dem/particle_wall_contact_force.h
```diff
@@ -1053,6 +1081,52 @@ class ParticleWallContactForce
...
 
+  /**
+   * @brief Shift particles that are in contact using their normal overlap. This
+   * contact model is used for the packed insertion method.
+   *
+   * @param[in] normal_overlap Contact normal overlap.
+   * @param[in,out] contact_info A container that contains the required
+   * information for calculation of the contact force for a particle-wall pair.
+   */
+
```
Julamipolymtl:
```suggestion
```

include/dem/particle_particle_contact_force.h
```diff
@@ -332,8 +359,11 @@ class ParticleParticleContactForce
    * particle two.
    * @param[out] rolling_resistance_torque Contact rolling resistance torque.
    */
+
+
```
Julamipolymtl:
```suggestion
```

include/core/parameters_lagrangian.h
```diff
@@ -43,7 +43,11 @@ namespace Parameters
       /// Hertz model with Johnson-Kendall-Roberts cohesive force
       hertz_JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift collision mode. This model shifts the particles using their
```
Julamipolymtl:
```suggestion
      /// Shift collision model. This model shifts the particles using their
```

include/core/parameters_lagrangian.h
```diff
@@ -59,7 +63,11 @@ namespace Parameters
       /// Johnson-Kendall-Roberts model
       JKR,
       /// Derjaguin-Muller-Toporov model
-      DMT
+      DMT,
+      /// Shift collision mode. This model shifts the particles using their
```
Julamipolymtl:
```suggestion
      /// Shift collision model. This model shifts the particles using their
```

include/dem/particle_particle_contact_force.h
```diff
@@ -100,7 +100,31 @@ class ParticleParticleContactForceBase
...
+   * @brief Reset the counter of number of contact for the next pseudo-time step.
+   */
+  void
+  reset_number_of_contacts()
+  {
+    n_contacts = 0;
+  }
+
 protected:
+  // Counter for the amount of contact
```
Julamipolymtl:
```suggestion
  // Counter for the number of contacts
```

source/dem/dem.cc
```diff
@@ -198,14 +198,37 @@ DEMSolver<dim, PropertiesIndex>::setup_functions_and_pointers()
...
                                              triangulation,
                                              parameters,
                                              maximum_particle_diameter);
 
+  if (parameters.insertion_info.insertion_method ==
+      InsertionInfo<dim>::InsertionMethod::packed)
+    {
+      pcout << "\nUsing packed insertion method. This method overloads the \n"
+               "regular contact model. Please consult the documentation for \n"
+               "information on how to appropriately use it.\n"
```
Julamipolymtl:
```suggestion
      pcout << "\nUsing packed insertion method. This method overrides the\n"
               "regular contact model. Please consult the documentation for\n"
               "information on how to appropriately use it.\n"
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
 .. warning::
     The critical Rayleigh time step is computed from the parameters in the ``particle type`` subsections, not the ``insertion info`` subsection. It is the user's responsibility to fill the ``particle type`` subsections correctly according to the diameter values stored in the insertion input file, otherwise Rayleigh time percentage displayed at the start of every DEM simulation may not be accurate.
 
+--------------------
+Packed
+--------------------
+The ``packed`` insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions **as is** for the initial configuration of a DEM simulation would result in an unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching other particles or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motions are not solved.
```
Julamipolymtl:
```suggestion
The ``packed`` insertion method inserts particles in an insertion box in an unstructured manner without guaranteeing that the particles are not overlapping on insertion. Using these particle positions **as is** for the initial configuration of a DEM simulation would result in an unstable simulation. To solve this problem, overlapping particles are shifted apart, relative to their normal contact vector, until all particles are no longer touching other particles or walls. This particle shifting is performed using pseudo time-stepping, where gravity is not considered and where the equations of motion are not solved.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -193,6 +195,20 @@ Each line is associated with a particle and its properties. The main advantage o
...
+
+.. note::
+    Currently, when using this insertion method, the simulation ends once no contact is detected, thus the particle positions need to be extracted in an insertion file and the simulation must be relaunched with the ``file`` insertion method.
+
+.. note::
+    Due to the use of pseudo time-stepping and the functioning of this insertion method, if multiple particle types are used, every particle type will be inserted on the first insertion time-step.
+
+.. warning::
+    The ``packed`` insertion method is experimental. It does not guarantee that the particles will reach a non-overlapping configuration in a reasonable amount of time, especially when total inserted particle volume over the total triangulation volume is over 50%.
```
Julamipolymtl:
```suggestion
    The ``packed`` insertion method is experimental. It does not guarantee that the particles will reach a non-overlapping configuration in a reasonable amount of time, especially when the total inserted particle volume over the total triangulation volume is over 50%.
```

include/dem/particle_particle_contact_force.h
```diff
@@ -100,7 +100,31 @@ class ParticleParticleContactForceBase
     this->combined_periodic_offsets = offsets;
   }
 
+  /**
+   * @brief Return the amount of contact that occurred in the
+   * present pseudo-time step when using the shift insertion method.
+   *
+   * @return The amount of contact that occurred in the present pseudo-time
+   * step.
+   */
```
Julamipolymtl, line 109:
```suggestion
  /**
   * @brief Return the number of contacts that occurred in the
   * present pseudo-time step when using the packed insertion method.
   *
   * @return The number of contacts that occurred in the present pseudo-time
   * step.
   */
```

include/dem/particle_wall_contact_force.h
```diff
@@ -70,6 +70,29 @@ class ParticleWallContactForceBase
     const double dt,
     const std::vector<std::shared_ptr<SerialSolid<dim - 1, dim>>> &solids,
     ParticleInteractionOutcomes<PropertiesIndex> &contact_outcome) = 0;
+
+
+  /**
+   * @brief Return the number of contacts that occurred in the
+   * present pseudo-time step when using the shift insertion method.
+   */
```
Julamipolymtl, line 78:
```suggestion
  /**
   * @brief Return the number of contacts that occurred in the
   * present pseudo-time step when using the packed insertion method.
   */
```

include/dem/particle_particle_contact_force.h
```diff
@@ -100,7 +100,31 @@ class ParticleParticleContactForceBase
...
+  unsigned int inline get_number_of_contacts() const
+  {
+    return n_contacts;
+  }
+
+  /**
+   * @brief Reset the counter of number of contact for the next pseudo-time step.
+   */
```
Julamipolymtl, line 117:
```suggestion
  /**
   * @brief Reset the counter of number of contacts for the next pseudo-time step.
   */
```

Julamipolymtl (approved):
Looks good to me! Just vocab/syntax I found in there

include/dem/particle_wall_contact_force.h
OGaboriault: lol what? How did this work?

Julamipolymtl: I think maybe the test has no particle-wall contact?

Issue #2014, opened 2026-06-05
DEM - Particle size distribution - Volume weighted PDF


OGaboriault:
The particle size distribution is not respected when using a interpolated volume weighted PDF.

blaisb:
This has been closed by  #2040

PR #2040, opened 2026-06-19
Dem fix custom distribution

OGaboriault:
### Description

Particle size sampling using the Volume based PDF with interpolation had a bug where the PDF wasn't actually interpolated during sampling. It was first converted into a CDF which was then sampled. This resulted  in a sampling that wasn't respecting the input PDF. This PR fixes this issue. 

This PR fixes issue #2014 

### Testing

Maybe I need to do a better test? The thing is  we did not want our unit test to print the diameter values of every particle created. 

Here are two examples. The first two graphs we produced by defining a volume based PDF with interpolation. We see at the top that the PDF is peace wise linear, which is logical since we used interpolation. particle_size_distribution

The second image was produced by defining a volme based CDF with interpolation. We see that the CDF is peace wise linear, which results in a stair case function for the PDF. 
particle_size_distribution

This means that it is preferable to use the PDF in our simulations since it results smoother curve (except when your PDF is actually a stair-case)  

Here is the same simulation run on master with a volume based PDF with interpolation.
particle_size_distribution
We see that the interpolation on the PDF was not respected. 

### Documentation

Nothing changed here

Code related list:

Pull request related list:
- [x] Labels are applied


source/dem/distributions.cc
```diff
@@ -477,27 +477,31 @@ CustomDistribution::CustomDistribution(
...
 
               // We normalize the new CDF.
               for (unsigned int i = 0; i < number_d_values; ++i)
                 number_based_cdf[i] /= number_based_cdf.back();
             }
-          else
+          else // No interpolation
             {
+              // d_probabilities represent the fractions of every
+              // diameter value over the total number. (n_d/N_tot)
```
antonellasuccar:
```suggestion
              // d_probabilities represents the fractions of every
              // diameter value over the total number. (n_d/N_tot)
```

source/dem/distributions.cc
```diff
@@ -512,82 +516,124 @@ CustomDistribution::CustomDistribution(
...
-              // relative particle count.
-              dfn[i] = this_bin_volume_fraction * weight;
-
-              // Sum the relative counts. This sum will be used as the
-              // denominator to normalize the final number-based CDF so it
-              // ends at 1.0.
-              number_fraction_sum += dfn[i];
-            }
+              // Normalize the input volume PDF so that ∫ fv dd = 1
+              // We use the trapezoidal method to do the integral.
```
antonellasuccar, line 533: Can't we impose that the input pdf be normalized at the input?

source/dem/distributions.cc
```diff
@@ -512,82 +516,124 @@ CustomDistribution::CustomDistribution(
...
+              for (unsigned int i = 0; i < number_d_values; ++i)
+                fv[i] = d_probabilities[i] / fv_integral;
 
-          // Reconstruct the CDF from the Number fractions while making sure
-          // to normalize every value.
-          number_based_cdf[0] = 0.0;
-          for (unsigned int i = 1; i < number_d_values - 1; ++i)
-            number_based_cdf[i] =
-              number_based_cdf[i - 1] + (dfn[i - 1] / number_fraction_sum);
+              // Use the exact integral of fv(d)/d³ over each segment/bin
```
antonellasuccar: I'm curious to why we don't just use the trapezoidal rule here to integrate fv/d3

source/dem/distributions.cc
```diff
@@ -512,82 +516,124 @@ CustomDistribution::CustomDistribution(
...
+                  // becomes (d_high - d_low)(d_high + d_low).
+                  // 5. Weight = [1 / (d_high - d_low)] * Integral, thus the
+                  // (d_high - d_low) term is cancelled.
+                  //
+                  // Final Weight Formula: (d_low + d_high) / (2 * d_low^2 *
+                  // d_high^2)
+                  const double weight =
+                    (d_low + d_high) / (2.0 * std::pow(d_low * d_high, 2));
+
+                  // Convert th volume fraction to a relative number fraction
```
antonellasuccar:
```suggestion
                  // Convert the volume fraction to a relative number fraction
```

source/dem/distributions.cc
```diff
@@ -512,82 +516,124 @@ CustomDistribution::CustomDistribution(
...
+                  // Weight = [1 / (d_high - d_low)] * Integral from d_low to
+                  // d_high of (1/x^3) dx
+                  //
+                  // Solving the integral:
+                  // 1. Integral of x^-3 dx = -1/(2x^2).
+                  // 2. Evaluate the integral: [1/(2*d_low^2) - 1/(2*d_high^2)]
+                  // 3. Simplify the evaluated integral: (d_high^2 - d_low^2) /
+                  // [2
+                  // * d_low^2 * d_high^2 ]
+                  // 4. Using (a^2 - b^2) = (a-b)(a+b), (d_high^2 - d_low^2)
```
antonellasuccar:
```suggestion
                  // 3. Simplify the evaluated integral: 
                  // (d_high^2 - d_low^2) /[2* d_low^2 * d_high^2 ]
                  // 4. Using (a^2 - b^2) = (a-b)(a+b), (d_high^2 - d_low^2)
```

source/dem/distributions.cc
```diff
@@ -512,82 +516,124 @@ CustomDistribution::CustomDistribution(
...
+                  //
+                  // Final Weight Formula: (d_low + d_high) / (2 * d_low^2 *
+                  // d_high^2)
+                  const double weight =
+                    (d_low + d_high) / (2.0 * std::pow(d_low * d_high, 2));
+
+                  // Convert th volume fraction to a relative number fraction
+                  // (dfn). Note: This value is not yet normalized; it's a
+                  // relative particle count.
+                  dfn[i] = this_bin_volume_fraction * weight;
```
antonellasuccar, line 644: So we are assuming that the volume CDF is constant over the bin?

source/dem/distributions.cc
```diff
@@ -662,11 +705,68 @@ CustomDistribution::particle_size_sampling(
...
+            }
+          else
+            {
+              const double inv_d_low2  = 1.0 / (d_low * d_low);
+              const double inv_d_high2 = 1.0 / (d_high * d_high);
+
+              sampled_diameter =
+                1.0 /
+                std::sqrt(inv_d_low2 - u_local * (inv_d_low2 - inv_d_high2));
+            }
```
antonellasuccar, line 821: Could you explain this more?

antonellasuccar (commented):
Nice work :) A few comments but I haven't tested it yet. I also think it would be a good idea to document how the interpolation is done for a custom distribution. We can do that in a subsequent PR if it is too much work for this one.

source/dem/distributions.cc
blaisb: It's better to renormalize it anyway to be safe.

source/dem/distributions.cc
OGaboriault: Because there is a transformation between volume base to number based using the PDF, and if you don't do that, you won't get linear segment on the volume based PDF like I have shown in the PR description.

antonellasuccar: Yes, you're right. We are getting the exact integral assuming that the function is linear.

OGaboriault: Basically, this in a trapezoidal integration in the volume based form that we transform into the number base form since that is the form we need to sample the PSD. :)

source/dem/distributions.cc
OGaboriault: no, it is not... it is the same code as previously, it is just that github did some shenanigan so it appears as new code. 

The CDF is linearly interpolated between the input data points.

antonellasuccar: We are assuming that the volume-based PDF is constant over each bin, so the volume-based CDF is then linear. My bad. I would clarify this more in the comment on line 588.

include/dem/distributions.h
```diff
@@ -376,6 +376,22 @@ class CustomDistribution : public Distribution
    * @brief Random number generator for the diameter selection.
    */
   std::mt19937 gen;
+
+  /**
+   * @brief
+   */
```
blaisb, line 383: Missing the member description :)

include/dem/distributions.h
```diff
@@ -376,6 +376,22 @@ class CustomDistribution : public Distribution
...
+  // Normalized volume-based PDF values at each diameter node.
+  // Only populated when weighting_type == volume_based &&
+  // interpolate_diameter_values.
+  std::vector<double> fv;
+
+  // Normalization constant for the volume-based interpolated CDF
+  // (sum of all per-segment integrals of fv(d)/d³).
+  // Only populated when weighting_type == volume_based &&
+  // interpolate_diameter_values.
+  double total_integral = 0.0;
```
blaisb: I would avoid putting the initial value in the .h

source/dem/distributions.cc
```diff
@@ -480,27 +481,30 @@ CustomDistribution::CustomDistribution(
...
+          // If the input probability function is already given as a
+          // number-based CDF, nothing needs to be done.
           number_based_cdf = d_probabilities;
         }
       else if (function_type == ProbabilityFunctionType::PDF)
         {
           if (interpolate_diameter_values)
             {
               number_based_cdf[0] = 0.;
+              // Trapezoidal method to do the integral.
```
blaisb: Could you write the analytical formula for the integral you are doing, it would just be a bit easier to follow :)

source/dem/distributions.cc
```diff
@@ -480,27 +481,30 @@ CustomDistribution::CustomDistribution(
...
 
               // We normalize the new CDF.
               for (unsigned int i = 0; i < number_d_values; ++i)
                 number_based_cdf[i] /= number_based_cdf.back();
             }
-          else
+          else // No interpolation
             {
+              // d_probabilities representsba the fractions of every
+              // diameter value over the total number. (n_d/N_tot)
```
blaisb:
```suggestion
              // d_probabilities represents the fractions of every
              // diameter value over the total number. (n_d/N_tot)
```

source/dem/distributions.cc
```diff
@@ -515,82 +519,123 @@ CustomDistribution::CustomDistribution(
...
+              total_integral = 0.0;
 
-          // Reconstruct the CDF from the Number fractions while making sure
-          // to normalize every value.
-          number_based_cdf[0] = 0.0;
-          for (unsigned int i = 1; i < number_d_values - 1; ++i)
-            number_based_cdf[i] =
-              number_based_cdf[i - 1] + (dfn[i - 1] / number_fraction_sum);
+              for (unsigned int i = 0; i < number_d_values - 1; ++i)
+                {
```
blaisb, line 559: Could you add some comments to describe this part? :)

source/dem/distributions.cc
```diff
@@ -665,12 +706,78 @@ CustomDistribution::particle_size_sampling(
...
-          const double inv_d_low2  = 1.0 / (d_low * d_low);
-          const double inv_d_high2 = 1.0 / (d_high * d_high);
+          double sampled_diameter;
 
-          const double sampled_diameter =
-            1.0 / std::sqrt(inv_d_low2 - u_local * (inv_d_low2 - inv_d_high2));
+          if (weighting_type == DistributionWeightingType::volume_based &&
+              function_type == ProbabilityFunctionType::PDF)
+            {
+              // The volume-based PDF is assumed piecewise linear:
```
blaisb:
```suggestion
              // The volume-based PDF is assumed to be piecewise linear:
```

source/dem/distributions.cc
```diff
@@ -665,12 +706,78 @@ CustomDistribution::particle_size_sampling(
...
+                sampled_diameter = root2;
+              else
+                {
+                  // Fallback: linear interpolation on the CDF.
+                  // Triggers only in degenerate/edge cases (e.g. A ≈ 0
+                  // with no valid analytic root, or floating point
+                  // edge effects at bin boundaries).
+
+                  // This should never happen.
+                  sampled_diameter = d_low + u_local * (d_high - d_low);
```
blaisb: Should we print a warning then? I think it would be a good idea.

blaisb (approved):
Super minor stuff. Good work. I will wait for the final ok from @antonellasuccar

antonellasuccar (approved):

blaisb (approved):

PR #2041, opened 2026-06-20
Dem fix distribution container

OGaboriault:
### Description

A std::vector<double> was kept in the distribution class, which could take a lot of ram when the number of particle inserted in a single time-step was large big. This is was it wasn't notice before. 

I also took the time to correct typos in the code and stuff.

### Solution

Store the particles in a std::vector that is passed by reference to the assign_properties function.

### Testing

Tests are still good. 

### Documentation

Nothing

Code related list:

Pull request related list:
- [x] Labels are applied
- [] There are at least 2 reviewers (or 1 if small feature)

### DIFF
--- a/include/dem/distributions.h
+++ b/include/dem/distributions.h
@@ -396,8 +396,8 @@ using namespace Parameters::Lagrangian;
 inline void
 setup_distributions(const LagrangianPhysicalProperties &lpp,
                     std::vector<std::shared_ptr<Distribution>>
-                                &size_distribution_object_container,
-                    double      &maximum_particle_diameter,
+                                      &size_distribution_object_container,
+                    double            &maximum_particle_diameter,
                     const unsigned int mpi_process_id,
                     const ConditionalOStream &pcout)
 {
--- a/include/dem/distributions.h
+++ b/include/dem/distributions.h
@@ -38,11 +38,9 @@ public:
    *
    * @param[in] number_of_particles Number of particles inserted at a given
    * insertion time step.
-   * @param[out] particle_sizes Generated particle diameter values.
    */
-  virtual void
-  particle_size_sampling(const unsigned int  &number_of_particles,
-                         std::vector<double> &particle_sizes) = 0;
+  virtual std::vector<double>
+  particle_size_sampling(const unsigned int &number_of_particles) = 0;
 
   /**
    * @brief Return the minimum diameter for a certain distribution.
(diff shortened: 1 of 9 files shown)


include/dem/distributions.h
```diff
@@ -37,10 +35,14 @@ class Distribution
...
-   * @param number_of_particles Number of particle inserted at a given insertion time step.
+   *
+   * @param[in] number_of_particles Number of particles inserted at a given
+   * insertion time step.
+   * @param[out] particle_sizes Generated particle diameter values.
    */
   virtual void
-  particle_size_sampling(const unsigned int &number_of_particles) = 0;
+  particle_size_sampling(const unsigned int  &number_of_particles,
+                         std::vector<double> &particle_sizes) = 0;
```
blaisb: Does the vector already have the right size? If so you don't need to pass the number of particles, it would be the size of the vectors.
If not, then it's better to return the vector instead of passing a non-initialized vector and filling it in the function.

include/dem/distributions.h
```diff
@@ -95,14 +97,16 @@ class NormalDistribution : public Distribution
...
    *
-   * @param[in] number_of_particles Number of particle inserted at a given
+   * @param[in] number_of_particles Number of particles inserted at a given
    * insertion time step.
+   * @param[out] particle_sizes Generated particle diameter values.
    */
   void
-  particle_size_sampling(const unsigned int &number_of_particles) override;
+  particle_size_sampling(const unsigned int  &number_of_particles,
+                         std::vector<double> &particle_sizes) override;
```
blaisb: same as above.

include/dem/insertion.h
```diff
@@ -191,9 +191,6 @@ class Insertion
   // Maximum particle diameter
   double maximum_diameter;
 
-  // Inserted number of particles at this step on this processor
-  unsigned int inserted_this_step_this_proc;
-
```
blaisb, line 196: Why is this parameter removed?

source/dem/distributions.cc
```diff
@@ -167,10 +167,11 @@ NormalDistribution::NormalDistribution(
 
 void
 NormalDistribution::particle_size_sampling(
-  const unsigned int &number_of_particles)
+  const unsigned int  &number_of_particles,
+  std::vector<double> &particle_sizes)
```
blaisb: I like this design, but instead of passing a reference to a vector and clearing it, just return the vector. In Modern C++ this will just trigger a move instead of a copy, so it won't take more memory, but it will be a lot less ambiguous about what is being done about the vector.

blaisb (changes requested):
Nice changes. One main thing though, return the vector instead of passing a reference to it and modifying it. In modern C++, this will trigger a move instead of a copy so there will be zero memory overhead, however it will be a lot less ambiguous

include/dem/insertion.h
OGaboriault: Because it is only used by the volume insertion, thus It should not be in the attributes of every insertion class. 

Also, this value is reset at every insertion time step, thus it does not need to be stored as an attribute. This is why the insertion_volume.cc file is modified in the PR.

blaisb (approved):

PR #2045, opened 2026-06-26
Dem volume insertion based on function criteria

OGaboriault:
### Description

The volume insertion in Lethe can only insert particle align in a rectangular shape. This limitation can be annoying when dealing with a cylindrical triangulation, since the insertion box corners cannot be outside of the cylinder at the risk of losing particles by inserting them directly outside of the simulation domaine. 

This pull request fixes this issue by adding a new feature, called the ``insertion acceptance function``. The user can define an insertion box as big as he wants. The insert point that do not respect the condition defined by the ``insertion acceptance function`` will not be used as an insertion point.

The condition function doesn't need to be a cylinder, it can represent any shape that can be defined by a function, like a sphere or the union of many geometry. 

### Note 

To keep the volume insertion parallel even with this feature on, I have to create a std::map<unsigned int, unsigned int> to store the filtered id associated to the ID inside the insertion box.  

### Testing

A new application test is added. In this test, we define an insertion box that is as big as the simulation domain. We then define a ``insertion acceptance function`` that represent a small cylinder that occupies a small region of the simulation domain. Only the insertion point that are inside this cylinder are used to insert particles. 

### Documentation

I need to do the documenation, (monday). Should not take long. 

Code related list:

Pull request related list:
- [x] Labels are applied


applications_tests/lethe-particles/insertion_acceptance_function.mpirun=2.output
```diff
@@ -0,0 +1,721 @@
+
+*********************
+Running on 2 rank(s)
+*********************
```
blaisb, line 4: it's going to be a bit tricky to track atest with 1k particles. Can you make it with less so that a human can parse the output. You can just increase the spacing between the particles.

applications_tests/lethe-particles/insertion_acceptance_function.prm
```diff
@@ -0,0 +1,107 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Test description: We insert particle using the volume insertion with the an ``insertion acceptance function``.
+# Particles that are outside of the small cylinder are not inserted into the triangulation.
```
blaisb:
```suggestion
# Test description: We insert particle using the volume insertion with the an ``insertion acceptance function``.
# This function corresponds to an cylinder with a radius of 0.08 and a height of 0.08. Particles can only be inserted at location for which the acceptance function is negative. Particles whose location leads to a positive value of the acceptance function are not inserted.
```

Be as descriptive as possible when describing the test. This always helps people understanding its purpose.

source/core/parameters_lagrangian.cc
```diff
@@ -664,6 +664,18 @@ namespace Parameters
                           Patterns::Double(),
                           "Distance threshold");
 
+
+        // By default, every point is accepted, thus we need a default value
+        // of 1. We need to declare the parameters manually since the default
+        // value of a function is 0.
```
blaisb: Ok so 1 is inserted or is it > 0 inserted? If so, what you need is a positive value.

blaisb (changes requested):
I am just starting the review. Will finish over the weekend. This will need a .rst documentation though!

source/dem/insertion_volume.cc
```diff
@@ -227,6 +228,142 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+      // Create the insertion point
+      find_insertion_location(
+        insertion_location, id, 0., 0., insertion_information);
+
+      if (acceptance_fct->value(insertion_location) > 0.)
+        {
+          filted_id_to_box_id.insert(std::make_pair(filtered_id_count, id));
+          ++filtered_id_count;
+        }
+    }
```
blaisb: So in essence only the particles without the perturbation are checked right? This will need to be documented clearly

source/core/parameters_lagrangian.cc
OGaboriault: Actually, bigger than 0 is accepted. I will adjust the comment

source/dem/insertion_volume.cc
OGaboriault: Exactly. This will be explain in the documentation. I will also add comments in the code. 

Otherwise, we need to check while creating the points if they respect the condition or not.

source/dem/insertion_volume.cc
```diff
@@ -227,6 +228,142 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+            axis_max[1] = insertion_information.insertion_box_point_2(1);
+            break;
+          case 2:
+            axis_min[2] = insertion_information.insertion_box_point_1(2);
+            axis_max[2] = insertion_information.insertion_box_point_2(2);
+            break;
+          default:
+            AssertThrow(false,
+                        ExcMessage("Insertion direction must be 0, 1 or 2"));
+        }
```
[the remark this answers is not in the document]
OGaboriault: In the ``declare_entry`` function associated to the ``insertion sequence`` parameter, I have added a min value of ``0`` .

applications_tests/lethe-particles/insertion_acceptance_function.mpirun=2.output
OGaboriault: 100 is enough?

OGaboriault:
I think I have fix the issue with the size of the container in parallel. Some application test results should change because of this sadly.

applications_tests/lethe-particles/insertion_acceptance_function.mpirun=2.output
blaisb: Yup

include/dem/insertion_volume.h
```diff
@@ -77,28 +77,79 @@ class InsertionVolume : public Insertion<dim, PropertiesIndex>
   }
 
 private:
+  /**
+   * @brief Carries out finding the maximum number of inserted particles based on the
+   * insertion box size. If the requested number of particles for insertion in
+   * each insertion step is larger than this maximum, it is limited to this
+   * value and a warning is printed.
```
blaisb:
```suggestion
   * @brief Find the maximum number of inserted particles based on the
   * insertion box size. If the requested number of particles for insertion in
   * each insertion step is larger than this maximum, it is limited to this
   * value and a warning is printed.
```

include/dem/insertion_volume.h
```diff
@@ -77,28 +77,79 @@ class InsertionVolume : public Insertion<dim, PropertiesIndex>
...
+
+  // Map used to transfer from the filtered to the pre-filted numbering of the
+  // insertion point inside the insertion box.
+  std::vector<unsigned int> filted_box_index;
+
+  unsigned int first_index_this_proc, last_index_this_proc;
+
+  unsigned int number_of_valid_insertion_point_global;
+
+  double maximum_diameter;
```
blaisb: Can you document these parameters

source/core/parameters_lagrangian.cc
```diff
@@ -664,6 +664,17 @@ namespace Parameters
...
 
+
+        // By default, every point is accepted, thus we need a default value
+        // bigger than 0 (we use 1 here).
+        auto insertion_acceptance_function_parsed =
+          std::make_shared<Functions::ParsedFunction<dim>>(1);
+        prm.enter_subsection("insertion acceptance function");
+        {
+          insertion_acceptance_function_parsed->declare_parameters(prm, 1, "1");
+        }
```
blaisb, line 675: Wait, above when describing the test case you said negative meant the particles were inserted. Positive is inserted right?

applications_tests/lethe-particles/insertion_acceptance_function.prm
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Test description: We insert particle using the volume insertion with the an ``insertion acceptance function``.
+# This function corresponds to an cylinder with a radius of 0.08 and a height of 0.08. Particles can only be inserted at location for which the acceptance function is negative. Particles whose location leads to a positive value of the acceptance function are not inserted.
```
blaisb:
```suggestion
# Test description: We insert particle using the volume insertion with the an ``insertion acceptance function``.
# This function corresponds to an cylinder with a radius of 0.08 and a height of 0.08. Particles can only be inserted at location for which the acceptance function is positive. Particles whose location leads to a positive value of the acceptance function are not inserted.
```

acceptance function should be negative or positive?

source/dem/insertion_volume.cc
```diff
@@ -177,54 +156,267 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+void
+InsertionVolume<dim, PropertiesIndex>::set_filtered_index(
+  const InsertionInfo<dim> &insertion_information)
+{
+  // If you don't insert particles in this simulation, we don't need to build
+  // the container.
+  if (insertion_information.inserted_this_step == 0)
+    return;
+
+  // Checking if the insertion directions are valid (no repetition)
```
blaisb:
```suggestion
  // Check if the insertion directions are valid (no repetition)
```

source/dem/insertion_volume.cc
```diff
@@ -177,54 +156,267 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+    {
+      axis_sum = insertion_information.direction_sequence[0] +
+                 insertion_information.direction_sequence[1] +
+                 insertion_information.direction_sequence[2];
+
+      AssertThrow(
+        axis_sum == 3,
+        ExcMessage(
+          "Invalid insertion directions: at least 2 directions are the same "));
+    }
```
blaisb, line 237: Should this not be done when the parameter is being parsed? It seems odd to do it here (although I agree it's good to do it)

source/dem/insertion_volume.cc
```diff
@@ -177,54 +156,267 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+            axis_max[0] = insertion_information.insertion_box_point_2(0);
+            break;
+          case 1:
+            axis_min[1] = insertion_information.insertion_box_point_1(1);
+            axis_max[1] = insertion_information.insertion_box_point_2(1);
+            break;
+          case 2:
+            axis_min[2] = insertion_information.insertion_box_point_1(2);
+            axis_max[2] = insertion_information.insertion_box_point_2(2);
+            break;
```
blaisb, line 273: Why a switch here?
You could just say axis_min[axis] = insertion_information.insetion_box_point_1[axis]

and assert that axis is between 0 and 2
That would be simpler to read. right now your switch essentially does nothing.

source/dem/insertion_volume.cc
```diff
@@ -177,54 +156,267 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+      find_insertion_location(
+        insertion_location, index, 0., 0., insertion_information);
+
+      // If the point respects the acceptance function, we insert the index in
+      // the vector.
+      if (acceptance_fct->value(insertion_location) > 0.)
+        filted_box_index.push_back(index);
+    }
+
+  // Each proces needs to know what is its first and last filtered indexes.
```
blaisb:
```suggestion
  // Each process needs to know what is its first and last filtered indexes.
```

blaisb (approved):
Few comments (minor). This is a really killer feature

include/dem/insertion_volume.h
OGaboriault: actually, this function does not calculate anything anymore...

source/core/parameters_lagrangian.cc
OGaboriault: The description of the test case was wrong. Good catch. 

I think it was because we were planning on using a sign distance function, which are defined as negative in the inside, but now we use a boolean function.

source/dem/insertion_volume.cc
```diff
@@ -24,11 +24,18 @@ InsertionVolume<dim, PropertiesIndex>::InsertionVolume(
...
+
+  // We need to fill the vector filted_box_index.
+  // To do this, we loop of every insertion point inside the box considering
+  // that the acceptance_fct accepts every point.
+  // For each point, we check if it respects the condition. If so, we insert
+  // the ID associated with this point inside the map.
+  set_filtered_index(dem_parameters.insertion_info);
 }
 
 // The main insertion function. Insert_global_function is used to insert
```
OGaboriault: I will remove this comment since the brief is used for that.

source/dem/insertion_volume.cc
OGaboriault: This was previously done in this class, but not in this function. I think we should keep it for now due to the size of this PR. But I agree, it should be in the parsing.

source/dem/insertion_volume.cc
OGaboriault: the switch is already present on the master branch, just in another function. I would keep it for now.

applications_tests/lethe-particles/insertion_acceptance_function.prm
```diff
@@ -0,0 +1,108 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Test description: We insert particle using the volume insertion with the an
+# ``insertion acceptance function``. This function corresponds to a cylinder with
+# a radius of 0.025 and a height of 0.05. Particles can only be inserted at
+# location for which the acceptance function is positive. Particles whose
+# location leads to a negative value of the acceptance function are not inserted.
```
blaisb:
```suggestion
# Test description: We insert particle using the volume insertion with an
# ``insertion acceptance function``. This function corresponds to a cylinder with
# a radius of 0.025 and a height of 0.05. Particles can only be inserted at
# location for which the acceptance function is positive. Particles whose
# location leads to a negative value of the acceptance function are not inserted.
```

doc/source/parameters/dem/insertion_info.rst
```diff
@@ -98,6 +101,8 @@ The ``volume`` insertion method uses an insertion box where particles will be in
 
     Generally, we recommend users to use a threshold in the range of 1.3-2.0, depending on the value of offset.
 
+* ``insertion acceptance function`` subsection is used to define a function used to accept or reject insertion points in the insertion box based on their position. This parameter allows to create an insertion box of a arbitrary shape, which is useful when inserting in curved geometries, like a cylinder. A insertion point is accepted if the function returns a value greater than ``0.`` relative to its coordinate, before applying the offset defined by the ``insertion maximum offset`` parameter. By default, every point is accepted.
```
blaisb:
```suggestion
* ``insertion acceptance function`` subsection is used to define a function used to accept or reject insertion points in the insertion box based on their positions. This parameter allows to create an insertion box of a arbitrary shape, which is useful when inserting in curved geometries, like a cylinder. A insertion point is accepted if the function returns a value greater than ``0.``, before applying the offset defined by the ``insertion maximum offset`` parameter. The default value of the function is ``1``  and leads to every point being accepted..
```

release_notes/current/2025_06_03_Gaboriault.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/05/29
+
+### Added
+
+- MINOR The ``insertion acceptance function`` feature has been added. When using the ``volume`` insertion, this feature allows to insert particle from an insertion box of an arbitrary shape by defining a function. Insertion points from the original insertion box are kept if the function returns a value greater then ``0.`` relative to its coordinates before applying the insertion offset. [#2045](https://github.com/chaos-polymtl/lethe/pull/2045)
```
blaisb:
```suggestion
- MAJOR The ``insertion acceptance function`` feature has been added. When using the ``volume`` insertion, this feature allows to insert particle from an insertion box of an arbitrary shape by defining a function. Insertion points from the original insertion box are kept if the function returns a value greater then ``0.`` relative to its coordinates before applying the insertion offset. [#2045](https://github.com/chaos-polymtl/lethe/pull/2045)
```

This is a very cool new feature. We will want to promote it. IMO this is a major new feature.

source/dem/insertion_volume.cc
blaisb: Maybe open an issue so we don,t forget about it, we can do it in a short follow-up PR.

source/dem/insertion_volume.cc
```diff
@@ -177,54 +155,267 @@ InsertionVolume<dim, PropertiesIndex>::find_insertion_location_volume(
...
+
+      AssertThrow(
+        axis_sum == 1,
+        ExcMessage("Invalid insertion directions: 2 directions are the same "));
+    }
+  if constexpr (dim == 3)
+    {
+      axis_sum = insertion_information.direction_sequence[0] +
+                 insertion_information.direction_sequence[1] +
+                 insertion_information.direction_sequence[2];
```
blaisb, line 231: That's not exactly a good test right, cause axis 1 1 1 would give you 3 still. I would check that the three axis are also not equal one to another

blaisb (approved):
Good to go, some very minor stuff

PR #2051, opened 2026-07-01
Add Chebychev smoother as alternative for ID/ASM

blaisb:
### Description

Deal.II now has an implementation of a Chebychev smoother. It seems to do a better job as a smoother than the inverse diagonal, while being less expensive than the additive schwarz method. For example, it has worked for all steady-state problems I have tested where inverse diagonal would fail, so this is an interesting avenue.

### Testing

Added one GCMG test that uses it as a smoother (`mms3d_fe1_gcmg`). This is a good change because that test right now was using a constant smoothing coefficient without eigenvalue estimations which IMO is not a good idea.

### Documentation

This is a new smoother that comes with even more parameters, so I have added the parameters and documented them.

Code related list:

Pull request related list:
- [x] Labels are applied


doc/source/parameters/cfd/linear_solver_control.rst
```diff
@@ -339,7 +344,9 @@ Different parameters for the main components of the two geometric multigrid algo
...
+
+* The ``chebyshev`` option is a Chebyshev-accelerated inverse-diagonal smoother. Instead of a fixed-parameter relaxation, it applies a single step of a Chebyshev polynomial of degree ``mg smoother chebyshev degree`` built on the inverse diagonal, which improves robustness on anisotropic/boundary-layer meshes at a comparable cost per V-cycle. The maximum eigenvalue is estimated internally (``mg smoother chebyshev eig cg n iterations`` iterations), so no relaxation parameter or manual eigenvalue estimation is needed. The polynomial acts on the spectral interval ``[lambda_max / smoothing_range, lambda_max]``, with ``mg smoother chebyshev smoothing range`` typically between 5 and 20. A degree of ``0`` recovers the damped inverse-diagonal (Jacobi) smoother. If the smoother diverges, the most likely cause is an underestimated maximum eigenvalue: increase ``mg smoother chebyshev eig cg n iterations`` before changing the degree.
```
brunaccampos: The last part of the explanation looks like it might fit better in a separate `..tip::` block

applications_tests/lethe-fluid-matrix-free/mms3d_fe1_gcmg.prm
```diff
@@ -118,11 +118,11 @@ subsection linear solver
     set verbosity         = quiet
 
     #MG parameters
-    set mg verbosity               = quiet
-    set mg smoother eig estimation = false
+    set mg verbosity                    = quiet
+    set mg smoother preconditioner type = chebyshev
+    set eig estimation verbosity        = quiet
```
brunaccampos, line 123: Would allowing some verbosity here make the test more robust?

brunaccampos (approved):
Nice addition! Just two very small comments.

doc/source/parameters/cfd/linear_solver_control.rst
blaisb: Sure thing, will do.

applications_tests/lethe-fluid-matrix-free/mms3d_fe1_gcmg.prm
blaisb: I thought about it, but what i liked is that the output of the test was not changed by the PR. Maybe in a follow-up PR when this new feature has been used more, we could change the test a bit :).

PR #2059, opened 2026-07-08
Impinging jet mesh

blaisb:
### Description

This PR introduces a new grid generator for an impinging jet mixer. This will be used in a new example I am currently working on just for fun.  The coarse grid generated is illustrated below in the Miscellaneous section. It uses manifolds on all curved surfaces so as you refine the topology is preserved. This is a relatively nice mesh and I tested it on heat transfer and fluid dynamics simulation at relatively high Re (like 2k) and it works well.

### Testing

A unit test for the grid has also been added just like for the others.

### Documentation

The new grid is fully documented in the mesh subsection. As the number of lethe mesh is growing, maybe this will need to become its own subsection in the future.

Disclaimer: A part of the code was made by AI (Claude Opus 4.8). I mostly verified the mesh and guided the agent and then reviewed and tested the code myself. The main code itself remains AI-written.

image

Code related list:

Pull request related list:
- [x] Labels are applied


release_notes/current/2026_07_08_Blais.md
```diff
@@ -0,0 +1,5 @@
...
... the `type = lethe` parameter). It builds a fully hexahedral, conforming mesh for an impinging-jet mixer: a vertical cylindrical mixing chamber closed at the top by a hemispherical dome, a conical reduction followed by a straight outlet pipe below the chamber, and two opposing horizontal inlet pipes entering the chamber side wall from `+x` and `-x`. Because a round pipe meeting the curved chamber wall is a saddle intersection that cannot be produced by merging library primitives, each inlet pipe is instead grown out of the refined chamber wall itself, so the junction stays watertight and conforming. The two inlets share a common axis and the mesh is datumed so that axis lies on the `z = 0` (impingement) plane. The `grid arguments` accepts either an empty string (default dimensions) or exactly eight colon-separated dimensions in meters, in the order `R_chamber : r_inlet : r_outlet : H_chamber : L_cone : L_outlet : L_inlet : z_inlet`; the discretisation is fixed internally and not exposed. Boundary IDs are 0 for the `+x` inlet, 1 for the `-x` inlet, 2 for the outlet, and 3 for all walls. This mesh will be featured in a future example. [#](https://github.com/chaos-polymtl/lethe/pull/)
```
Julamipolymtl:
```suggestion
- MAJOR A new `impinging_jet_mixer` mesh lethe grid generator was added (selectable through the `type = lethe` parameter). It builds a fully hexahedral, conforming mesh for an impinging-jet mixer: a vertical cylindrical mixing chamber closed at the top by a hemispherical dome, a conical reduction followed by a straight outlet pipe below the chamber, and two opposing horizontal inlet pipes entering the chamber side wall from `+x` and `-x`. Because a round pipe meeting the curved chamber wall is a saddle intersection that cannot be produced by merging library primitives, each inlet pipe is instead grown out of the refined chamber wall itself, so the junction stays watertight and conforming. The two inlets share a common axis and the mesh is datumed so that axis lies on the `z = 0` (impingement) plane. The `grid arguments` accepts either an empty string (default dimensions) or exactly eight colon-separated dimensions in meters, in the order `R_chamber : r_inlet : r_outlet : H_chamber : L_cone : L_outlet : L_inlet : z_inlet`; the discretisation is fixed internally and not exposed. Boundary IDs are 0 for the `+x` inlet, 1 for the `-x` inlet, 2 for the outlet, and 3 for all walls. This mesh will be featured in a future example. [#2059](https://github.com/chaos-polymtl/lethe/pull/2059)
```

source/core/grids.cc
```diff
@@ -195,6 +196,19 @@ attach_grid_to_triangulation(Triangulation<dim, spacedim> &triangulation,
...
+          AssertThrow(
+            !mesh_parameters.simplex,
+            ExcMessage(
+              "Unsupported mesh type - impinging jet mixer mesh with simplex is not supported"));
+
+          GridImpingingJetMixer<dim, spacedim> grid(
+            mesh_parameters.grid_arguments);
+          grid.make_grid(triangulation);
+
           GridTools::scale(mesh_parameters.scale, triangulation);
```
Julamipolymtl, line 244: `GridTools::scale` seems to moves vertices but not manifolds, and since the dome SphericalManifold is centred off-origin at (0, 0,offset), *scaling the mesh* also implies displacing the manifold center, which `GridTools::scale` (more precisely `GridTools::transform`) does not do. 

Therefore any further refined mesh would position the new nodes on a sphere about the unscaled center. The other manifolds (cylindrical) have no issues since their axis are not offset from the origin.

Julamipolymtl: I don't really see any solution that isn't case specific (no manifold scaling function on dealii's side), so maybe a quickfix for this mesh is to use the mesh scaling as a parameter within the grid script itself. That way all vertices are already scaled, and the manifold gets the proper scaled offset (`scale * offset`) from the start. In that case we would drop `GridTools::scale` for this particular mesh.

source/core/grid_impinging_jet_mixer.cc
```diff
@@ -0,0 +1,615 @@
...
+GridImpingingJetMixer<dim, spacedim>::GridImpingingJetMixer(
+  const std::string &grid_arguments)
+{
+  if constexpr (!(dim == 3 && spacedim == 3))
+    {
+      AssertThrow(
+        false,
+        ExcMessage(
+          "The impinging-jet mixer mesh is only supported in 3D space with 3D elements."));
+      return;
```
Julamipolymtl:
```suggestion
```

source/core/grid_impinging_jet_mixer.cc
```diff
@@ -0,0 +1,615 @@
...
+  // Sanity checks: all dimensions positive, the inlet/outlet pipes fit inside
+  // the chamber cross-section, and the inlet axis lies strictly inside the
+  // cylindrical part of the chamber wall (where the port patch is grown).
+  const MixerGeometry &g = this->geometry;
+  AssertThrow(g.R_chamber > 0.0 && g.r_inlet > 0.0 && g.r_outlet > 0.0 &&
+                g.H_chamber > 0.0 && g.L_cone > 0.0 && g.L_outlet > 0.0 &&
+                g.L_inlet > 0.0,
+              ExcMessage("All impinging-jet mixer dimensions must be strictly "
+                         "positive."));
+  AssertThrow(g.r_inlet < g.R_chamber && g.r_outlet < g.R_chamber,
```
Julamipolymtl, line 234: This check makes total sense geometrically, but is probably looser than what can be properly described by the discrete coarse geometry. I don't know if we need to get fancy about the exact condition for it to be properly defined, or simply be aware of it when generating meshes with large inputs/outlets.

include/core/grid_impinging_jet_mixer.h
```diff
@@ -0,0 +1,139 @@
...
+using namespace dealii;
+
+/**
+ * @brief Geometry dimensions of the impinging-jet mixer (SI units, metres).
+ *
+ * This is the single container for the mixer's geometric dimensions; the
+ * defaults below are used when the grid-argument string is empty, and are
+ * overridden by the constructor otherwise. The member functions expose the
+ * axial coordinates derived from these dimensions in the construction datum
+ * (bottom of the mixer at z = 0), in which the vessel is built before
```
Julamipolymtl: is the construction dattum at the bottom of the mixer or at the bottom of the cylindrical chamber (at the cylinder/conical interface)

Julamipolymtl (commented):
Looks good on the whole! I see one potential issue on the mesh scaling with a non-centred spherical manifold though.

source/core/grids.cc
blaisb: Yeah that's a good point. What I can do to change / fix this is put the center of the sphere at z=0 x=0 and y=0, this way the scaling won't have an impact. This is how I had done it, but I felt it was harder to set nice inlet/outlet boundary conditions because of that. However, better approach to have the manifold be robust.
Let me change that. Good catch.

source/core/grid_impinging_jet_mixer.cc
blaisb: Yeah this is a very tricky mesh to make a "general version". Like I said in the comments of the mesh, it's really only tested with a few set of parameters hence I'm really not sure what it will give with random parameters. I t's a bit overly complicated already as a mesh (and a surprise this can be made) :)

source/core/grids.cc
blaisb: Ok so the manifold is correct, it's the scale like we discussed. I will add adefensive assertions for that!

blaisb:
This version should be better now @Julamipolymtl :). I added the warnings like we discussed.
In a future PR I will add the same for translations, but it changed too many tests and I did not want to go there here since this grew the PR into something bigger than it is.

Julamipolymtl (approved):
Looks good I see nothing else:)

PR #2072, opened 2026-07-27
Allow pure DEM simulations within the CFD-DEM solver by disabling fluid dynamics

blaisb:
### Description

The CFD-DEM solver could not be used with fluid dynamics = false. It is still an interesting behaviour to be able to do these FEM-DEM simulation without fluid dynamics (e.g. for coupling with microwave). This PR adds this option.

### Testing

Testing on an eexample and it works just fine. I don,t think this requires it's own application test, since this is a simple if.

### Documentation

Nothing to document here that parameter is already documented, it is just efficiently generalized now.

Code related list:

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/include/solvers/navier_stokes_base.h
+++ b/include/solvers/navier_stokes_base.h
@@ -403,6 +403,18 @@ protected:
   update_multi_stage_solution(double time_step);
 
   /**
+   * @brief Specify the fluid dynamics solution instead of solving for it.
+   *
+   * When the fluid dynamics is disabled (fluid dynamics = false in the
+   * multiphysics subsection), the velocity and the pressure fields are not
+   * solved for, but rather imposed. They are either taken from the average
+   * velocity profile or re-established from the initial condition, which may
+   * be time-dependent.
+   */
+  virtual void
+  set_specified_fluid_dynamics_solution();
+
+  /**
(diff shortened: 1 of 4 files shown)


OresteMarquis (approved):
Nothing to add; it's rather simple indeed! With this, will the particle be unaffected by a constant null initial velocity profile, for exemple will it still feel the drag because of a difference in relative velocity or will the result be like a pur DEM simulation? Maybe a test that compare both output if they are suppose to be the same could be usefull at some point?

blaisb:
> Nothing to add; it's rather simple indeed! With this, will the particle be unaffected by a constant null initial velocity profile, for exemple will it still feel the drag because of a difference in relative velocity or will the result be like a pur DEM simulation? Maybe a test that compare both output if they are suppose to be the same could be usefull at some point?

Drag has to be explicitely disabled to ensure that the particles do not feel drag. The particle will feel the forces from the fluid velocity that are specified in the parameter file.  I'd rather keep it that way since it makes the solver very clear. Forces are forces, you choose them this way :).

OresteMarquis:
Oh perfect, I just wanted to know if it was possible to do it! But as it is, it's even more general!

PR #2074, opened 2026-07-29
Fix three annoying bugs with velocity averaging

blaisb:
### Description

  Three independent bugs affecting the time-averaged velocity and the resolved Reynolds
  stresses.

  1. The Reynolds shear stresses were never written to the output files.
  `NavierStokesBase::gather_output_results` builds `reynolds_shear_stress_names` and the
  matching component interpretation vector, but never adds the corresponding output struct,
  so only the normal stresses and the turbulent kinetic energy appeared in the results. The
  two vectors were left as dead code, which is why no warning was ever emitted. This is a
  regression from #1624  before that commit there were two `data_out.add_data_vector()` calls, 
and only the first one was
  converted to an `OutputStructSolution`.

  2. The averaging accumulators were reset at every load balance in the CFD-DEM solvers.
  `CFDDEMSolver::load_balance` and `CFDDEMMatrixFree::load_balance` transfer the fluid
  solutions, the void fraction and the particle handler to the new partition, but the
  averaging vectors were never part of that transfer. The `setup_dofs` call that follows the
  repartitioning reinitializes them, so `sum_velocity_dt`, `sum_reynolds_normal_stress_dt`
  and `sum_reynolds_shear_stress_dt` were reset to zero at every load balance while
  `real_initial_time` and `has_started_averaging` were left untouched. The numerator of the
  average therefore only covered the time elapsed since the last load balance whereas
  `total_time_for_average` still spanned the whole averaging window, giving
  `<u> ~ u * dt_since_last_load_balance / (t - t_avg_start)`. The average velocity and the
  Reynolds stresses consequently decayed towards zero as the simulation progressed, as a
  sawtooth with a `1/t` envelope whose jumps coincide with the repartitioning steps.

  3. The checkpointed averages were discarded on restart in the CFD-DEM solvers.
  `CFDDEMSolver::read_checkpoint` and `CFDDEMMatrixFree::read_checkpoint` duplicate the
  restart logic of `NavierStokesBase` and were never updated when #1300 
   made `AverageVelocities::read` deserialize into
  the vectors with ghost cells. `sanitize_after_restart`, which copies those back into the
  locally owned vectors the averaging routines accumulate into, was only added to
  `NavierStokesBase`. The locally owned sums were therefore left at zero after a restart,
  with the same consequence on the averages as above, and the corrupted state was propagated
  to every subsequent checkpoint. The same two functions were also missing the
  reinitialization of the average when the initial averaging time has not been reached yet.

  The bugs were found while auditing `AverageVelocities` after noticing that only the normal
  stresses were present in the output files.

### Solution

1. The output struct for the Reynolds shear stresses is added back in
   `NavierStokesBase::gather_output_results`, next to the one for the normal stresses.

2. Both `load_balance` implementations now call
   `AverageVelocities::prepare_for_mesh_adaptation` before `repartition()` and
   `AverageVelocities::post_mesh_adaptation` after `setup_dofs` and after the fluid vectors
   have been interpolated back, which is what the mesh adaptation of the other fluid
   solvers already did. `post_mesh_adaptation` also reevaluates the average velocity and
   the Reynolds stresses from the transferred sums.

3. The restart logic shared by the three solvers is factored into the new
   `NavierStokesBase::restore_average_velocities_after_checkpoint`, which performs the
   reinitialization check and the sanitization, and is called by the three
   `read_checkpoint` implementations.

### Testing

No test covered any of these paths, which is why all three went unnoticed:
- The `reynolds_stress_01/02` unit tests call `get_reynolds_shear_stresses()` directly
  rather than going through the output path, so they kept passing while the VTU files were
  missing the components.
- No application test combines averaging with a CFD-DEM simulation, so neither the load
  balance nor the restart path was ever exercised with `calculate average velocities` on.
  The restart tests that do use averaging (`poiseuille_restart`,
  `cavity_ht_average_velocity_as_initial_condition`) are `lethe-fluid` cases and go through
  `NavierStokesBase::read_checkpoint`.

I grew tired of this so I added two unit tests average_velocities_04 and average_velocities_05 that
check if the velocity averages are ok after restart and after load balancing/ mesh changes.

### Documentation

Nothing to document here

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/navier_stokes_base.h
```diff
@@ -473,6 +473,47 @@ class NavierStokesBase : public PhysicsSolver<VectorType>
...
+  }
+
+  /**
+   * @brief Restore the state of the time-averaged velocity object after its
+   * vectors have been deserialized.
+   *
+   * The deserialization fills the vectors with ghost cells, since only those
+   * are serialized, whereas the averaging routines accumulate into the
+   * locally owned vectors. The locally owned vectors must therefore be reset
+   * from their ghosted counterparts, otherwise the averaging silently
```
antonellasuccar: I don't find this very clear: are only values of the time-averaged velocity in the ghost cells serialized? Is it just the ghost cells or the locally relevant cells?

source/solvers/navier_stokes_base.cc
```diff
@@ -2315,15 +2304,10 @@ NavierStokesBase<dim, VectorType, DofsType>::read_checkpoint()
...
-      this->average_velocities->calculate_average_velocities(
-        this->local_evaluation_point,
-        simulation_parameters.post_processing,
-        simulation_control->get_current_time(),
-        simulation_control->get_time_step());
-    }
+  // The averaged fields, which are used by the average velocity profile initial
+  // condition, are reevaluated from the checkpointed sums by
+  // restore_average_velocities_after_checkpoint, which is called by
+  // set_solution_from_checkpoint above.
```
antonellasuccar: Is it only if an average velocity profile is used as an initial condition or also when required by the postprocessing?

tests/solvers/average_velocities_04.cc
```diff
@@ -0,0 +1,357 @@
...
+ *   1. accumulates the averaged fields over five time steps,
+ *   2. uniformly refines the mesh,
+ *   3. accumulates the averaged fields over five more time steps.
+ *
+ * The continuous L2 norm of the three averaged fields is printed after each of
+ * these stages. The norms printed before and after the refinement must be
+ * identical.
+ *
+ * The refinement is carried out with refine_mesh_uniform(), which is the
+ * production path the dynamic mesh adaptation of the solvers goes through, so
```
antonellasuccar: what's a production path?

tests/solvers/average_velocities_04.cc
```diff
@@ -0,0 +1,357 @@
...
+  NSparam.parse(prm);
+
+  // Manually alter some of the default parameters of the solver
+  NSparam.simulation_control.method =
+    Parameters::SimulationControl::TimeSteppingMethod::bdf1;
+  NSparam.simulation_control.dt       = 0.1;
+  NSparam.simulation_control.time_end = 1.0;
+
+  // A second order element is used so that the averaged fields, which are
+  // quadratic in space, belong to the finite element space and are therefore
```
antonellasuccar, line 295: quadratic at most. The averaged velocity is linear

antonellasuccar (approved):
Nice catch and nicely explained :)

tests/solvers/average_velocities_04.cc
blaisb: If the average velocity is linear, the Reynolds stresses, which are part of the average fields, are quadratic. I 'll adapt the comment to be clearer here.

include/solvers/navier_stokes_base.h
blaisb: Here it meant locally relevant, I adapted the comment to use locally_relevant.

source/solvers/navier_stokes_base.cc
blaisb: That comment was not really helpful now. I removed it.

tests/solvers/average_velocities_04.cc
blaisb: It was to say that this is how the solver does it (since this is a unit test that mimics a solver) I removed that part to clarify that.

tests/solvers/average_velocities_04.cc
blaisb: This was actually explained in the brief of the test (that the Reynolds stresses are quadratic). See Paragraph at line 31.

PR #2079, opened 2026-07-31
Restrict expanded particle-wall contact search to convex boundary faces

antonellasuccar:
### Description

Currently, the ``expand particle-wall contact search`` in the DEM solver allows the calculation of the overlap between a particle and the walls from neighboring cells. This is needed in a convex domain. However, using this feature in a  in a concave geometry leads to the calculation of erroneous contact, therefore this feature should not be activated in these geometries. The problem arises with domains that have some faces that form an convex inner part and other ones that form a concave part. 

This PR adresses this issue in the following way, in the function ``add_boundary_neighbors_of_boundary_cells``:
1) For each boundary face already identified in boundary_cells_info, a loop is performed over its neighboring cells using the vertex of the boundary faces. 

2) For each boundary face of a given neighboring cell, if it is neither an outlet nor a periodic boundary face, check if it shares an edge (two vertices) with the main boundary face. If the two faces share an edge, check if they delimit a concave or convex part of the domain by calculating the following:

$$
\mathbf{n}_1 \cdot (\mathbf{p}_2-\mathbf{p}_1)
$$

where $\mathbf{n}_1$ is the normal to the main boundary face (pointing outwards of the domain) and $\mathbf{p}_1$ and $\mathbf{p}_2$ are the center of the main boundary face and the neighboring boundary face, respectively. If this product is negative, it means that the faces form a convex geometry. Thus, the main boundary cell is added with the information of the neighbor cell boundary face to the ``boundary_cells_info``container. This allows the calculation of the contact between the particle and the boundary face in the neighboring cell when the particle is in the main cell. **This is not done if the faces form a concave geometry.**

If this PR is approved, the ``expand particle-wall contact search`` feature will be removed in a subsequent PR. Nonetheless, since it is not yet compatible with ``check diamond cells`` feature, it is better to make it compatible in a PR, before completely removing this feature.

### Testing

This has been tested with two tests: a particle on a horizontal surface followed by an ascending slope (convex configuration) and another test with a particle on a horizontal surface followed by a descending slope (concave geometry). When setting ``expand particle-wall contact search`` to ``true``, the particle velocity is no longer diverging in the concave case and gives the same result as master with no  ``expand particle-wall contact search``, indicating that only the appropriate neighboring boundary faces are tested for an overlap when the particle is in the main cell. 

Also, using the convex geometry and setting ``expand particle-wall contact search`` to ``true`` gives the same result of master with ``expand particle-wall contact search`` set to ``true``, if the code is modified to include faces that have the same orientation of a given boundary face. The code does not include this type of faces in the final version, which gives results that are slightly different from master, which is an improvement.

image
image
PR_fig_2

### Use of LLMs
They were mainly used to inquire about certain parts of the code and to count number of certain faces in the output.

Code related list:

Pull request related list:
- [x] Labels are applied


source/dem/find_boundary_cells_information.cc
```diff
@@ -745,44 +757,96 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                          TriaIterator< TriaAccessor< dim - 1, dim, dim > > neighbor_face_iterator;
+                          for (unsigned int f = 0; f < GeometryInfo<dim>::faces_per_cell; ++f)
+                            {
+                                if (boundary_neighbor_information.cell->face_index(f) ==
+                                    boundary_neighbor_information.global_face_id)
+                                {
+                                    neighbor_face_iterator = boundary_neighbor_information.cell->face(f);
+                                    break;
+                                }
+                            }
```
blaisb: I don't understand that loop you are doing here. You are already iterating over tehf aces at line 747, why loop again over the faces?

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,44 +757,96 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                          unsigned int common_vertices = 0;
+
+                          for (auto main_vertex_id : main_face_iterator->vertex_indices()) 
+                             for (auto neighbor_vertex_id : neighbor_face_iterator->vertex_indices()) 
+                                if (main_face_iterator->vertex_index(main_vertex_id) == neighbor_face_iterator->vertex_index(neighbor_vertex_id)) 
+                                   {
+                                      common_vertices++;
+                                   }
+                          if (common_vertices == 2)
+                              share_edge = true;
```
blaisb: This is only true in 3D. This check should not occur when dim==2

blaisb: Also won't work in meshes that are adapted. Worth a comment at the very least.

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,44 +757,96 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                              // needed in geometries with both concave and convex 
+                              // parts of the boundaries. 
+
+                              // Get the normal to the face of the main boundary cell,
+                              // pointing outwards from the cell
+                              auto normal_1 = -boundary_cells_info.normal_vector;
+                              // Get the point on the face of the main boundary cell
+                              auto point_1 = boundary_cells_info.point_on_face;
+                              // Get the point on the face of the neighbor boundary cell
+                              auto point_2 = boundary_neighbor_information.point_on_face;
```
blaisb: These should be const references.

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,44 +757,96 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                          if (share_edge)
+                            {
+                              // Check if the boundary faces of the main cell and
+                              // the neighbor cell form a convex surface. This is
+                              // needed in geometries with both concave and convex 
+                              // parts of the boundaries. 
+
+                              // Get the normal to the face of the main boundary cell,
+                              // pointing outwards from the cell
+                              auto normal_1 = -boundary_cells_info.normal_vector;
```
blaisb: This one should be const.

source/dem/find_boundary_cells_information.cc
```diff
@@ -717,17 +717,29 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
   for (auto &&boundary_cells_info :
        boundary_cells_information | boost::adaptors::map_values)
```
blaisb, line 720: This is tricky. You are mutating the map (adding entries to it) while looping through it. This was an issue in the past but I don,'t think it caused much troubles, but here it's a bit weirder I don't think you should be adding to the same map you are iterating on. Maybe store things in a vector or another container and at the end merge the two containers.

applications_tests/lethe-particles/expand_particle_wall_concave.prm
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
+
+set dimension = 3
```
blaisb, line 17: Can you add a paragraph that describes what the application test aims to accomplish

applications_tests/lethe-particles/expand_particle_wall_convex.prm
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 The Lethe Authors
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception OR LGPL-2.1-or-later
+
+# Listing of Parameters
+#----------------------
+
+set dimension = 3
```
blaisb, line 16: Can you add a paragraph that explains what the application test aims to accomplish

release_notes/current/2026_08_06_Succar.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/08/06
+
+### Fixed
+
+- MINOR The `expand particle-wall contact search` option in DEM allows the calculation of particle-wall overlaps with the boundary faces of the cell containing the particle, as well as with the boundary faces of neighboring cells. This is required for convex geometries. However, when this feature is used in a concave geometry, it can lead to the calculation of unphysical contacts. The issue particularly arises in domains that contain both convex and concave boundary regions. This PR addresses the problem by adding only the boundary faces of neighboring cells that form a convex boundary with the boundary face of the cell containing the particle. [#2079](https://github.com/chaos-polymtl/lethe/pull/2079)
```
blaisb:
```suggestion
- MAJOR The `expand particle-wall contact search` option in DEM allows the calculation of particle-wall overlaps with the boundary faces of the cell containing the particle, as well as with the boundary faces of neighboring cells. This is required for convex geometries. However, when this feature is used in a concave geometry, it can lead to the calculation of unphysical contacts. The issue particularly arises in domains that contain both convex and concave boundary regions. This PR addresses the problem by adding only the boundary faces of neighboring cells that form a convex boundary with the boundary face of the cell containing the particle. [#2079](https://github.com/chaos-polymtl/lethe/pull/2079)
```

This is a major change since it alters simulation results.

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +706,41 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+        {
+          if (boundary_cells_info.cell->face_index(f) ==
+              boundary_cells_info.global_face_id)
+            {
+              main_face_iterator = boundary_cells_info.cell->face(f);
+              found_face         = true;
+              break;
+            }
+        }
+      AssertThrow(found_face,
```
blaisb: I think there's an issue here when diamond cells are enabled.
Diamond cells are added as artificial cells with a wrapped unsigned int (like taking the negative value of an unsigned int) So if you had a mesh with a diamond cell (and diamond cell enabled) AND this feature set to on, you would automatically assert here.
I am not sure this can really happen, but what I would do is I would try to take one of the test case we have with diamond cell on and check what will happen here (even if there's no wall to be expanded in that test case).
It might be that for the time being, both options are incompatible together. It's not a big deal because diamond cell + convex cell in a same mesh are not something I have seen in the past, but if that is the case, we should AssertThrow that both features are not "on" at the same time at the beggining of the simulation instead of randomly throwing later on in the simulation.

blaisb: Furthermore, since this is in a "Loop" I would use Assert() here instead of AssertThrow. You don't want this assertion all the time in release mode, it's more a defensive protection.

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,50 +758,121 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                          // This part assumes a mesh with a uniform refinement
+                          // level
+                          if constexpr (dim == 2)
+                            {
+                              share_edge = (common_vertices == 1);
+                            }
+                          else if constexpr (dim == 3)
+                            {
+                              share_edge = (common_vertices == 2);
+                            }
```
blaisb, line 793: Yup, we will have to think what happens if we one day have adaptive mesh refinement, but this is not really the case often in DEM/CFD-DEM.

source/dem/find_boundary_cells_information.cc
```diff
@@ -65,10 +65,6 @@ BoundaryCellsInformation<dim>::build(
         {
           pcout
             << "Warning: expansion of particle-wall contact list is enabled. "
-            << std::endl
-            << "This feature should be used only in geometries with convex boundaries. "
-               "(For example, particles flowing inside a cylinder or sphere). For geometries with "
-               "concave boundaries, this feature MUST NOT be activated."
             << std::endl;
           display_pw_contact_expansion_warning = false;
```
blaisb, line 71: I am not sure it's worth it to print this warning?
I mean, it used to be worth it because there were action that were derived from it. Maybe I would still only say that "This feature should be used only in geometries with convex boundaries. ". Otherwise this feature is not useful....
If it should "always" be used, then it should not be an option.

blaisb (changes requested):
A few changes there and there and things to be careful about.
Please rebase the branch since the half-step Velocity-Verlet merge will change some of your test results.

doc/source/parameters/dem/mesh.rst
```diff
@@ -20,13 +20,9 @@ The mesh subsection of DEM simulations is almost identical to the `CFD <https://
...
 .. image:: images/expand_particle_wall.png
     :alt: Schematic
     :align: center
-    :width: 400
-
-
-.. warning:: 
-     In geometries with convex boundaries, this feature MUST NOT be activated.
\ No newline at end of file
+    :width: 400
```
blaisb, line 28: I think the image needs to be updated no?
It does not show the correct new definition of the wall expansion

source/dem/find_boundary_cells_information.cc
antonellasuccar: I guess we would have hanging nodes? It would be interesting to look at it later

blaisb:
I'll wait also for @OGaboriault 's review

doc/source/parameters/dem/mesh.rst
```diff
@@ -20,13 +20,9 @@ The mesh subsection of DEM simulations is almost identical to the `CFD <https://
...
+* ``expand particle-wall contact search`` enables adding the boundary faces of cells neighboring the boundary cell containing the particle, to the particle-wall contact search list. Only neighboring boundary faces that form a convex boundary with the current cell are added; boundary faces forming a non-convex boundary are excluded. The following figure illustrates this concept. When a particle is located in a boundary cell (shown in red), the boundary faces of neighboring cells that form a convex boundary with the current cell (neighboring cells shown in green) are also considered during the particle-wall contact search. This is necessary because a particle whose centroid lies within the red cell can simultaneously contact both the boundary face of the current cell and the boundary face of a neighboring cell when the two faces form a convex boundary. In contrast, boundary faces of neighboring cells that form a concave boundary (neighboring cells shown in orange) are not considered during the particle-wall contact search.
```
OGaboriault: The first sentence is weird.  "of cells neighboring" ? 

Do you mean "enables adding the boundary faces of neighboring cells" ?

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,41 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+          if (boundary_cells_info.cell->face_index(f) ==
+              boundary_cells_info.global_face_id)
+            {
+              main_face_iterator = boundary_cells_info.cell->face(f);
+              found_face         = true;
+              break;
+            }
+        }
+      Assert(found_face,
+             ExcMessage("Boundary face not found in boundary cell."));
```
OGaboriault: I am not sure I understand what is happening here. 

You loop on boundary cell info, then you loop on ``f``, but what is f? faces_per_cell is a constant? I though the code with any type of cell, thus this is not necessarily the same number for every cell.. 

I would had comments here.

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,50 +760,121 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                                  // applied to the boundary cell & store in map
+                                  // with imaginary key.
+                                  boundary_cells_info_struct<dim>
+                                    boundary_information =
+                                      boundary_neighbor_information;
+                                  boundary_information.cell =
+                                    boundary_cells_info.cell;
+
+                                  imaginary_boundary_faces.insert(
+                                    {imaginary_face_id, boundary_information});
```
OGaboriault: This is more of an esthetic comment: 

 Right now, you have 4 "if" nested into each other. This makes the number of characters per line really small. You could use 
`if (!current_condition)
    continue` instead

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,50 +760,121 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                          // This part assumes a mesh with a uniform refinement
+                          // level
+                          if constexpr (dim == 2)
+                            {
+                              share_edge = (common_vertices == 1);
+                            }
+                          else if constexpr (dim == 3)
+                            {
+                              share_edge = (common_vertices == 2);
+                            }
```
OGaboriault:
```suggestion
                          if constexpr (dim == 2)
                            
                              share_edge = (common_vertices == 1);
                            
                          else if constexpr (dim == 3)
                            
                              share_edge = (common_vertices == 2);
                            
```

OGaboriault (commented):
I need to read the find_boundary_cells_information.cc file again. First thing tomorrow.

doc/source/parameters/dem/mesh.rst
blaisb:
```suggestion
* ``expand particle-wall contact search``  extends the boundary walls in a cell by appending the boundary walls of the neighboring cells to the particle-wall contact list.  This adds the boundary faces of the neighboring cells and allow particles to collide with boundary walls that are not faces of the cell in which they reside. Only neighboring boundary faces that form a convex boundary with the current cell are added; boundary faces forming a non-convex boundary are excluded. The following figure illustrates this concept. When a particle is located in a boundary cell (shown in red), the boundary faces of neighboring cells that form a convex boundary with the current cell (neighboring cells shown in green) are also considered during the particle-wall contact search. This is necessary because a particle whose centroid lies within the red cell can simultaneously contact both the boundary face of the current cell and the boundary face of a neighboring cell when the two faces form a convex boundary. In contrast, boundary faces of neighboring cells that form a concave boundary (neighboring cells shown in orange) are not considered during the particle-wall contact search.
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,41 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
-      // This loop will check for all the cell neighbors located at vertices of
-      // a boundary cell and will add the proper information in the boundary
-      // cell list if faces are not at outlet nor periodic boundary.
+      // The boundary_cells_info is local to each processor by construction
+      // Each entry in boundary_cells_info corresponds to a boundary face
+      // We start by getting the face iterator of the boundary face in the
+      // boundary cell
+      TriaIterator<TriaAccessor<dim - 1, dim, dim>> main_face_iterator;
+      bool                                          found_face = false;
+      for (unsigned int f = 0; f < GeometryInfo<dim>::faces_per_cell; ++f)
```
blaisb: You should not use GeometryInfo<dim>::faces_per_cell since we will one day want to be able to use Simplex meshes. 

Using either these:
```
for (const unsigned int f : cell->face_indices())
  if (cell->face(f)->at_boundary()) ...

for (const auto &face : cell->face_iterators())
  if (face->at_boundary()) ...
```

This is the better way of doing it.
GeometryInfo should be deprecated

blaisb: It's also what you are using below

source/dem/find_boundary_cells_information.cc
blaisb: Yeah I agree
Instead of if (share_edge)
I would use if (!share_edge) continue;

Then continue the code. It would remove one full indentation level.

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,50 +760,121 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                            }
+                          else if constexpr (dim == 3)
+                            {
+                              share_edge = (common_vertices == 2);
+                            }
 
-                          boundary_cells_information.insert(
-                            {imaginary_face_id, boundary_information});
+                          if (share_edge)
+                            {
```
blaisb: here I would use if (!share_edge) continue;

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,50 +760,121 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                                point_2 - point_1;
+                              vector_main_to_neighbor /=
+                                vector_main_to_neighbor.norm();
+                              const double tolerance = 1e-12;
+                              bool         is_convex =
+                                (normal_1 * vector_main_to_neighbor) <
+                                -tolerance;
+
+                              if (is_convex)
+                                {
```
blaisb: here I would use if (!is_convex) continue


This way you will have removed two indentations levels

blaisb (approved):
Good to go on my side after my changes

doc/source/parameters/dem/mesh.rst
antonellasuccar: I implemented Bruno's suggestion

source/dem/find_boundary_cells_information.cc
antonellasuccar: `f` is the local index of the face on the cell. The goal of this loop is to get the face iterator so i can use it to get the vertices of the face. I removed ``GeometryInfo<dim>::faces_per_cell`` and used Bruno's first suggestion

source/dem/find_boundary_cells_information.cc
antonellasuccar: I applied the first suggestion. The code below did not change unless i misunderstood something

source/dem/find_boundary_cells_information.cc
antonellasuccar: I did this for is_convex and share_edge

source/dem/find_boundary_cells_information.cc
```diff
@@ -745,25 +761,89 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
 
                       if (face_is_wall)
                         {
```
blaisb: Here also you could continue if the face is not wall instead.
like if(!face_is_wall) continue
It would make the code a lot easier to read by removing indentations.

blaisb (approved):
One last comment. I will wait for the final approval also of @OGaboriault 
I would like to havet hat in the next release.

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
   const std::map<int, boundary_cells_info_struct<dim>>
     &global_boundary_cells_information)
 {
+  // Store imaginary faces here and add them to boundary_cells_information after
```
OGaboriault: what is an imaginary face?

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+  // the loop to avoid modifying the map while iterating over it
+  std::map<int, boundary_cells_info_struct<dim>> imaginary_boundary_faces;
+
   // Create a vector of a set of adjacent cells of all the vertices
```
OGaboriault:
```suggestion
  // Create a vector of a sets containing the adjacent cells of all the vertices
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
-      // This loop will check for all the cell neighbors located at vertices of
-      // a boundary cell and will add the proper information in the boundary
-      // cell list if faces are not at outlet nor periodic boundary.
+      // The boundary_cells_info is local to each processor by construction
+      // Each entry in boundary_cells_info corresponds to a boundary face
+      // We start by getting the face iterator of the boundary face in the
+      // boundary cell
+      TriaIterator<TriaAccessor<dim - 1, dim, dim>> main_face_iterator;
+      bool                                          found_face = false;
+      for (const unsigned int f : boundary_cells_info.cell->face_indices())
```
OGaboriault:
```suggestion
     // Loop on all the face a the boundary cell ? 
      for (const unsigned int f : boundary_cells_info.cell->face_indices())
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+        {
+          if (boundary_cells_info.cell->face_index(f) ==
```
OGaboriault:
```suggestion
          // Check if the current face is ...
          if (boundary_cells_info.cell->face_index(f) ==
```
what is  boundary_cells_info.global_face_id ?

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+              break;
+            }
+        }
+
+      Assert(found_face,
+             ExcMessage("Boundary face not found in boundary cell."));
 
-      // Iterate over the vertices of each boundary cell
-      for (auto vertex_id : GeometryInfo<dim>::vertex_indices())
+      for (auto vertex_id : main_face_iterator->vertex_indices())
```
OGaboriault, line 751:
```suggestion
      // Loop on the vertex of the main cell
      for (auto vertex_id : main_face_iterator->vertex_indices())
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
         {
           // Iterate over the neighbors of each boundary cell having vertex_id
```
OGaboriault:
```suggestion
          // Iterate over the neighboring cells linked to this vertex
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,24 +708,42 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
           for (const auto &neighbor :
-               v_to_c[boundary_cells_info.cell->vertex_index(vertex_id)])
+               v_to_c[main_face_iterator->vertex_index(vertex_id)])
             {
               // Iterate over the faces of the neighbor cell
```
OGaboriault:
```suggestion
              // Iterate over the faces of the current neighboring cell
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +759,108 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
                         (!outlet_boundaries.contains(face->boundary_id()) &&
                          !neighbor->has_periodic_neighbor(face_id));
```
OGaboriault: I think that this is wrong. 

a periodic *cell* can have a concave/convex face with another cell. 

The condition should be, if the *face* is periodic, not if the *cell* is. (I think)

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +759,108 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
-                          boundary_information.cell = boundary_cells_info.cell;
-
-                          boundary_cells_information.insert(
-                            {imaginary_face_id, boundary_information});
-                        }
+                      if (!face_is_wall)
+                        continue;
+
+                      // Get the boundary face in neighbor cell info
+                      const boundary_cells_info_struct<dim>
```
OGaboriault: what is the boundary_cells_info_struct?

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +759,108 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                      for (auto main_vertex_id :
+                           main_face_iterator->vertex_indices())
+                        for (auto neighbor_vertex_id : face->vertex_indices())
+                          if (main_face_iterator->vertex_index(
+                                main_vertex_id) ==
+                              face->vertex_index(neighbor_vertex_id))
+                            {
+                              common_vertices++;
+                            }
+                      // This part assumes a mesh with a uniform refinement
```
OGaboriault:
```suggestion
                      // In 2D, faces are neighbors if they share 1 vertex. 
                      // In 3D, ...
                      // Note:This part assumes a mesh with a uniform refinement
```

OGaboriault (commented):
I will continue later, but I think there is a problem regarding if the cell is periodic vs is the face is.

source/dem/find_boundary_cells_information.cc
blaisb: They are the new "virtual faces" that are added in the contact list for the particles due to the expand particle wall contact. That's always been there. Maybe it could be clarified b etter.

antonellasuccar: I added a comment to clarify this

source/dem/find_boundary_cells_information.cc
antonellasuccar: It is the global id of the face corresponding to the current entry in ``boundary_cells_info``. Each entry in this map corresponds to a boundary face and gives information about the boundary cell linked to it as well as the normal to it and a point on the face. I added more comments.

source/dem/find_boundary_cells_information.cc
antonellasuccar: It is actually a loop over the vertices of the face

source/dem/find_boundary_cells_information.cc
antonellasuccar: I don't think so. This is the dealii description of the function
`If the cell has a periodic neighbor at its ith face, this function returns true, otherwise, the returned value is false.`

source/dem/find_boundary_cells_information.cc
antonellasuccar: I'm not sure I understand. It is the structure defined in ``boundary_cells_info_struct.h``. I'm not sure if i reexplain it here? It may become redundant

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+      // face and find the neighboring cells
+      TriaIterator<TriaAccessor<dim - 1, dim, dim>> main_face_iterator;
+      bool                                          found_face = false;
+      // Loop over the faces of the boundary cell to get the face iterator of 
+      // the current boundary face
+      for (const unsigned int f : boundary_cells_info.cell->face_indices())
+        {
+          // Check if the global index of this face in the boundary cell matches
+          // the global index of the boundary face we are currently at in
+          // the loop over boundary_cells_info. If so, get the face iterator
```
OGaboriault:
```suggestion
          // Check if the global index of this face matches the global index of 
          // the boundary we are inspecting. If so, get the face iterator.
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+              break;
+            }
+        }
+
+      Assert(found_face,
+             ExcMessage("Boundary face not found in boundary cell."));
 
-      // Iterate over the vertices of each boundary cell
-      for (auto vertex_id : GeometryInfo<dim>::vertex_indices())
+      // Loop over the vertices of the boundary face in the main boundary cell
```
OGaboriault:
```suggestion
      // Loop over the vertices of the boundary face.
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
   const std::map<int, boundary_cells_info_struct<dim>>
     &global_boundary_cells_information)
 {
-  // Create a vector of a set of adjacent cells of all the vertices
+  // Store imaginary boundary-face entries here and add them to
```
OGaboriault: what is a imaginary boundary-face?

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+  // with a different cell, we need to generate unique imaginary keys for them.
+
+  std::map<int, boundary_cells_info_struct<dim>> imaginary_boundary_faces;
+
+  // Create a vector of sets containing the adjacent cells of all the vertices
   std::vector<std::set<typename Triangulation<dim>::active_cell_iterator>>
     v_to_c = GridTools::vertex_to_cell_map(triangulation);
 
   for (auto &&boundary_cells_info :
        boundary_cells_information | boost::adaptors::map_values)
```
OGaboriault, line 724: What are we looping on here? 

what is ``boundary_cells_info`` and ``boundary_cells_information`` ? Not sure if the variable name is adequate considering they are besicaly the same.

OGaboriault: information takes an ``s``? 

maybe boundary_cells_info could become this_boundary_cell_infos ?

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
-      // Iterate over the vertices of each boundary cell
-      for (auto vertex_id : GeometryInfo<dim>::vertex_indices())
+      // Loop over the vertices of the boundary face in the main boundary cell
+      for (auto vertex_id : main_face_iterator->vertex_indices())
         {
-          // Iterate over the neighbors of each boundary cell having vertex_id
+          // Iterate over the neighboring cells linked to this vertex
           for (const auto &neighbor :
-               v_to_c[boundary_cells_info.cell->vertex_index(vertex_id)])
+               v_to_c[main_face_iterator->vertex_index(vertex_id)])
```
OGaboriault, line 755: what is the difference when using : 

"main_face_iterator->vertex_index(vertex_id)"
and 
"main_face_iterator->vertex_index()"    ?

source/dem/find_boundary_cells_information.cc
OGaboriault: boundary_cells_info_struct "Handle the information related to the boundary cells which will be used for particle-wall contact detection." I get that... 

What I means is, which cell are we talking about here. The main cell? One of it's neighbors? 

// Get the boundary information of the neighboring cell. 

Something like that.

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                      // level. In 2D, two boundary faces share an edge if 
+                      // they have one common vertex. In 3D, the boundary faces
+                      // share an edge if they have two common vertices.
+                      if constexpr (dim == 2)
+
+                        share_edge = (common_vertices == 1);
+
+                      else if constexpr (dim == 3)
+
+                        share_edge = (common_vertices == 2);
```
OGaboriault:
```suggestion
                      if constexpr (dim == 2)
                        share_edge = (common_vertices == 1);

                      else if constexpr (dim == 3)
                        share_edge = (common_vertices == 2);
```

I also think that Clang Tiddy will not like this since share_edge is not define for dim == 0, 1, 4, etc. 

It is not a mistake since we are always compiling in dim == 2 and 3, but you might get a waring.

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+
+                      // Exit the loop if the boundary faces of the main
+                      // cell and the neighbor cell do not share an edge
+                      if (!share_edge)
+                        continue;
+
+                      // Check if the boundary faces of the main cell and the
+                      // neighbor cell form a convex surface. This is needed in
+                      // geometries with both concave and convex parts of the
+                      // boundaries.
```
OGaboriault:
```suggestion
                      // Check whether the two boundary faces form a convex surface.
                      // This is needed for geometries with both concave and convex
                      // regions.
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
         {
-          // Iterate over the neighbors of each boundary cell having vertex_id
+          // Iterate over the neighboring cells linked to this vertex
           for (const auto &neighbor :
-               v_to_c[boundary_cells_info.cell->vertex_index(vertex_id)])
+               v_to_c[main_face_iterator->vertex_index(vertex_id)])
             {
-              // Iterate over the faces of the neighbor cell
+              // Iterate over the faces of the current neighboring cell
               for (const auto &face : neighbor->face_iterators())
```
OGaboriault: You should call face --> neighboring_face

This will help distinguish between the boundary face of the main cell and its neighbor(s).

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,26 +708,55 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+      // The boundary_cells_info is local to each processor by construction
+      // Each entry in boundary_cells_info corresponds to a boundary face
+      // We start by getting the face iterator of the boundary face in the
+      // boundary cell. This is needed to access the vertices of the boundary
+      // face and find the neighboring cells
+      TriaIterator<TriaAccessor<dim - 1, dim, dim>> main_face_iterator;
+      bool                                          found_face = false;
+      // Loop over the faces of the boundary cell to get the face iterator of 
+      // the current boundary face
+      for (const unsigned int f : boundary_cells_info.cell->face_indices())
```
OGaboriault: You should call f --> main_face

This will help distinguish between the boundary face of the main cell and its neighbor(s)

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                        continue;
+
+                      // Get the boundary face in neighbor cell info
+                      const boundary_cells_info_struct<dim>
+                        &boundary_neighbor_information =
+                          global_boundary_cells_information.at(
+                            neighbor->face_index(face_id));
+
+                      // Check if boundary face in neighbor cell shares an
+                      // edge with the boundary face in the main cell.
```
OGaboriault:
```suggestion
                      // Check if the neighboring face shares an edge 
                      // with the main face.
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                      if constexpr (dim == 2)
+
+                        share_edge = (common_vertices == 1);
+
+                      else if constexpr (dim == 3)
+
+                        share_edge = (common_vertices == 2);
+
+                      // Exit the loop if the boundary faces of the main
+                      // cell and the neighbor cell do not share an edge
```
OGaboriault:
```suggestion
                      // Exit the loop if the main face and the 
                      // neighboring face do not share an edge
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                      if (!share_edge)
+                        continue;
+
+                      // Check if the boundary faces of the main cell and the
+                      // neighbor cell form a convex surface. This is needed in
+                      // geometries with both concave and convex parts of the
+                      // boundaries.
+
+                      // Get the normal to the face of the main boundary cell,
+                      // pointing outwards from the cell
```
OGaboriault:
```suggestion
                      // Normal of the main face ( pointing outwards from the cell) 
```
Maybe pointing outside of the triangulation?

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                      const auto &point_1 = boundary_cells_info.point_on_face;
+                      // Get the point on the face of the neighbor boundary cell
+                      const auto &point_2 =
+                        boundary_neighbor_information.point_on_face;
+                      // Create a vector from point_1 to point_2
+                      Tensor<1, dim> vector_main_to_neighbor =
+                        point_2 - point_1;
+                      vector_main_to_neighbor /= vector_main_to_neighbor.norm();
+                      const double tolerance = 1e-12;
+                      bool         is_convex =
```
OGaboriault:
```suggestion
                      // If the scalar product is negative, the face are convex.
                      bool         is_convex =
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                        (normal_1 * vector_main_to_neighbor) < -tolerance;
+
+                      // Exit the loop if the boundary faces of the main cell
+                      // and the neighbor cell do not form a convex surface
```
OGaboriault:
```suggestion
                      // Exit the loop if the two boundary faces 
                      // do not form a convex surface
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -743,52 +770,110 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
+                        {imaginary_face_id, boundary_information});
                     }
                 }
             }
         }
     }
+
+  // Add the imaginary boundary faces to the boundary_cells_information map
+  // after the loop is finished
+  boundary_cells_information.merge(imaginary_boundary_faces);
```
OGaboriault, line 874: imaginary_boundary_faces is not define explained.

antonellasuccar (commented):
I think Gabo's comments were on point. That part of the code is clearer now :)

blaisb:
Yeah I agree, it's much better than it has ever was. I'll still wait for @OGaboriault final ok before merging just to be sure

release_notes/current/2026_08_06_Succar.md
```diff
@@ -0,0 +1,5 @@
+## [Master] - 2026/08/06
+
+### Fixed
+
+- MAJOR The `expand particle-wall contact search` option in DEM allows the calculation of particle-wall overlaps with the boundary faces of the cell containing the particle, as well as with the boundary faces of neighboring cells. This is required for convex geometries. However, when this feature is used in a concave geometry, it can lead to the calculation of unphysical contacts. The issue particularly arises in domains that contain both convex and concave boundary regions. This PR addresses the problem by adding only the boundary faces of neighboring cells that form a convex boundary with the boundary face of the cell containing the particle. [#2079](https://github.com/chaos-polymtl/lethe/pull/2079)
```
OGaboriault:
```suggestion
- MAJOR The `expand particle-wall contact search` option in DEM allows the calculation of particle-wall overlaps with the boundary faces of the cell containing the particle, as well as with the boundary faces of neighboring cells. This is required for convex geometries. However, when this feature is used in a concave geometry, it can lead to the calculation of unphysical contacts. The issue particularly arises in domains that contain both convex and concave boundary regions. This PR addresses the problem by adding only the boundary faces of neighboring cells that form a convex region with the boundary face of the cell containing the particle. [#2079](https://github.com/chaos-polymtl/lethe/pull/2079)
```

source/dem/find_boundary_cells_information.cc
```diff
@@ -710,85 +708,170 @@ BoundaryCellsInformation<dim>::add_boundary_neighbors_of_boundary_cells(
...
        boundary_cells_information | boost::adaptors::map_values)
     {
-      // The boundary cell is local by definition
-      // This loop will check for all the cell neighbors located at vertices of
-      // a boundary cell and will add the proper information in the boundary
-      // cell list if faces are not at outlet nor periodic boundary.
+      // Each entry in boundary_cells_information corresponds to a boundary face
+      // We start by getting the face iterator of the boundary face in the main
+      // boundary cell. This is needed to access the vertices of that face and
+      // find the neighboring cells
```
OGaboriault, line 729: I would write an issue, about this. 

boundary_cell_info should be the info about a boundary cell, not a boundary face like it is explain. The name of the variable is not appropriate.

OGaboriault (approved):
There is still one thing that is bothering me, but I think it is ok. 

boundary_cell_info is not an appropriate variable name if it is related to the boundary face... that it all.

source/dem/find_boundary_cells_information.cc
antonellasuccar: I opened an issue to revisit this. I feel it is information about both the faces and the cells. I think it is debatable though. I see it as we are adding boundary face information pertinent to the cells. Especially after adding the imaginary keys, we are adding neighboring face information to the cell. Also, when finding the contacts, the loop will be over the particles in a given cells and the boundary face information will be appended to it.

PR #2090, opened 2026-08-11
Add probes to postprocessing

AmishgaAlphonius:
### Description

This PR adds probes to the postprocessing module. This allows evaluation of solved variables at remote points of the domain (not necessarily a DoF). Values are outputted in ``.dat`` files (1 per probing point).  Each probing point is characterized by a unique ID and a user-specified ``location``. Multiple variables can be evaluated at a same probing point and values will be written in the same output file. The first column of the file is the time and the subsequent columns are either evaluated scalar fields or components of evaluated vector fields. Here is an example for a 2D simulation:

```
    time      velocity_x   velocity_y     pressure        phase     
0.000000e+00 0.000000e+00  0.000000e+00  0.000000e+00  1.000000e+00 
1.060000e-01 4.922532e-09 -3.504747e-03  1.034432e-02  1.002319e+00 
3.196684e-01 3.963123e-08 -1.155384e-02  9.928013e-03  1.021603e+00 
4.744774e-01 8.129337e-08 -1.736564e-02  4.190060e-03  9.907638e-01 
```

Remote point evaluation is done by the function ``evaluate_values_at_points()`` using Utilities::MPI::RemotePointEvaluation<dim, dim>. It was added to the utilities of Lethe. It is templated for both scalar and vector evaluations.

This implementation introduces a new PostprocessingProbes<dim> class which evaluates quantities of interest at probing points and holds all the data tables of probes. It is instantiated as a member of the MultiphysicsInterface<dim> class. Evaluation of each physics' probed quantities is done within the ``postprocess()`` (or ``postprocess_fd()``) attribute of the physics by communicating to PostprocessingProbes<dim> through the MultiphysicsInterface<dim> object. The probe tables' management (i.e., writing in output file, serializing, deserializing) is handled by the PostprocessingProbes<dim> object.

This implementation required points to describe the location of probes, as a consequence the Parameters::PostProcessing struct needed to be templated with ``dim``.  Appropriate updates due to this change were made in the code.

### Testing

- A unit test was added to test both the scalar and vector evaluation of ``evaluate_values_at_points()``: ``evaluate_values_at_points.cc``
- Two ``lethe-fluid`` application tests were updated to test the new feature (``cls-velocity-extrapolation`` and ``heat_transfer_stefan``).
 
### Documentation

- Doxygen comments were added to the code.
- User documentation on postprocessing parameters was updated. 

- ChatGPT was used for coming up with a simple scalar solution field for the ``evaluate_values_at_points.cc`` unit test.
- Claude was used for debugging purposes during implementation.

At the moment, the probes are only implemented for the following variables: velocity, pressure, phase, and temperature. Other variables will follow in a future PR.

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+         ++variable_it)
+      {
+        const auto &probing_points_of_variable =
+          probing_points_parameters.probing_points_per_variable.at(
+            *variable_it);
+
+        // Go trough tables and prepare them
+        for (unsigned int i = 0; i < probing_points_of_variable.ids.size(); ++i)
+          {
+            TableHandler     &current_table = probe_tables[i];
```
Julamipolymtl: `probe_tables` is indexed by probe ID, but `i` here is a position in the variable's `ids` list, not the probe id itself. For now tests work because probes with more variables come before probes with less varialbes, making `i` fit `id`. 

Ex:
```
probe 0 variables = pressure, temperature, phase
probe 1 variables = pressure, temperature, phase
probe 2 variables = phase
``` 
leading to `pressure ids = {0,1}`, `temperature ids = {0,1}`, `phase ids = {0,1,2}` (So `i=id`). But if we were to define:
```
probe 0 variables = phase
probe 1 variables = pressure, temperature, phase
probe 2 variables = pressure, temperature, phase
```
we get `pressure ids = {1,2}`, `temperature ids = {1,2}`, `phase ids = {0,1,2}`, which no longer fits i for pressure and temperature. This seems to happen at a few places in the code, I'll try to track them down. Quick fix for this loop would be to simply loop on the ids themselves.
```suggestion
        for (const unsigned int &id : probing_points_of_variable.ids)
          {
            TableHandler     &current_table = probe_tables[id];
```

applications_tests/lethe-fluid/cls-velocity-extrapolation.prm
```diff
@@ -153,6 +155,24 @@ subsection post-processing
...
+      set location               = 0,0.01
+      set variable               = pressure, phase, velocity
+      set probing point filename = probe_wave_top
+    end
+    subsection probe 2
+      set location               = -0.85,-0.85
+      set variable               = phase
+      set probing point filename = probe_in_fluid_1
+    end
+  end
```
Julamipolymtl, line 175: This test could switch probe 2 and probe 0 to reproduce the behavior my other comment addressed about having the stricter probes at lower index.
```suggestion
  subsection probing points
    set number of probing points = 3
    subsection probe 0
      set location               = -0.85,-0.85
      set variable               = phase
      set probing point filename = probe_in_fluid_1
    end
    subsection probe 1
      set location               = 0,0
      set variable               = pressure, phase, velocity
      set probing point filename = probe_origin
    end
    subsection probe 2
      set location               = 0,0.01
      set variable               = pressure, phase, velocity
      set probing point filename = probe_wave_top
    end
  end
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,313 @@
...
+        {
+          pcout << "At probe " << id << " ("
+                << Patterns::Tools::Convert<Point<dim>>::to_string(
+                     probing_points_of_the_variable.points[i])
+                << ") - " << table_name << std::endl;
+          pcout << " Evaluated " << variable_name
+                << " value: " << evaluated_value << std::endl;
+        }
+
+      TableHandler &current_table = probe_tables[i];
```
Julamipolymtl: Same issue about indexing probe table with index `i` (ith probe id in the variable's list) vs its probe `id` itself.
```suggestion
      TableHandler &current_table = probe_tables[id];
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,313 @@
...
+                     probing_points_of_the_variable.points[i])
+                << ") - " << table_name << std::endl;
+          pcout << " Evaluated " << variable_name << " vector: "
+                << Patterns::Tools::Convert<Tensor<1, dim, double>>::to_string(
+                     evaluated_value)
+                << std::endl;
+        }
+
+
+      TableHandler &current_table = probe_tables[i];
```
Julamipolymtl: See above.
```suggestion
      TableHandler &current_table = probe_tables[id];
```

applications_tests/lethe-fluid/cls-velocity-extrapolation.prm
Julamipolymtl: Your output would probably need to be updated due to the order change

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+  const std::set<Variable> implemented_scalar_variables =
+    {Variable::pressure, Variable::phase, Variable::temperature};
+
+  /// Set of valid vector variables
+  const std::set<Variable> implemented_vector_variables = {Variable::velocity};
+
+private:
+  /**
+   * @brief Checks if the table of a specified probing point (@p probe_id) has
+   * already a row entry for the current time step.
```
Julamipolymtl:
```suggestion
   * @brief Checks if the table of a specified probing point (@p probe_id)
   * already has a row entry for the current time step.
```

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+  PostProcessing<dim>::ProbingPoints::declare_parameters(ParameterHandler &prm)
+  {
+    prm.enter_subsection("probing points");
+    {
+      prm.declare_entry("number of probing points",
+                        "0",
+                        Patterns::Integer(0),
+                        "Number of probing points");
+
+      const std::string default_point_entry_sting =
```
Julamipolymtl:
```suggestion
      const std::string default_point_entry_string =
```

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+        (dim == 2) ? "0., 0." : "0., 0., 0.";
+
+      for (unsigned int i = 0; i < max_number_of_probing_points; ++i)
+        {
+          prm.enter_subsection("probe " + Utilities::int_to_string(i));
+          {
+            prm.declare_entry(
+              "location",
+              default_point_entry_sting,
```
Julamipolymtl:
```suggestion
              default_point_entry_string,
```

include/solvers/simulation_parameters.h
```diff
@@ -520,6 +520,28 @@ class SimulationParameters
...
+
+    // For postprocessing probes
+    if (post_processing.probing_points.probing_points_per_variable.contains(
+          Variable::velocity) ||
+        post_processing.probing_points.probing_points_per_variable.contains(
+          Variable::pressure))
+      AssertThrow(
+        multiphysics.fluid_dynamics,
+        ExcMessage(
+          "You are attempting to evaluate velocity or pressure vales using at least one postprocessing probe, but 'fluid dynamics = false' in the 'multiphysics' subsection."));
```
Julamipolymtl:
```suggestion
          "You are attempting to evaluate velocity or pressure values using at least one postprocessing probe, but 'fluid dynamics = false' in the 'multiphysics' subsection."));
```

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+    auto keys_range = std::views::keys(probing_points_per_variable);
+
+    for (auto variable_it = keys_range.begin(); variable_it != keys_range.end();
+         ++variable_it)
+      {
+        const auto &probing_points_of_variable =
+          probing_points_parameters.probing_points_per_variable.at(
+            *variable_it);
+
+        // Go trough tables and prepare them
```
Julamipolymtl:
```suggestion
        // Go through tables and prepare them
```

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+              Patterns::List(Patterns::Double()),
+              "Probe point location in the mesh's reference. "
+              "The different components of the point must be separated comma (e.g., ``set location = 0.0, 0.0, 0.0``). ");
```
Julamipolymtl:
```suggestion
              "The different components of the point must be separated by a comma (e.g., ``set location = 0.0, 0.0, 0.0``). ");
```

include/core/multiphysics.h
```diff
@@ -95,4 +95,96 @@ get_physics_id(std::string physics_name)
...
+  else if (variable == Variable::pressure)
+    return "pressure";
+  else if (variable == Variable::phase)
+    return "phase";
+  else if (variable == Variable::temperature)
+    return "temperature";
+  else if (variable == Variable::phase_cahn_hilliard)
+    return "phase cahn hilliard";
+  else if (variable == Variable::chemical_potential_cahn_hilliard)
+    return "chemical potential cahn_hilliard";
```
Julamipolymtl:
```suggestion
    return "chemical potential cahn hilliard";
```

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+   * obtain the current time and current iteration number.
+   * @param[in] probing_points_parameters Set of parameters describing the
+   * specified probing point and variables of interest.
+   * @param[in] output_folder Path to the output folder where the .dat files
+   * are saved.
+   *
+   * @remark At the moment, probing for only the velocity, pressure, phase, and
+   * temperature variables are implemented. Implementation for other variables
+   * will follow.
+   */
```
Julamipolymtl, line 34: Missing documentation for a few parameters (`output_frequency`, `verbosity`)

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+  PostprocessingProbes(std::shared_ptr<SimulationControl> &simulation_control,
+                       const Parameters::PostProcessing<dim>::ProbingPoints
+                                                   &probing_points_parameters,
+                       const double                 output_frequency,
```
Julamipolymtl:
```suggestion
                       const int                 output_frequency,
```

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,281 @@
...
+#include <core/vector.h>
+
+#include <ranges>
+
+template <int dim>
+class PostprocessingProbes
+{
+public:
+  /**
+   * @brief Object that evaluated variables (e.g., velocity) at specified points
```
Julamipolymtl:
```suggestion
   * @brief Object that evaluates variables (e.g., velocity) at specified points
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -425,4 +437,23 @@ This feature can be used to evaluate heights, depths, or lengths when quantities
...
\ No newline at end of file
+* ``bounding box filename``: filename(s) for outputted isocontour(s). When multiple isocontours are defined, the different filenames must be separated by commas (e.g., ``set bounding box filename = interface_bounding_box, solidus_bounding_box, liquidus_bounding_box``) and follow the same order as ``variable`` and ``isovalue``.
+
+Probing points
+~~~~~~~~~~~~~~
+
+The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a `.dat` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe is saved in the same output file. Each scalar field represents a column and the same goes for components of vector fields.
```
Julamipolymtl:
```suggestion
The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a ``.dat`` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe is saved in the same output file. Each scalar field represents a column and the same goes for components of vector fields.
```

Julamipolymtl (changes requested):
First look through. There is an issue that would appear if the first occurrence of a variable was not at probe 0. Otherwise nice implementation!

include/solvers/postprocessing_probes.h
AmishgaAlphonius: Good catch! You are totally right! I made the changes :)

include/core/parameters.h
```diff
@@ -1103,9 +1106,82 @@ namespace Parameters
...
+      /// Number of probing points
+      unsigned int number_of_probing_points;
+
+      /// Maximum number of probing points
+      unsigned int max_number_of_probing_points = 25;
+
+      /// Probing point output names ordered by ID
+      std::vector<std::string> probing_points_output_names;
+
+      /** Map that regroups all probing points information of a same variable
```
hepap:
```suggestion
      /** Map that regroups all probing points information (ID and coordinates) of a same variable
```

include/core/utilities.h
```diff
@@ -1180,4 +1183,102 @@ struct cut_cell_comparison
...
+ *
+ * @param[in] triangulation Triangulation object.
+ * @param[in] mapping Mapping of the domain.
+ * @param[in] dof_handler DoF handler associated to the solution field.
+ * @param[in] solution_field Vector containing the solution field.
+ * @param[in] evaluation_points Vector of points where the values of the
+ * solution field are to be evaluated.
+ * @param[in,out] evaluated_scalar_values Vector of evaluated scalar values.
+ * @param[in] first_selected_component First component of the solution to select
+ * if the solution field is a vector field.
```
hepap: Can you clarify what first_selected_component means?

blaisb:
Ok this is exactly what I needed for another project, you the beast. Will review now.

include/core/multiphysics.h
```diff
@@ -95,4 +95,96 @@ get_physics_id(std::string physics_name)
...
+    return "tracer";
+  else if (variable == Variable::electric_field)
+    return "electric_field";
+  else if (variable == Variable::magnetic_field)
+    return "magnetic_field";
+  else if (variable == Variable::electromagnetic_fields)
+    return "electromagnetic_fields";
+  else
+    AssertThrow(false,
+                dealii::ExcMessage("An unknown Variable string was requested"));
```
blaisb, line 187: Sounds a bit "hardcore" but why not the call the get_variable_string and replace the " " with underscore? This way if you change the name at one place, you change it at the other.


```
#include <algorithm>
#include <string>

std::string s = "hello world foo";
std::replace(s.begin(), s.end(), ' ', '_');
```

include/core/parameters.h
```diff
@@ -1103,9 +1106,82 @@ namespace Parameters
...
+       * @brief IDs and locations associated with probe points of a given
+       * variable.
+       */
+      struct ProbingPointsPerVariable
+      {
+        /// Identifier of the probe points
+        std::vector<unsigned int> ids;
+
+        /// Point locations where we want to evaluate values
+        std::vector<Point<dim>> points;
```
blaisb, line 1127: I just have a quick question. Why do you need an ID for the probing point? Why not instead use the "id" in the std::vector? I don't understand why you need a separate id (thus it is worth documenting why) :)

include/core/utilities.h
```diff
@@ -1180,4 +1183,102 @@ struct cut_cell_comparison
...
+ * @param[in] remote_point_evaluator RemotePointEvaluation object used to
+ * evaluate values.
+ *
+ * @return @p true if all points are within the simulated domain.
+ */
+template <int dim>
+inline bool
+points_are_in_domain(
+  const std::vector<Point<dim>>                         &evaluation_points,
+  const Utilities::MPI::RemotePointEvaluation<dim, dim> &remote_point_evaluator)
```
blaisb, line 1265: Good idea to add such a check, lovely.

include/solvers/multiphysics_interface.h
```diff
@@ -876,8 +890,62 @@ class MultiphysicsInterface
...
+   * @param[in] dof_handler DoF handler associated to the solution field.
+   * @param[in] present_solution Vector containing the scalar solution field.
+   * @param[in] variable Variable of interest.
+   * @param[in] pcout Parallel console output stream.
+   */
+  template <typename VectorType>
+  void
+  postprocess_probes(const Triangulation<dim> &triangulation,
+                     const Mapping<dim>       &mapping,
+                     const DoFHandler<dim>    &dof_handler,
```
blaisb, line 913: You can get the triangulation from the dof_handler (get_triangulation()). So I don't think you need to pass it.

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+   *
+   * @remark At the moment, probing for only the velocity, pressure, phase, and
+   * temperature variables are implemented. Implementation for other variables
+   * will follow.
+   */
+  PostprocessingProbes(std::shared_ptr<SimulationControl> &simulation_control,
+                       const Parameters::PostProcessing<dim>::ProbingPoints
+                                                   &probing_points_parameters,
+                       const unsigned int           output_frequency,
+                       const std::string           &output_folder,
```
blaisb, line 39: Output_Frequency and output_folder are stored inside the simulation_control. You could get them directly from there instead of passing them as arguments.

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+  /**
+   * @brief Evaluates scalar variables (e.g., pressure) at specified points
+   * in the domain using evaluate_values_at_points and fills appropriate probe
+   * tables.
+   *
+   * @tparam VectorType Type of vector of the solution vector.
+   *
+   * @param[in] triangulation Triangulation object.
+   * @param[in] mapping Mapping of the domain.
+   * @param[in] dof_handler DoF handler associated to the solution field.
```
blaisb, line 113: Triangulation here is also a redundant parametr. It is inside the dof handler.

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+  /** Set of parameters describing the specified probing point and variables of
+   * interest */
+  const Parameters::PostProcessing<dim>::ProbingPoints
+    probing_points_parameters;
+
+  /// Postprocessing output frequency
+  const unsigned int output_frequency;
+
+  /// Path to the output folder where all .dat are saved
+  const std::string output_folder;
```
blaisb, line 266: These are both contained in the simulation control. It's not a "bad idea" ot make a const local copy of them, but I would avoid doing that to be honest.

source/core/utilities.cc
```diff
@@ -868,3 +870,189 @@ delete_vtu_and_pvd_files(const std::string &output_path)
         std::filesystem::remove(filename.path());
     }
 }
+
+template <int dim, typename VectorType>
+void
+evaluate_values_at_points(const Triangulation<dim>      &triangulation,
+                          const Mapping<dim>            &mapping,
+                          const DoFHandler<dim>         &dof_handler,
```
blaisb, line 878: Here also the superflous triangulation.

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+         ++variable_it)
+      {
+        const auto &probing_points_of_variable =
+          probing_points_parameters.probing_points_per_variable.at(
+            *variable_it);
+
+        // Go through tables and prepare them
+        for (const unsigned int &id : probing_points_of_variable.ids)
+          {
+            TableHandler     &current_table = probe_tables[id];
```
blaisb, line 71: Something is off here and I think it also stems from this double id you have going on. probe_tables is indexed by global probe ID it is sized number_of_probing_points in the constructor, and both write_probing_points_tables() and gather_tables() pair probe_tables[i] with probing_points_output_names[i] over the full range. But i here is the position within this variable's ids list, not the probe ID. I think it should be be:
`
TableHandler &current_table = probe_tables[probing_points_of_variable.ids[i]];
`
With probe 0 = pressure / probe 1 = velocity, the velocity list is ids = {1}, so i = 0 declares velocity_x/velocity_y on table 0, and table 1 keeps only the time column.
Could you add a case where the probes request different variable lists , like a third probe with only velocity after two pressure-only probes?

More importantly, do we really need this double ID? I am not sure to be honest... It's worth talking about it for sure.

source/core/utilities.cc
```diff
@@ -868,3 +870,189 @@ delete_vtu_and_pvd_files(const std::string &output_path)
...
+                          const VectorType              &solution_field,
+                          const std::vector<Point<dim>> &evaluation_points,
+                          std::vector<double> &evaluated_scalar_values,
+                          const unsigned int   first_selected_component)
+{
+  // Update ghost values of the scalar solution field
+  VectorType solution_field_owned_copy;
+  solution_field_owned_copy.reinit(solution_field);
+  solution_field_owned_copy = solution_field;
+  solution_field_owned_copy.update_ghost_values();
```
blaisb, line 888: I am not sure I understand why this is done this way.
Why not instead ensure that the input of the problem is a ghost vector (locally relevant) instead of making a local copy.
You make solution_field_owned_copy but then you ghost it. Essentially every time you probe you do a vector copy. Can't you just pass / request that the function uses a vector that is already ghosted?

source/core/utilities.cc
```diff
@@ -868,3 +870,189 @@ delete_vtu_and_pvd_files(const std::string &output_path)
...
+
+  // Evaluate scalar values at points using RemotePointEvaluation if they are
+  // found within the domain
+  Utilities::MPI::RemotePointEvaluation<dim, dim> remote_point_evaluator;
+  remote_point_evaluator.reinit(evaluation_points, triangulation, mapping);
```
blaisb, line 893: The call to reinit in remote_point_evaluator is the expensive part.
Yet, as long as the mesh has not been refined, you do not need to reinit it.
What you could do is connect the post-processing tool for probing to the signal of the triangulations and only when the triangulation is refined or altered, this is when you would call reinit.
Maybe this does not need to be done now, but it really worth looking into, because then you would pay the price only once per mesh adaptation for this functionnality.
What is expensive here i think is the bounding box and everything.

tests/core/evaluate_values_at_points.output
blaisb, line 1: The unit test should really be parallel. The challenging thing about this feature is parallelization, make the unit test parallell IMO

tests/core/evaluate_values_at_points.cc
```diff
@@ -0,0 +1,342 @@
...
+      deallog << "************************************************************"
+              << std::endl;
+      deallog << "Running dim == 3, linear functions" << std::endl;
+      test<3>(8, LinearScalarField<3>(), LinearVectorField<3>());
+      deallog << "************************************************************"
+              << std::endl;
+      deallog << "Running dim == 3, cosine functions" << std::endl;
+      test<3>(32,
+              Functions::CosineFunction<3>(1),
+              Functions::CosineGradFunction<3>());
```
blaisb, line 315: This is a  very big mesh for a unit test. it contains 32^3 cells no? Can you amke it smaller. I know you are trying to catch the cosine well, but maybe use non linear elements or assume the cosine is not perfectly caught?

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+    const ConditionalOStream            &pcout,
+    std::vector<Tensor<1, dim, double>> &evaluated_vector_values);
+
+  /**
+   * @brief Writes probe table into output .dat file.
+   */
+  inline void
+  write_probing_points_tables()
+  {
+    if (simulation_control->get_iteration_number() % output_frequency == 0)
```
blaisb, line 163: I would also check if the simulation control is at the end of the simulation. Otherwise, this means that you won't flush the probe on the last time step if hte post-processing output frequency is not 0. I would add a check for fo rthe simualtion control end so that at the last time step this is also flushed out.

blaisb (changes requested):
First round of comments. I have more coming. This is a very nice feature.

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+  void
+  PostProcessing<dim>::ProbingPoints::parse_parameters(ParameterHandler &prm)
+  {
+    prm.enter_subsection("probing points");
+    {
+      number_of_probing_points = prm.get_integer("number of probing points");
+      AssertThrow(
+        number_of_probing_points <= max_number_of_probing_points,
+        ExcMessage(
+          "The current maximum of probing points allowed is 25. Please adjust 'number of probing points'."));
```
blaisb, line 2394: If you can here I would use teh parametr value  instead of hardcoding the number

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+                    add_probing_point(Variable::temperature, id, point);
+                  }
+                else
+                  throw std::invalid_argument(
+                    "Error, the only valid variables for a 'probing point' are: "
+                    "'velocity', 'pressure','phase' and 'temperature'.");
+              }
+
+            const std::string filename = prm.get("probing point filename");
+            probing_points_output_names.emplace_back(filename);
```
blaisb, line 2432: Since every probe gets it's own filename, somebody could put the same filename twice and then the probe files would be silently overwritten.
It's not a big deal, because this is very unlikely, but anyway.
I'm not sure, either I would NOT allow user to specify the probe files and call them instead automatically probe_X or otherwise I would assert very quickly that there is no duplication in the filenames :)

include/solvers/postprocessing_probes.h
blaisb: Or are those teh output freuqency and folder of the post-processing?

blaisb (changes requested):
Some more small comments.
I thought about it  and I think you need to keep the ID, I don't think there's a way around it.

include/core/parameters.h
blaisb: Ok I think I understand why now. But i would still add a bit of a documentation explaining why.

source/core/utilities.cc
blaisb: Actually it is even simpler than that.
If you pass the remote point evaluator or keep it as a member instead of remaking it. You can actually check if it needs to be reinit using this:
```
  if (!remote_point_evaluator.is_ready())
    remote_point_evaluator.reinit(evaluation_points, triangulation, mapping);
```

since each variable has their own point set, you could have one of this class per variable pretty much

include/core/multiphysics.h
```diff
@@ -95,4 +95,96 @@ get_physics_id(std::string physics_name)
...
+    return "tracer";
+  else if (variable == Variable::electric_field)
+    return "electric field";
+  else if (variable == Variable::magnetic_field)
+    return "magnetic field";
+  else if (variable == Variable::electromagnetic_fields)
+    return "electromagnetic fields";
+  else
+    AssertThrow(false,
+                dealii::ExcMessage("An unknown Variable string was requested"));
```
hepap, line 143: would you like to add the variable_string as a attribute of the class and just return it in this function?

include/solvers/multiphysics_interface.h
```diff
@@ -876,8 +890,62 @@ class MultiphysicsInterface
       {
         iphys.second->read_checkpoint();
       }
+    this->probe_postprocessor.read_checkpoint();
   };
 
+  /**
+   * @brief Checks validity of the variable and if valid, calls probe
+   * postprocessing function of the PostprocessingProbes object.
```
hepap:
```suggestion
   * @brief Checks validity of the variable and if valid, calls 
   * postprocess_probes() function for the given variable of the PostprocessingProbes object.
```

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+  /** Contains IDs of probe for which the current time has been added to the
+   * table to avoid duplication of time entries when a probing point contains
+   * multiple variables */
+  std::set<unsigned int> probe_tables_with_current_time;
+
+  /// Tables of probe measures
+  std::vector<TableHandler> probe_tables;
+
+  /// Precision of table entries
+  const unsigned int table_precision = 6;
```
hepap: I would switch this for the log_precision parameter that @blaisb implemented, because maybe in some  situations (like convergence studies), a parameter control on the precision is nice :)

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+
+#include <ranges>
+
+template <int dim>
+class PostprocessingProbes
+{
+public:
+  /**
+   * @brief Object that evaluates variables (e.g., velocity) at specified points
+   * in the domain.
```
hepap, line 20: Can you add the break down of the structure of this object and the probe tables? Like each probe point has an associated table with all requested variables.

release_notes/current/2026_08_11_Alphonius.md
```diff
@@ -0,0 +1,12 @@
...
+  At the moment, the probes are only implemented for the following variables: velocity, pressure, phase, and temperature. Other variables will follow in a future PR.
+  A unit test was added to test both the scalar and vector evaluation of ``evaluate_values_at_points()``: ``evaluate_values_at_points.cc``. Two ``lethe-fluid`` application tests were updated to test the new feature (``cls-velocity-extrapolation`` and ``heat_transfer_stefan``).
+
+  This implementation required points to describe the location of probes, as a consequence the Parameters::PostProcessing struct needed to be templated with ``dim``.  Appropriate updates due to this change were made in the code.
+  [#2090](https://github.com/chaos-polymtl/lethe/pull/2090)
```
hepap, line 12: Is the postprocessing of the probes called at each time step, or it is compatible with the calculation and output frequency you added recently?

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+                    add_probing_point(Variable::phase, id, point);
+                  }
+                else if (variable == "temperature")
+                  {
+                    add_probing_point(Variable::temperature, id, point);
+                  }
+                else
+                  throw std::invalid_argument(
+                    "Error, the only valid variables for a 'probing point' are: "
+                    "'velocity', 'pressure','phase' and 'temperature'.");
```
hepap, line 2551: I don't mind this structure (a table per variable with every probe points), it's just that in the PostProcessingProbes, it the opposite (a table per probe point with every variables). Is there a reason why?

include/solvers/postprocessing_probes.h
AmishgaAlphonius: For the frequency it is the post processing one (I've clarified it in the doc now), but the folder is the one specified in SimulationControl indeed. I've made the changes for the latter now.

I don't think we have a specific output folder parameter for post-processing.

hepap (commented):
First batch while github works haha

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+        (dim == 2) ? "0., 0." : "0., 0., 0.";
+
+      for (unsigned int i = 0; i < max_number_of_probing_points; ++i)
+        {
+          prm.enter_subsection("probe " + Utilities::int_to_string(i));
+          {
+            prm.declare_entry(
+              "location",
+              default_point_entry_string,
+              Patterns::List(Patterns::Double()),
```
Julamipolymtl: It looks like this is never checked against the expected size (`dim`).
```suggestion
              Patterns::List(Patterns::Double(), dim, dim),
```

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+          prm.enter_subsection("probe " + Utilities::int_to_string(id));
+          {
+            const std::string        variables_list = prm.get("variable");
+            std::vector<std::string> variables_vec =
+              Utilities::split_string_list(variables_list);
+
+            const Point<dim> point(
+              value_string_to_tensor<dim>(prm.get("location")));
+
+            for (const std::string &variable : variables_vec)
```
Julamipolymtl, line 2530: Here it would seem that it would be possible to ahve the same variable be defined multiple times in the same list and it seems to lead to a failure in debug. Not a big deal as its unlikely but I don't know if you want to guard against that.
```
set variables = velocity, velocity, velocity 
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -425,4 +437,23 @@ This feature can be used to evaluate heights, depths, or lengths when quantities
...
\ No newline at end of file
+* ``bounding box filename``: filename(s) for outputted isocontour(s). When multiple isocontours are defined, the different filenames must be separated by commas (e.g., ``set bounding box filename = interface_bounding_box, solidus_bounding_box, liquidus_bounding_box``) and follow the same order as ``variable`` and ``isovalue``.
+
+Probing points
+~~~~~~~~~~~~~~
+
+The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a ``.dat`` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe is saved in the same output file. Each scalar field represents a column and the same goes for components of vector fields.
```
Julamipolymtl:
```suggestion
The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a ``.dat`` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe are saved in the same output file. Each scalar field represents a column and the same goes for components of vector fields.
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -425,4 +437,23 @@ This feature can be used to evaluate heights, depths, or lengths when quantities
...
+
+* ``number of probing points``: number of probes defined.
+
+  .. attention::
+    A maximum of 25 probes can be defined.
+
+* ``location``: location of the probing point. Separate components of the points with comma.
```
Julamipolymtl:
```suggestion
* ``location``: location of the probing point. Separate components of the points with commas.
```

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -425,4 +437,23 @@ This feature can be used to evaluate heights, depths, or lengths when quantities
...
+
+* ``variable``: variable or list of variables (separated by comma) to evaluate at the probing location.
```
Julamipolymtl:
```suggestion
* ``variable``: variable or list of variables (separated by commas) to evaluate at the probing location.
```

include/solvers/postprocessing_probes.h
```diff
@@ -0,0 +1,283 @@
...
+    const Triangulation<dim>            &triangulation,
+    const Mapping<dim>                  &mapping,
+    const DoFHandler<dim>               &dof_handler,
+    const VectorType                    &present_solution,
+    const Variable                       variable,
+    const ConditionalOStream            &pcout,
+    std::vector<Tensor<1, dim, double>> &evaluated_vector_values);
+
+  /**
+   * @brief Writes probe table into output .dat file.
```
Julamipolymtl:
```suggestion
   * @brief Writes the probe table into output .dat file.
```

source/core/parameters.cc
```diff
@@ -2335,8 +2337,109 @@ namespace Parameters
...
+
+      for (unsigned int i = 0; i < max_number_of_probing_points; ++i)
+        {
+          prm.enter_subsection("probe " + Utilities::int_to_string(i));
+          {
+            prm.declare_entry(
+              "location",
+              default_point_entry_string,
+              Patterns::List(Patterns::Double()),
+              "Probe point location in the mesh's reference. "
```
Julamipolymtl:
```suggestion
              "Probe point location in the mesh's reference frame. "
```

Julamipolymtl (approved):
I had another go at it, other than knitpicks and unlikely user inputs, this all looks good to me! Nice feature!

include/core/utilities.h
AmishgaAlphonius: I've added: 

> This is used to extract the correct field when the solution vector contains multiple fields (e.g., velocity and pressure are stored in a same solution vector).

include/solvers/postprocessing_probes.h
AmishgaAlphonius: Good idea, it can help with expanding the feature also! :)

include/solvers/postprocessing_probes.h
AmishgaAlphonius: Isn't 
```
        for (const unsigned int &id : probing_points_of_variable.ids)
          {
            TableHandler     &current_table = probe_tables[id];
```

equivalent to  `TableHandler  &current_table = probe_tables[probing_points_of_variable.ids[i]];` ?

I've added comments to make it less confusing.

I've adjusted the `cls-velocity-extrapolation` application test so that the last probe only has the velocity field like you requested to capture better possible issues :)

include/solvers/postprocessing_probes.h
AmishgaAlphonius: For the output folder, I'm keeping a constant copy of the string here to avoid making a new copy every time I request it with `get_output_path()`.

(For the frequency, it's the post-processing one:))

include/solvers/postprocessing_probes.h
AmishgaAlphonius: Nice thinking! I also thought about it. 
Actually, regarding this, I previously opened an issue https://github.com/chaos-polymtl/lethe/issues/2039 there is no user-prescribed precision value for postprocessing, but only one for the console output (which would affect other values like the printed CFL, where you don't really need 8 digits).

I'm adding an in-code `todo` refering to the issue to solve this.

release_notes/current/2026_08_11_Alphonius.md
AmishgaAlphonius: It is compatible with the `calculation_frequency` specified in the postprocessing parameters since it is called through the postprocess_fd() and postprocess() methods of physics. :)

https://github.com/chaos-polymtl/lethe/blob/067e14f65baf3c3965f3d546d14109a50d462a0b/include/solvers/navier_stokes_base.h#L204-L223

I'm adding a sentence here to specify it.

source/core/parameters.cc
AmishgaAlphonius: Good catch! There is an assertion now :)

source/core/parameters.cc
AmishgaAlphonius: I've changed it now for a combination of both "probe_XX" (with 'XX' being the padded ID number) and the user specified name: `probe_XX_filename`. This way it is unique and easy for the user to find the files :)

include/core/utilities.h
```diff
@@ -1180,4 +1183,103 @@ struct cut_cell_comparison
...
+ * @param[in] mapping Mapping of the domain.
+ * @param[in] dof_handler DoF handler associated to the solution field.
+ * @param[in] solution_field Vector containing the vector solution field.
+ * @param[in] evaluation_points Vector of points where the values of the
+ * solution field are to be evaluated.
+ * @param[in,out] evaluated_vector_values Vector of evaluated vector values.
+ * @param[in] first_selected_component First component of the solution to
+ * select. This is used to extract the correct fields when the solution vector
+ * contains multiple fields (e.g., velocity and
+ * pressure are stored in a same solution vector).
```
hepap:
```suggestion
 * select. This is used to extract the correct fields when the solution vector
 * contains multiple fields (e.g., velocity and pressure are stored in a same
 * solution vector).
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,295 @@
...
+
+  const double current_time = simulation_control->get_current_time();
+
+  if (verbosity == Parameters::Verbosity::verbose)
+    announce_string(pcout, variable_name + " probes");
+
+  // For each probing point fill the correct table
+  for (unsigned int i = 0; i < probing_points_of_the_variable.ids.size(); ++i)
+    {
+      const unsigned int &id = probing_points_of_the_variable.ids[i];
```
hepap, line 84:
```suggestion
      // Get the ID of the probe
      const unsigned int &id = probing_points_of_the_variable.ids[i];
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,295 @@
...
+        {
+          pcout << "At probe " << id << " ("
+                << Patterns::Tools::Convert<Point<dim>>::to_string(
+                     probing_points_of_the_variable.points[i])
+                << ") - " << table_name << std::endl;
+          pcout << " Evaluated " << variable_name
+                << " value: " << evaluated_value << std::endl;
+        }
+
+      TableHandler &current_table = probe_tables[id];
```
hepap, line 102:
```suggestion
      // Access the table associated to the probe
      TableHandler &current_table = probe_tables[id];
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,295 @@
...
+                << Patterns::Tools::Convert<Point<dim>>::to_string(
+                     probing_points_of_the_variable.points[i])
+                << ") - " << table_name << std::endl;
+          pcout << " Evaluated " << variable_name << " vector: "
+                << Patterns::Tools::Convert<Tensor<1, dim, double>>::to_string(
+                     evaluated_value)
+                << std::endl;
+        }
+
+      TableHandler &current_table = probe_tables[id];
```
hepap, line 231:
```suggestion
      // Access the table associated to the probe
      TableHandler &current_table = probe_tables[id];
```

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,295 @@
...
+
+  const double current_time = simulation_control->get_current_time();
+
+  if (verbosity == Parameters::Verbosity::verbose)
+    announce_string(pcout, variable_name + " probes");
+
+  // For each probing point fill the correct table
+  for (unsigned int i = 0; i < probing_points_of_the_variable.ids.size(); ++i)
+    {
+      const unsigned int &id = probing_points_of_the_variable.ids[i];
```
hepap, line 210:
```suggestion
      // Get the ID of the probe
      const unsigned int &id = probing_points_of_the_variable.ids[i];
```

include/core/parameters.h
```diff
@@ -1103,9 +1106,96 @@ namespace Parameters
       parse_parameters(ParameterHandler &prm);
     };
 
+    /**
+     * @brief Probing points parameters. This is used to monitor the evolution
+     * of variables (e.g., temperature) at specified points in the domain.
+     *
+     * @remark The points defined must be part of the domain.
+     */
```
hepap, line 1129: I think that additional information on the struct could help because it is not just the Probing points parameters, it's the Probing points themselves (like I don't think the coords or ids are stored anywhere else). It would help with the readability

source/core/utilities.cc
```diff
@@ -868,3 +870,181 @@ delete_vtu_and_pvd_files(const std::string &output_path)
...
+                          const unsigned int   first_selected_component)
+{
+  // Get triangulation
+  const auto &triangulation = dof_handler.get_triangulation();
+
+  // Check if solution vector is relevant
+  assert_vector_has_ghost_elements<VectorType>(
+    solution_field,
+    dof_handler.get_mpi_communicator(),
+    "evaluate_values_at_points");
```
hepap: Usually, dealii checks if the vector is in a readable state, do you need to explicitly check it? Or you can just call update_ghost_values() instead of throwing.

hepap (commented):
Here are some additional comments, I'm not completely done yet :)

source/core/utilities.cc
AmishgaAlphonius: I thought the same, but it does not always work. For TrilinosWrappers::MPI::Vector, `update_ghost_values()` does not actually update the ghost element solutions, it is a dummy function. 

Image

I tested using the unit test with 2 processes and nothing actually threw (in Release) even if the computation was wrong for some evaluation points. This is why I added this function.

source/core/utilities.cc
AmishgaAlphonius: This is great, I made the change! :)

source/core/utilities.cc
blaisb: Yeah but it will never throw in release. These things are Assert and not AssertThrow. You need to test in Debug if you want to know what happens.

AmishgaAlphonius: Exactly, so I implemented a function that has an AssertThrow so that it is easier (quicker) to catch. Additionally, I prefer having the entering solution vector as a const reference rather than allowing it to be modified within the `evaluate_values_at_points` method. This is why I added the AssertThrow function rather than updating the ghost elements inside the method when required.

blaisb: As long as you don't AssertThrow in a loop (which is expensive) things should be fine.

include/core/vector.h
```diff
@@ -4,6 +4,8 @@
 #ifndef lethe_vector_h
 #define lethe_vector_h
 
+#include <core/utilities.h>
+
```
blaisb: Is this include needed? Because if that's the case this will bring utilities in so many places and will have a relatively important impact on compilation efficiency.

I think the two real includes you wanted to use are  deal.II/base/mpi.h and deal.II/base/exceptions.h

tests/core/evaluate_values_at_points.cc
```diff
@@ -0,0 +1,395 @@
...
+}
+
+
+int
+main(int argc, char *argv[])
+{
+  try
+    {
+      mpi_initlog();
+      Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
```
blaisb: The mpi_initlog() should be after the MPI_InitFinalize and not before

tests/core/evaluate_values_at_points.cc
```diff
@@ -0,0 +1,395 @@
...
+                           solution_field_scalar_owned);
+  solution_field_scalar_relevant = solution_field_scalar_owned;
+
+
+  TrilinosWrappers::MPI::Vector solution_field_vector_owned(
+    dof_handler_vector.locally_owned_dofs(), mpi_communicator);
+  TrilinosWrappers::MPI::Vector solution_field_vector_relevant(
+    dof_handler_vector.locally_owned_dofs(),
+    DoFTools::extract_locally_relevant_dofs(dof_handler_vector),
+    mpi_communicator);
```
blaisb: Can you use a GlobalVecrorType here? Otherwise the test will automatically not work if Lethe uses LDV

source/solvers/postprocessing_probes.cc
```diff
@@ -0,0 +1,309 @@
...
+  std::string prefix = output_folder;
+  std::string suffix = ".checkpoint";
+
+  for (unsigned int i = 0; i < probe_tables.size(); ++i)
+    {
+      const auto &filename =
+        probing_points_parameters.probing_points_output_names[i];
+
+      table_output_structs.emplace_back(probe_tables[i],
+                                        prefix + filename + suffix);
```
blaisb, line 315: I think there is an issue right?
Does filename contain the "number" of the probe here or not? Like should it? Because otherwise all of the probes will be checkpointed under the same name if they share the same filename (do they by default?) Can you double check this part.

include/core/utilities.h
```diff
@@ -1180,4 +1183,112 @@ struct cut_cell_comparison
...
+ *
+ * @return @p true if all points are within the simulated domain.
+ */
+template <int dim>
+inline bool
+points_are_in_domain(
+  const std::vector<Point<dim>>                         &evaluation_points,
+  const Utilities::MPI::RemotePointEvaluation<dim, dim> &remote_point_evaluator)
+{
+  if (!remote_point_evaluator.all_points_found())
```
blaisb, line 1277: Is it possible that this place is reached without remote_point_evaluator having been reinited? If not, no worries then, if yes, then I would try to defend from that.

include/core/multiphysics.h
```diff
@@ -95,4 +97,77 @@ get_physics_id(std::string physics_name)
...
+ *  - @p "tracer"
+ *  - @p "electric_field"
+ *  - @p "magnetic_field"
+ *  - @p "electromagnetic_fields"
+ */
+inline std::string
+get_variable_string_with_underscores(Variable variable)
+{
+  std::string variable_string = get_variable_string(variable);
+  std::ranges::replace(variable_string, ' ', '_');
```
blaisb, line 169: std::ranges is not included in string. On my machine this did not compile because of that. It is defined in header <algorithm>(https://en.cppreference.com/cpp/header/algorithm)

blaisb (changes requested):
Some additional changes. Can you double check this prefix thing (is the number of the probe contained in the filename we put in, or is the number of the probe appended automatically by the code? This seems to be confused right now.

include/core/utilities.h
AmishgaAlphonius: Good point, I've added the reinitialization check within the method now.

source/solvers/postprocessing_probes.cc
AmishgaAlphonius: Nice catch! I changed it now! :)

doc/source/parameters/cfd/post_processing.rst
```diff
@@ -425,4 +437,23 @@ This feature can be used to evaluate heights, depths, or lengths when quantities
...
\ No newline at end of file
+* ``bounding box filename``: filename(s) for outputted isocontour(s). When multiple isocontours are defined, the different filenames must be separated by commas (e.g., ``set bounding box filename = interface_bounding_box, solidus_bounding_box, liquidus_bounding_box``) and follow the same order as ``variable`` and ``isovalue``.
+
+Probing points
+~~~~~~~~~~~~~~
+
+The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a ``.dat`` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe are saved in the same output file. Each scalar field represents a column and the same goes for components of vector fields.
```
hepap:
```suggestion
The ``probing points`` subsection defines probes that can evaluate ``variable`` (e.g., velocity) values at remote locations of the domain throughout the simulation. Each probe has its own subsection ``probe x``, where ``x`` is the ID of the probe ranging from ``0`` to  ``number of probing points - 1``. The monitored time-evolving values are saved in a ``.dat`` file (``probing point filename``). Each probe has its own ``location``, list of variables and output file. A given probe can evaluate multiple variables as specified below. All variable values of a same probe are saved in the same output file. Each scalar field represents a column and the same goes for each component of vector fields.
```

hepap (approved):
One small thing :) Great tool!

blaisb (approved):
GOOD TO GO READY FOR LE MERGEGEGGEGEGE

PR #2102, opened 2026-08-24
CG iterative solver initial guess implementation

OresteMarquis:
### Description
 
The CG solver used to solve the time-harmonic Maxwell equations requires an initial guess before starting the convergence process. Before this PR, the initial guess was always a zero vector, which had no significant impact when solving the physics only once. However, in coupled multiphysics simulations that require multiple electromagnetic solves, the solution from the previous solve can provide a better initial guess for the next solve, particularly when the electromagnetic properties have not changed significantly between iterations. This can reduce the number of iterations required for convergence.

### Testing

Using a test .prm file I could see that this new approach changes the number of iteration from 744 to 268. 

### Documentation

No documentation need to be added. 

### Use of LLMs  (Large Language Models)

No LLM were used for this PR.

Code related list:
- [x] An entry describing the change has been added in `/release_notes/current/` following the instructions of `/release_notes/template.md`

Pull request related list:
- [x] Labels are applied

### DIFF
--- a/source/solvers/time_harmonic_maxwell.cc
+++ b/source/solvers/time_harmonic_maxwell.cc
@@ -2057,8 +2057,6 @@ TimeHarmonicMaxwell<dim>::solve_linear_system()
     ExcMessage(
       "The time harmonic electromagnetism physics does not support a preconditioner. Please set: subsection linear solver -> subsection electromagnetics -> set preconditioner = none."));
 
-  auto mpi_communicator = this->triangulation->get_mpi_communicator();
-
   // Define the linear solver tolerance
   const double absolute_residual =
     simulation_parameters.linear_solver.at(PhysicsID::electromagnetics)
--- a/source/solvers/time_harmonic_maxwell.cc
+++ b/source/solvers/time_harmonic_maxwell.cc
@@ -1668,6 +1668,10 @@ TimeHarmonicMaxwell<dim>::setup_dofs()
   this->system_rhs.reinit(this->locally_owned_dofs_trial_skeleton,
                           mpi_communicator);
 
+  // We reinitialize the initial guess vector
+  this->initial_guess_iterative_solver.reinit(
+    this->locally_owned_dofs_trial_skeleton, mpi_communicator);
+
   // Define constraints
   define_constraints();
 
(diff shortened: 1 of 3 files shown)


source/solvers/time_harmonic_maxwell.cc
```diff
@@ -2080,12 +2082,10 @@ TimeHarmonicMaxwell<dim>::solve_linear_system()
                                linear_solver_tolerance);
 
   // Solve
-  GlobalVectorType completely_distributed_solution(
-    this->locally_owned_dofs_trial_skeleton, mpi_communicator);
   TrilinosWrappers::SolverCG solver(solver_control);
 
   solver.solve(this->system_matrix,
-               completely_distributed_solution,
+               this->initial_guess_iterative_solver,
```
blaisb, line 2088: Would it not be better instead to give the present solution as the initial_guess_iterative_solver here instead of assuming it already has it inside
So many things could go wrong right now in the implementation
- What if I restart (initial guess is zero)
- What if I adapt the mesh (initial guess is zero)
- and so on and so forth
I think it would be better to just start with initial_guess = present_solution and then use it to solve. What od you think?

blaisb (changes requested):
There's a bit of a pickle in the way the initial guess is managed IMO

source/solvers/time_harmonic_maxwell.cc
OresteMarquis: It was on purpose that the initial guess for restart and mesh adaptation was zero. I thought it would not be the end of the world to do it like this to avoid checkpointing or projecting the initial_guess. It is not possible to use the present_solution directly (at least for all multiphysic simulation) because the solution vector is rescaled after being solved to recover the units defined by the user. The rescaled solution is generally a bad initial guess.

In my opinion, we have 2 other options besides the one I implemented here:
- We create another function that does the inverse of scaling the solution, and we use it to remake the skeleton_solution into a dimensionless form. Personally, I don't like this one; it seems rather expensive vs only storing a vector.
- We could only rescale the interior solution and never the skeleton one, since it's only the interior solution that is output in the .vtu and in multiphysics coupling. This would work but would make it confusing, as the different parts of the solution (skeleton and interior) would be in different states.

blaisb: Yeah you are right, good arguments.
Thenlet's do this.
We don't checkpoint it, it's just an iterative solver anyway. However, we document clearly that the iteration count may change when restarting due to this feature. Is it a good compromise?

OresteMarquis: Where shoud I document it? In the linear solver section of the documentation?

blaisb: In the  brief of the solver linear system for this physics. It's not so much a user facing detail, just something so that we realise what's the impact. It belongs more in the doxygen than anywhere else

blaisb (approved):

PR #2103, opened 2026-08-24
Set precision in the save function

OGaboriault:
### Description

The precision used to checkpoint the simulation control object was limited to five decimal places. As a result, when using time steps smaller than this precision, the time stored in the checkpoint file could be significantly rounded. 

### Solution

The `std::setprecision(17)` function is now used in the save function. 

### Testing

I have run a application test with this change and the number of decimal places was coherent with the change.  

### Documentation

No. 

ChatGPT was asked what precision should be used. The answer was 17.

Code related list:

Pull request related list:
- [X] Labels are applied
- [X] There are at least 2 reviewers (or 1 if small feature)
- [X] The PR description is clean and is ready to be used as the commit message when merging the PR


source/core/simulation_control.cc
```diff
@@ -209,6 +209,11 @@ SimulationControl::save(const std::string &prefix)
   std::string   filename = prefix + ".simulationcontrol";
   std::ofstream output(filename.c_str());
   output << "Simulation control" << std::endl;
+
+  // We se the precision to 17. If the time-step is significantly smaller then
+  // the end time of the simulation, round off errors can become problematic.
```
blaisb:
```suggestion
  // We set the precision to 17. If the time-step is significantly smaller then
  // the end time of the simulation, round off errors can become problematic.
```

Just a small thing

blaisb (approved):
Small thing to fix then good to go!

PR #2117, opened 2026-09-10
Prevent pressure rescaling in matrix-free

blaisb:
### Description

The pressure rescaling approach is used in the matrix-based approach to improve the condition number of the linear solver. But it is not implemented in the matrix-free solver. Yet right now it could still be used, but this would lead to a wrong pressure update.

### Solution

Added adequate assertions to prevent this option from being used at all when we are using a mtrix-free solver. This is done by asserting that the value is not different from unity.

### Testing

Tested locally on my machine, but no need for unit testing in this regard.

### Documentation

I don't think this is needed since that option is not something we wish to expose a lot to the users anyway.

None except spelling mistake.

Code related list:
- [ ] An entry describing the origin of the bug as well as its fix has been added in `/release_notes/current/` following the instructions of `/release_notes/template.md`

Pull request related list:
- [ ] Labels are applied
- [ ] The PR description is clean and is ready to be used as the commit message when merging the PR


Julamipolymtl (approved):
Pretty straight forward. Nice catch!

PR #131, opened 2021-02-03
Moving sharp ib v2

Luckabarbeau:
Add the capacity to simulate moving IB particles using an implicit Euler shame. Add modification to the linear and non-linear solver to make the process more robust.

PR #2125, opened 2026-09-15
Report the real reason a linear solve failed

karlwilson:
### Description

When a linear solve failed, every affected application aborted with the single word `std::exception` and no further information.

The exception handlers around the linear solve caught the exception by reference but re-threw it by value:
```cpp
catch (std::exception &e)
  {
    ...
    throw e;      // copies only the std::exception subobject
  }
```

`throw e;` copy-initializes a new exception from the static type of `e`, so the object was sliced down to its `std::exception` base, losing all useful information : `dealiii::ExceptionBase`'s file, line and message, and `SolverControl::NoConvergence::last_step` and `last_residual`. The user received no information about what went wrong with the linear solver.

Affected handlers:

- `FluidDynamicsMatrixBased::solve_system_GMRES`, `solve_L2_system`, `solve_system_BiCGStab`
- `FluidDynamicsMatrixFree::solve_system_GMRES`
 
Long standing issue: first introduced in [#131](https://github.com/chaos-polymtl/lethe/pull/131) (2021/03/04), and reproduced in the matrix-free GMRES handler in [#836](https://github.com/chaos-polymtl/lethe/pull/836) (2023/08/15).

### Solution

Permanent fix, in 3 parts.

**Re-throw original exception**: `throw e;` becomes `throw;`, which re-throws the exception object unchanged. The handler still catches `const std::exception &`. A `dynamic_cast` to `SolverControl::NoConvergence` recovers the iteration and residual when the failure is a non-convergence, and each handler now prints them next to the tolerance the solver was aiming for.

**Pulled the preconditioner validation out of the `try` block**, so an unsupported preconditioner is reported as a configuration error instead of being mistaken for a solver failure.

**Report why BiCGStab failed**: `SolverControl::NoConvergence` is raised both for genuine non-convergence and for an algorithmic breakdown. The distinction matters because a breakdown is not fixed by a larger iteration budget or a higher preconditioner fill level. The two solvers need different treatments:

- *Matrix-based (Trilinos)*: AztecOO returns control as soon as it breaks down, and `trilinos_solver.cc` then calls `solver_control.check(solver.NumIters(), solver.TrueResidual())` once. An unexhausted iteration budget is then reported as indicating a breakdown.
- *Matrix-free (deal.II)*: `SolverBicgstab::iterate()` returns on breakdown with its state still set to `SolverControl::iterate`, and `solve()` loops `while (state.state == SolverControl::iterate)`, so it silently restarts rather than terminating. `NoConvergence` is only thrown once `SolverControl::check` returns `failure`, i.e. at `step >= maxsteps` or on a NaN residual. A real breakdown therefore always exhausts the budget and is indistinguishable from slow convergence.

Both solvers additionally report a non-finite residual explicitly.

The matrix-free timer section is now held by a `TimerOutput::Scope` so it is closed on every path out of the solve, including the re-throw.

### Testing

This has been tested locally with non-converging simulations. No new application test.

### Documentation

Updated `doc/source/parameters/cfd/linear_solver_control.rst` to reflect the more detailed linear solver failure output messages.

Claude Code (Opus 5) was used throughout this PR to identify the concerned files, the various linear solver failure scenarios and to implement the fixes. It was also used for an initial review of the PR and to produce parts of its description. I have reviewed and verified all LLM-generated content, and I accept responsibility for the proposed changes.

The first release-note entry was tagged MAJOR, as it targets the linear solvers used by the main Lethe applications, but perhaps it should be changed to MINOR based on the indications from `release_notes/template.md`

Code related list:

Pull request related list:
- [x] Labels are applied


source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -4022,92 +4035,112 @@ FluidDynamicsMatrixFree<dim>::solve_system_BiCGStab(
...
+        // of poisoning the solution. The l2_norm() call is collective and
+        // triggers the MPI reduction required for a global check across all
+        // processes.
+        if (!std::isfinite(this->newton_update.l2_norm()))
+          {
+            this->pcout
+              << " The Newton update contains non-finite values; resetting it to zero."
+              << std::endl;
+            this->newton_update = 0.0;
+          }
```
blaisb: I don't see what's the point of this
If you reset the Newton update to zero, you will essentially stagnate the whole solver. There's no point in this because you will get a zero update and just... continue to zero update., I don't think this is a good idea to do this. I would remove this check to be honest
If it NaNs, let it Nans and let it crash. Otherwise it will be impossible to debug.,

source/solvers/fluid_dynamics_matrix_free.cc
```diff
@@ -3887,92 +3891,101 @@ FluidDynamicsMatrixFree<dim>::solve_system_GMRES(const double absolute_residual,
...
```
blaisb: I don't see what's the point of this
If you reset the Newton update to zero, you will essentially stagnate the whole solver. There's no point in this because you will get a zero update and just... continue to zero update., I don't think this is a good idea to do this. I would remove this check to be honest
If it NaNs, let it Nans and let it crash. Otherwise it will be impossible to debug.,

blaisb (changes requested):
Just something to change with this is_finite check that fixes the newton update to zero. l would remove that alltogether it's not a good idea.

OresteMarquis (approved):
Nothing much to say, its pretty clear! Go to go for me.

karlwilson:
Merged Bruna's fix for resetting the Netwton update. Changed `fluid_dynamics_matrix_based.cc:1784` from this merge from `throw e;` to `throw;`, as was done in the rest of this PR

blaisb (approved):
Great
Thanks!

PR #2126, opened 2026-09-15
DEM periodic particle-particle refactoring

OGaboriault:
### Description

The periodic particle-particle fine search used to find the nearest periodic image of a particle by scanning a combinatorial list of periodic translations (`combined_periodic_offsets`, up to 26 entries in 3D) and picking the closest one. Since periodic directions are always axis-aligned and independent (perpendicular to each other), that's overkill: the nearest image can be computed directly, one direction at a time, with the minimum image convention. This is now done in `nearest_periodic_translation()` in the `particle_particle_fine_search.cc` file, a fixed amount of work (N scan per number of periodic boundary condition) per pair instead of a variable-length scan.

On top of that, for a pair that's already an established periodic contact, the fine search now reuses the periodic translation found on the previous call instead of recomputing it from scratch every time. It only falls back to the full minimum image computation if that cached translation no longer puts the pair within the neighborhood threshold (contact ending, or the rare case where the nearest image changed). Positions are still fetched fresh every call, so this is bit-identical to a full recomputation whenever the cache is still valid.

To support this, `PeriodicBoundariesManipulator::combined_periodic_offsets` (a `std::vector<Tensor<1, dim>>`) is replaced by `periodic_offset_per_direction`, a `std::array<double, dim>` holding the signed period of the domain per direction (0 if that direction isn't periodic). Same rename/simplification in `DEMContactManager`. Building this array by summing offsets per direction only works if there's at most one periodic boundary pair per direction, so `set_periodic_boundaries_information()` now throws if two pairs share a direction.

This is all internal and behavior-preserving, no impact on simulation results. I measured a ~1-2% speed up with these changes.

### Testing

- Replaced the unit test `tests/dem/combined_periodic_offsets` with `tests/dem/periodic_offset_per_direction`: instead of just printing every combinatorial offset, it now drives the real `set_periodic_boundaries_information` + `map_periodic_cells` path on a periodic hyper_cube and checks the resulting `periodic_offset_per_direction`.
- Removed `applications_tests/lethe-particles/multiperiodic_anisotropic_domain_2d.prm`. It was a regression test for a bug specific to the old combinatorial offset construction (order-dependent, anisotropic-magnitude edge case). That whole bug class is structurally impossible with the new per-direction computation, so the test isn't needed anymore.
- Updated comments in `multiperiodic_single_axis_collisions_3d.prm`, `multiperiodic_edge_contact_3d.prm` and `multiperiodic_parallel_contact_2d.prm` to describe what they cover in terms of the new `periodic_offset_per_direction`/minimum image convention instead of the old combined-offset construction. No physics/parameter changes in these.
- Fixed `multiperiodic_parallel_contact_2d`'s golden file: it was named `.output` (implying 1 rank) but the test is meant to run with `mpirun=2` (the test description was saying that the whole point of the test was catching the periodic contact across the rank boundary). Renamed it to `multiperiodic_parallel_contact_2d.mpirun=2.output`. 

### Documentation

Didn ot touch the documentation. 

### Use of LLMs  (Large Language Models)

Claude Code was used substantially here: it proposed and implemented the minimum-image-convention rewrite of the fine search and the periodic-offset caching, and wrote/updated the associated tests. I reviewed what Claude code proposed, rewrote comments to make the code readable.

Claude Code also identified a bug in the code regarding periodic corners. The force is double when the connected corners are linked by two different ranks. This causes the particle to be at the same time in the local_ghost and ghost_local containers. Thus, the force is applied twice on the particle. Another branch is already open regarding this. I will work on it after this PR is merged, since the application test results will be affected by this change. 

Code related list:
- [x] An entry describing the change has been added in `/release_notes/current/` following the instructions of `/release_notes/template.md`

Pull request related list:
- [x] Labels are applied


applications_tests/lethe-particles/multiperiodic_anisotropic_domain_2d.prm
OGaboriault, line 1: I think we could remove the `applications_tests/lethe-particles/multiperiodic_collision_2d.prm` application test, since they are basicly the same... only the triangulation changes (hyper rectangle vs hyper_square)

OGaboriault: also, this test doesn't have a .ouput

applications_tests/lethe-particles/multiperiodic_parallel_contact_2d.prm
OGaboriault, line 6: The test is designed to be run on 2 ranks, but is ran on 1 rank...

applications_tests/lethe-particles/multiperiodic_parallel_contact_2d.mpirun=2.output
OGaboriault, line 1: The test description was saying that this test is suppose to be ran with 2 rank, but was only being run on one...

blaisb: Well if the output was mpirun=2, it was ran on two no? or you changed the file name?

OGaboriault: the output did not have mpirun=2... I have change the output file name.

blaisb:
@jatred and @karlwilson if you could look at this before I review it

include/dem/particle_particle_fine_search.h
```diff
@@ -45,8 +45,12 @@ using adjacent_pairs_for_contact_t = std::conditional_t<
...
  * contact pair candidates.
  * @param neighborhood_threshold A value which defines the neighbor particles.
- * @param combined_periodic_offsets A vector of tensors of the periodic offsets
- * to change the location of the particles crossing periodic boundaries.
+ * @param periodic_offset_per_direction An array whose component d holds the
+ * signed period of the domain along direction d (0 if d is not periodic),
+ * used to find the nearest periodic image of particles crossing periodic
+ * boundaries via the minimum image convention. This is a plain array of
+ * independent scale factors, not a Tensor, since it is only ever used
+ * component-wise.
```
blaisb: This feels a lot like an AI documentation.
You don't need to say what it is not, but what it is.
```suggestion
 * @param periodic_offset_per_direction An array whose component d holds the
 * signed period of the domain along direction d (0 if d is not periodic),
 * used to find the nearest periodic image of particles crossing periodic
 * boundaries via the minimum image convention. This is a plain array of
 * independent scale factors for each periodic component.
```

include/dem/periodic_boundaries_manipulator.h
```diff
@@ -211,10 +197,14 @@ class PeriodicBoundariesManipulator
   std::unordered_map<types::boundary_id, Tensor<1, dim>> periodic_offsets;
 
   /**
-   * @brief Storage for all 9 (2D) or 27 (3D) precomputed periodic translation
-   * vectors. Calculated from periodic_offsets.
+   * @brief Signed period of the domain along every direction. Component d is
```
blaisb: What does it mean period? You mean periodic distance I guess?

source/dem/particle_particle_fine_search.cc
```diff
@@ -12,20 +12,89 @@
...
+   * particle two.
+   * @param periodic_offset_per_direction An array whose component d holds
+   * the signed period of the domain along direction d (0 if d is not
+   * periodic).
+   *
+   * @return The nearest translation, and whether any periodic direction
+   * required a nonzero translation. A pair whose nearest image requires no
+   * translation on any periodic direction is not a periodic contact (it is
+   * already handled by the non-periodic contact types).
+   */
```
blaisb: Why is there a brief in the .cc file? I don't get it?
This function is only declared in the .cc file i presume, but we don't do that in Lethe. We always declare the prototype in the .h for documentation purpose.

source/dem/particle_particle_fine_search.cc
```diff
@@ -12,20 +12,89 @@
...
+#include <array>
+#include <cmath>
+#include <limits>
 #include <unordered_map>
+#include <utility>
 
 using namespace dealii;
 
+namespace
+{
```
blaisb: What is this unnamed namespace here?

blaisb (changes requested):
A few things. Some .cc defined function. Furthermore, could you double check some of the comments. They are super verbose (clearly AI) and are not really that useful.

blaisb:
@OGaboriault I am working on a PR to fix that test that is unstable

blaisb (approved):
Ok on my side, but I will wait for the others to review before merging

applications_tests/lethe-particles/multiperiodic_edge_contact_3d.prm
```diff
@@ -4,17 +4,17 @@
...
-# +/-z), i.e. crossings of the second AND third registered periodic
-# pairs simultaneously, with no x-crossing involved. The y-z combined
-# offset is precisely one of the entries omitted by the original buggy
-# compute_combined_periodic_offsets, which only generated combined offsets
-# that involved the first registered pair (x).
+# unit cube: their nearest image requires a translation on BOTH the y and z
+# directions simultaneously (0, +/-y, +/-z), i.e. crossings of the second AND
+# third registered periodic pairs at once, with no x-crossing involved. This
+# exercises the minimum image convention's per-direction independence
+# (see particle_particle_fine_search): each periodic axis must produce its
```
jatred, line 11: What's the minimum image convention? You reference it a couple of times in this PR and refer to particle_particle_fine_search here, but I have not found an explanation in any file and I'm still not sure what it means concretely.

applications_tests/lethe-particles/multiperiodic_single_axis_collisions_2d.prm
```diff
@@ -1,13 +1,15 @@
...
 # Two stationary particles are placed in slight overlap across the
-# x-periodic boundary, in a fully periodic unit square. The y-periodic BC
-# is registered first and the x-periodic BC second, so a correct contact
-# requires the pure +/- offset of the SECOND registered pair to be present
-# in combined_periodic_offsets. The Hertzian repulsion pushes them apart along x.
+# x-periodic boundary, in a unit square that is periodic in both x and y.
+# periodic_offset_per_direction must hold a correct, independent, nonzero
+# entry for x (not just for y) for the minimum image convention to detect
+# this contact -- a regression that dropped or zeroed either direction's
+# offset would leave this contact undetected. The Hertzian repulsion pushes
```
jatred, line 11: I thought the offsets were calculated only once at the beginning with the id and direction in the .prm, in what case would you have "a regression that dropped or zeroed either direction's offset"?

Is it linked to the minimum image convention?

applications_tests/lethe-particles/multiperiodic_single_axis_collisions_3d.prm
```diff
@@ -1,14 +1,15 @@
...
-# z-periodic boundary, in a fully periodic unit cube. The periodic BCs are
-# registered in the order x, y, z so that a correct contact requires the
-# pure +/- offset of the THIRD registered pair to be present in
-# combined_periodic_offsets, which was the case missed most aggressively
-# by the original implementation bug. The Hertzian repulsion pushes them apart along z.
+# z-periodic boundary, in a fully periodic unit cube (periodic in x, y and
+# z). periodic_offset_per_direction must hold a correct, independent,
+# nonzero entry for z alongside the x and y ones for the minimum image
+# convention to detect this contact -- a regression that dropped or zeroed
+# any one direction's offset would leave this contact undetected. The
```
jatred, line 11: Same comment as above here

include/dem/dem_contact_manager.h
```diff
@@ -188,16 +188,14 @@ class DEMContactManager
     const double                                      neighborhood_threshold);
 
   /**
-   * @brief Set the combined offsets for periodic boundaries.
+   * @brief Set the periodic offset (signed period) for every direction.
    *
-   * @param[in] offsets Combined periodic offsets for periodic
-   * boundaries, used for determining periodic contacts.
+   * @param[in] offset_per_direction Signed period of the domain along every
+   * direction, used for determining periodic contacts.
```
jatred, line 194: On lines 102-103 of particle_particle_fine_search.cc in this PR, you use "signed periodic distance" instead of only "signed period". 
In my opinion the first one is a lot more descriptive than the second, so I would replace all the "signed period" instances with "signed periodic distance".

If you wanted to keep the "signed period" notation, I would at least replace lines 102-103 with that for consistency.

include/dem/particle_particle_fine_search.h
```diff
@@ -45,8 +45,10 @@ using adjacent_pairs_for_contact_t = std::conditional_t<
  * @param contact_pair_candidates The output of broad search which shows
  * contact pair candidates.
  * @param neighborhood_threshold A value which defines the neighbor particles.
- * @param combined_periodic_offsets A vector of tensors of the periodic offsets
- * to change the location of the particles crossing periodic boundaries.
+ * @param periodic_offset_per_direction An array whose component d holds the
+ * signed distance associated with each periodic directions. If the given d
```
jatred, line 49: Is signed distance used interchangeably with signed period in this context?

source/dem/particle_particle_fine_search.cc
```diff
@@ -81,57 +86,77 @@ particle_particle_fine_search(
...
             {
-              // Getting contact information and particle 2 as local variables
+              // Getting the contact information and particle 2 as local
+              // variables
               auto &adjacent_pair_information =
                 adjacent_particles_list_iterator->second;
               auto &particle_two = adjacent_pair_information.particle_two;
 
-              // Finding the properties of the particles in contact
+              // Finding the properties of particles 2.
```
jatred, line 95:
```suggestion
              // Finding the properties of particle 2.
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -81,57 +86,77 @@ particle_particle_fine_search(
...
+              // distance.
+              Tensor<1, dim> cached_translation =
+                adjacent_pair_information.periodic_offset;
+              const double cached_square_distance =
+                particle_one_location.distance_square(
+                  particle_two_real_location + cached_translation);
+
+              // Using the caches periodic offset, if the shared distance
+              // respect the neighborhood threshold, this means that the
+              // particles stayed on their respective side of the PBC.
```
jatred, line 112:
```suggestion
              // Using the cached periodic offset, if the shared distance
              // respects the neighborhood threshold, this means that the
              // particles stayed on their respective side of the PBC.
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -81,57 +86,77 @@ particle_particle_fine_search(
...
+                  // list.
+                  // The nearest_translation is the periodic offset.
+                  const auto [nearest_translation, found_periodic_translation] =
+                    nearest_periodic_translation<dim>(
+                      particle_one_location,
+                      particle_two_real_location,
+                      periodic_offset_per_direction);
+
+                  // If a periodic offset was found, the min_square_distance is
+                  // computed. Otherwise, we know that the particle are far
```
jatred, line 133:
```suggestion
                  // computed. Otherwise, we know that the particles are far
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -81,57 +86,77 @@ particle_particle_fine_search(
...
+
+                  // If the min_square_distance does not respect the
+                  // neighborhood threshold at this point, this means that the
+                  // particle officially moved apart. We can remove the
+                  // potential contact from the list.
+                  if (min_square_distance > neighborhood_threshold)
+                    adjacent_particles_list_iterator =
+                      second_particles.erase(adjacent_particles_list_iterator);
+
+                  // Otherwise, this means that one or both particle crossed the
```
jatred, line 150:
```suggestion
                  // Otherwise, this means that one or both particles crossed the
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -185,37 +210,30 @@ particle_particle_fine_search(
...
-                      nearest_translation = translation;
-                    }
-                }
-
-              // If simulation is well defined, there should be at most one
-              // translation from combined_periodic_offsets that brings a
-              // periodic image within a neighborhood threshold.
+              // nearest_periodic_translation returns periodic translation that
+              // brings two particle the closest together. (if they are close
+              // enough)
```
jatred, line 215:
```suggestion
              // nearest_periodic_translation returns the periodic translation that
              // brings two particles the closest together. (if they are close
              // enough)
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -185,37 +210,30 @@ particle_particle_fine_search(
...
+                nearest_periodic_translation<dim>(
+                  particle_one_location,
+                  particle_two_real_location,
+                  periodic_offset_per_direction);
+
+              // If is possible that the two particle moved far appart so much
+              // that the round operation in the nearest_periodic_translation
+              // function return a 0. In this case, found_periodic_translation
+              // will be at false. Thus, we need to make sure that this isn't
+              // the case by imposing min_square_distance to max().
```
jatred, line 226: I may have wrongly interpreted what happens in the code, but I would rewrite this comment as something like this

```suggestion
              // If two particles move far enough apart, the rounding operation in 
              // the nearest_periodic_translation function may return 0, which 
              // would leave found_periodic_translation as false. In that case, we 
              // set min_square_distance to std::numeric_limits<double>::max() so 
              // that it never falls within the neighborhood_threshold.
```

source/dem/particle_particle_fine_search.cc
```diff
@@ -185,37 +210,30 @@ particle_particle_fine_search(
...
+              const double min_square_distance =
+                found_periodic_translation ?
+                  particle_one_location.distance_square(
+                    particle_two_real_location + nearest_translation) :
+                  std::numeric_limits<double>::max();
+
+              // If the neighborhood_threshold is respected, we add particle 2
+              // to particle 1 potential contact list. Otherwise, we do nothing.
```
jatred, line 234:
```suggestion
              // If the neighborhood_threshold is respected, we add particle 2
              // to particle 1's potential contact list. Otherwise, we do nothing.
```

source/dem/particle_particle_fine_search.cc
jatred: I'm not sure in what situation a particle might move enough in a single time step to be simultaneously in the contact_pair_candidates and have a found_periodic_offset round to 0, but I don't want to hold my review because I don't fully understand the contact detection.

I might have another question on that part of the code when I have a better understanding.

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -227,39 +275,25 @@ PeriodicBoundariesManipulator<dim>::check_and_move_particles(
...
+  // Since Lethe only support PBC that are align the cartesian coordinates
+  // (x,y,z), each entry of periodic_offsets is nonzero in exactly one component
+  // (the direction of the associated periodic boundary pair). Further more,
+  // Lethe does not support multiple PBC in the same direction. Instead of
+  // storing the entire Tensor<1,dim>, we only store the norm of that tensor,
+  // which is equivalent to the distance between linked periodic boundaries, or
+  // period of the domain for a given direction. Since the PBC can be declare
+  // in any order (not necessarily x first, y second and z last) we use the
+  // += instead of .norm()
+  //
```
jatred, line 291:
```suggestion
```

source/dem/periodic_boundaries_manipulator.cc
```diff
@@ -227,39 +275,25 @@ PeriodicBoundariesManipulator<dim>::check_and_move_particles(
...
 void
-PeriodicBoundariesManipulator<dim>::compute_combined_periodic_offsets()
+PeriodicBoundariesManipulator<dim>::compute_periodic_offset_per_direction()
 {
-  this->combined_periodic_offsets.clear();
-
+  std::array<double, dim> offset_per_direction{};
+
+  // Since Lethe only support PBC that are align the cartesian coordinates
+  // (x,y,z), each entry of periodic_offsets is nonzero in exactly one component
```
jatred, line 283:
```suggestion
  // Since Lethe only supports PBC that are aligned with the cartesian coordinates
  // (x,y,z), each entry of periodic_offsets is nonzero in exactly one component
```

jatred (commented):
Nice refactoring, it makes it easier to follow the offsets when they are combined in this fashion.

I mostly have questions about the code and changes/corrections to comments, nothing to change in the structure itself.

source/dem/particle_particle_fine_search.cc
jatred: Okok I think I get it, it's for an edge case where you only have two cells in the periodic direction

PR #2130, opened 2026-09-20
Sequentialize writing the cell status

blaisb:
### Description

The mobility application test is unstable sometimes on more than one core. This is because the output, even though they have an MPI wait command, can be non-deterministic

### Solution

Serialize the output using processor 0  instead of having every processor output their own information. This is essentially what is already done for the DEM position and velocity.

### Testing

Testing on my machine and it gives the same results, as expected. Should make the CI more robust.

### Documentation

Nothing to document

Just spelling mistakes and a double check with Claude.

Code related list:

Pull request related list:
- [x] Labels are applied


source/dem/visualization.cc
```diff
@@ -167,30 +166,33 @@ Visualization<dim, PropertiesIndex>::print_intermediate_format(
...
-          // contains some deal.II version information that may change
-          std::string  line;
-          unsigned int counter = 0;
-          while (std::getline(out, line))
-            {
-              if (counter++ > 4)
-                std::cout << line << std::endl;
-            }
-        }
+      if (counter++ > 4)
```
OGaboriault, line 181: Maybe add that this is how we drop the header lines explicitly.

OGaboriault (approved):

source/dem/visualization.cc
blaisb: Yeah true good point

PR #2132, opened 2026-09-22
Update linear solver tolerance in ``solve_linear_system_and_update_solution``

antonellasuccar:
### Description
 
This PR fixes two things in ``solve_linear_system_and_update_solution``: the solver now takes into account a relative tolerance and multiplies it by the L2-norm of the rhs of the linear system to get the solver's tolerance. It is also possible to scale the residual by dividing it by the square root of the triangulation volume, when ``rescale residual=true``. The tolerance printed corresponds to the scaled residual multiplied by the relative tolerance passed in the prm, while the minimum tolerance limits this value. In reality, the linear solver uses the unscaled residual and multiplies the tolerance obtained in the previous step by the square root of the volume of the triangulation, when ``rescale residual=true``. Nonetheless, this feature is not used currently for the void fraction linear solver.

Claude found the ``rescale residual`` bug in ``solve_linear_system_and_update_solution``. 

I followed what is done with the linear tracer solver. However, I'm not sure if it is valuable to use the relative residual for a linear solver, the same way it is for the nonlinear problems that also use a linear solver. I also think the output of the tests did not change a lot because the relative residual was small enough that the minimum residual is mostly taken (before, the relative residual was not used at all and only the minimum residual was used). We should always be careful to select a very small relative residual. I am also not sure if we need extra tests for this.

Code related list:

Pull request related list:
- [x] Labels are applied
- [ ] The PR description is clean and is ready to be used as the commit message when merging the PR


blaisb (approved):

blaisb (approved):
Good. this is a nice change,.

PR #2133, opened 2026-09-23
Multiphysics cfd dem

OresteMarquis:
### Description

MultiphysicsInterface is now VANS-aware. FluidDynamicsVANS/FluidDynamicsVANSMatrixFree::iterate() previously never called multiphysics->solve(), so auxiliary physics enabled under a VANS/CFD-DEM solver were allocated, post-processed, and checkpointed, but never actually solved (fixed as part of this PR). Now that they run, MultiphysicsInterface needs to know whether it's driven by a standard or a volume-averaged (VANS) Navier-Stokes solver. None of the heat transfer, tracer, CLS, and Cahn-Hilliard has a volume-averaged form implemented yet so enabling one of them together with lethe-fluid-vans, lethe-fluid-vans-matrix-free, lethe-fluid-particles, or lethe-fluid-particles-matrix-free now throws . Electromagnetics is exempt, since the time-harmonic Maxwell equations don't change form in a VANS framework, only their magnetic property cahnge to be dependent on the void fraction (this will be part of future PR). The intgration has been done using a is_vans flag is threaded through the constructors from NavierStokesBase → FluidDynamicsMatrixBased/FluidDynamicsMatrixFree → MultiphysicsInterface, defaulted to false so every non-VANS solver is unaffected.

When creating the test to verify the implementation, I notice that the CFD-DEM/VANS solvers did not supported mesh adaptation. This is a feature that I will want to use when performing MMS using the void fraction to test my electromagnetic coupling. Therefore, part of this PR I added the support for mesh refinement. The particle handler and the void fraction solution were never synchronized across an actual triangulation change for these solvers, but the machinery to do this already existed for load balancing. Consequently, this PR refactor the part of the load_balancing function useful for the mesh adaptation into prepare_particles_for_mesh_adaptation()/unpack_particles_after_mesh_adaptation() on FluidDynamicsVANS/FluidDynamicsVANSMatrixFree, and used by the CFDDEMSolver/CFDDEMMatrixFree while also rebuilding the DEM contact-detection caches. These function are also reuses in load_balance(). Note that I kept the mesh-refinement synchronization contained in the VANS/CFD-DEM class hierarchy to keep the NavierStokesBase's own refine_mesh() logic is untouched, but I am not convince this it the most robust approach. 

### Testing

- tests/solvers/multiphysics_interface_01.cc was extended to check that heat transfer, tracer, CLS, and Cahn-Hilliard each throw when enabled together with is_vans = true, and that electromagnetics does not.
- New application tests, time_harmonic_maxwell_waveguide.prm, were added under both applications_tests/lethe-fluid-vans/ and applications_tests/lethe-fluid-particles/. Both solve the same time-harmonic Maxwell waveguide problem also covered by the lethe-fluid tests and check that the L2 errors match, while also exercising mesh refinement: the VANS version refines the mesh via number mesh adapt in a steady run; the CFD-DEM version (which doesn't support steady simulations) instead uses bdf1 with mesh adaptation enabled and the time-harmonic Maxwell time coupling strategy set to resolve every time step, reproducing the same two-mesh-level convergence check through time-stepping instead.
- applications_tests/lethe-fluid-particles/spouted_bed_load_balancing.mpirun=2.output needed regeneration: reusing the new particle/void-fraction synchronization hooks inside load_balance() changed console print ordering (a diamond-shaped-cell warning now prints after "Setup DOFs" instead of before, matching the new call order) and caused a negligible last-digit floating-point difference in aggregated particle statistics (summation-order effect, not a physical change). 

Claude Code (Sonnet 5 / Opus 5) was used throughout this PR:  understanding the code, creating the new tests, and asking for review passes.

Code related list:

Pull request related list:
- [x] Labels are applied


include/solvers/fluid_dynamics_matrix_free.h
```diff
@@ -419,8 +419,12 @@ class FluidDynamicsMatrixFree
    * according to simulation parameters.
    *
    * @param[in] nsparam Relevant parameters for the solver.
+   * @param[in] p_is_vans Whether the solver solves the volume-averaged
+   * Navier-Stokes (VANS) equations, as opposed to the standard
+   * Navier-Stokes equations.
    */
-  FluidDynamicsMatrixFree(SimulationParameters<dim> &nsparam);
+  FluidDynamicsMatrixFree(SimulationParameters<dim> &nsparam,
+                          const bool                 p_is_vans = false);
```
blaisb: maybe i'd call it fluid_solver_is_vans
or something like that.

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -831,49 +784,60 @@ CFDDEMSolver<dim, PropertiesIndex>::load_balance()
...
-    1 + this->particle_projector.previous_void_fraction.size());
-
-  GlobalVectorType vf_distributed_system(
-    this->particle_projector.locally_owned_dofs, this->mpi_communicator);
-
-  vf_system[0] = &(vf_distributed_system);
+  // Restore the void fraction solution, unpack the particle handler,
+  // rebuild the vertex-to-cell map and the DEM contact-detection caches,
+  // exactly as done for an actual mesh refinement.
+  this->unpack_particles_after_mesh_adaptation();
```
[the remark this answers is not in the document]
blaisb: Yeah that's a good point.

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -829,49 +784,60 @@ CFDDEMSolver<dim, PropertiesIndex>::load_balance()
...
```
blaisb: To be more clear, I would advise that you use a different function to "solution transfer" the void fraction, than to transfer the particles. The two are not the same thing exactly.

source/fem-dem/cfd_dem_coupling.cc
```diff
@@ -829,49 +784,60 @@ CFDDEMSolver<dim, PropertiesIndex>::load_balance()
...
+}
 
-  std::vector<GlobalVectorType> vf_distributed_previous_solutions;
+template <int dim, typename PropertiesIndex>
+void
+CFDDEMSolver<dim,
+             PropertiesIndex>::rebuild_dem_caches_after_triangulation_change()
```
blaisb: Is it really a rebuild, or could it just be "build_dem_data_structures" or something similar that you would call whenever you do the triangulation6
The reason for calling the function does not need to reside in the name of the function in my opinon.

source/fem-dem/cfd_dem_coupling_matrix_free.cc
```diff
@@ -779,31 +781,10 @@ CFDDEMMatrixFree<dim, PropertiesIndex>::load_balance()
...
-  vf_system_trans_vectors.prepare_for_coarsening_and_refinement(
-    vf_set_transfer);
-
-  // Let's now prepare the particle handler for the load balancing
-  // Prepare particle handle for serialization
-  this->particle_handler.prepare_for_coarsening_and_refinement();
+  // Prepare the void fraction solution and the particle handler for the
+  // upcoming triangulation change, exactly as done for an actual mesh
+  // refinement.
+  this->prepare_particles_for_mesh_adaptation();
```
blaisb: I would have two functions. prepare_particles and prepare_void_fraction.
They are (IMO) two different things and should be treated as such.

source/fem-dem/fluid_dynamics_vans.cc
```diff
@@ -187,6 +186,107 @@ FluidDynamicsVANS<dim, PropertiesIndex>::vertices_cell_mapping()
...
+  void_fraction_solution_transfer =
+    std::make_unique<SolutionTransfer<dim, GlobalVectorType>>(
+      particle_projector.dof_handler);
+  void_fraction_solution_transfer->prepare_for_coarsening_and_refinement(
+    vf_set_transfer);
+
+  // The particle handler must also be prepared before the triangulation
+  // changes, so particles are correctly relocated to their new cells.
+  particle_handler.prepare_for_coarsening_and_refinement();
+}
```
blaisb: Split into two function
Prepare_particles_for_mesh_adaptation

prepare_void_fraction_for_mesh_adaptation

Otherwise it's confusing.

blaisb (changes requested):
Some initial comments i will continue tonight / tomorrow

